PR #309, opened 2014-07-14
Use clang-format to format C++ code.

take-cheeze:
See builds/run-clang-format.sh and http://clang.llvm.org/docs/ClangFormatStyleOptions.html .


fdelapena:
Clang formatted added a mechanism to disable formatter for specific blocks:
http://lists.cs.uiuc.edu/pipermail/cfe-commits/Week-of-Mon-20141006/116106.html

This approach pollutes the source code a bit, but maybe is worth to consider.

fdelapena:
Can't be merged long ago, please reopen it if it gets rebased.

PR #320, opened 2014-07-25
CMakeLists.txt: Fix boost paths and C++11 compilation

Ancurio:
@take-cheeze review please?

The 'string(TOUPPER ${i} i)' directive ensures correct inclusion
of all found packages except Boost, where the variable name has
to stay in its original mixed case form (not 'BOOST' but 'Boost').
To generalize this, just include both uppercase path variables
and normal case ones.

Before, I just added 3 extra lines that find_package()'d and included boost manually, but this approach looks more general.

This patch also includes passing the `-std=c++0x` flag to the compiler, which for some reason doesn't happen on my system automatically.


Ancurio:
Removed the C++11 flag per take-cheeze's request and rebased.

fdelapena:
Thanks!

PR #387, opened 2015-01-03
Fix bug with variable indexing.

ProGM:
Hi,
This is a bug fix for a problem with "Variable Reference" functionality of rpgmaker 2003.
This one:
RPG Maker 2003 variables

Here's an example project that shows the bug:

https://dl.dropboxusercontent.com/u/55372319/TestVariables.zip

Talking to the dog in RPG_RT.exe returns this:
RPG_RT output
In easyrpg-player returns this:
EasyRPG output


fdelapena:
Nice find, thanks for the fix!

PR #412, opened 2015-02-03
Localisation addition (French)

BlisterBoy:
Android build : Add French Translation


Ghabry:
Will this "just work" when this file is available?

Don't be sad but nearly nobody will see the French translation. According to Google Play Stats (They are collecting a lot) the installs in French are less then 1%.

Country | Percentage
------------ | -------------
Japan | 55.44%
USA | 9.27%
Korea | 6.82%
Germany | 5.50%
Mexiko | 2.73%

The Japanese extreme high percentage is because of a news on GigaZine.

BlisterBoy:
Yep of course it work :p. Excuse me I should have provided an image : 
http://img11.hostingpics.net/pics/510800Capturedcran20150203094759.png

Android selects the more appropriate ressources according to the name of the directory. In this case, if the system is fr-fr, fr, fr-be etc. Android will take the string in `values-fr` in priority, and the other in the default directory (if all the string are not in string.xml).
If the system is in another langage, Android take the default directory (the one without option in its name).

Maybe the french community is not enormous but easyRPG interests them :). The last time a saw a thread in rpg-maker.fr, this app was not totally functional, this can be an explanation.

I am also moving some text from the code to the .xml, in Android we have to put string in string.xml to improve misspelling checking and localization :).

fdelapena:
Though game configuration settings could be eventually moved to a cross platform menu some day, I think having this on Android earlier is good as this is one of most popular (if not the first) port nowadays :+1:. Translation strings could be reused later for the portable C++ app. I'll provide the `values-es/string.xml` translation on another pull request.

BlisterBoy:
Okay, so I'll transfer all the ouput instructions on the code to string.xml. There are a lot of little ones remaining in the code :).

builds/android/res/values/strings.xml
```diff
@@ -20,4 +20,5 @@
 	<string name="thai">Thai</string>
 	<string name="vietnamese">Vietnamese</string>
 	<string name="cancel">Cancel</string>
-</resources>
+	<string name="no_game_found">No game found in $PATH</string>
```
fdelapena: games has been replaced with game here. Could you push another commit to master adding the missing "s"?

BlisterBoy:
Scuse, the no 's' is a french grammar rule :p.

Ghabry:
If you want you can also translate the other strings that are hardcoded in the code because I was lazy ;)

BlisterBoy:
No problem, I will do that :).

BlisterBoy:
String transfers and translation terminated :).

builds/android/AndroidManifest.xml
```diff
@@ -4,8 +4,8 @@
 -->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
       package="org.easyrpg.player"
-      android:versionCode="1"
-      android:versionName="1.0"
+      android:versionCode="2533"
+      android:versionName="0.2.2.70"
```
carstene1ns: You accidentally overwrote a recent change here, please `git pull/rebase` before submitting the changes to avoid this in the future.

Ghabry: That wasnt an incorrect merge. This happened because he used the build script we use in Jenkins. This replaces the versionCode and versionName.

BlisterBoy: Yep I forgot to ignore this file in the first commit, should I have to do anything ?

Ghabry: Just make another commit with the lines changed to
android:versionCode="1"
android:versionName="1.0"

BlisterBoy:
That's done :)

fdelapena:
Thanks, almost done, you have committed a new file (.orig.xml) instead.

BlisterBoy:
Ahahah wtf what is this file, probably the github app, FIX.

PR #417, opened 2015-02-16
Emscripten playground

fdelapena:
Trying to play it at https://easy-rpg.org/jenkins/job/player-js-pr/ws/easyrpg-player.html

### DIFF
--- a/src/audio.cpp
+++ b/src/audio.cpp
@@ -23,8 +23,12 @@
 
 AudioInterface& Audio() {
 	static EmptyAudio default_;
+#ifdef SUPPORT_AUDIO
 	if (Player::no_audio_flag || !DisplayUi) {
 		return default_;
 	}
 	return DisplayUi->GetAudio();
+#else
+	return default_;
+#endif
 }
--- a/src/graphics.cpp
+++ b/src/graphics.cpp
@@ -351,10 +351,13 @@ void Graphics::Transition(TransitionType type, int duration, bool erase) {
 				screen1 = screen2;
 		}
 
+#ifndef EMSCRIPTEN
+		// Fixme: Refactor how transitions work, they should return to the main loop
 		for (int i = 1; i <= transition_duration; i++) {
 			Player::Update();
 			InternUpdate1();
 		}
+#endif
 	}
 
 	if (!erase) frozen_screen = BitmapRef();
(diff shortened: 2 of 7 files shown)


fdelapena:
jenkins test this please

fdelapena:
PR builder works again:
https://easy-rpg.org/jenkins/job/player-js-pr/ws/easyrpg-player.html

Currenty shows this exception message:
```Uncaught exception: 0```

Maybe exit(1) throws this. Looks promising.

Ghabry:
Actually it seems to work except the rendering. When you press enter you can see that it writes "Warning: No start position" in the Javascript console which indicates that Scene::Title is active.

fdelapena:
I haven't checked Player code but SDL emscripten tests use 32 bit sw surface for screen and SDL_flip. Maybe requires some guards. By the way, they changed EMSCRIPTEN to __EMSCRIPTEN__.

fdelapena:
Maybe getting files from https://github.com/EasyRPG/TestGame/ (or the desired game) and putting them from the Jenkins job is better for fetching from the same domain (preventing CORS issues).

The function requires 2 parameters and looks like the 1st and 2nd could be the same if placed in the same relative path, the first the real file system and the second the virtual file system If I understand the documentation correctly.

Ghabry:
This port would need a refactoring of the code because the file loading must be asyncronous. I don't see a reason for downloading everything at the beginning, most files are never used anyway during one play session (and music can be huge).
This basicly means:
No file can be used directly after it was requested (e.g. via a Cache::-function). Instead you have to deal for one frame without the file. Then it returns to the main loop where you pause the main loop until all download requests are finished. And on the 2nd frame you can finally work with the file.
(Of course loading can be still done syncrhonous on other systems, but the code should be async-friendly)

Ghabry:
Jenkins, retest this please <3

Ghabry:
Jenkins, retest this please <3

Ghabry:
Jenkins, retest this please <3

Ghabry:
Jenkins, retest this please <3

Ghabry:
Jenkins retest this please <3

Ghabry:
This is ready for now... Audio doesn't work and FPS are not really 60...

PR #418, opened 2015-02-17
Add enemy drops after battle

### LINKED ISSUE #398: Battle: Dropped items are ignored
Items dropped from enemies are not added to the inventory of the player.

This is especially bad in Vampires Dawn because you need souls to create save stones that allow saving.

Test case: Basicly any battle in VD
> After a fight the hero should get souls according to the opponent: if it was a weak one, then a weak soul, a strong one then a strong soul, but the souls are never dropping unfortunately.

Reported by Darkztar

scurest:
This should fix #398. Also fixes a related problem where the Gold/EXP/Level Up messages don't show at the end of an RPG2k3 battle. I'm not sure if there's a better way to do that last part though.

### DIFF
--- a/src/game_enemy.cpp
+++ b/src/game_enemy.cpp
@@ -153,6 +153,14 @@ int Game_Enemy::GetMoney() const {
 	return enemy->gold;
 }
 
+int Game_Enemy::GetDropId() const {
+	return enemy->drop_id;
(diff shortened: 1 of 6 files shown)


Ghabry:
Jenkins: ok to test

Issue #399, opened 2015-01-27
Multi-target spells cost n*cost (n = target amount)


Ghabry:
Multi-target spells should only cost once the cost and not multiplied with all targets.

I wonder how this even happened, wasnt on purpose.

Test case: Any spell that targets multiple enemies in battle

> If you fight more than one opponent in the fight and you use a spell which attack all opponents this spell takes the resource costs twice for one attack, respectively so often how much opponents you have.
> E.g. a spell costs 7 blood if you have 1 opponent and it takes 14 for one attack if you have 2, but it should take still 7. The same happens with items, which you can use for attacking all opponents. It uses 2 items for 2
>opponents for one attack, but it should use only 1.
>In the example above: If the hero has something between 7 and 14 blood left and you use the spell for 7 which attacks all opponents, so after the round you have 0 blood left.

Reported by Darkztar

PR #420, opened 2015-02-18
Implement multi-use items.

scurest:
When using an item, only one "use" should be used up now.

A quick test in RM2K showed
* the number of uses resets when an item is removed
* the number of uses stays the same when an item is added, even if the player already had x99 of the item

Also, for the same reason as #399, using a special item in battle consumes a number of uses equal to the number of targets, currently.

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -562,7 +562,7 @@ void Game_BattleAlgorithm::Skill::Apply() {
 	AlgorithmBase::Apply();
 
 	if (item) {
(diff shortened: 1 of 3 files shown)


Ghabry:
Jenkins: ok to test

fdelapena:
Thanks @scurest! In order to get automatic Jenkins pull requests building for automated tests in multiple platforms, please accept the "Player team" invitation from GitHub to get your user whitelisted in the Jenkins Continuous Integration service. This will prevent the need of the "okay to test" comment for every opened pull request.

Issue #356, opened 2014-11-20
Map transition make wait the parallel process/auto event


gadesx:
Problem: If you have a map, when you start and it haves in an event:
Change color: 0 0 0 0 (black) 0.0s wait.
When you play this map, in RPG Maker you don't see the map but in easy
only is processed when the animation of the transition is finished.

Test case added, start in "Test case 1 map"

http://dl.dropboxusercontent.com/u/22524283/temporal/Tests%20varios.rar

fdelapena:
This issue seems to be fixed already.

Issue #344, opened 2014-11-01
Interpreter hangs when running "continue processing events" command after "move event" command execution on parallel interpreter


fdelapena:
How to reproduce:

1. Create a parallel event with a move event command.
2. On the same map, create an automatic event with a continue processing events command.
3. This command must be set after some behavior specific commands, like wait command or move event. Otherwise, after e.g. a show message will work fine (?).

Here is a [test case](https://fran.cr/owncloud/index.php/s/KHwW2DiMk4IZoz2/download) to verify this weird race condition.

Verified affected games: 
- Desolate City (initial cutscene)

scurest:
Has this one changed, maybe? I couldn't find Desolate City (link please? :) but I don't think I got the hang in Rainbow Treasure (you don't mean this one,
![hang](https://cloud.githubusercontent.com/assets/11024420/6281193/2d7c5950-b881-11e4-8f4b-72f63418c358.png),
do you?).

RPG_RT will hang on the testcase if the movement command takes time to complete. Seems like it only goes through to the next message box if the movement route completes instantly, like a "Face Right" or an empty route or something.

fdelapena:
Not sure if 8cb3d84 changed the behaviour of the test case but it is still hanging at least for Desolate City. Here is a link to the [Desolate City](https://www.sendspace.com/file/a8n0g5) game. The game is unreleased yet, but authors contribute with Player with bug reporting.

About the Rainbow Treasure image hang yes, the hang I was reporting is in this cutscene point, just after these "!" appear.

scurest:
The hang with the ！ is actually unrelated: it's caused like this

1. a switch goes on
2. the ！ events' page condition is met, so they come into existence.
3. all the characters except Nickel jump and land
4. because Nickel is on top of the location of the bottom most ！, the game won't let her land (in `isLandable`, the first loop that checks whether there's an event at the same location won't go through) so she gets stuck in the air.

ec7f07de let's you get past that spot in the cutscene.

Thanks for the link for Desolate City, I'll look at it later.

fdelapena:
Wow, neat find, I was guessing it was the same issue because this operation was using the "continue processing events" too and when removing some commands was continuing.

About Desolate City: recently Zegeri said the game may have some relation with the mirror event.

scurest:
Confirmed it's the mirror event, just like in the testcase.

[...]

Issue #333, opened 2014-09-07
Passable upper layer tiles with "star" terrain attribute shows player and events character head


fdelapena:
Character's head is visible when it is in movement when walking in vertical, from up to down and from down to up when passing below "star" tiles.

Looks like RPG_RT always shows the whole player and charset-based events below star tiles (even when the "invisible" first upper tile is opaque in some tilesets).

[Test case](https://fran.cr/owncloud/index.php/s/MHE6leSXZcfoCH8/download) download available.

This also happens with events with character graphics moving around these star tiles.

Ghabry:
It's possible that this was fixed by the looping maps stuff merged yesterday. Modified some Z-values.

fdelapena:
Tested, works better: now the head is not visible when the character stops, but it is still visible when it is in movement when walking in vertical, from up to down and from down to up.

PR #428, opened 2015-02-26
Fixes for Rainbow Treasure OP

scurest:
Fixes the errors I noticed in the Rainbow Treasure OP.

##### Character placement off  f1fe079

tear

Some events are drawn off by one pixel. Caused by the formula in `Game_Character::GetScreenX()` and `Y()`. Changed to take the difference in pixels, not real coordinates, which I _think_ is correct. Can someone check?

##### Pan speed too slow  be05554

pan

The pan drags behind the characters during the OP. Changed just by increasing the pan speed, but also changed to use `SCREEN_TILE_WIDTH / 128` instead of `2`, like for characters, I think it goes here too: @Ghabry, you made that change for character speed, what do you think?

##### Pan jumps after map teleport  8efd1a5

copper

Caused by the pan in `Game_Character` being reset too early. I think this also fixes a bug where the pan is lost when teleporting within the same map (?).

##### Parallax BGs don't scroll backwards  28cedf8

manganese

Manganese's OP uses a parallax background to make it look like she's walking, but it goes the wrong way! Just a missing sign in `UpdateParallax` but I also used `SCREEN_TILE_WIDTH / 128` here and sped it up.

##### Jumping in place 716623d

jump

Discussed [here](https://github.com/EasyRPG/Player/issues/344#issuecomment-75355015).

--------------

There's also #356 and one more I just noticed—the old man in front of the chest's sweat drop is drawn behind him when it should be drawn in front—that I didn't fix.

### DIFF
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -1018,7 +1018,11 @@ int Game_Character::BeginJump(const RPG::MoveRoute* current_route, int current_i
 int Game_Character::EndJump(const RPG::MoveRoute* current_route, int current_index) {
 	jumping = false;
 
-	if (!IsLandable(jump_x + jump_plus_x, jump_y + jump_plus_y)) {
+	if (
+		// A character can always land on a tile they were already standing on
+		!(jump_plus_x == 0 && jump_plus_y == 0) &&
+		!IsLandable(jump_x + jump_plus_x, jump_y + jump_plus_y)
+	) {
 		// Reset to begin jump command and try again...
 		move_failed = true;
 
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -142,11 +142,11 @@ void Game_Character::MoveTo(int x, int y) {
 }
 
 int Game_Character::GetScreenX() const {
-	return (real_x - Game_Map::GetDisplayX() + 3) / (SCREEN_TILE_WIDTH / TILE_SIZE) + (TILE_SIZE/2);
+	return real_x / (SCREEN_TILE_WIDTH / TILE_SIZE) - Game_Map::GetDisplayX() / (SCREEN_TILE_WIDTH / TILE_SIZE) + (TILE_SIZE/2);
 }
 
 int Game_Character::GetScreenY() const {
-	int y = (real_y - Game_Map::GetDisplayY() + 3) / (SCREEN_TILE_WIDTH / TILE_SIZE) + TILE_SIZE;
+	int y = real_y / (SCREEN_TILE_WIDTH / TILE_SIZE) - Game_Map::GetDisplayY() / (SCREEN_TILE_WIDTH / TILE_SIZE) + TILE_SIZE;
 
 	int n;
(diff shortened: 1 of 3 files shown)


src/game_character.cpp
```diff
@@ -142,11 +142,11 @@ void Game_Character::MoveTo(int x, int y) {
 }
 
 int Game_Character::GetScreenX() const {
-	return (real_x - Game_Map::GetDisplayX() + 3) / (SCREEN_TILE_WIDTH / TILE_SIZE) + (TILE_SIZE/2);
+	return real_x / (SCREEN_TILE_WIDTH / TILE_SIZE) - Game_Map::GetDisplayX() / (SCREEN_TILE_WIDTH / TILE_SIZE) + (TILE_SIZE/2);
 }
 
 int Game_Character::GetScreenY() const {
-	int y = (real_y - Game_Map::GetDisplayY() + 3) / (SCREEN_TILE_WIDTH / TILE_SIZE) + TILE_SIZE;
+	int y = real_y / (SCREEN_TILE_WIDTH / TILE_SIZE) - Game_Map::GetDisplayY() / (SCREEN_TILE_WIDTH / TILE_SIZE) + TILE_SIZE;
```
Ghabry: This can be simplified to
```
(real_y - Game_Map::GetDisplayY()) / (SCREEN_TILE_WIDTH / TILE_SIZE) + TILE_SIZE;
```
(same for x)

Zegeri: a / (b / c) = a * c / b. So that can be rewritten as:
```
((real_y - Game_Map::GetDisplayY()) / SCREEN_TILE_WIDTH + 1) * TILE_SIZE;
```

Ghabry: Finally a practical use for school math.

fdelapena:
Thanks scurest, about the swear drop issue is possibly related with #333.

src/game_character.cpp
scurest: Unfortunately integer division doesn't distribute over addition, eg. take `SCREEN_TILE_WIDTH/TILE_SIZE`=16, `GetDisplayY()`=1, and `real_y`=16. Then 16/16 - 1/16 = 1, but (16-1)/16 = 0.

Ghabry:
+1 for this really detailed pull request.
Are you aware of any further open issue that are fixed by this?

scurest:
I'm pretty sure no open issues are closed by this one, haha.

By the way... (I've never used Github before) would it have been better to open individual issues for all these and put details there?

fdelapena:
I think there are a good collection of visible bugs not reported in the issue tracker which can be fixed "on the fly" without needing to open a separated bug before and it saves development time. They can be reported before, but I think it is not necessary in the current development stage. A lot of code contributed didn't have an issue previously opened and it can be discussed in pull requests too :) .

Ghabry:
I usually only open issues for problems where I'm certain that there won't be a fix from my side soon.
When you fix some unreported problems just do it like in this PR

PR #436, opened 2015-03-18
Minor vehicle fixes

scurest:
Some very minor fixes on the heels of #434.

2dfa0d5 adds a "stepping speed" to decouple the rate that characters do their left-middle-right-middle animation from their movement speed, but since it's actually the _player_'s movement speed that's faster when riding a vehicle, an alternative would be to set all the vehicle's move speeds to 1/8 and move the switch to choose higher move speeds when riding vehicles to the line changed by e7880d8. I don't think the vehicles' move speed actually affect anything but their stepping animation does it?

Was everyone ok with 10455da?

### DIFF
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -68,6 +68,10 @@ Game_Character::~Game_Character() {
 	}
 }
 
+int Game_Character::GetSteppingSpeed() const {
+	return GetMoveSpeed();
+}
+
 bool Game_Character::IsMoving() const {
 	if (move_count > 0) return false; //Jumping
 
@@ -181,7 +185,7 @@ void Game_Character::Update() {
 		}
 	}
 
-	if (anime_count > 36.0/(GetMoveSpeed()+1)) {
+	if (anime_count > 36.0/(GetSteppingSpeed()+1)) {
 		if (IsSpinning()) {
 			SetPrelockDirection((GetPrelockDirection() + 1) % 4);
 		} else if (!IsContinuous() && IsStopping()) {
--- a/src/game_character.h
+++ b/src/game_character.h
@@ -141,6 +141,14 @@ public:
 	virtual void SetLayer(int new_layer) = 0;
 
 	/**
+	 * Gets character stepping speed: the speed of the left-middle-right-middle
+	 * walking animation. Same as the movement speed, by default.
+	 *
+	 * @return stepping speed (the same units as movement speed)
+	 */
+	virtual int GetSteppingSpeed() const;
+
+	/**
 	 * Gets character movement speed.
 	 *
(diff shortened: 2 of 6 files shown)


fdelapena:
> Was everyone ok with 10455da?

Looks good to me if this fixes that particular offcanvas issue unless there is a better approach.
@Zegeri did you have a planned alternate way to solve this in #434 as commented in the PR description?

Zegeri:
I agree this is the simplest solution, I was thinking of refreshing the sprites of vehicles after SetVehicleLocation, like [ChangeMapTileset does with tileset changes](https://github.com/EasyRPG/Player/blob/6158831773f2a99ed965866e2b4f7540cde4b9be/src/game_interpreter_map.cpp#L1398). But that would mean to have another vector of sprites just for the vehicles, not a pretty solution. 

By the way, this is not the issue commented in the PR description.
>Board a vehicle outside the borders of the map
That means to board a vehicle that it's in the current map, but outside the range of valid positions. For example, if the player is in (1,1) facing left and a ship is in (0,1) (not a valid position), the player should be able to get on the vehicle. This is how RPG_RT behaves.

Ghabry:
I don't think such optimizations are worth the effort for 3 sprites. They won't induce a huge cpu penalty.

Another question is if the Draw-code of off-screen Sprites draws anything or just skips. That would be a general optimization that all sprites profit from. No idea if we already do this.

scurest:
Thanks Zegeri. Can I take out [these lines](https://github.com/EasyRPG/Player/blob/75b405c3525d7d0d7577c7da874d9a828a4a5f13/src/game_player.cpp#L260-L264) too?

@Ghabry [looks like](https://github.com/EasyRPG/Player/blob/master/src/sprite.cpp#L72) a visibility test is the first thing out the door for sprite drawing, but they still participate in z-ordering and stuff. One more thing is the airship shadow is basically one or two more sprites in the spriteset too, and they'll have the same issue.

fdelapena:
About airship shadow: there is a blend of two semitransparent sprites from system.png.

I've prepared two red and blue squares there for testing:
default

You can compare them with these Cairo (pixman) operations pictures to check which blend operation is closer. If you swap the colors, the colour blend result will be different:
http://cairographics.org/operators/

Zegeri:
@scurest Yeah, we won't need those lines.

scurest:
Alright, if everyone's good, that's all the stuff I had for this branch. I have code for the airship shadow too but I'll do it in another PR.

fdelapena:
Looks good to me, thanks :+1:.

Ghabry:
so this is rdy?

scurest:
Yeah.

Issue #438, opened 2015-03-19
Android : Touch Control to reconsider ?


BlisterBoy:
Hi everyone,
On my way to work on fix #437 and #288 I found that button management on the android port need to be totally reconsider.

For example, The EasyRPGPlayer Activity only draw the cross and the button, but the touch control is controller by an extern libraby.. But I didn't found anywhere where the code consider that the (x,y) position correspond to A, B or whatever button, this is just not handled by the android code, which is problematic.
This programming pattern create many problems, such as no possibility to modify a button location on another place, or add new buttons. And this makes the code hardly comprehensible (I took many days to understand the way this thing work).

My proposition (and I'm working on it) is to have a Button extending the View class : 
- :+1: The code would be clearer
- :+1: The button takes care of his own drawing (which is the way component should work on Android)
- :+1: The button takes care of touch event on its own bound
- :+1: The model will facilitate a Button Mapping System (which is part of my final plan mwahaha)
- :+1: It would be easier to debug the program, including this issue of course

Ghabry:
It actually makes more sense to handle the touching in the c++ code (we do it in SdlUi::ProcessFingerEvent) because this code can be shared with other touch devices. E.g. Windows tablets, yes we don't do this yet but think about the feature.
Imo the problem is that the button rendering is not in the C++ code ;)

BlisterBoy:
I understand your point. C++ dev is more portable and directly modify the core engine will have many advantages.

But for now the android port has some problems, the code is not clear and hardly understandable (please don't consider that as blame, the whole program is just amazing). As I'm an Android dev and there is no Window's Tablet port for now, I'm able to fix this issue, clear the code and provide a button mapping system (which is the most popular demand on the port).

If we adopt an OOP approach of the control button on the android port : 
:+1: The code will be **more modular**, more easily modifiable (compare to now)
:+1: **The port will quickly have a button mapping system**, which would interest a lot of people and would make the android port more attractive
:+1: The port will be **compatible with way more RPG Maker game** until a C++ developper add a button mapping system in the core engine (**and we don't know when such a developper would do that**)
:+1: If a C++ dev decide to program all of this in the core engine (in C++), it will just need some minutes to restore the way the program handled touch event before (because the code would be more clear and modular)

If I modify the android port in this way, I just see advantages, does it interest you ?

Ghabry:
You make valid points.
If you feel more used to programming with Java implement the touch-code Android specific for now.
[...]

PR #440, opened 2015-03-21
Creation of the Button Mapping System

BlisterBoy:
Hi guys !
I've implemented the button mapping system #438.
You can summon it in the Game Browser Activity. I had to change a lot of things but now we are abble to :
:+1: Change button disposition.
:+1: Add buttons.

It's kind basic for now, but I hope you'll like it !


Zegeri:
It works like a charm. Thank you. :+1:
I think the F2 (toggle FPS) button is not needed, because you can already toggle the FPS display using the contextual menu. F5/Toggle zoom seems more fitting in the contextual menu too (or removed).

Ghabry:
Awesome! Wohooo.
Busy today write yiu more in the èvening

BlisterBoy:
I'm happy that you like it ^^, it took me the whole day yesterday but I learned a lot of things on Android (I was totally a nooby).
I focused on efficiency, the user experience doesn't satisfy me at 100% but I have a lot of college projects (EasyRPG is one of them) so I will improved the system when I will have the time in the next months : 
:mag: A "save on exit" context menu (for the moment you have to save before exit the activity).
:mag: The possibility to delete a button on a long-press touch
:mag: A zoom system for boutton.
:mag: Slightly extend the hitbox cross direction to improve the taking in hand.
:mag: The possibility to have several Button Mapping and associate one to a specific game

Zegeri I agree that those buttons will not be usefull for everyone but as we say in French "that doesn't eat bread" :p. It may be usefull for debugging purpose so for the moment I prefer not touch that.
But as you can see F2 and F5 are not operational, and the "*" button neither (which is more preoccupying), they are supported after API 11. I left a "//TODO" to remember that if we update the API.

Zegeri:
What if we used an API 1 Keycode like KEYCODE_STAR for "*"?

BlisterBoy:
Thanks Zegeri you make me found an issue : + / - * correspond to the KEYCODE_NUMPAD_... family, keycodes such as KEYCODE_STAR don't work.
I fixed the problem but it brings another one : KEYCODE_NUMPAD family are not supported until API 11, I don't really understand if we build the project on API 12 or 10 but it's maybe the good moment to migrate to API 12.

Less than 8% devices are under 12 and that number tends to quickly reduce since few month = https://developer.android.com/about/dashboards/index.html

I finally deleted F2 and F5 buttons as Zegeri suggested (I faced up the fact that my argumentary was just hiding laziness hahaha).

BlisterBoy:
I fixed a minor bug and increased the virtual cross feeling, unless you found bugs I think I finished for this pull request :).

Ghabry:
Instead of F2 or F11 it's probably better to show the purpose of the button

F2 - Toggle FPS (this one is really pretty useless)
F3 - Show log console. Nice to have
F9 - Show debug menu, only on map scene. Needs Debugmode, you can't toogle this mode under Android at the moment -> useless
F11 - Show save scene: Needs debugmode - same problem as above
F10 - Take screenshot -  Better as menu option
Ctrl left - When pressed: Walk through walls, disable random encounter -> Needs debug
F12 - Back to title (dangerous)

See the SDLActivity. Android provides functions to check the device API level

``
Build.VERSION.SDK_INT >= 12
``

__Before merging at least deleting of buttons would be useful.__

Nice to have:
Per game config -> Edit while game is running
"burger menu button" that toggles the  menu to make it more obvioua that there is functionality in the menu.
Per engine default (rpg2000 and 2003), engine detectable via jni callback to android.
Store button mapping in android settings file.
Rezisable buttons

fdelapena:
For JNI we might differentiate between 2000 and 2000 Value too.

BlisterBoy:
Huuum I don't know why I'm not able to detect a long press in the activity... For now it's possible to reset the mapping, I'll try latter.
The test on API version has been pushed.

As I told few posts before, I deleted Fx buttons :).

Ghabry:
What do devs think about pushing this to play store soon
Users will be happy I guess.

BlisterBoy:
I think so too :).
Vehicules was a huge improvement for the Player !

fdelapena:
:+1: 
What about a global 0.3 early release after the Android build around April 1st?
With vehicles and maybe map loop, then fixes on 0.3.x for thr remaining major bugs

Ghabry:
Okay lets wait another 1-2 weeks. GIves some more time for fixes...

PR #441, opened 2015-03-22
Add shadows to the airship.

scurest:
This adds shadows beneath the airship. I added a new sprite class that's visible when you're on the ship and put an instance in `Spriteset_Map`. The only annoying bit is updating it when the System graphic changes, which I did by having the interpreter notify the spriteset just like for chipset changes.

Are the only files I need to change after adding a file `Makefile.am` and the two `vcxproj` files? Did I do the `vcxproj` files right?

--------

The colors are also wrong, though its hard to tell, especially over a real map. I'm blending the two shadows at 26% opacity, left one first. Here's the System graphic fdelapena posted in #436, drawn against a white background in RPG_RT, and again in EasyRPG with this branch. This also shows the effect of switching the order of the two sprites. The frame that shows the original System graphic in the corner is from RPG_RT.

compare

For comparison, this is using 25% opacity.

com25

The blending method is actually _not_ separable: the blue in the system graphic has 0 for both its R and G components, but when blended to the map, the R and G components are disequal (189 vs 190). 26% was picked to get the R component correct.

(I'm not going to mess with this any more in this PR, this is just for documentation.)

------

The Z-value is also wrong, and has a #333-like effect. I'm not really sure how the Z values work; anyone know what it should be?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -180,6 +180,8 @@ libeasyrpg_player_la_SOURCES = \
 	src/sprite_battler.h \
 	src/sprite_character.cpp \
 	src/sprite_character.h \
+	src/sprite_shadow.h \
+	src/sprite_shadow.cpp \
 	src/sprite_timer.cpp \
 	src/sprite_timer.h \
 	src/sprite.cpp \
--- a/Makefile.am
+++ b/Makefile.am
@@ -176,12 +176,12 @@ libeasyrpg_player_la_SOURCES = \
 	src/sdl_audio.h \
 	src/sdl_ui.cpp \
 	src/sdl_ui.h \
+	src/sprite_airshipshadow.h \
+	src/sprite_airshipshadow.cpp \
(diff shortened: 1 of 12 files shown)


fdelapena:
About the color value imprecision: RPG_RT is not using true color, seems to use RGB565, so don't worry about this.
About the z value, I don't know. There was some chat about current z values for all layers, but I don't remember where was, IRC or issue number.

Ghabry:
Minor nitpick: Maybe call it Sprite_AirshipShadow instead of Sprite_Shadow to make the purpose a bit more obvious. Otherwise nothing to complain :+1:.

Z-Value is a general problem fixing this is out-of-scope for this PR.

You altered the files correclty, PR builder passed.

scurest:
Agreed. Done. (I actually called it that in my first version, but I thought it was a little long :)

PR #443, opened 2015-03-24
Android : Fix the SDK compatibility of the Button Mapping System

BlisterBoy:
As suggested by Ghabry, I used directly the value of keycode. I tested on a test project : everything is alright.


Ghabry:
Can you add some // API11: KeyEvent.KEYCODE_NUMPAD_ADD comments like we do in other places?

To prevent such merge commits you can use "rebase" instead.

BlisterBoy:
Huhu to be honest I use the Github app

PR #445, opened 2015-03-25
Improve frame skipping code. Fixes #287

### LINKED ISSUE #287: Improve Frame Skip (Turbo Bug)
When the FPS drop below 60 the player character movement is not smooth anymore but "jumps"

Ghabry:
The new code takes into account that the non-graphic updates take some time, too and only when there is time in the current frame left the graphic is rendered. The game logic always runs at 60 updates per sec.

Don't merge before I verified that this works under emscripten ;)

### DIFF
--- a/src/game_system.cpp
+++ b/src/game_system.cpp
@@ -55,7 +55,6 @@ void Game_System::BgmPlay(RPG::Music const& bgm) {
 		Audio().BGM_Stop();
 	}
 	data.current_music = bgm;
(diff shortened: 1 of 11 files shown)


Ghabry:
Emscripten doesn't blackscreen but it's not really possible to frameskip due to how the main_loop with requestAnimationFrame works. But all other ports get smooth frameskipping except if the renderer is really overloaded (like with that zoom in gadesx old bugreport :/)

PR #450, opened 2015-04-09
Looping maps

### LINKED ISSUE #314: Looping Maps
RPG Maker features horizontal and/or vertical loop when scrolling, but Player does not have this implemented yet.

Required for some games like Yume Nikki or The Longing Ribbon.

Ghabry:
Fixes #314 

Havn't tested this in any real game yet, please test Yume Nikki & co.

I wasted days with these GetScreenX-Z code to get it right but by looking at the changes it wasn't really difficult in the end :(.

Imo there is some flaw with the Z code, but I wasn't motivated to investigate the tilemap z-layer code. Tilemap rendering doesn't work correctly when over a map boundaries (tile rendered too late, but this is only a visual glitch)

- [x] Rendering issues on small maps (20x15 and 20x16) -> Needs refactor in rendering code, WONT FIX
- [ ] Panoramas rendered wrong
- [ ] Pan makes sprites dissapear

### DIFF
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -148,11 +148,35 @@ void Game_Character::MoveTo(int x, int y) {
 }
 
 int Game_Character::GetScreenX() const {
-	return real_x / (SCREEN_TILE_WIDTH / TILE_SIZE) - Game_Map::GetDisplayX() / (SCREEN_TILE_WIDTH / TILE_SIZE) + (TILE_SIZE/2);
+	int x = real_x - Game_Map::GetDisplayX();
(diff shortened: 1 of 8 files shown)


Ghabry:
There are further rendering issues with events when the map has a width of 20 or a height of 15 or 16.
This is because events can have a height of 2 tiles. On the smallest possible map size RPG_RT wraps the sprite (head at bottom, body at top).
I don't plan fixing this one, is a corner case.

Parallalax is broken in Player, same for Screen Pane (always only when over boundaries)

Ghabry:
superseeded by #455

PR #459, opened 2015-04-27
Silence invalid index messages. Fixes #458

### LINKED ISSUE #458: "Switch/Variable index %d is invalid" warnings should be debug messages
Some games have bugs, e.g. when trying to enable switches/variables from a variable value they may reference to an invalid switch/variable index value, e.g. 0000.

The default behaviour of RPG_RT is just ignore this defect. Player is showing them and this makes these games display annoying warnings in the on-screen console.

fdelapena:


src/battle_animation.cpp
```diff
@@ -39,7 +39,7 @@ BattleAnimation::BattleAnimation(int x, int y, const RPG::Animation* animation)
 		graphic = Cache::Battle2(name);
 	}
 	else {
-		Output::Warning("Couldn't find animation: %s", name.c_str());
+		Output::Debug("Called battle animation without graphic file");
```
Ghabry: Is this really acceptable?
What if a name is supplied but is just not found.

fdelapena: The missing file warning is handled in the FileFinder code, it will try to find the file only when the string parameter is non-empty.

Ghabry: I mean when name of animation is e.g. "Monster" and there is no "monster.png" in "Battle" and "Battle2" this silences that warning now. Instead of reporting "Monster" not found.
When the animation is empty "" then silencing is okay I guess.

PR #466, opened 2015-05-07
Fix Tilemap z-layer bug in async-loading.

Ghabry:
Yeah I moved two lines around with horrible consequences... :(

They are now at there old position :)

Please test (the emscripten) games if this unbroke everything.

Don't merge yet, want to fix some potential race conditions in the async code tomorrow.

### DIFF
--- a/src/game_picture.cpp
+++ b/src/game_picture.cpp
@@ -33,7 +33,8 @@
 Game_Picture::Game_Picture(int ID) :
 	data(Main_Data::game_data.pictures[ID - 1]),
 	old_map_x(0),
-	old_map_y(0)
+	old_map_y(0),
+	request(NULL)
 {
 	SetTransition(0);
 }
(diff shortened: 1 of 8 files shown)


Ghabry:
oops dont merge forgot to remove some debug code

fdelapena:
Looks like BattleCharSet images are failing to load in async mode. It shows the "BUG: ..." debug message. This 2k3 game has a battle with three monkeys after first maps , a couple of minutes after starting. Allied battlers are not shown and they load in desktop version:

https://easy-rpg.org/play/index-pr.html?game=ehr

There is also an issue with some "missing" audio files in this game. Debug messages say they are not found, however they exist and the php file finder are returning them without issues. I've regenerated the cache, however they still fail to load. Other audio files in the same folder work. Maybe is the upper capital case (?).

Ghabry:
Oops just checked the code before BattleCharset is ```FileRequestAsync* request = AsyncHandler::RequestFile("Title", Data::system.title_name);``` which is obviously a copypaste error. Fix it when home.

And how did you get the "All is well — 2 successful checks " with default and player-js-pr. Usually there is only "default" with each pr-build overwriting the other pr results.

Issue #477, opened 2015-05-14
Pointing a variable after the variable 5000 doesn't work


BlisterBoy:
Hi guys :) ! Here's a tricky bug that a friend found.
RPG Maker has recently increase to 9999 the number of switchs and variables, Easy RPG has done the same but it seems that the referencing system hasn't been changed.
For exemple : 
I do V[1] = 7000
But if I do V[V[1]] = 42 that just doesn't work, the variable 7000 isn't changed.
But it works if V[1] = 20.

NB : It works on RPG Maker
NB2 : If a directly do V[7000] = 42 in Easy RPG, it works

BlisterBoy:
Fun fact, it was a trick used to bypass the 5000 variables restrictions in french community.

fdelapena:
A question: is this working when you do \V[1] = 3000? Current code is not limited to 5000.

Ghabry:
Referencing an incorrect variable will return a reference to a dummy-var. So accessing 0, 9999 and 7000 points to the same :).
I think the problem is that RPG_RT does on-the-fly resizing when you do a out-of-bounds access. So if you declared only 100 vars in the editor and you access 200 RPG_RT resizes the array to 200 elements. We don't... does anybody really depend on this? -.-

BlisterBoy:
Unfortunatly... Yes hahaha. 
Please notice that this is not anymore an incorrect use of variables referencing because RPG Maker does support 99999 variables.

fdelapena : I'm currently on Mac OS, I'll tell you that and upload the test project when I'll go back to Windows (I'm writing a presentation of Easy RPG for french community ^^), which website do you use to upload test projects ?

fdelapena:
@BlisterBoy do you mean projects for testing from the HTML5 player? they are uploaded by hand to the web server, if you have the game list for downloading and trying. They need a conversion from midi to ogg and minify assets for better performance. There are however some issues not fixed yet (e.g. black lower layer tileset)
You could also get the HTML5 player and put all games in another host. It requires to run a php script for the online file finder.

BlisterBoy:
Excuse me I didn't explained well, I wanted to upload the testproject (google drive, mega or whatever) so you can test difference with RPG Maker Player and Easy RPG.

But thanks for the explanations, some frenchies ask me questions about the web port today ^^.

carstene1ns:
[...]

Issue #330, opened 2014-08-16
Monsters not shown in the "The Observer" game


fdelapena:
[The Observer](http://www.hellsoft.net/descargas/juegos/rm-2000/completos/TheObserverF.7z) uses monster group battle commands to spawn all monsters in battle scene with some style.

Looks like battle interpreter is not supporting this yet, but battles are still playable with invisible monsters.

fdelapena:
Test case:
https://easy-rpg.org/play/?game=ae&battle-test=1

Ghabry:
In Observer it works. AE has a different interpreter problem. Opening a new issue to track this.

PR #482, opened 2015-05-17
Battle improvments

Ghabry:
Adds:
- Single Target skills for enemy
- States respect probability (A-E)
- States are respected
- Transparent msgbox for rpg2k3
- Hidden enemies
- Enemy self destruct
- enemy escape
- Enemy charge
- Enemy defend
- Only consume multi-target skills once

Unrelated: Resizing of switch/variable array when oob

Missing: Autoheal of states after time / after hit

#477,  #330 and #399

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -344,6 +344,10 @@ int Game_Actor::GetNextExp(int level) const {
 	}
 }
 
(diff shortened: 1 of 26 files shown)


BlisterBoy:
Hey really nice commit Ghabry gg ^^

PR #495, opened 2015-06-01
Fix "Change Tilemap"-command and emscripten tilemap

Ghabry:
Always regenerate the tile cache. Should fix "Change Tilemap"-command and emscripten black tilemap. Make tilemap "important" to make fade-in more beautiful (not black)


Ghabry:
That one is ready

PR #494, opened 2015-05-31
Make Bitmap::ToneBlit more closely resemble RPG_RT behavior

glynnc:

PR #496, opened 2015-06-01
Fix screen saturation

### LINKED ISSUE #357: Wrong tile and panorama colors when changing screen tint
Problem: Start the game in a map with a panorama and a process of
change palette of screen to 0 0 0 0,
In easyrpg the screen will look like if it made it a 8bit fade of colours.
I think that could be because the chrome its wrong.
3a

test case:
Start in the map TEST panorama color
http://dl.dropboxusercontent.com/u/22524283/temporal/Tests%20varios.rar

Ghabry:
Basicly same as #494 but also works when oversaturated

Fixes #357

*Note* Don't merge yet. We are discussing currently faster alternatives (at least for under-saturate)

### DIFF
--- a/src/bitmap.cpp
+++ b/src/bitmap.cpp
@@ -895,24 +895,49 @@ void Bitmap::ToneBlit(int x, int y, Bitmap const& src, Rect const& src_rect, con
 								 x, y,
 								 src_rect.width, src_rect.height);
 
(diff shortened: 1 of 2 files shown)


fdelapena:
Caching a zero saturation then blending with it looks the current candidate to go.

Ghabry:
Merging as asked in irc because this wont  be improved soon, opening a new issue for faster solution.

PR #498, opened 2015-06-07
Fix #493 : Screen doesn't shutdown in game anymore

### LINKED ISSUE #493: App returns to "GameBrowser" when screen turns off
I suggest to put the option "screen always on" in Android.
ty.

BlisterBoy:
Of course this works only in game, not in other activities.

(Android programmation is so easy, just one xml parameter oO).

### DIFF
--- a/builds/android/res/layout/main.xml
+++ b/builds/android/res/layout/main.xml
@@ -3,5 +3,6 @@
 	android:id="@+id/main_layout"
     android:layout_width="fill_parent"
     android:layout_height="fill_parent"
+    android:keepScreenOn="true"
     >
 </RelativeLayout>


fdelapena:
This looks effective, however when the user presses accidentally the power off button, the activity is reset yet. I guess the best solution should be more like this:

http://stackoverflow.com/a/14059121/1244033

And not requiring additional battery draining permissions. You can still go screen off and keep the game activity alive.

BlisterBoy:
Oh nice, I thought it was a different problem I'll have to deal with separately, I'll study your link.

BlisterBoy:
Hum I read the URL's discussion, the Power Button problem is a different problem. 
I'm not fan of the fix they propose (it's a unreality directive with seems to resolve the problem, by side-effect). But I'll try it, I'll push it if it works.

Ghabry:
Turning off the screen with power does not reset the activity for me.
Maybe this is different on Samsung devices because it uses a different Launcher (TouchWiz)

BlisterBoy:
So as I expected the fix suggested in the stack overflow url doesn't work (for me at least) (and required to upgrade to API 15).

In fact I highly doubt that we will be able to fix this problem because this is how Android is supposed to work : it destroy an application when it wants. Moreover the lib we use (the game player extends it) overload functions such as onDestroy and onResume, I tried to modify them but that didn't change the way it works and it could have random side-effects.

A solution would be to implement a quick save system in the core engine, we would do a quick save in onDestroy and a quick load in onResume (but it would be a big amount of work just to fix this issue which can be considered as the normal Android behavior).

As suggested Ghabry this behavior depends on the version of Android you run.

What I suggest : 
   - Merge this pull request (keep the screen ON is a huge improvement for user experience)
   - Don't close the issue, maybe someone more competent than me will have a solution.

Users will enjoy playing game without having to randomly push the screen during cut-scenes with the fear to loose their session haha.

Ghabry:
The way Android handles applications is sooo annoying.

BlisterBoy:
Hahaha yes it's true, in fact I'm glad there is a clean way to keep the screen on. A lot of emulators seems to forget this flag.

Issue #442, opened 2015-03-23
Menu button not working under Android possibly after setting target SDK 12


fdelapena:
Though both software menu (from the on screen button bar) and hardware menu button are available on the tested Android 3.2 device, neither of them are working This happens at least since recent 0.2.2 updates. Show FPS and other in-game options are not accessible. Targetting SDK 12 is needed to have USB/Bluetooth keyboard/joystick support from SDL2.

New menu option added by @BlisterBoy shows the "Change the default button mapping" from hardware button properly, however it is not possible to use the hardware button to add, modify or save options due to the same issue, making impossible to set them from the device.

I guess it is somewhat related with [this behaviour](http://techblog.rokoder.com/android-missing-menu-button/).

Maybe adding a hidden action bar for these activities could fix the button issue for Android < 4.x.

BlisterBoy:
Hum I'm trying to obtain the same issue but everything was alright on my tests, do you use a VM or a real device for your tests ?

Ghabry:
Maybe render a hamburger icon (≡) in the top-right corner that triggers that menu... Most newer apps have such an icon.

fdelapena:
The hamburger or triple vertical dot icon is shown already for the menu option added and it works, I can use the hardware button and opens it. I use a real Android 3.2 device.

Ghabry:
Soo is it now broken or working? :)

fdelapena:
I mean only works for the menu which shows the "Change the default button mapping menu item. It does not work for the in game menu (show fps and such) and for the menu in the button layout for the add, save and such options.

BlisterBoy:
OOoookay I was not sure but now I understand that you're facing a problem I had when I was developping the button mapping system on a test project working with SDK 20. The problem was solved when I downgraded the SDK.

The GameBrowser's menu works because it's a type of Activity displaying an action bar (the menu is automatically place in it).
When you use a simple Activity, Android seems to act differently depending on the API (surely because of the successive modifications of Android GUI Mantras). I'm not sure at 100% but placing a menu icon doesn't seems to be wanted by Google (no mention in http://developer.android.com/guide/topics/ui/menus.html and I never saw this on another app).

I'll try to solve the problem, but if I don't find a solution I can use a trick I saw on a lot of emulators : use the back button to display the menu
Pro:
:+1: Compatible with all API, no more problem for now and ever
:+1: We can't decently use an action bar on Player and Button Mapping System (ergonomic purpose)
[...]

PR #511, opened 2015-07-05
Back button call menu in ButtonMapping Activities

BlisterBoy:


BlisterBoy:
I remembered that in Player Activity, back button ask a confirmation to quit the game.

Should I delete this behavior to call the menu too ?
Pro : 
- Less complicated GUI
- Fix definitively #442

fdelapena:
Because some recent devices don't have hardware menu buttons (only home and back buttons), I think any hardware involved menu button should be removed. Adding an action bar for this and stealing space in the game screen is ugly so this feature could be added to the back button instead.

A good device example without this button: HTC One.

BlisterBoy:
@fdelapena there is no hardware button involved in the android port :p . (My phone doesn't have such a button and I have no problem).
When we associate a menu to an activity, it creates an hamburger icon in the dock (software button) to reach the menu .

But I think I'll associate the menu to back button to the Player Activity, because back button is already associated to a pop-up box (asking for stopping the game).

BlisterBoy:
Done !

Ghabry:
Thanks for the fix. Just got back from a short weekend trip. Will check tomorrow :)

fdelapena:
Tested here, back button works for adding/editing from the screen button layout editor with latest pull request generated .apk :+1:.

Ghabry:
Sorry that this took so long, was without internet the last days.
Works perfect, thx.

Ghabry:
One issue is still that the config function of the Game Browser is not obvious.
Similiar for long press to select region.

fdelapena:
These UX issues have been reported at #512 some days ago.

PR #530, opened 2015-08-04
Android: try adding a settings icon to menu bar

fdelapena:
Trying to work around an issue not allowing to access to change button mapping from the screen (the properties hardware button works yet):

missing

(Samsung Android 4.0.x device)

### DIFF
--- a/builds/android/res/menu/game_browser.xml
+++ b/builds/android/res/menu/game_browser.xml
@@ -1,14 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
 <menu xmlns:android="http://schemas.android.com/apk/res/android" >
-    <item android:id="@+id/game_browser_refresh"
-        android:title="@string/refresh"
-        android:icon="@drawable/ic_action_action_autorenew" 
-        android:showAsAction="ifRoom"
-        />
     <item android:id="@+id/game_browser_menu_change_mapping"
         android:title="@string/change_default_mapping"
         android:icon="@drawable/ic_action_action_settings"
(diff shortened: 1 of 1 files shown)


fdelapena:
After trying, it didn't work, I'm doing something wrong. This keeps the button mapping hidden in the hardware context menu :cry:.
@BlisterBoy any tip to improve the UX for this? There are still reports about this mostly hidden feature not being found for the most recent Google Play version.-

BlisterBoy:
Whaaat da faaack :octopus: 
The Refresh menu is shown but not the hamburger. That make no sense, I followed all the official instructions to have maximum supported devices :@ !!

BlisterBoy:
Oookayyyy
Soo apparently the problem comes from Samsung devices. I found this : 
http://stackoverflow.com/questions/9739498/android-action-bar-not-showing-overflow

I'll study that in the next days ^^.

BlisterBoy:
I'm not fan of this fix because it adapts on Samsung bullshit mantra ("yeaaah guyz letz go on 4.0 but letz keep the old mantra loool, gimme more hardwar butt and weed guyz")... But Samsung devices are the most important part of the android market x). Damn you Android !

yaminogame:
Hey ! It's a really good job !

Ghabry:
Samsung... I also had to workaround a bug that only affects Samsung that results in a red color when SDL2 runs... except if the volume control is open then colors are temporary correct oO

fdelapena:
Fixed via #533.

PR #532, opened 2015-08-09
.nomedia file + directory creation at startup

BlisterBoy:
.nomedia file added, works like a charm ! Thanks to the user who send the message :).

The app now also create all the necessary folders, this will be usefull for everyone :).


fdelapena:
Thanks! However, /easyrpg/games/ path was already being generated. Is there any difference for this?

PR #533, opened 2015-08-09
Try to force Samsung's device to show the action overflow button

BlisterBoy:
@fdelapena , does it resolve the problem ?


fdelapena:
It works like a charm! Thanks!

screenshot_2015_08_09_10_55_02

Ghabry:
Making a private field public ... seems legit. Thanks! My Galaxy Note is happy now.:)

PR #561, opened 2015-09-03
Fix #557

### LINKED ISSUE #557: Battle: back button when targetting allied when using skill target shows item window
Tested in 2k3 battle only, needs 2k testing. You need some skill targetting a single allied, e.g. some curation skill.
How to reproduce:
- Start a battle
- Select skill menu
- Select some cure to single allied
- The allied selectable will be shown. Press cancel/back button

Expected:
- Shows the skill window

What really happens:
- Shows the item window

scurest:
Fixes #557, by just setting the state to the last state, instead of always the item select one.


Ghabry:
Thanks. Don't worry about the compile errors.
They are because liblcf is currently out of sync with the Player code.

If you want you can fix this by commiting a search->replacing "pdef_f -> physical_rate" and "mdef_f -> magical_rate"

PR #567, opened 2015-09-05
Fix openOptionsMenu on Samsung devices

### LINKED ISSUE #536: Android: back button in button mapping activity not working in some devices
There is a comprehensive report by Inkuna in a [forum post](https://easy-rpg.org/forums/viewtopic.php?f=4&t=600) with Samsung :trollface: Galaxy Tab 4 and Android 5.0.2.

Seems like there is also a recent [comment](https://play.google.com/store/apps/details?id=org.easyrpg.player&hl=ru) in Google Play reporting the same issue. I could get device details from the Google Play developer console for this one if needed.

BlisterBoy:
Fixes #536 
It should work now :).

### DIFF
--- a/builds/android/src/org/easyrpg/player/button_mapping/ButtonMappingActivity.java
+++ b/builds/android/src/org/easyrpg/player/button_mapping/ButtonMappingActivity.java
@@ -9,6 +9,7 @@ import android.app.Activity;
 import android.app.AlertDialog;
 import android.content.Context;
 import android.content.DialogInterface;
+import android.content.res.Configuration;
 import android.os.Bundle;
 import android.util.Log;
 import android.view.KeyEvent;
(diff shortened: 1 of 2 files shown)


BlisterBoy:
Sorry for the 19 days, that was just a little fix to copy/past !

fdelapena:
Neat :) does it fix #536?

BlisterBoy:
#536 ! Yep x).

Ghabry:
I have no idea why your code works :D

BlisterBoy:
Hahaha this is the solution given in the @fdelapena 's threads. It's a similar problem than for the previous Samsung's troll : we have to override an already implemented method to be sure that Samsung didn't modify their rom to cook a shitty code. I have absolutely no idea why they broke the Android API, maybe to promote a specific way to use their product (to use hardware button like in previous mantra ?). 
I have no Samsung device but I'm pretty sure this will work. At least it don't bring bug in normal android device.

Oh, thanks @fdelapena for the URLs !

PR #569, opened 2015-09-06
Add a manual page

### LINKED ISSUE #298: Packaging: write a manual page (man 6 easyrpg-player)
As there are command line options now and also some used environment variables, it is good practice to have a manual page on Linux platform (and probably others).

The file would be easyrpg-player.6, compressed to easyrpg-player.6.gz (6 stands for section "Games")

As the manpage format is a bit "special" to tinker with, I think it would be a good idea to write the page in a simple markup like Markdown and convert it with [ronn](http://rtomayko.github.io/ronn/). This would also have the additional benefit, that a generated html page can also be rendered at easy-rpg.org.

carstene1ns:
Fixes #298.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ resources/player.aps
 /builds/android/android
 /builds/android/local.properties
 /builds/android/assets/
+/builds/easyrpg-player.6
 /builds/vs20*/Debug*
 /builds/vs20*/Release*
 /builds/vs20*/*.*sdf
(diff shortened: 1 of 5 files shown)


carstene1ns:
That really was something... with all the corner cases. Sometimes the build just errored out.
Hopefully this covers most cases. The cmake method is not thoroughly tested.

fdelapena:
Pretty well detailed thanks!

Issue #365, opened 2014-11-22
Battle animation vertical center offset is not working


fdelapena:
Test case: run [Wadanohara and the Great Blue Sea](http://vgperson.com/games/wadanohara.htm) game with `--start-map-id 4` (Sea Witch)

Battle animation ID 132 (SeaWitchBlink) has the Y coordinate parameter value set as (centered).

![Without animation](https://cloud.githubusercontent.com/assets/174885/5155125/184e5e8a-723f-11e4-85c4-75aff27e250d.png)

![With animation](https://cloud.githubusercontent.com/assets/174885/5155129/35bed756-723f-11e4-92ed-08e11ad36a24.png)

Ghabry:
Fixed by #574

Issue #349, opened 2014-11-12
SDL2: Use AudioCVT functions to support digital sound and music speed change.


fdelapena:
Since SDL2, AudioCVT functions might be useful to change audio speed without resampling the source, by providing a different source audio spec sampling rate than detected.

For the OpenAL backend could be implemented in its own way, if not implemented yet (not checked really).

This is required to implement audio speed properties for Music and Sound playback from 50% to 150% of the original speed.

MIDI works on a diffent way, using tempo change, not related with this issue.

SDL 1.2 does not work with non-power-of-2 frequency changes, so this can't work for the Wii port, as it currently lacks SDL2.

fdelapena:
Superseded by #895, SDL has long standing bugs which make unreliable to implement this with it.

PR #574, opened 2015-09-07
Battle animations

scurest:
Implements some missing features for battle animations. In particular

* both the screen and the target (on either the map or in battle) should flash correctly now  
    battler_flash
* battle animations are not fixed to the point on the screen where they first occurred, but follow their targets around  
    follow
* the position of animations targeting map characters is fixed (#365)  
    position
* "Show on entire map" animations work, I think

Shaking isn't implemented, but there's an obvious path to doing it once shaking is actually working :)

What I'm planning to do before I call this finished:

- [x] ~~`Game_Screen::ShowGlobalBattleAnimation` was completely removed. Need to check exactly what RPG_RT does in this case (should be used for targeting the whole enemy party, I think)~~ Done for RM2k3, need to check 2k now
- [x] Need to check I didn't break 809b381

(Also, @Ghabry, I backed out the `GetX`/`SetX`/etc. functions from `BattleAnimation` you just put in. Hope that's okay...)


Ghabry:
Cool, working battle animations. Seems we need a 0.3.2 hotfix soon.

Yeah, change whatever you want, only the result matters :)

BlisterBoy:
Cool commits guy !
The project is really increasing in quality, it's a shame that I don't have time to get back to C++ !

Ghabry:
Don't worry @BlisterBoy you can help us by improving the Android port by using Java :D

BlisterBoy:
The external storage thing broke my heart hahaha. I don't really know what to do except this one : [, I'm open to any proposition !

scurest:
So, I decided _not_ to fix "Show on entire map animations" on looping maps. The problem is that the animations are located at the vertices of a 20x10(11?) grid which is independent of the map. So after looping around, the animations play at different spots on the map. Illustration: (the white grid shows the map edges, and the red spots are where animations play)

entiremap

So you can't know where to put them just from `GetDisplayX`/`Y` and would have to complicate `Game_Map` with code to track the position against the grid of animations. OTOH, the graphical glitch from this tiny: just a slight jump in the animation grid position when the camera crosses a looping map edge. So I kind of decided it's not worth it.

scurest:
That said, this is ready.

Ghabry:
The same graphical glitch happens with Panoramas at the moment when crossing map boundaries. As you said: Not worth it.

Have no time until Thursday, somebody else can review. (fdela: when 0.3.2 release?)

fdelapena:
Thanks scurest, great fixes and improvements :+1:. Also 2k3 battlers flickering was annoying :smile:.

I have an exam tomorrow, same here until Thursday.

Ghabry:
Okay the flickering was then not observable for me because the VS-compiler probably inserted a 0. Good find.

fdelapena:
It seems like the animation sound playing when attacking is not OK anymore in OFF. The attack sound plays the same as the hit sound. Other things worked like a charm :+1:.

scurest:
My bad. Of course there can be multiple `AnimationTiming` for one frame (eg. flash both the target and screen, though OFF's playing two SEs in one frame doesn't make a lot of sense). Try this.

fdelapena:
Thanks! now it plays the splat audio combined. It seems to behave correctly now.
(I guess it sounds different than RPG_RT because of #349, but this one won't be soon probably).

scurest:
Urk, you're right. I changed it to only play the last SE if there are multiple in one frame... Sorry, one more try.

Ghabry:
I have an empty frame now in rpg2k3 battles when the battler is a battle animation. This wasn't happening before.
My fix to this was "return GetFrame()+1 >= GetFrames();", but you have this code already...

There is one corner case, but I assume almost no game uses this because it's nonsense:
When a battler has a battle animation as a pose flash effects have the actor as a target and screen effects are ignored.

fdelapena:
Thanks @scurest, however RPG_RT seems to play both sounds in the same frame, I guess 029710c was OK already. I think it just sounds different because the playback speed adjust is not implemented yet (#349).

Ghabry:
Concerning the one blank from of Battle Animation battlers.
A potential fix is changing back IsDone to the old behaviour:
```cpp
// battle_animation.cpp
// Battle animation line 99
bool BattleAnimation::IsDone() const {
	return GetFrame() >= GetFrames();
}
```

Completly avoid the function in sprite_battler and reset frame count one tick earlier to avoid the blank frame rendered:
```cpp
// sprite_battler.cpp line 107 
// old if (animation->IsDone())
if (animation->GetFrame()+1 >= animation->GetFrames()) {
```

scurest:
Thanks for being patient guys.

@Ghabry, how about reverting `IsDone` and then moving the `Update` call on [L119](https://github.com/EasyRPG/Player/blob/master/src/sprite_battler.cpp#L119) up before the `IsDone` check?

Ghabry:
@scurest  Yeah that's the best solution.

Ghabry:
Thanks for fixing all issues, excellent job.

Issue #575, opened 2015-09-07
Idea for the Android port


BlisterB:
Hi guys !

I open this issue to centralize suggestions about how improve the Android port, don't hesitate to give me job :) !

Idea : 
- [x] Multiple game directory
- [ ] Hi res splash screen while game browser parses the directories (I think this one is not possible has it depend from the C++ code)
- [x] Thumbnail support in GameBrowser (ask to find a quick way to store title screenshot)
- [ ] Key mapping for gamepads
- [x] Better configuration file for projects
- [x] Re do the settings activity
- [x] Different game displaying system depending on device screen
- [ ] A setting button in "select game layout"
- [ ] Ask for permission to add .nomedia in new game folder.

Ghabry:
- Support for multiply game directories (configurable in the settings)
  - The Game Browser itself should just merge all directories
  - Problem: How to deal with duplicates (same game in multiple directories)
  - Problem: On some Android devices (Android 4.4) the SD card is without rooting only writable by system apps (= only google)
     - Workaround: New command line option --save-path to overwrite the save directory (and save in the data directory of the apk instead). This should be transparent for the user, no interaction required.
- Button mapping for individual games
 - Reasonable defaults for RPG2k and RPG2k3 and some popular games, e.g. Yume Nikki preset
 - Moving buttons around while ingame (with running game rendered semi-transparent in the background)
 - Resizing of buttons
- A shiny, high-res "Easy RPG Player" splashscreen while the game browser parses the directories ;)

I don't think that file management (delete game e.g.) is required, but maybe "open game directory" and "open save directory", which launches ES File Explorer or whatever is installed.

BlisterBoy:
Thanks for the ideas :) ! I'll start to improve the button mapping system (the easiest xp). I think I'll switch to an xml format, that will be way more flexible for the futur.

I thought of this : 
- Use tabs to switch 2k/2k3/all game
   - https://www.google.com/design/spec/components/tabs.html#tabs-usage
   - Require to find how to know the Rm version of a game x))).

I also thought of displaying a thumbnail, that would be a big UX improvement BUT there are those .LDB parsing and folder's name thing (different depending on version/location). If I find which function are involved in the C++ code I maybe will be able to use them with JNI (after learning how to use it hahaha).

Ghabry:
[...]

PR #591, opened 2015-09-21
Android : Multiple InputLayout + Vibration support + UX improvements

BlisterBoy:
Hi guys, that was a big one. For the moment this is a functional update, UX improvements will become later. #575 

- Multiple Input Layouts
- Possibility to change layout's organisations
- Possibility to attach an Input Layout to a game


BlisterBoy:
Now the system indicate the default layout and you can directly edit a layout, you can now test it :).

Ghabry:
Wow, that's so awesome. And works really excellent. Good job.

The only thing I miss is that the "Choose layout" dialog for a game does not indicate the current layout (radio buttons like "Change Region" dialog).

BlisterBoy:
Thank you Ghabry ^^.
Yeah I agree, I was a bit lazy and was hopping no one would notice haha.

I'll do that, and a little surprise !

BlisterBoy:
EasyRPG now supports vibrations ^^.

Ghabry:
Hmmm, this needs a JNI callback to vibrate when damage occurs ;)

BlisterBoy:
The core have vibration support oO ?
If it's the case I'll maybe see how to add it to the android port (but I lack of time before my last school year begin :SS). For the moment the android port vibrate when we touch an on screen input. (We can enable/disable it in settings).

Talking about that. Android provide a Settings' system, it's the shitiest thing I've seen so far : no slider, we can't add normal view. It brings a lot of UX problem (useless clicks), I think I'll just make a custom activity.

BlisterBoy:
We can now modify the input layout transparency !

BlisterBoy:
The android port is now more beautiful, I'm not a designer (and supporting API 10 doesn't help) but I think it will be enough :p.

@Ghabry I'll use radio button to choose and input layout for a game, and I think it will be enough for this pull request ^^.

Ghabry:
Don't worry, it's really nice.

Some more feedback:
Don't place a space between the last word and the ":".

The default setting for vibrations is questionable. I think vibrations for button presses are annoying, maybe others like it... We will see what the reviews say :D. But I don't think that the Cross should vibrate at all (even when On)

The button mapping edit scene is now broken for me, I only see black, no buttons are added/displayed (but ingame it works)

BlisterBoy:
Haaaaa english typo rules ^^".

For the vibration I was thinking the same thing :p. I hate when vibration is too present but some user like it (they don't need a visual feedback, some others play without the input layout). Now they have the choice !

I also fixed the button mapping bug.

BlisterBoy:
Now it indicates the game's layout.
Pfiuuu if you don't find a problem, I'm done with this pull request !

Ghabry:
Reset button mapping clears the screen completly instead of reverting. In my opinion it should be the same as "Exit without Save" -> "Edit layout".

And "Vibrate when sliding to another direction" is not really obvious, how about "Vibrate when changing direction on the D-Pad"?

And a minor nitpick ;) (sorry): Can you move the config option "Manage Input layouts" to the top? Is the most important function in the settings.

BlisterBoy:
For "Manage Input Layout" the problem is that it can take a long place (and hide other option I found usefull), and having a scrollbar in a scrollbar is not good for android UX. And honestly that doesn't really shock me :p.

You found a bug for the "reset mapping button" button oO. I don't know why it does that, it's supposed to display the standard 2000 layout. I'll fix that.

Thanks for the UX problem with Change Region, I didn't noticed. Android provide gui service reaaaally confusing.
My intention was to limit useless clicks, with this approach I don't like the presence of "Ok" button (I prefer directly validate the user choice). But you can't do that with radio buttons without doing an ugly design pattern x).
But well it's Android, I'll make the region button dialog look like the change layout (or Samsung's troll will come back).

Ghabry:
The region change dialog was my first attempt to make a popup-dialog with Android, and I couldn't figure out the correct way to do this.
Not sure if auto closing is the best solution, afterwards I display a Toast to indicate what was really touched. So I think OK/Cancel is the better way. Is an extra click but more save that you get was you wanted (and not missed the option with your fat finger, especially on small displays).

BlisterBoy:
I agree :).

builds/android/src/org/easyrpg/player/button_mapping/VirtualButton.java
```diff
@@ -8,112 +9,200 @@
...
 	}
-	
+
+	public char getAppropriateChar(int keyCode) {
+		char charButton;
+
+		if (keyCode == ENTER) {
+			charButton = 'A';
+		} else if (keyCode == CANCEL) {
+			charButton = 'B';
```
fdelapena: About "A" and "B" for Enter and Cancel... what about Z and X for them?

BlisterBoy:
Z and X oô ? This is awfull xpppp

Ghabry:
I think a configuration for this is not worth it :D
A and B because I grown up with Nintendo :D

BlisterBoy:
Yep, moreover A and B are respectively touch for Validation et Cancelation on RPG Maker. Maybe I didn't understood some thing.

BlisterBoy:
I did the region dialog thing, it's over !

Btw sorry guy but you probably will have to update translation in spanish and german. Hablo un poco de español pero no lo suficiente :p.

Ghabry:
You can delete the toast "region_modification_success" now. Before it was useful to indicate what was really touched because the dialog was auto closing.
Now it's superfluous.

BlisterBoy:
Done

Ghabry:
I ask all translators to update there translations then slooowly :)
@BlisterBoy @fdelapena @myself

PR #594, opened 2015-09-22
Fix crash when empty chipset is used

Ghabry:
Problem was that autotile cache read out of bounds, a black tilemap is now generated to work around all issues.

My fix in 0.3.2 had some bugs :/

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -224,7 +224,7 @@ BitmapRef Cache::Tile(const std::string& filename, int tile_id) {
 			sub_tile_id = tile_id - 96;
 			rect.x += 384;
 			rect.y += 128;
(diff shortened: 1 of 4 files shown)


fdelapena:
It is more complete than Zegeri's a8440f8, thanks!

Ghabry:
Oh I didn't saw zegeris commit

Zegeri:
It shouldn't be a black tilemap, but a transparent empty tilemap, so it shows the background picture if there's any.

Left: EasyRPG. Right: RPG_RT.exe

Ghabry:
hmm... right. :/
And because I added the 3rd arg (Color() which is transparent) it doesn't even work by accident because this constructor disables the transparency setting.

PR #596, opened 2015-09-24
Android: update Spanish translation

fdelapena:


Ghabry:
Entry "region_modification_success" can be removed everywhere, not used anymore.

Let's wait before merging, strings for configuring the position in portrait mode come probably soon ^^

builds/android/res/values-es/strings.xml
```diff
@@ -1,6 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <string name="app_name">EasyRPG Player</string>
+	<!-- Preferences keys, do not translate -->
+	<string name="pref_enable_vibration">PREF_ENABLE_VIBRATION</string>
+	<string name="pref_layout_transparency">PREF_LAYOUT_TRANSPARENCY</string>
+	<string name="pref_vibrate_when_sliding_direction">PREF_VIBRATE_WHEN_SLIDING</string>
+
```
Ghabry: I think you can omit these "Do not translate" lines

Ghabry:
The string "ignore_size_settings" is missing.
I don't really like the English text
```Ignore buttons\' size settings and use this one for every buttons:```, in the German version I translated it to ```Use the following button size instead of the default size:```

File: builds/android/res/values-es/strings.xml (hunk 1 above)
Comment (carstene1ns) [inline builds/android/res/values-es/strings.xml]: I wonder if we could remove them from this file altogether. It would make the translation interface tidier.

Comment (carstene1ns) [inline builds/android/res/values-es/strings.xml]: If we cannot remove them, i will add the `translate` property soon. Transifex ignores such strings.
```
<string name="name" translate="false">Untranslated</string>
```

BlisterBoy:
The thing is that it will be possible in futur to set different size for each button (in button mapping activity).
Although I'm not sure anymore to implement that thing, the option in Settings is enough.

carstene1ns:
I think `app_name` should not be translated.

PR #609, opened 2015-09-30
Android : possibility to change the main folder

BlisterBoy:
It's now possible to change the application folder (in Settings). Works with no problem for me. I prefer to do this PR now to quickly know if the functionality poses problem on your devices (that was TOO simple, i smell something like android trolling in the air).

What I have to do : 
   - [x] Nice UX integration
   - [x] A button "Create a directory"
   - [x] An automatic refresh on returning on GameBrowser
   - [x] Discuss about the possibility to hide some folder at the root : I saw that I've some directory like "sdcard" (which is my internal memory, which is not the "internal memory" in android but anyway), "sdcard1", "usbdisk" etc. It's maybe a good thing to only show those folder, but I'm not sure that every Android device works the same way.


BlisterBoy:
I did the three first points, now let's talking ^^.

Ghabry:
Sorry! You can get one storage path using Environment.getExternalStorageDirectory() and for the µSD card you must manually parse the files /proc/mounts and /system/etc/vold.conf
The file chooser is not alphabetical sorted

I don't really like the concept of moving the easyrpg folder around. Because now when I select a new folder it creates the whole directory hierarchy again. I think this is not nice, because it creates folders everywhere and is an unexpected side effect.

Imo the only option really needed is to add multiple game folders (like for "Manage Input Layouts" but "Manage game folders" and "Add game folder").

The dir browser on it's own works well, nice.

Ghabry:
At my system the vold.conf contains nothing of interest, the relevant line for my extSdCard is:
```/storage/extSdCard /storage/extSdCard ecryptfs rw,nodev,relatime,ecryptfs_sig=3702882fe05cd852,ecryptfs_cipher=aes,ecryptfs_key_bytes=32,ecryptfs_passthrough 0 0```
What do other devs get here?

**Comment by BlisterBoy** [comment]
Hello Ghabry !
I'm not really sure to understand this part's implications: 
"Sorry! You can get one storage path using Environment.getExternalStorageDirectory() and for the µSD card you must manually parse the files /proc/mounts and /system/etc/vold.conf"
Does it means you don't like the DirChooser system and prefer of a list of choice between Environment.getExternalStorageDirectory, but not the system's root ?
If it's the case, the problem is that Environment.getExternalStorageDictory can either return the root of your "internal memory" (the root of user session) or the root of a SD card.
Doc quote : 
Note: don't be confused by the word "external" here. This directory can better be thought as media/shared storage. It is a filesystem that can hold a relatively large amount of data and that is shared across all applications (does not enforce permissions). Traditionally this is an SD card, but it may also be implemented as built-in storage in a device that is distinct from the protected internal storage and can be mounted as a filesystem on a computer

Please note that in my devices, SD card folder can be accessed at the root system, in /sdcard1. So for a lambda user, the user start to / and click to "sdcard" or "sdcard1".

"I don't really like the concept of moving the easyrpg folder around. Because now when I select a new folder it creates the whole directory hierarchy again. I think this is not nice, because it creates folders everywhere and is an unexpected side effect."
I agree that the "creating directory logic" of the system is a flaw but I don't see how to prevent it, knowing there's user for whose is apparently difficult to figure out that A button is Enter and B Cancel (it's not an attack on @fdelapena  :p, it's an attack on users xDDD).
Users don't understand the system logic and I thought that create the automatic folder's creation is easier for them (they select the dir in the app, plug their phone to the pc, discover the pre defined folders).

Have multiple game's folder would be excellent, but brings other problems I don't know how to solve : 
- what to do of RTPs ? 
   - Use multiple RTP folder -> I don't think I can in the Android port
   - Use just one folder -> which one ?
      - Possible solution : Have multiple game folder, select one RTP dir 
      - Remember there is 2000 and 2003 RTP. Choosing just a folder that contains both, or select separately for the two ? 
         -> Possible new problem : multiple game folders, other preferences for RTP (one or two, whatever)... How to do with dumb users ? Some of them can't even understand how to find the default folder according to some review :/, it's too difficult for them, even for confirmed user, it's just too heavy.

Considering this aspect, it's become to be a little tricky to find a definitive good solution. That's why I choosed the actual configuration : Apple's style, don't let user chooser some aspect to not loose them under a lot of preferences (which are, moreover, separately a lot sources of bugs).

My actual solution considering your post:
   - Multiple game's folder
   - One modifiable RTP folder (that contains 2000 and 2003 RTP), whith automatic creation of 2000 and 2003's folder in it, if they don't exist. To let indications for users.
   - A default folders' configuration by default, the actual.
   - Ignore complains about how the system is complicated (because when you want to create video game, you must know of to deals with folder AND because this system is not too heavy).

Don't hesitate to correct me if I'm wrong

**Comment by Ghabry** [comment]
I was talking about filtering some folders when the directory browser is currently in /. I think this should be abstracted away from the users and only list the available storage paths (internal flash storage which is called externalStorage by Android and SdCard storage) and after selecting a storage the real directory choosing begins. Most of them will only plug in into there PC and then they get 1 or 2 devices (depending on if the device has a SD slot) and are confused when seeing the content of "/".
But this is complecated to implement correctly so I would just stick on presenting "/" directly. Maybe this could be made less scary by starting the navigation from the default folder [externalStorage]/easyrpg/games instead of "/"

I think the RTP should just stay at [externalStoragePath]/easyrpg/rtp and shouldn't be movable. It's not that big, don't see a reason to move it around for saving space.
All games access the same RTP.

Your summary:
Multiple game folder: YES (but it shouldn't create a folder "games" automatically but choose the one the user really selected, when the selected folder contains a valid RPG Maker game the chooser is clever and adds the directory one level up instead)
One modifiable RTP folder: Nice to have but not really neeeded
A default folder config. YES, nothing to do here
Ignore complains: YES :P

**Comment by BlisterBoy** [comment]
Hi Ghabry !
(Aaah I was sure to have answered to this messages but then I remembered I had to go to a party before finishing it x) )

To be short I agree with you and I now know how to do this thing... But it will have to wait few months. I'm already under an heavy load of works for my last year and I spend too much time in front of pc :xxx. But I'll do that ^^.

I wish you bon courage for the project ! Make it perfect for my come back haha !

**Comment by BlisterBoy** [comment]
Time to terminate this thing ^^

File: builds/android/res/values/strings.xml [hunk 1]
```diff
@@ -6,6 +6,7 @@
 	<string name="pref_vibrate_when_sliding_direction" translate="false">PREF_VIBRATE_WHEN_SLIDING</string>
 	<string name="pref_ignore_size_settings" translate="false">PREF_IGNORE_SIZE_SETTINGS</string>
 	<string name="pref_size_every_buttons" translate="false">PREF_SIZE_EVERY_BUTTONS</string>
+    <string name="pref_directory" translate="false">PREF_DIRECTORY</string>
```
carstene1ns: Please use a `TAB` here instead of spaces to keep the file consistent.

builds/android/res/values/strings.xml
```diff
@@ -71,6 +72,11 @@ Special features:\n
 	<string name="vibrate_when_sliding_direction">Vibrate when sliding to another direction</string>
 	<string name="input_layout_transparency">Input Layout transparency:</string>
 	<string name="ignore_size_settings">Ignore buttons\' size settings and use this one for every buttons:</string>
+	<string name="dir_does_not_exist">The directory %s$1 doesn\'t exist</string>
+	<string name="no_write_access_on_dir">No write access on %1$s</string>
+	<string name="no_read_access_on_dir">No read access on %1$s</string>
+	<string name="easyrpg_folder">EasyRPG folder</string>
+
```
carstene1ns: This stray newline could be removed as well :grinning:...

builds/android/src/org/easyrpg/player/game_browser/GameBrowserActivity.java
```diff
@@ -122,6 +129,8 @@ public void displayGameList(){
 			adapter = new GameListAdapter(this, project_list);
 		}
 		
+		Log.i("Browser", error_list.size() + " jeux trouvés");
```
carstene1ns: This seems like a debug leftover, if it is needed please use english, thanks!

BlisterBoy:
Oh sorry for the space line ! I quickly merge via sublime text, I'm correcting that ^^.

I'm not sure to understand the comment about the new lines, the file chooser I'm coding use them :p.

carstene1ns:
This was a misunderstanding, I did not mean the added (new) lines, just the blank one.

builds/android/src/org/easyrpg/player/game_browser/GameBrowserActivity.java
BlisterBoy: Yep no problem ! This PR is kind of a WIP (to discuss about how should the system should be), the final version will be cleaned of logs ^^. Don't hesitate to interact with further pushes if you think the system should be different :).

BlisterBoy:
Game Browsers now support multiple games folders !
I implemented a flexible system based on our discussion : 
- There is just one directory for EasyRPG, containing the folders "games" and "rtp", none of these are modifiable
- User can add as many games folders as he wants

I think this system is a good balance between all the problematic we came through.

NB : There is some UX things to fix (find a "delete" image, fix font size)

BlisterBoy:
This should be good !
(There is a little problem with icon overlapping a long text (not new), this will be fixed in further UX fix.)

Ghabry:
That's really awesome blisterboy.
May I request three minor improvments?

1. Add a 2nd text line to the "no games found in /easyrpg/games". Something like "You can add additional game directories in the settings"
2. When the user selects a directory containing a game select automatically the folder above.
3. Don't allow selecting of folders that are not writable (because saving will not work -> useless)

BlisterBoy:
Thank you, no problem I did the 3), and I'll do the 1) and 2) :).

Ghabry:
This class could be useful to provide some "Quick jump" functionality in the directory browser:
https://gist.github.com/loop/5805820
Tries to auto detect the secondary (SD card) storage. Some devices also use the environment variable "SECONDARY_STORAGE" (split at ":", can contain multiple paths), could be also used as a data source.

BlisterBoy:
Thank you for the tip Ghabry. I'll study this link.
Sorry for the lack of reactivity doing the 2), it demands to restructure the code.
For the moment the lists in game browser only show game, or only show error messages (such as "no game found"). List view in Android doesn't like to mix different type of content so I have to remade one or two things.
I have a job interview to work on for Monday but I'll try to do it this week end :).

BlisterBoy:
PS : Some friends has the opportunities to work on the project for school. I'll ask them to migrate the project on Android Studios.
This will hugely facilitate the development of the Android port and graddle will permit to use android support library.
After this migration I should be able to merge the current game browser and the legacy one (making the code easier to maintain), and remake the interface (I'm planning to use thumbnail).

Ghabry:
Yeah I only meant to show this 2nd error message when "No games found" occured. In case games are found don't display it.
This should work without refactoring.

BlisterBoy:
Oh I thought you wanted a message for each empty game folders, I'll do it as soon as possible !

BlisterBoy:
Done !
Your classe is really interesting, I'll use it later to improve the directory chooser

Ghabry:
Can I merge this for now and you create a new PR for improved directory chooser?

BlisterBoy:
You can merge

PR #622, opened 2015-10-05
Implement new FPS display behaviour

### LINKED ISSUE #558: Hide FPS by default
FPS are shown by default in full screen and in the window title in window mode. When pressing F2 only alternates title and screen FPS displaying.

New behaviour proposal:

Windowed mode:
- Don't display FPS in title or in screen by default.
- F2 (show FPS): display FPS in title only. When switching to desktop full screen, don't display FPS in screen.

Desktop full screen mode:
- Don't display FPS in screen by default
- F2 (show FPS): display FPS in screen. When switching to windowed mode, display FPS in title only.

carstene1ns:
Closes #558.

### DIFF
--- a/builds/android/jni/src/org_easyrpg_player_player_EasyRpgPlayerActivity.cpp
+++ b/builds/android/jni/src/org_easyrpg_player_player_EasyRpgPlayerActivity.cpp
@@ -33,7 +33,7 @@ extern "C" {
 JNIEXPORT void JNICALL Java_org_easyrpg_player_player_EasyRpgPlayerActivity_toggleFps
   (JNIEnv *, jclass)
 {
-	Graphics::show_fps = !Graphics::show_fps;
+	Player::fps_flag = !Player::fps_flag;
 }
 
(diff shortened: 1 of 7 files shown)


carstene1ns:
Maybe it would be better to join both flags together.
However, the whole thing with global flags is bad, so it does not hurt to have one more. :grinning:

Ghabry:
Patches to reduce global-flags are welcome ;)

Does pressing F2 still work to toggle FPS even without passing the command line?
Is it still possible to toggle the fps in Android?
https://github.com/EasyRPG/Player/blob/master/builds/android/jni/src/org_easyrpg_player_player_EasyRpgPlayerActivity.cpp

carstene1ns:
Yes, F2 still toggles, but not the "mode" (title bar vs. overlay), but instead on/off.
The mode is depending on fullscreen state.

I see no reason why it would not work on android, as I changed the flag in every source file, but it is untested.

carstene1ns:
Now only one global variable remains. This should be ready now.

PR #639, opened 2015-10-20
Ignore "." and ".." instead of any dotfile (fixes #637)

### LINKED ISSUE #637: File finder ignores hidden files
I notice this in one of the endings in .flow, the player fails to find the ".flowbeet" music file. It's there, I can confirm it.

Maybe it just can't because it's a dotfile and as such it's considered hidden.

fdelapena:


Ghabry:
Uhm one line below is
```std::string const name = Utils::FromWideString(ent->d_name);```
so you can also use ```name == ".." || name == "."```

PR #643, opened 2015-10-26
Emscripten: init SDL mixer frequency based on Web Audio API information

### LINKED ISSUE #641: Emscripten: init SDL mixer frequency based on Web Audio API information
This should prevent some well known audio playback noise issues because the Web Audio API does not allow to change this setting.

This can be performed by [Calling JavaScript from C/C++](https://kripken.github.io/emscripten-site/docs/porting/connecting_cpp_and_javascript/Interacting-with-code.html#interacting-with-code-call-javascript-from-native) and [related API functions](https://kripken.github.io/emscripten-site/docs/api_reference/emscripten.h.html#calling-javascript-from-c-c).

fdelapena:
(fixes #641)

### DIFF
--- a/src/platform/sdl_audio.cpp
+++ b/src/platform/sdl_audio.cpp
@@ -20,6 +20,9 @@
 #include "filefinder.h"
 #include "output.h"
 
+#ifdef EMSCRIPTEN
+#  include <emscripten.h>
+#endif
 
 #ifdef _WIN32
 #  include "util_win.h"
@@ -39,6 +42,16 @@ SdlAudio::SdlAudio() :
 	}
 #ifdef GEKKO
 	int const frequency = 32000;
+#elif EMSCRIPTEN
+	int const frequency = EM_ASM_INT_V({
+		var context;
+		try {
+			context = new AudioContext();
+		} catch (e) {
+			context = new webkitAudioContext();
+		}
+		return context.sampleRate;
+	});
 #else
 	int const frequency = 44100;
 #endif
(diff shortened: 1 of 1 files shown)


Ghabry:
Didn't work for me yesterday.
Is better with Chrome or under Linux?

fdelapena:
It does not seem to improve anything here and still sounds choppy. Is the patch returning a different freq there when you change it on Windows?
It looks like the problem is SDL itself because the audio file needs to resample to the browser freq anyways. Wave sounds from menu actions seem to work well in most tests.

Ghabry:
The frequency reported by Web Audio always changes depending on the Windows settings (this was always the case even before this patch).

Midi plays now correctly, too (at least under 44.1khz), awesome!

Maybe this is a bit silly but we could also provide instructions how to change the frequency under the common operation systems... You can't really hear a difference between 44.1 and 48k, so nobody will probably about this small decrease (in other programs).

fdelapena:
Could you retry with 48 kHz? Maybe now sounds different (or it does not sound at all).

Ghabry:
It still opens at 44.1khz and fails. BUT combined with carstene1ns patch is WORKS!!!

```diff
--- a/src/platform/sdl_audio.cpp
+++ b/src/platform/sdl_audio.cpp
@@ -42,6 +42,16 @@ SdlAudio::SdlAudio() :
        }
 #ifdef GEKKO
        int const frequency = 32000;
+#elif EMSCRIPTEN
+       int const frequency = EM_ASM_INT_V({
+               var context;
+               try {
+                       context = new AudioContext();
+               } catch (e) {
+                       context = new webkitAudioContext();
+               }
+               return context.sampleRate;
+       });
 #else
        int const frequency = 44100;
 #endif
```

fdelapena:
Ready, though ogg files should be adjusted for the most common web audio freq setup to prevent buffer issues.

PR #647, opened 2015-11-07
Fix Issue 453 (parallel interpreter resume) and common event save writing

Ghabry:
" 	When the start condition (switch) of a parallel common event is not met the interpreter is now suspended instead of destroyed. Parallel common events continue where they got suspended the last time. "


Ghabry:
Jenkins: Test this please

Zegeri:
:+1:

How unfortunate that there are games that depend on this weird behaviour. It seems that XP is more sane and parallel common events are restarted instead of suspended.

PR #664, opened 2015-12-06
Game browser

### LINKED ISSUE #53: Implement project browser / loader
When player executable is not in a game folder, allow to browse and load different games from the same executable, like ScummVM does.

This is important for systems without easy to edit file browser, like embedded devices.

A GUI design and a cross file system capable file browser implementation will be needed.

Ghabry:
Fixes #53 , almost 3 years old

Don't merge yet, there are still some issues I want to resolve.

This also renamed ProjectTree to DirectoryTree (because it's general purpose) and changes how the dummy bitmaps are rendered. In case of missing system it uses now an embedded System.png (from our TestGame).

Not in this release:
Navigation, it always uses the startup/working directory for searching games (but better then before ;)).
Is really only suited for systems without a simple way to browse, this is no replacment for the awesome Android Ui (yet?).

Bugs I won't fix:
Won't work with emscripten (too difficult to implement, use a game browser in the browser directly)
Windows can't load directories that are not in the current codepage (Reader issue, hard to fix). Can be worked around by renaming the folder. Imo Desktop systems should get a nicer Browser with GTK/Qt.

screenshot_0
screenshot_1


Ghabry:
screenshot_0

Almost ready, just need to fix emscripten ;)

BlisterBoy:
Hahaha great Ghabry !
But i have to let you know that my skill in Android are wayyy better than before, you shall definitvely not pass on the Android UI when I'll come back in April ;)

Ghabry:
@BlisterBoy Great, looking forward to your new skills

Should be okay now.

I already updated the php file on the server, bad error handling bug...

BlisterBoy:
We will have to migrate the android port to Android Studio, graddle will permit to use some library such as the Support Library which will be usefull for the Android Gamebrowser (and will permit to clean code and have just one version of the GameBrowser class for all api).
I don't have the time to read what you've done but I assume you did some usefull function in C++ to detect wheter a directory is game folder or not etc. Some of them will maybe be usefull using JNI in the android port.

Issue #674, opened 2015-12-15
Battle: Pages with condition "actor/event has turn" execute each turn


Ghabry:
Happens e.g. in OFF battle ID 4

Issue #608, opened 2015-09-29
Monigote Fantasy: Battle: Map encounters with an empty monster group


fdelapena:
Found in Monigote Fantasy, sometimes an empty battle appears and there is nobody to fight with. Escape is disabled, then the game gets into a deadlock. Happened in the first world map playable area. I can't specify where is this happening exactly (because of #607).
2190988 (#604 fix) needs to be merged from to go to the world map.
**Note:** with --engine rpg2k does not seem to happen.

Not tested but I guess the affected encounter area might contain some broken reference to removed battle groups. Or maybe not, the log was showing an encounter vs battler 123 and it works properly with BattleTest 123.

Ghabry:
Hmmm, I'm sure that I tested 0 Actor and Enemy groups.
There must be something special about this one. Or I broke it :D

Ghabry:
Besides that the MP3 music doesn't work because it uses 48kHz

Ghabry:
The battle is not empty, the enemy just looks like a tree :P

fdelapena:
Yeah, there is a battler shown as a tree, maybe the battler was really there but the enemy menu looked empty. The hang was happening anyways, thanks for the fix ^^.

Issue #601, opened 2015-09-27
Embric of Wulfhammer's Castle: PC orders in battle not loaded correctly


Siltala:
In this game, in combat the PC orders (not sure if that's the correct term. Attack/defend/item and the like.) don't seem to be loaded at all. In their place you are presented with an empty list of options, and pressing enter instantly crashes the game.

Steps to reproduce: 
* Download game (portable version of the Deluxe edition) from http://aeresland.net/ 
* Download my savegame from http://www.heavylobster.com/files/Save03.lsd 
* Take one step north to enter battle
* Select "Strategize"

Screenshot of correct behavior using RPG_RT.exe: http://www.heavylobster.com/pics/embric_working.png
Screenshot of broken behavior using easyRPG: http://www.heavylobster.com/pics/embric_broken.png

fdelapena:
Crash stack trace:
```
Program received signal SIGSEGV, Segmentation fault.
0x0000000000537731 in Scene_Battle_Rpg2k3::CommandSelected (this=0x452cab0)
    at src/scene_battle_rpg2k3.cpp:722
722		active_actor->SetLastBattleAction(command.ID);
#0  0x0000000000537731 in Scene_Battle_Rpg2k3::CommandSelected (this=0x452cab0)
    at src/scene_battle_rpg2k3.cpp:722
        command = @0x1802e16a90: <error reading variable>
#1  0x00000000005374b9 in Scene_Battle_Rpg2k3::ProcessInput (this=0x452cab0)
    at src/scene_battle_rpg2k3.cpp:638
No locals.
#2  0x00000000005260f4 in Scene_Battle::Update (this=0x452cab0)
    at src/scene_battle.cpp:175
        timer1 = 0
        timer2 = 0
        events_finished = true
#3  0x0000000000534d89 in Scene_Battle_Rpg2k3::Update (this=0x452cab0)
    at src/scene_battle_rpg2k3.cpp:90
No locals.
#4  0x000000000043dd4f in Player::Update (update_scene=true)
    at src/player.cpp:253
        framerate_interval = 16,666666666666668
        cur_time = 23002
#5  0x0000000000455856 in Scene::MainFunction (this=0x452cab0)
    at src/scene.cpp:96
        init = true
        __PRETTY_FUNCTION__ = "virtual void Scene::MainFunction()"
#6  0x000000000043da51 in Player::MainLoop () at src/player.cpp:181
No locals.
[...]

PR #675, opened 2015-12-15
Fix #419, #601, #608, #673, #674 (and more)

### LINKED ISSUE #419: Crash while displaying monsters' names in RM2k battle
If you enter a battle in an RM2k game and press F12 to return to the title screen while the monsters' names are being displayed (eg. "Slime encountered"), and then start or load a game, the *next* time you enter a battle, the Player will crash when it tries to display the monsters' names.

Probably caused by the static vars in `DisplayMonstersInMessageWindow` not being reset, so the enemy iterator is bad. I think making them member variables would fix it.

### LINKED ISSUE #673: Windows: Japanese and Korean locales can't find ./RPG_RT.ldb
It seems to be an encoding issue when trying to find files with path separators in the string, crashing Player when Windows is configured with Japanese or Korean locales, which use a different directory separator (Yen an Won).

This happens always, and can't be solved by forcing the `--encoding` parameter.

To test this on Windows, follow this guide to change the locale:
http://windows.microsoft.com/en-us/windows/change-system-locale

You can check the easyrpg_log.txt also to verify the bad encoding output, even when opened as UTF-8.

Ghabry:
Also Implement some RPG2k3 v1.11 features

Use correct value for heal values

### DIFF
--- a/builds/android/jni/src/Android.mk
+++ b/builds/android/jni/src/Android.mk
@@ -29,6 +29,6 @@ LOCAL_SHARED_LIBRARIES := SDL2 SDL2_mixer iconv #smpeg2 #freetype2-static pixman
 LOCAL_LDLIBS := -L$(EASYDEV_ANDROID)/libs/armeabi-v7a -lGLESv1_CM -llog -lz -lfreetype2-static -lpixman -lpng -licui18n -licuuc -licudata -lcpufeatures
 
 LOCAL_CFLAGS := -O2 -Wall -Wextra -fno-rtti -DUSE_SDL -DHAVE_SDL_MIXER -DLCF_SUPPORT_ICU
-LOCAL_CPPFLAGS	=	$(LOCAL_C_FLAGS) -fno-exceptions -std=c++0x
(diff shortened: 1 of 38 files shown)


carstene1ns:
Used in the android toolchain is boost 1.58.0, seems like boost 1.59.0 fixed some `std:` declarations.

Ghabry:
Open for review...

fdelapena:
It fails when loading savegames in emscripten, it shows:

    Debug: Loading save Save01.lsd
    Error: Couldn't find  save file

captura de pantalla de 2015-12-17 08-01-21

Note the 2 spaces between find and save, the filename is empty in the error message.
The save appears in the load window, so the file can be read from IndexedDB.

Could this issue have a fix in this PR or it is better on a separate PR?
It is failing already in master, so merging this PR won't affect with this.

Ghabry:
This should work, will check

Ghabry:
Is fixed, forgot to remove old EMSCRIPTEN related load code :D

PR #697, opened 2015-12-24
Android: Update to new SDK

Ghabry:
*Ready for merge*

What's new:
-New toolchain compiles almost all libs via configure && make && make install.
Exception is SDL because of the JNI-interaction
-liblcf is an extra lib now
-Support for x86


Ghabry:
jenkins: test this please

Ghabry:
Needs more CPU then emscripten, but runs great :)

intel

PR #706, opened 2015-12-30
autoconf: check for specific liblcf version

fdelapena:
This prevents building Player with outdated or incorrect liblcf versions, as [reported in forums](https://easy-rpg.org/forums/viewtopic.php?p=1335).


fdelapena:
As @carstene1ns pointed in chat, we should better bumping the version after tagging releases, as we break liblcf API later, to not match with official release versions.

PR #727, opened 2016-01-15
Android: Fix write detection

### LINKED ISSUE #726: Android: Detecting if writing on SdCard is possible does not work
I'm so pissed now of Android that I have to write a report just for your amusement.

Got some report on GooglePlay, have not verified if this is the cause but very likely is:

Last PR added redirecting of save and config when the directory is not writable. This was to work around the Android >=4.4 limitation that only system apps can write the SD card.

I use the Java-Api File.canWrite() to detect if writing works. The Api returns true, but the actual writing fails. I bet this is because .canWrite only uses stat which checks the linux access rights but Android does some other magic to prevent writing anyway.
Next time I will try to write a file instead and catch the exception.

The case for canWrite == false works, I tested that one...

Ghabry:
Fix #726 

Also adds logic to copy saves from game dir to save dir if save dir has no saves yet. (otherwise people will complain that saves got lost after copying games to external storage)

Also adds bug Report function to ingame menu

### DIFF
--- a/builds/android/res/values/strings.xml
+++ b/builds/android/res/values/strings.xml
@@ -11,9 +11,6 @@
 
 	<!--GameBrowser Activity-->
 	<string name="app_name" translate="false">EasyRPG Player</string>
-	<string name="toggle_fps">Toggle FPS</string>
-	<string name="toggle_ui">Toggle Virtual Buttons</string>
-	<string name="end_game">End Game</string>
 	<string name="autodetect">Autodetect Region (Recommended)</string>
 	<string name="west_europe">Western European</string>
 	<string name="east_europe">Central/Eastern European</string>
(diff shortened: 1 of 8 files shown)


carstene1ns:
Is there a way to check if the email could be sent? Would make sense to display a message about manually reporting a bug if the user aborts the mail or something did not work.

Ghabry:
No. The only thing you can detect is if the user pressed back in the "select mail program" dialog.
The mail sending is not detectable.

builds/android/res/values/strings.xml
```diff
@@ -85,5 +95,8 @@ Once there, tap the back button to access to an additional menu to add more butt
 	<string name="add_a_button">Add a button</string>
 	<string name="reset_button_mapping">Reset button mapping</string>
 	<string name="exit_without_saving">Exit without saving</string>
-	<string name="save_and_quit">Save and quit</string>	
+	<string name="save_and_quit">Save and quit</string>
+	<string name="key_enter">Enter (Z)</string>
+	<string name="key_cancel">Cancel (X)</string>
```
fdelapena: Please consider mentioning `Esc` too.

carstene1ns:
Well, then maybe an explanatory message before is better.
> "This will try to open you mail program to send us the bug report. If you do not have an active internet connection or do not want to send a mail, press cancel now. You can also open a new issue at https://github.com/EasyRPG/Player anytime."
> OK / CANCEL

fdelapena:
Maybe is a bit obvious the fact about sending an email requires Internet connection.

Ghabry:
@carstene1ns done

Ghabry:
Jenkins: Test this please

PR #733, opened 2016-01-22
Reminder: merge in translations from transifex 

carstene1ns:
Just a friendly reminder what happens when you directly change the language source files and do not care about external translations…

Seems like the "How to Use" message completely disappeared from some languages.

### DIFF
--- a/builds/android/res/values-de/strings.xml
+++ b/builds/android/res/values-de/strings.xml
@@ -2,9 +2,6 @@
 <resources>
   <!--Preferences Keys-->
   <!--GameBrowser Activity-->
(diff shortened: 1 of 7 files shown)


Ghabry:
Thanks for showcasing this.
The old messages were out of date though, so the text wasn't 100% fitting anymore.

Gimme a sec to update the German translation on transifex...

Ghabry:
Done. Could you also add the russian and the ukrainian one?

carstene1ns:
I have manually added back the old instruction text for now. Also this string is marked on transifex, to let translators know they should check it.

About the Cyrillic languages, I think it is better to wait until @rohkea fills the new strings in.

PR #737, opened 2016-01-24
Specify C++11 for CMake

scurest:
Building with CMake is broken, because C++11 is used, but `CXX_STANDARD` isn't specified.

This does bump the minimum required version of CMake though.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,7 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.1)
+
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED on)
 
 project(EasyRPG_Player CXX C)
 


carstene1ns:
_cough_ #320
Maybe we can get it through this time... :grinning:

scurest:
Oh, actually I do remember that!  :)

But that was before. for example, b0b7152. Requring C++11 to build the codebase seems to constitute a de facto C++11 policy.

carstene1ns:
yes, i am okay with it, but maybe we can go without requiring a newer cmake version and gcc>=4.7?
No idea what is used now on debian, but this approach still works with 2.8: https://github.com/EasyRPG/Player/pull/320#issuecomment-50289313

carstene1ns:
well, whatever, they can use autotools if they want, lol

fdelapena:
Yeah, C++11 code started landing, so the old policy is over. Even with pretty recent cmake requirement looks good to me, distributions with old package version still may use autotools when needed IMO, as points Carsten.

PR #760, opened 2016-02-09
Misc

carstene1ns:
Tidied the log, skipped the registry on Wii and Android and fixed two oversights related to the gamebrowser.


Ghabry:
Great :+1: . Only needs emscripten codefix

carstene1ns:
Yeah, seems like compilers still do not evaluate early returns without looking at the whole function :/
I will provide some fix-up soon™

carstene1ns:
Should be ready to merge.

PR #772, opened 2016-02-21
Get rid of platform subfolder

fdelapena:
There are already platform specific files in the src folder, e.g. input_buttons_*.
The purpose of this pull request is normalize these cases and use proper preprocessor guards instead of using dummy files just for including them.

This PR has nothing to do with a proper subfolder redesign, if considered in the future to make source code navigation cleaner could be OK, this should not be considered as a step back.


src/sdl_ui.cpp
```diff
@@ -156,9 +163,9 @@ SdlUi::SdlUi(long width, long height, const std::string& title, bool fs_flag) :
 	ShowCursor(false);
 #endif
 
-#if defined(HAVE_SDL_MIXER)
+#ifdef HAVE_SDL_MIXER
 	audio_.reset(new SdlAudio());
-#elif defined(HAVE_OPENAL)
+#elif HAVE_OPENAL
```
carstene1ns: These checks are not the same. You check if `HAVE_SDL_MIXER` is defined at all (may be set `0` => disabled or even some random string `"foobar"`) and then check if `HAVE_OPENAL` is defined _and_ set to true (i.e. `1`).
I do not consider the way it was done before really better, but it should be consistent.

src/sdl_ui.cpp
```diff
@@ -16,35 +16,42 @@
...
-	#define NOMINMAX
-	#endif
-	#include <windows.h>
-	#include "SDL_syswm.h"
+#  define WIN32_LEAN_AND_MEAN
+#  ifndef NOMINMAX
+#    define NOMINMAX
+#  endif
+#  include <windows.h>
+#  include "SDL_syswm.h"
```
carstene1ns: We really need a tab vs. spaces policy somehow.

src/audio_al.cpp
```diff
@@ -605,3 +607,5 @@ void ALAudio::SE_Stop() {
 	}
 	se_src_.clear();
 }
+
+#endif
```
carstene1ns: Also a policy to include or not include comments about closing preprocessor conditionals ` // HAVE_OPENAL`

src/sdl_ui.cpp
```diff
@@ -1211,4 +1218,4 @@ void GekkoResetCallback() {
 }
 #endif
 
-#endif
+#endif // USE_SDL
```
carstene1ns: (added here for example)

carstene1ns:
The good thing about this pull request is that the build systems like autotools now detect if things in the formerly hidden (i.e. not timestamped) `platform/*` files changed :+1:.

src/sdl_ui.cpp
fdelapena: Some years ago we changed these to the way I did because of http://stackoverflow.com/a/789156/1244033

PR #773, opened 2016-02-22
Don't free from Mix_LoadMUS_RW() and play ADPCM with BGS_Play

### LINKED ISSUE #710: Support MS ADPCM playback for files in Music/ game folder
SDL can play MS ADPCM WAV files, however SDL_mixer is not supporting them for music playback, however they can be played in a channel.

Because current file format detection is using `SDL_Mixer` to play music and files must be loaded first to detect the music type with `SDL_LoadMUS_RW`, mixer will return `SDL_error` about `unknown WAVE data format`. SDL_LoadWAV does not have audio type format

There is also a known double free crash bug in `SDL_mixer` when passing `1` value to automatically free the data source when calling `SDL_LoadMUS_RW` to detect the file format, which is the reason of the **Yume 2kki** or the game [星の海で](http://www.geocities.jp/viprpg_gw2007/entry/025.html) crash as these games use MS ADPCM WAV files for music.

A simple workaround could be reading the first four bytes of the file to check WAV format before doing the bogus `SDL_LoadMUS_RW` detection then directly do the `BGS_Play` call when those first four bytes are "RIFF".

fdelapena:
Fix #710. Also, remove remaining stray `\n` symbols in warning messages.

### DIFF
--- a/src/audio_sdl.cpp
+++ b/src/audio_sdl.cpp
@@ -158,13 +158,17 @@ void SdlAudio::BGM_Play(std::string const& file, int volume, int /* pitch */, in
 	bgm.reset(Mix_LoadMUS_RW(rw), &Mix_FreeMusic);
 #endif
 
+#if SDL_MIXER_MAJOR_VERSION>1
+	// SDL2_mixer bug, see above
+	if (bgs_playing) {
+		BGS_Stop();
+	}
+#endif
+
 	if (!bgm) {
 #if SDL_MIXER_MAJOR_VERSION>1
 		if (strcmp(Mix_GetError(), "Unknown WAVE data format") == 0) {
-			// SDL2_mixer does not support ADPCM playback but SDL2 does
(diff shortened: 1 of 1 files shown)


Ghabry:
I'm just spotting some unrelated bugs that could be quickly fixed:

Line 176
```
	// SDL2_mixer bug, see above
	if (bgs_playing) {
		BGS_Stop();
	}
```
Should be before the ```if (!bgm) {``` check at 161. (Because music should always stop when changed, also in error case).

And ```BGM_Volume``` should call ```BGS_Volume``` but this function does not exist :(

Ghabry:
At least the first one could be fixed. The 2nd is just nit-picking ;)

fdelapena:
Jenkins: retest this please

fdelapena:
This one looks ready now.
@Ghabry your slave seems to be down right now.

Issue #756, opened 2016-02-08
F12 / end game should stop music and channels playback


fdelapena:
This is easy to reproduce when a game does not use music in the title scene, it will keep existing music and channels playing. In RPG_RT it will "play" silence.
This shouldn't be handled from the title scene because 2k3 1.11 can skip it, this should be handled in the end game.
Test with a 2k3 1.11 game skipping title scene and returning to a map without music: https://easy-rpg.org/play/?game=itwr

carstene1ns:
This only applies to games which skip the title scene.
If the title scene is enabled, it will stop the music if it is empty.

Issue #731, opened 2016-01-19
Skip `*.script.wav` files from Ineluki's Key Patch


carstene1ns:
Currently, the log is spammed with
```
[2016-01-19 13:36:50] Warning: Couldn't load MidiTickOutput.script SE.
Unrecognized sound file type
[2016-01-19 13:36:50] Warning: Couldn't load KeyList.script SE.
Unrecognized sound file type
[2016-01-19 13:36:50] Warning: Couldn't load KeySupportOn.script SE.
Unrecognized sound file type
[2016-01-19 13:36:51] Warning: Couldn't load MidiTickOutput.script SE.
Unrecognized sound file type
[2016-01-19 13:36:51] Warning: Couldn't load KeyList2.script SE.
Unrecognized sound file type
[2016-01-19 13:36:51] Warning: Couldn't load KeySupportOn.script SE.
Unrecognized sound file type
[2016-01-19 13:36:51] Warning: Couldn't load MidiTickClear.script SE.
```
It is because the sound effects are used as triggers for scripts. While we cannot easily support the whole key patch (maybe we could write some - likely game specific - key wrapper logic after #491 is implemented), we should at least give a better message and only display it once per game launch.

This could be done like the infamous '(Brak)' check, "if SE ends with .script", then set a flag and display:
"Warning: Game uses Ineluki's Key patch, additional keys and scripts will not work!"

Besides the additional keys, this patch adds mouse support (was this even used by a popular game?)
and a buggy mp3 support (through `*.link.wav` files like disharmony? - not worth it)

---

An example game using this patch is "Harvest Moon - Mondlichts Kinder", issue reported by mail.

Issue #624, opened 2015-10-06
Use a better EasyRPG logo at startup


carstene1ns:
The EasyRPG logo at startup is good and should be kept, as it clearly brands the application.

However, there are some minor flaws:
- The colors of the text logo are too dark
- The URL does not contain the `s` for secure access and has an additional, unneeded slash `/` at the end

Here is the current logo (as everybody knows it :grinning:):
![oldlogo](http://picload.org/image/paarolw/logo.png)
_I actually had to write an unxxd tool to extract it from source as-is._

And here is a quick mockup to address both issues:
![mockup](http://picload.org/image/paarocr/new_logo.png)
Do not mind the realignment, it was for testing purposes. Also the font may be a bit too big in comparison.

About the URL: Maybe drop the `http://` or `https://` altogether?
**web: easy-rpg.org**
About the two logos: Maybe drop the text logo and just use a fullscreen logo?

Here is another mockup:
![mockup2](http://picload.org/image/paarcda/new_logo2-fs8.png)

BlisterBoy:
Your first mockup is really nice ;).

vinipsmaker:
> About the URL: Maybe drop the `http://` or `https://` altogether?

If we're going to be pedantic, shouldn't we maintain the `/` at the end? I vote to remote the protocol prefix altogether.

The mockup I like the most is:

[![](https://camo.githubusercontent.com/833e8d13676d24b92b21d6ab92542309a2fb433b/687474703a2f2f7069636c6f61642e6f72672f696d6167652f706161726f63722f6e65775f6c6f676f2e706e67)](https://camo.githubusercontent.com/833e8d13676d24b92b21d6ab92542309a2fb433b/687474703a2f2f7069636c6f61642e6f72672f696d6167652f706161726f63722f6e65775f6c6f676f2e706e67)

Ghabry:
Btw, I'm interested in 640x480 versions of these logos.

Ghabry:
As an alternative: Use the original logo and remove the URL (it is displayed in the about tab of the game browser anyway.
And use the gained space to make the icon better :)

fdelapena:
A text prefix like "Powered by [logo]" might be useful.

BlisterBoy:
I agree with Ghabry : displaying the URL is kind of useless for a lot a reasons

fdelapena:
Remote picture links in OP post are broken now.

Here's a splash logo proposal for 0.6.0:

[...]

Issue #383, opened 2014-12-19
Player doesn't generate some window properties when started in full screen mode


fdelapena:
This makes some display issues at least on Linux and Windows 8.1 when switching to window mode.
Related: #179

carstene1ns:
This one got mostly fixed by #792, the rest is `WontFix` (→ #179 - SDL2 specific bug) for now.
Maybe when we have another UI besides SDL someday this will not be an issue anymore.
IMO we could close this and reopen #179 instead.

carstene1ns:
This one should be fixed by #1252.

PR #792, opened 2016-03-02
Fix #700, #731, #756. Prepare for #87, #383, #624.

### LINKED ISSUE #700: End game goes to title with --new-game and does not show the selectable cursor
It does not refresh title scene system and text, maybe didn't should go to the title scene at all.
Not tested how it should behave with rm2k3 1.11.

carstene1ns:


carstene1ns:
This is ready for review.

Ghabry:
End Game restarts the game now. That's on purpose? (makes sense, just making sure)

carstene1ns:
Yes, is on purpose. I was not sure how to handle it correctly. It depends on what `--new-game` should be used for. See #700.
IMO it is better to skip it every time, like RPG_RT does with the 2k3 feature enabled.
Btw. this makes it also consistent with the (Better)AEP patches or others which can skip the title scene.
Second possibility is to only use it on first launch and resetting the flag, so the title screen is shown later.

PR #625, opened 2015-10-06
Better system integration

carstene1ns:
This pull request currently is only used to get some player executable to test the changes under windows/WINE.
Do *not* merge yet.

PR #793, opened 2016-03-03
Add a window icon on Linux and OS X

carstene1ns:
...and try to work around issues on Windows with it.
(This is a refactor of #625)


Ghabry:
Nice will try it under Windows

Ghabry:
Awesome, works :+1:

fdelapena:
Works, flawlessly :+1:, however I'd like to get a high res icon when doing alt+tab instead of a 32x32 one (zopfli'd 192x192 or 256x256 maybe?).

By the way, to solve the application name in GNOME Shell and other few freedesktop compliant desktops depending on WM_CLASS, before initializing SDL video on X11, this should do the trick (setenv requires `<cstdlib>`):

`setenv("SDL_VIDEO_X11_WMCLASS", "EasyRPG Player", 0);`

carstene1ns:
Adding a bigger icon would be possible, however it might then look bad when used in the title bar at a small size. There will be always scaling involved if we only use one for all cases.
It would be possible to use icons with different sizes like on windows, however this is not supported by SDL (and will bloat the executable size of course). [Here](https://gist.github.com/carstene1ns/973bcb80907f6557ac7d) is a small demo using X11 directly, you can check with `xprop`.

Ghabry:
I wonder why the WM does not handle this automatically, like Windows does.

PR #810, opened 2016-03-08
fix spell drain #809

### LINKED ISSUE #809: Failed SP draining Spells seem to be successfull
In "Vampires Dawn" spell "Blutsaugen" should be only successfull, if enemy has blood(SP).
Otherwise it should fail. Currently it allways shows "Absorbs x Blood"

Tondorian:
now drain sp spells can fail if there is no more sp on target
also it shows currentTargets  SP if its lower then spells SP-Effect


Ghabry:
Jenkins: test this please

Ghabry:
Jenkins: test this please

PR #814, opened 2016-03-09
Fix the windows build.

carstene1ns:
Now my question is, why did MSVC not blow up on the PR builder for this one?
It should really not have cached the `filefinder.cpp`, as the file changed in cc2d2e67453caa7d3db32c8b85635089b801285b.

### DIFF
--- a/src/filefinder.cpp
+++ b/src/filefinder.cpp
@@ -27,15 +27,14 @@
 
 #include <boost/optional.hpp>
 
-#ifdef __MINGW32__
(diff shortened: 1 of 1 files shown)


Ghabry:
The problem is that dirent_win.h does not include windows.h but some other winapi headers. And some stuff for x64 is not defined when you do this. But enough for x86 :(

carstene1ns:
Ah, and the pull request builder is 32 bit only... makes sense^^

PR #815, opened 2016-03-12
Conditions in Battles

Tondorian:
Conditions are applied to battle_actors now.
GainMoneyMessage only shown if money >0

I only tested it with rpg Maker 2k but should work on 2k3...

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -312,21 +312,6 @@ void Game_BattleAlgorithm::AlgorithmBase::SetTarget(Game_Battler* target) {
 }
 
 void Game_BattleAlgorithm::AlgorithmBase::Apply() {
-	std::vector<int16_t> inflictedStates = source->GetInflictedStates();
-	if (inflictedStates.size() != 0) {
-		for (int i = 0; i<inflictedStates.size(); ++i) {
-			RPG::State state = Data::states[inflictedStates[i]];
-			int hp = state.hp_change_val;
-			int sp = state.sp_change_val;
-			int source_hp = source->GetHp();
-		    int source_sp = source->GetSp();
-			int src_hp = std::min(source_hp + 1, IsPositive() ? hp : -hp);
-			int src_sp = std::min(source_sp, IsPositive() ? sp : -sp);
-			source->ChangeHp(src_hp);
-			source->ChangeSp(src_sp);		
-		}
-	}
-
 	if (GetAffectedHp() != -1) {
 		int hp = GetAffectedHp();
 		int target_hp = (*current_target)->GetHp();
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -482,6 +482,10 @@ bool Game_BattleAlgorithm::Normal::Execute() {
 		}
 
 		int effect = (source->GetAtk() / 2 - (*current_target)->GetDef() / 4);
+
+		if (effect < 0)
+			effect = 0;
+
 		int act_perc = (rand() % 40) - 20;
 		// Change rounded up
 		int change = (int)(std::ceil(effect * act_perc / 100.0));
@@ -614,8 +618,14 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 
 			// TODO: Phys/Magic attribute: Phys.Attribute /100 x Magic.Attribute /100
 			// see #480
+			if(effect < 0) {
+				effect = 0;
+			}
 
 			effect += rand() % (((effect * skill.variance / 10) + 1) - (effect * skill.variance / 20));
+
+			if (effect < 0)
+				effect = 0;
 
 			if (skill.affect_hp) {
 				this->hp = effect / ((*current_target)->IsDefending() ? 2 : 1);
@@ -977,6 +987,7 @@ bool Game_BattleAlgorithm::SelfDestruct::Execute() {
 	// Like a normal attack, but with double damage and always hitting
 	// Never crits, ignores charge
 	int effect = source->GetAtk() - (*current_target)->GetDef() / 2;
+
 	if (effect < 0)
 		effect = 0;
 
@@ -984,6 +995,10 @@ bool Game_BattleAlgorithm::SelfDestruct::Execute() {
 	int act_perc = (rand() % 40) - 20;
 	int change = (int)(std::ceil(effect * act_perc / 100.0));
 	effect += change;
+
+	if (effect < 0)
+		effect = 0;
+	
 	this->hp = effect / ((*current_target)->IsDefending() ? 2 : 1);;
 
 	if ((*current_target)->GetHp() - this->hp <= 0) {
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -312,6 +312,21 @@ void Game_BattleAlgorithm::AlgorithmBase::SetTarget(Game_Battler* target) {
 }
 
 void Game_BattleAlgorithm::AlgorithmBase::Apply() {
+	std::vector<int16_t> inflictedStates = source->GetInflictedStates();
+	if (inflictedStates.size() != 0) {
+		for (int i = 0; i<inflictedStates.size(); ++i) {
+			RPG::State state = Data::states[inflictedStates[i]];
+			int hp = state.hp_change_val;
+			int sp = state.sp_change_val;
+			int source_hp = source->GetHp();
+		    int source_sp = source->GetSp();
+			int src_hp = std::min(source_hp + 1, IsPositive() ? hp : -hp);
+			int src_sp = std::min(source_sp, IsPositive() ? sp : -sp);
+			source->ChangeHp(src_hp);
+			source->ChangeSp(src_sp);		
+		}
+	}
+
 	if (GetAffectedHp() != -1) {
 		int hp = GetAffectedHp();
 		int target_hp = (*current_target)->GetHp();
@@ -644,7 +659,7 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 		for (int i = 0; i < (int) skill.state_effects.size(); i++) {
 			if (!skill.state_effects[i])
 				continue;
-			if (!healing || rand() % 100 >= skill.hit)
+			if (!healing && rand() % 100 >= skill.hit)
 				continue;
 
 			this->success = true;
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -457,7 +457,6 @@ bool Game_BattleAlgorithm::Normal::Execute() {
 	Reset();
 
 	int to_hit;
-
 	if (source->GetType() == Game_Battler::Type_Ally) {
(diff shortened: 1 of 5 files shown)


Tondorian:
Jenkins: test this please

fdelapena:
Jenkins: test this please

src/game_battlealgorithm.cpp
```diff
@@ -312,6 +312,21 @@ void Game_BattleAlgorithm::AlgorithmBase::SetTarget(Game_Battler* target) {
...
 
 void Game_BattleAlgorithm::AlgorithmBase::Apply() {
+	std::vector<int16_t> inflictedStates = source->GetInflictedStates();
+	if (inflictedStates.size() != 0) {
+		for (int i = 0; i<inflictedStates.size(); ++i) {
+			RPG::State state = Data::states[inflictedStates[i]];
+			int hp = state.hp_change_val;
+			int sp = state.sp_change_val;
+			int source_hp = source->GetHp();
+		    int source_sp = source->GetSp();
```
carstene1ns: Stray spaces detected.

Tondorian: Should be fixed

Ghabry:
Jenkins: Test this please

Ghabry:
One reason speaking against ::Apply is that under RPG2k3 all actors get damage at the begin of a battle action (doesn't matter if ally or enemy). With Apply only the current one doing the action gets damage.

I suggest moving this to a function in Game_Battler and when invoking that function it directly alters SP and MP of that Battler.

For RPG2k you can place it in ... maybe BattleActionState_ConditionHeal (for the FirstAttack case)

So for RPG2k3 I suggest placing the call to this function (and displaying the damage number (for HP only!) as a floating number) in the BattleActionState_Start for the "IsFirstAttack" case.

The RPG2k3 battle system also shows that damaging due to states happens before the action is executed.

Tondorian:
git commit --amend failed again :-1:

Ghabry:
git reset HEAD~1   :D should work

fdelapena:
Mr. Jenkins, test this please.

fdelapena:
Jenkins: Test this please

fdelapena:
Jenkins: retest this please

Ghabry:
Jenkins: Test this please

src/scene_battle_rpg2k3.cpp
```diff
@@ -532,6 +532,15 @@ bool Scene_Battle_Rpg2k3::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBas
 
 		if (action->IsFirstAttack() && action->GetStartSe()) {
 			Game_System::SePlay(*action->GetStartSe());
```
Ghabry: The damageTaken codepath depends on ```action->GetStartSe()```

src/scene_battle_rpg2k3.cpp
```diff
@@ -532,6 +532,15 @@ bool Scene_Battle_Rpg2k3::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBas
...
+			int damageTaken = action->GetSource()->ApplyConditions();
+			if (damageTaken != 0) {
+				DrawFloatText(
+					action->GetTarget()->GetBattleX(),
+					action->GetTarget()->GetBattleY(),
+					0,
+					boost::lexical_cast<std::string>(damageTaken),
+					30);
+			}
 		}
```
Ghabry: Everytime anybody does his turn all allies and enemies get state heal/damage. So should be something like this:
```
std::vector<Game_Battler*> battlers;
Main_Data::game_party->GetBattlers(battlers);
Main_Data::game_enemyparty->GetBattlers(battlers);

for (auto& b : battlers) {
  // Your damageTaken code
}
```

src/game_battler.cpp
```diff
@@ -327,6 +327,40 @@ static bool non_permanent(int state_id) {
 	return Data::states[state_id - 1].type == RPG::State::Persistence_ends;
 }
 
+int Game_Battler::ApplyConditions()
+{
+	int damageTaken = 0;
+	std::vector<int16_t> inflictedStates = this->GetInflictedStates();
+	if (inflictedStates.size() != 0) {
+		for (int i = 0; i<inflictedStates.size(); ++i) {
+			RPG::State state = Data::states[inflictedStates[i]];
```
Ghabry: The loop can be easier written as:
```
for (int16_t inflicted : inflictedStates) {
    RPG::State state = Data::states[inflicted - 1];
```
and the ```if (inflictedStates.size() != 0) {``` check is not needed

src/game_battler.cpp
```diff
@@ -327,6 +327,40 @@ static bool non_permanent(int state_id) {
...
+			int hp = state.hp_change_val;
+			int sp = state.sp_change_val;
```
Ghabry: This calculation ignores the "Percentage of max HP/SP". In other sections of code where this is required we do something like this:
```int hp = state.hp_change_val + (int)(std::ceil(GetMaxHp() * state.hp_change_max / 100.0));```

File: src/game_battler.cpp [hunk 6]
```diff
@@ -327,6 +327,40 @@ static bool non_permanent(int state_id) {
...
+			}
+			else {
+				src_hp = std::min(source_hp, hp);
+			}
+			if (state.sp_change_type == state.ChangeType_lose) {
+				 source_sp = -std::min(source_sp, sp);
+			}
+			else {
+				source_sp = std::min(source_sp, sp);
+			}
```
Ghabry: RPG2k3 added ```RPG::State::ChangeType_gain```, which is what you handle in your ELSE case. But ```RPG::State::ChangeType_none``` (No change at all) is not handled.

src/game_battler.cpp
```diff
@@ -327,6 +327,40 @@ static bool non_permanent(int state_id) {
...
+			this->ChangeHp(src_hp);
+			this->ChangeSp(src_sp);
+			damageTaken += src_hp;
+		}
+	}
+	return damageTaken;
```
Ghabry: The return value must be negated (return a negative value for damage currently)

Ghabry:
Test case for RPG Maker 2003 (add a RPG_RT.exe):
BattleTest with Monster Party 2.

[rpg2k3 testcase.zip](https://github.com/EasyRPG/Player/files/170635/rpg2k3.testcase.zip)

Command Line for Player: "BattleTest 2"

Ghabry:
Test case for RPG Maker 2000
BattleTest with Monster Party 1.

[rpg2k battletest.zip](https://github.com/EasyRPG/Player/files/170637/rpg2k.battletest.zip)

src/scene_battle_rpg2k.cpp
```diff
@@ -450,12 +461,12 @@ bool Scene_Battle_Rpg2k::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBase
...
 
 			if (action->TargetNext()) {
-				battle_action_state = BattleActionState_Start;
+				battle_action_state = BattleActionState_ConditionHeal;
 				return false;
 			}
 
 			// Reset variables
-			battle_action_state = BattleActionState_Start;
+			battle_action_state = BattleActionState_ConditionHeal;
```
Ghabry: You changed all the ```battle_action_state``` transitions. This is **super super risky**, pls don't do this unless you are 100% certain this is required. Because this must be tested with every available battle action available (Normal, different items, different skills, defend) to proof that it doesn't break anything.

One problem I already noticed is, that you introduced some delay before everybodys turn.

Tondorian: Yeah i needed to change them to get poisen effect working before the battlers attac

Ghabry: I reverted all the state changes and I only took your code from line 370 to 388 and it still works fine for me.

Tondorian: so it shows conditioneffct, Clean enemyAttacks, Damage clean to you? 
Takes me 2h yesterday to get it shown correct

Ghabry: In my tests your state changes broke the message box. At the beginning it shows an empty message box with a delay and between battle actions, too.

And after an enemy finished his battle animation the delay got removed.

src/scene_battle_rpg2k3.cpp
Tondorian: Oh there is also an bug in my line should be action->GetSource()->GetBattleX()
but your solution is even better

src/game_battler.cpp
Tondorian: in ChangeType_none case i only need to set src_s/hp = 0?

Ghabry: Yeah, 0 change

Ghabry:
The delay is still wrong here.

Normal attacks without any states have a 30 frame delay before anything happens -> Must be 0.

And after battle animations the feedback is now instant instead of a delay. (After showing the attack animation, e.g. the sword slash, the old version had a 30 frame delay).

Tondorian:
I think it should work as expected now
if you can confirm, i'll clean up my branch

Ghabry:
I will prepare an emscripten test case for both battle systems this evening.
Then it's easier to verify this (also for future pull requests).

Ghabry:
Jenkins: Test this please

Test cases:

https://easy-rpg.org/play/index-pr.html?game=battle_test&battle-test=1&engine=rpg2k

https://easy-rpg.org/play/index-pr.html?game=battle_test&battle-test=1&engine=rpg2k3

VS. Master:

Test cases:

https://easy-rpg.org/play/index-master.html?game=battle_test&battle-test=1&engine=rpg2k

https://easy-rpg.org/play/index-pr.html?game=battle_test&battle-test=1&engine=rpg2k3

Ghabry:
@Tondorian  The PR builder says std::ceil needs a <cmath> include

carstene1ns:
Needs `#include <cmath>`, http://www.cplusplus.com/reference/cmath/ceil/

Ghabry:
Jenkins: Test this please

@fdelapena  Maybe we should whitelist him ;)

Tondorian:
> And sometimes enemy attacks give me now a negative amount and heal me :O
See this on your Emscripten Testcase first time 
I think i did not change anything on this part game_battlealgoithm.cpp

fdelapena:
> I think i did not change anything on this part game_battlealgoithm.cpp

Yeah, Ghabry commented on IRC this also happens with the current upstream master version.

Tondorian:
but i messed up dead enemys in rpg2k3 :D if they are dead they get 1hp ^^ 
will fix this part now

Ghabry:
Okay, looks fine now. The negative HP after attacks and poison are strange but this already happens in Master... :(.
@Tondorian  Is the PR ready?

Ghabry:
The problem for that "-5" is the skill damage code in ```Game_BattleAlgorithm::Skill::Execute()```

Before the "TODO comment" we need a "effect < 0 -> = 0" check.

And the effect with the rand() modulo can be only be executed, when effect is > 0 because this will Div by 0.

Ghabry:
Hm, no. When effect is 0 it will % with 1. Just add a < 0 then = 0 check.

Issue #812, opened 2016-03-09
Android: "horizontal" bug after returning into game from Google Launcher


Tondorian:
If you "minimize" easyRPG in horizontal-orientation and go back into application. App shows only 5 pixels of gamescreen on upper part of screen.

I think application calculates displaysettings wrong. Perhabs it trys to render vertical mode.

Ghabry:
This is too generic. Every android device has different bugs (hello, Samsung!)

Name of device and Android version please.

BlisterBoy:
I know this bug, you just have to flip the screen to recalculate the screen displaying.
This should be easy to fix, forcing the app the recalculate the screen displaying when it comes back to "foreground" (onResume callback ?)
Ghabry you can assign me on this task, I finish my exam in 2 weeks and I fix it :).

Tondorian:
Device: Samsung S4 
Android Version: 6.0.1

Ghabry:
Samsung ~.~

@Tondorian 
Do you know someone with a Samsung S6 (Edge)? We need to debug a black screen issue on start up :(

Tondorian:
No but will ask around,
But my parents got some sony devices 
s5 maybe ^^

Ghabry:
Nope, is a S6 exclusive issue :(

BlisterBoy:
@Tondorian : here is a little reminder of our friendly encounters with Samsung, it's actually funny to read:
https://github.com/EasyRPG/Player/issues/776
https://github.com/EasyRPG/Player/pull/567
https://github.com/EasyRPG/Player/issues/536
https://github.com/EasyRPG/Player/pull/533
https://github.com/EasyRPG/Player/pull/530
https://github.com/EasyRPG/Player/pull/392
https://github.com/EasyRPG/Player/issues/386

Tondorian:
Using CyanogenMod... propably my device is not best for testing

BlisterBoy:
Nope don't worry, your ROM is better than a Samsung rom. 
Your issue should be noticeable on every devices, I didn't already fix it because if I start to code for another thing that college stuff, I leave those college stuff xp.

Ghabry:
Fixed by #817, wasn't auto closed >.<

PR #817, opened 2016-03-12
Fix for the restarted screen display

FaithFeather:
So, that's the fix for #812.
I've add the function OnRestart() which is called each time the user return on the app after exiting the application (without destroying it). So the display is now in full and not only five pixels on the top.

### DIFF
--- a/builds/android/src/org/easyrpg/player/player/EasyRpgPlayerActivity.java
+++ b/builds/android/src/org/easyrpg/player/player/EasyRpgPlayerActivity.java
@@ -89,7 +89,7 @@ public class EasyRpgPlayerActivity extends SDLActivity {
 		surface = mSurface;
 		mLayout = (RelativeLayout) findViewById(R.id.main_layout);
 		mLayout.addView(surface);
-		udpateScreenPosition();
+		updateScreenPosition();
 
 		// Open the proper input_layout
 		bmm = ButtonMappingModel.getButtonMapping(this);
@@ -339,7 +339,7 @@ public class EasyRpgPlayerActivity extends SDLActivity {
 		}
 	}
 
-	public void udpateScreenPosition() {
+	public void updateScreenPosition() {
 		RelativeLayout.LayoutParams params = new RelativeLayout.LayoutParams(LayoutParams.WRAP_CONTENT,
 				LayoutParams.WRAP_CONTENT);
 		params.leftMargin = 0;
@@ -357,7 +357,16 @@ public class EasyRpgPlayerActivity extends SDLActivity {
 	public void onConfigurationChanged(Configuration newConfig) {
 		super.onConfigurationChanged(newConfig);
 
-		udpateScreenPosition();
+		updateScreenPosition();
 		updateButtonsPosition();
 	}
+	
+	/** Called after the activity is being re-displayed */
+	@Override
+	public void onRestart(){
+		super.onRestart() ;
+		updateScreenPosition();
+		updateButtonsPosition();
+	}	
+	
 }


carstene1ns:
jenkins: test this please

carstene1ns:
IMO this event is wrongly named, because the app is not actually restarted, but otherwise LGTM.

BlisterBoy:
Actually this seems to be the right function, I agree some Android's function names are not intuitive : 
http://developer.android.com/reference/android/app/Activity.html#onRestart()

Ghabry:
@FaithFeather 
Could you rename udpateScreenPosition to updateScreenPosition ? :D

BlisterBoy:
Hahaha sorry guys, I sometimes swap 2 consonants, I had to c/p the two functions names to figure out which was the difference x).
FaitFeather, you just have to do a commit, and push it. No special command to do :).

FaithFeather:
That should do the trick. :)

Ghabry:
Now the commit appears multiple times. :(
Please rebase (git rebase -i HEAD~2) and read in the internet about "squash"

BlisterBoy:
It seems that those are 2 different commit, but with the same name.

FaithFeather:
Uh, that's two different commits, but i forgot to change the name.

Ghabry:
Yeah, just squash them to one. The 3rd commit only because you forgot something in the 2nd :P

FaithFeather:
Ok, so... i did the rebase thing. That's ok now ?

fdelapena:
Looks good to me, thanks :), however you could also do a `git rebase upstream/master` then `git push -f` to keep the history clean.

Ghabry:
Your rebase did not work. Look at the history, you have now 3 times the same commit :(

So imo this is the perfect time for him to learn about Git :P

BlisterBoy:
So many lessons about git, and yet this is the tool we're not able to master x).

Ghabry:
Because the commits are the newest ones, instead of rebasing he can also just reset the commits via "git reset HEAD~N" where N is the amount of commits to remove, in that case "4".

FaithFeather:
I think that Git and I are in a complicated relationship. 
But... NOW. The thing is (normally) clean, so i'm gonna make all the changes and do the LAST commit.

Ghabry:
Now you created 3 revert commits and only made it worse. :(

fdelapena:
@FaithFeather don't panic, I cried a lot my first months with Git. Of course still can be squashed into 1 commit with few actions.

FaithFeather:
urf. okay.
So, (to be sure and not doing something bad/stupid again). Git reset HEAD~8 (now) and git push -f ?

Ghabry:
Yepp you do a ```git log```, count how many commits you want to remove (8 looks fine).

Then ```git reset HEAD~8```. Now you check ```git log``` if this is what you wanted ^^. Instead of "HEAD~8" you can also provide the SHA1 of the has btw.

Because you removed all but one commit you must commit your modified files again now (```git commit```).

And then force push :)

Ghabry:
You should consider using a GUI btw ;)

carstene1ns:
yay :+1:

FaithFeather:
okay, after a little mess with my mouse with the close/open...
i think it's good now ?

Ghabry:
Jenkins: Test this please

carstene1ns:
@FaithFeather: Yes, looks good now. And do not get upset with all the git-fu, it can be frustrating to learn and has a steep learning curve, but it will save you some time to know the advanced commands in case you need it later.

BlisterBoy:
Yay congratulationn miss (y)

Ghabry recommending to use a GUI :O, you tried the Github one ?

PR #823, opened 2016-03-15
Don't trust the SaveGameActor::ID anymore because of strange savegames

Ghabry:
Weather Forecast: Bit cloudy with the sun shining!


fdelapena:
jenkins: retest this please <3

Issue #821, opened 2016-03-15
Battle System: Remaining, not implemented stuff


Ghabry:
This is a meta bug to track the remaining battle stuff that is not implemented.
When fixing bugs please reference #821 in your pull request :)
- [x] Battle Action: Defend command: Not working at all in 2003, not sure about 2000. (Animation is wrong in 2003 but this only visual)
- [x] Actor: Dual Wield: "Equip Weapon with both hands". When doing a normal attack the attack animations of both weapons will play and the damage is the sum of both. Also needs testing if the Equipment Scene handles this **since 0.6**
- [x] Actor: Mighty Guard: "The effect of the [Defend] command will be doubled, further reducing received damage."
- [x] Actor: Element Guard: Multiplier for damage receiving
- [x] Skills: Ignore defense: "the specified stat will be reduced without applying the opponent's defence."
- [x] Skills: States. The 2nd non-default option (Heal instead of Inflict) is not handled. **since 0.6.1**
- [x] Skills: Attack Element is not respected
- [x] Skills: Reduce Resistance "If enabled, the resistance to the checked types in (14) of the target characters will decrease or increase by 1 stage." (so from C to D, or C to B)
- [x] Skills: MP cost based on percentage
- [x] Items: Two handed (needs changes to Equipment scene?) **since 0.6**
- [x] Items: MP cost (What is this? You can use Items as skills, maybe they mean this)
- [x] Items: Critical rate: We only handle the actor critical rate
- [x] Items: Effect: Preemptive Advantage during battle "Ignores agility and attacks at the beginning of the turn." **Partly, not for 2k** **full since 0.6**
- [x] Items: Effect: Attack twice at the same time " Each normal attack hits the enemy twice." **since 0.6**
- [x] Items: Effect: Attack all enemies at the same time "Each normal attack hits the entire troop."
- [x] Items: Effect: Ignore enemy dodge rate "Attack accuracy ignores the enemy's agility." **since 0.6**
- [x] Items: Effect: Player can't unequip. The equipment cannot be removed. (Needs a change to the Equipment scene)
- [x] Items: Attack element
- [x] Items: State handling: needs tests
- [x] Items (Weapons): The 2nd option (heal instead of inflict) for states **since 0.6**
- [x] Items (Other): The 2nd option (inflict instead of resist) for states **since 0.6.1**
- [x] Items: State Infliction % (is that handled???)
[...]

PR #826, opened 2016-03-17
BattleFixes 

### LINKED ISSUE #820: Attribute multipliers of conditions not implemented
Title. In an RPG Maker game I'm making, I have skills that should inflict conditions. They don't work; if the skill only inflicts the condition, it always misses, if it does damage as well, it only does damage.

Attributes don't work either. If I want, for example, a Slime to be weak to Fire (A) and immune to Ice (E), using a skill/attack with that attribute won't cause a damage boost/deduction.

Tondorian:
#821 
fixes #820 
Items: Consuming SP
Items: crit
Items: Conditions
Items: Ignore Dogerate
Items: Attributes increase/decrease Damage
Defending/Charged: On Dead stoped
Interpreter: ConditionChange Dead reduces HP to 0 now
Spells: Attributes increase/decrease Damage
Conditions: Silence and Hit fixed

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -118,12 +118,6 @@ bool Game_Actor::IsSkillUsable(int skill_id) const {
 	if (skill_id <= 0 || skill_id > (int)Data::skills.size()) {
 		return false;
 	}
-	for (int inflictedState : this->GetInflictedStates()) {
-		if ((Data::states[inflictedState - 1].restrict_magic && Data::skills[skill_id - 1].magical_rate > Data::states[inflictedState - 1].restrict_magic_level) || ((Data::states[inflictedState - 1].restrict_skill && Data::skills[skill_id-1].hit > Data::states[inflictedState - 1].restrict_skill_level))) {
-			return false;
-		}	
-	}
-	
 
 	const RPG::Skill& skill = Data::skills[skill_id - 1];
 
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -119,7 +119,7 @@ bool Game_Actor::IsSkillUsable(int skill_id) const {
 		return false;
 	}
 	for (int inflictedState : this->GetInflictedStates()) {
-		if ((Data::states[inflictedState - 1].restrict_magic && Data::skills[skill_id].magical_rate > Data::states[inflictedState - 1].restrict_magic_level) || ((Data::states[inflictedState - 1].restrict_skill && Data::skills[skill_id-1].hit > Data::states[inflictedState - 1].restrict_skill_level))) {
+		if ((Data::states[inflictedState - 1].restrict_magic && Data::skills[skill_id - 1].magical_rate > Data::states[inflictedState - 1].restrict_magic_level) || ((Data::states[inflictedState - 1].restrict_skill && Data::skills[skill_id-1].hit > Data::states[inflictedState - 1].restrict_skill_level))) {
 			return false;
 		}	
 	}
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -114,6 +114,12 @@ bool Game_Actor::IsSkillUsable(int skill_id) const {
 	if (skill_id <= 0 || skill_id > (int)Data::skills.size()) {
 		return false;
 	}
+	for (int inflictedState : this->GetInflictedStates()) {
+		if ((Data::states[inflictedState - 1].restrict_magic && Data::skills[skill_id].magical_rate > Data::states[inflictedState - 1].restrict_magic_level) || ((Data::states[inflictedState - 1].restrict_skill && Data::skills[skill_id-1].hit > Data::states[inflictedState - 1].restrict_skill_level))) {
+			return false;
+		}	
+	}
+	
 
 	const RPG::Skill& skill = Data::skills[skill_id - 1];
 
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -899,6 +899,8 @@ void Game_Actor::ChangeHp(int hp) {
 	if (GetData().current_hp == 0) {
 		// Death
 		RemoveAllStates();
+		SetDefending(false);
+		SetCharged(false);
 		AddState(1);
 	} else {
 		// Back to life
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -94,6 +94,10 @@ bool Game_Actor::UseItem(int item_id) {
 	return Game_Battler::UseItem(item_id);
 }
 
+std::vector<uint8_t> Game_Actor::GetAttributeRanks() const {
+	return Data::actors[GetId()-1].attribute_ranks;
+}
+
 bool Game_Actor::IsItemUsable(int item_id) const {
 	const RPG::Item& item = Data::items[item_id - 1];
 
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -612,10 +612,11 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 
 			int effect = skill.power +
 				source->GetAtk() * skill.physical_rate / 20 +
-				source->GetSpi() * skill.magical_rate / 40 -
-				(*current_target)->GetDef() * skill.physical_rate / 40 -
-				(*current_target)->GetSpi() * skill.magical_rate / 80;
-
+				source->GetSpi() * skill.magical_rate / 40;
+			if (!skill.ignore_defense) {
+				effect -= (*current_target)->GetDef() * skill.physical_rate / 40 -
+					(*current_target)->GetSpi() * skill.magical_rate / 80;
+			}
 			// TODO: Phys/Magic attribute: Phys.Attribute /100 x Magic.Attribute /100
 			// see #480
 			if(effect < 0) {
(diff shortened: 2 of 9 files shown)


src/game_actor.cpp
```diff
@@ -114,6 +114,12 @@ bool Game_Actor::IsSkillUsable(int skill_id) const {
 	if (skill_id <= 0 || skill_id > (int)Data::skills.size()) {
 		return false;
 	}
+	for (auto inflectedState : this->GetInflictedStates()) {
+		if ((Data::states[inflectedState - 1].restrict_magic && Data::skills[skill_id].magical_rate > Data::states[inflectedState - 1].restrict_magic_level) || ((Data::states[inflectedState - 1].restrict_skill && Data::skills[skill_id].hit > Data::states[inflectedState - 1].restrict_skill_level))) {
+			return false;
+		}	
+	}
+
```
Ghabry: inflicted, not inflected :)
Because the data type is simple to write you can also use "int" instead of "auto".

skill_id is off-by-one, needs a -1

src/game_battlealgorithm.cpp
```diff
@@ -321,6 +321,12 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 			int src_hp = std::min(target_hp, IsPositive() ? -hp : hp);
 			source->ChangeHp(src_hp);
 		}
+		if(source->GetType() == Game_Battler::Type_Ally) {
+			int weaponID = (static_cast<Game_Actor*>(source))->GetWeaponId() - 1;
+			if (weaponID != -1) {
+				source->SetSp(source->GetSp() - Data::items[weaponID].sp_cost);
+			}
+		}
```
Ghabry: You can use ChangeSp here

src/game_enemy.cpp
```diff
@@ -68,6 +68,10 @@ int Game_Enemy::GetId() const {
 	return enemy_id;
 }
 
+std::vector<uint8_t> Game_Enemy::GetAttributeRanks() const {
+	return Data::enemies[GetId()-1].attribute_ranks;
+}
+
```
Ghabry: I see an off-by-one error in line 53. (needs a -1)

src/game_battlealgorithm.cpp
```diff
@@ -362,7 +362,7 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 		if(source->GetType() == Game_Battler::Type_Ally) {
 			int weaponID = (static_cast<Game_Actor*>(source))->GetWeaponId() - 1;
 			if (weaponID != -1) {
-				source->SetSp(source->GetSp() - Data::items[weaponID].sp_cost);
+				source->ChangeSp(-Data::items[weaponID].sp_cost);
```
Ghabry: This will consume SP even if the attack was not with the sword

src/game_battlealgorithm.cpp
```diff
@@ -518,6 +518,7 @@ bool Game_BattleAlgorithm::Normal::Execute() {
 			hit_chance = weapon.hit;
 			crit_chance = crit_chance += weapon.critical_hit;
 			multiplier = GetAttributeMultiplier(weapon.attribute_set);
+			source->ChangeSp(source->GetSp() - Data::items[weaponID].sp_cost);
```
Ghabry: You make a permanent change in Execute()

src/game_actor.cpp
```diff
@@ -114,6 +118,12 @@ bool Game_Actor::IsSkillUsable(int skill_id) const {
 	if (skill_id <= 0 || skill_id > (int)Data::skills.size()) {
 		return false;
 	}
+	for (int inflictedState : this->GetInflictedStates()) {
+		if ((Data::states[inflictedState - 1].restrict_magic && Data::skills[skill_id].magical_rate > Data::states[inflictedState - 1].restrict_magic_level) || ((Data::states[inflictedState - 1].restrict_skill && Data::skills[skill_id-1].hit > Data::states[inflictedState - 1].restrict_skill_level))) {
+			return false;
+		}	
+	}
```
Ghabry: This is still of-by-one for skill_id, you use it twice (it's btw hard to find changes after reviewing when you completely rewrite the history^)

src/game_battlealgorithm.cpp
```diff
@@ -518,7 +512,7 @@ bool Game_BattleAlgorithm::Normal::Execute() {
 			hit_chance = weapon.hit;
 			crit_chance = crit_chance += weapon.critical_hit;
 			multiplier = GetAttributeMultiplier(weapon.attribute_set);
-			source->ChangeSp(source->GetSp() - Data::items[weaponID].sp_cost);
+			sp = - Data::items[weaponID].sp_cost;
```
Ghabry: This will not work because the "sp = " value is for the SP change of the target.

src/game_battlealgorithm.cpp
```diff
@@ -764,7 +758,7 @@ void Game_BattleAlgorithm::Skill::Apply() {
 	}
 	else {
 		if (first_attack) {
-			source->SetSp(source->GetSp() - source->CalculateSkillCost(skill.ID));
+			source->ChangeSp(- source->CalculateSkillCost(skill.ID));
```
Ghabry: You can do it like here in the first_attack-if instead (but in Normal::Apply ^^)

Ghabry:
Otherwise your PR looks okay. I can't test it currently but by just looking over it is fine.

I'm curious how magic weapons behave in RPG_RT, have to test this ^^

src/game_actor.cpp
```diff
@@ -119,7 +119,7 @@ bool Game_Actor::IsSkillUsable(int skill_id) const {
 		return false;
 	}
 	for (int inflictedState : this->GetInflictedStates()) {
-		if ((Data::states[inflictedState - 1].restrict_magic && Data::skills[skill_id].magical_rate > Data::states[inflictedState - 1].restrict_magic_level) || ((Data::states[inflictedState - 1].restrict_skill && Data::skills[skill_id-1].hit > Data::states[inflictedState - 1].restrict_skill_level))) {
+		if ((Data::states[inflictedState - 1].restrict_magic && Data::skills[skill_id - 1].magical_rate > Data::states[inflictedState - 1].restrict_magic_level) || ((Data::states[inflictedState - 1].restrict_skill && Data::skills[skill_id-1].hit > Data::states[inflictedState - 1].restrict_skill_level))) {
```
Ghabry: This code is already implemented in Game_Battler::IsSkillUsable. So I was wrong when I said that silence is not working yet. Sorry :(

Tondorian:
I tested it .. my version of rpg2k3 says: you will not be able to attack if sp is to low but tests shows it consumes sp to 0 and attacks allways works ...

src/game_actor.cpp
Tondorian: But i tested it befor i implement this and i was able to cast spells ..

Ghabry:
Imo it would be very stupid if you could not attack normally anymore when SP are 0 so this makes sense ^^

src/game_actor.cpp
Ghabry: I didn't say that the code was wrong but that you should merge it with the restriction code of Game_Battler::IsSkillUsable ^^

src/game_enemy.cpp
```diff
@@ -193,6 +193,9 @@ int Game_Enemy::GetDropProbability() const {
 }
 
 bool Game_Enemy::IsActionValid(const RPG::EnemyAction& action) {
+	if(action.kind == action.Kind_skill) {
+		return IsSkillUsable(action.skill_id);
+	}
 	switch (action.condition_type) {
```
Tondorian: Maybe this is not the best way to check
need feedback

src/game_battler.cpp
```diff
@@ -61,6 +61,10 @@ int Game_Battler::GetSignificantRestriction() {
 	return RPG::State::Restriction_normal;
 }
 
+int Game_Battler::GetId() const {
+	return Game_Battler::GetId();
```
Ghabry: warning C4717: "Game_Battler::GetId": Rekursiv für alle Steuerelementpfade. Die Funktion verursacht einen Stapelüberlauf zur Laufzeit.

src/game_battler.cpp
```diff
@@ -128,6 +132,12 @@ bool Game_Battler::IsSkillUsable(int skill_id) const {
 
 	const RPG::Skill& skill = Data::skills[skill_id - 1];
 
+	for (int inflictedState : this->GetInflictedStates()) {
+		if ((Data::states[inflictedState - 1].restrict_magic && Data::skills[skill_id - 1].magical_rate > Data::states[inflictedState - 1].restrict_magic_level) || ((Data::states[inflictedState - 1].restrict_skill && Data::skills[skill_id - 1].hit > Data::states[inflictedState - 1].restrict_skill_level))) {
+			return false;
+		}
+	}
```
Ghabry: Please provide a test case where your code works. Because for me it fails.

Tondorian: [rpg2k3 testcase.zip](https://github.com/EasyRPG/Player/files/180349/rpg2k3.testcase.zip)

Ghabry:
1st row RPG_RT
2nd row old code
3rd row your new code

unbenannt

PR #833, opened 2016-03-19
Replace PHP file finder

Ghabry:
Because everybody uses CDNs now ...

This downloads an index.json from the game directory and parses it via picojson.
Then the parsed result is used to generate requests :)

**This changes the location of the cache**
Because I want to share the request code the index file is now in the game directory and called "index.json"


Ghabry:
This is open for review :)

exe for windows: https://www.sendspace.com/file/7n7aol

fdelapena:
Amazing, it works like a charm and the latency saving due to the php call removal is clearly perceivable. :+1:.

PR #837, opened 2016-03-20
Fix the Wii port compilation and warnings

carstene1ns:
- Using `vsnprintf()` and `snprintf()` with g++ 4.8.2 is possible, but because of a quirk only with gnu extensions enabled.
- RTTI is not a C feature, so move it to `CXXFLAGS`
- Debug information might be interesting sometimes and the `boot.dol` gets them stripped anyway, so enabling (`-g`)
- `<cstdio>` is needed for `FILE*`, the jenkins script did that change for quite some time, so was forgotten :expressionless:
- Some variables were not used, so moving them or at least silence the compiler

### DIFF
--- a/builds/wii/Makefile
+++ b/builds/wii/Makefile
@@ -27,9 +27,9 @@ INCLUDES	:=	src
 # options for code generation
 #---------------------------------------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


carstene1ns:
This one should be ready.

PR #838, opened 2016-03-20
Fix Battle system regressions :/

Ghabry:
A bug in the newly added Attribute damage Multiplier returned 0 under certain circumstances :(


Ghabry:
This is not ready, I also want to fix another battle crash before...

PR #841, opened 2016-03-21
fixes #840

### LINKED ISSUE #840: Abilities deal like 10% of the DMG they dealt prior to this patch.
Hey, just updated my EasyRPG Player on Android and and i'm currently playing unterwegs in düsterburg and all magical spells to 4-5 dmg instead of 90-120~

the patch notes said "fixed dmg bugs" ,, i didn't have any dmg bugs prior.

Tondorian:
attributes_set.size could be very hight 12> even if there is just one attribute "true"
So it needs to count how much attributes are "true"


Ghabry:
The difference is noticable ;)

Issue #788, opened 2016-03-01
Calling a common event from battle uses the battle interpreter and breaks most common events


Ghabry:
TestCase:
Battle 198 in AraFell.

Displays "Seri Kesu casts powerful spells, but her Magic 
does not regenerate during combat unless she 
defends. Her Magic is restored after battles." but shouldn't.

Reason is using a ConditionalBranch. This is only implemented for the Map interpreter, not for Battle interpreter. Maybe we should move almost all commands to Game_Interpreter (except the ones that really don't make sense to use in battle).

Ghabry:
See also our wiki: https://easy-rpg.org/wiki/development/technical-details/common-events-battle-2003

PR #852, opened 2016-03-29
Fix battle bugs

Ghabry:
Two further oversights in the attribute multiplier :(

And that "Feuer um Mitternacht EX" problem with the non usable skill.


carstene1ns:
:+1:

Ghabry:
I also wanted to fix #788 but the weekend is over again and I don't want to postbone this PR again :)
Open for review.

PR #855, opened 2016-04-02
Make HardLight, Saturation, Fill and FileFinder faster

### LINKED ISSUE #846: Wii & 3DS: Startup is super slow
This is either a problem with parsing the database or with the file finder and needs investigation.

And super slow means more then 30seconds up to minutes!

Ghabry:
See the commit messages for detailed info.

I also measured hard light with and without lookup table. The lookup is cache unfriendly (64kb) but is faster then direct calculation because Hard light has a 50% branch prediction miss chance per color...

For saturation would be more tricky because RGB are combined here, I "only" saved an alloc and a blit here.

Fill (AddBackground, called once per frame) is also faster now by using OP_SRC which doesn't handle alpha correctly but that doesn't matter in this case.

Branch Pred: https://stackoverflow.com/questions/11227809/why-is-processing-a-sorted-array-faster-than-an-unsorted-array

Fixes #846

### DIFF
--- a/src/bitmap.cpp
+++ b/src/bitmap.cpp
@@ -832,10 +832,16 @@ static pixman_color_t PixmanColor(const Color &color) {
 
 void Bitmap::Fill(const Color &color) {
 	pixman_color_t pcolor = PixmanColor(color);
-	pixman_rectangle16_t rect = {
-	0, 0, static_cast<uint16_t>(width()), static_cast<uint16_t>(height())};
+	Rect src_rect(0, 0, static_cast<uint16_t>(width()), static_cast<uint16_t>(height()));
 
-	pixman_image_fill_rectangles(PIXMAN_OP_OVER, bitmap, &pcolor, 1, &rect);
+	pixman_image_t* timage = pixman_image_create_solid_fill(&pcolor);
+
+	pixman_image_composite32(PIXMAN_OP_SRC,
+		timage, (pixman_image_t*)NULL, bitmap,
+		src_rect.x, src_rect.y,
+		0, 0,
+		0, 0,
+		src_rect.width, src_rect.height);
 
 	RefreshCallback();
 }
@@ -846,7 +852,7 @@ void Bitmap::FillRect(Rect const& dst_rect, const Color &color) {
 	static_cast<int16_t>(dst_rect.x),
 	static_cast<int16_t>(dst_rect.y),
 	static_cast<uint16_t>(dst_rect.width),
-	static_cast<uint16_t>(dst_rect.height), };
+	static_cast<uint16_t>(dst_rect.height)};
 
 	pixman_image_fill_rectangles(PIXMAN_OP_OVER, bitmap, &pcolor, 1, &rect);
 
@@ -859,6 +865,15 @@ void Bitmap::Clear() {
 	0, 0, static_cast<uint16_t>(width()), static_cast<uint16_t>(height())};
 
 	pixman_image_fill_rectangles(PIXMAN_OP_CLEAR, bitmap, &pcolor, 1, &rect);
+
+	RefreshCallback();
+
+	pixman_image_composite32(PIXMAN_OP_CLEAR,
+		bitmap, (pixman_image_t*)NULL, bitmap,
+		0, 0,
+		0, 0,
+		0, 0,
+		width(), height());
 
 	RefreshCallback();
 }
--- a/src/bitmap.cpp
+++ b/src/bitmap.cpp
@@ -876,6 +876,22 @@ void Bitmap::ClearRect(Rect const& dst_rect) {
 	RefreshCallback();
 }
 
+// Hard light lookup table mapping source color to destination color
+static int hard_light_lookup[256][256];
+
+static void make_hard_light_lookup() {
+	for (int i = 0; i < 256; ++i) {
+		for (int j = 0; j < 256; ++j) {
+			int res = 0;
+			if (i <= 128)
+				res = (2 * i * j) / 255;
+			else
+				res = 255 - 2 * (255 - i) * (255 - j) / 255;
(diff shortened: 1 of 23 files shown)


Ghabry:
Speed tests on my old Android system (TestGame Tone map):
Tone: 42 FPS
Sat: 49 FPS
Both: 33 FPS

This PR:
Tone: 60 FPS
Sat: 59-60 FPS
Both: 58 FPS

BlisterBoy:
Nice! People are very interested in the engine optimization since the official RPG Maker update, this surely is a good step forward!

src/bitmap.cpp
```diff
@@ -876,6 +891,22 @@ void Bitmap::ClearRect(Rect const& dst_rect) {
 	RefreshCallback();
 }
 
+// Hard light lookup table mapping source color to destination color
+static int hard_light_lookup[256][256];
```
Zegeri: Could this be using uint8_t for less memory footprint?

Ghabry: Ugh, very good find.

Zegeri: I'm not sure if we would want this, but this could use constexpr to create the table at compile time like this:
```
struct hard_light_table {
	uint8_t table[256][256];
	constexpr hard_light_table() : table() {
	    for (int i = 0; i < 256; ++i) {
			for (int j = 0; j < 256; ++j) {
				int res = 0;
				if (i <= 128)
					res = (2 * i * j) / 255;
				else
					res = 255 - 2 * (255 - i) * (255 - j) / 255;
					table[i][j] = res > 255 ? 255 : res < 0 ? 0 : res;
			}
		}
	}
};
constexpr hard_light_table hard_light_lookup;
```
And replace hard_light_lookup with hard_light_lookup.table.

Ghabry:
Sorry I also added a non-blit operation to the speedups:
A gigantic performance gain in the FileFinder. readdir already tells you if the file is a directory so extra stat calls are nonsense.

For zegeris old test case this reduces the parsing time on Windows for me from 0.35 seconds to 0.000287

This needs a test on all systems because readdir doesn't give this guarantee that the field is populated with useful data.
Can somebody test this under Linux? But I assume it works there.
Also Wii needs a test obviously.

At least on the 3DS it doesn't work but I already have a solution by looking at other sourcecode.

Ghabry:
For @Rinnegatamante the required changes for the filefinder based on https://github.com/EasyRPG/Player/pull/855/commits/101ac37f8242feae3f5ea048b20a7c448ebb2f2c

The 3DS needs this before the ::readdir-loop:

```cpp
#ifdef _3DS
	sdmc_dir_t* sdmc_dir = (sdmc_dir_t*)dir->dirData->dirStruct;
#endif
```

and inside the loop

```cpp
#ifdef _3DS
		// ctrulib does not populate the d_type field
		bool is_directory = sdmc_dir->entry_data.attributes & FS_ATTRIBUTE_DIRECTORY;
#else
		bool is_directory = ent->d_type == S_IFDIR;
#endif
```

carstene1ns:
http://stackoverflow.com/a/10376245 and also the first comment of it is interesting why we do it the way we did before this PR.

src/filefinder.cpp
```diff
@@ -593,8 +595,8 @@ FileFinder::Directory FileFinder::GetDirectoryMembers(const std::string& path, F
 #endif
 		if (name == "." || name == "..") { continue; }
 
-		bool is_directory = ent->d_type == S_IFDIR;
-
+		bool is_directory = S_ISDIR(ent->d_type);
```
carstene1ns: `ent->d_type == DT_DIR` would work for me

Ghabry:
I added a fallback mechanism now. So when testing you have to check the log now if you get a debug log about the system not supporting fast dir checks.

Tondorian:
now wii buld finds game and is fast, too

Ghabry:
Great, I have two further optimisations under development that will give even more speed.
With them I reach 50 FPS on the old 3DS :)

src/message_overlay.cpp
```diff
@@ -63,6 +63,10 @@ void MessageOverlay::Draw() {
 			}
 			dirty = true;
 		}
+	} else {
+		if (!show_all) {
+			return;
+		}
 	}
```
carstene1ns: great little tweak!

Ghabry:
About the new commits:
When blitting, the code checks now if the src bitmap is completely opaque and will do a OP_SRC instead of OP_OVER here. This also applies to single map tiles, most of them are blitted now via OP_SRC.

Additionally the system graphic is not rendered anymore for title and map scenes.

Sprites that would be rendered out of bounds are not rendered anymore.

Ghabry:
Don't merge yet. Found already an visual issue when the chipset is missing or has an invalid tile and doesn't render ^^ (AB blocks in testgame)

Ghabry:
This and the battle PR are now finished (finally).
Expect enjoyable performance on all devices soon :D

carstene1ns:
Is your filefinder change tested with games that have nested subdirectories? Otherwise LGTM.

Ghabry:
I didn't try running any game. For this case I only checked zegeris subdirectory benchmark test case and the date structure showed the expected result.

Ghabry:
I tested now: Chipset under "aaa\basis" is found and transparency of tiles in Embric: Works

Ghabry:
[RPG_RT.zip](https://github.com/EasyRPG/Player/files/204947/RPG_RT.zip)

Just checkout testgame 2000 and replace the stuff.

BlisterBoy:
I tried this PR with the first town of Aedemphia.
Before : 9-25 fps
Now : 35-45 fps.

Improvements are clearly here, congratulation :) !

Ghabry:
Great, this reflects my basic benchmarks which showed up to +33%

PR #868, opened 2016-04-16
MSVC: First steps for UWP support

Ghabry:
This moves all but the main.cpp in a new project and compiles it into a lib.
Then the player project with the main.cpp links against the lib to get an exe.

(probably same trick automake does for unit testing)

I will need this later to make adding Universal Windows Platform support easier because the main project of UWP apps has lots of strange settings and otherwise this will be a mess.


Ghabry:
This needs a merge of https://github.com/EasyRPG/liblcf/pull/179 before it will compile

Ghabry:
Jenkins: test this please

Ghabry:
Don't expect too much. I only added a UWP skeleton which compiles and is a good base for further development. It will just start the game browser and find nothing because UWP runs in a sandbox and needs a special Api to access the outside world...
Also Audio is disabled because this gives lots of linker errors (no game = no audio, doesn't hurt yet ;)).
Besides some new ifdefs to disable unavailable WinApi the whole magic happens in builds\vs2015_uwp.

This is ready for now, will provide functionality another day.

PR #869, opened 2016-04-19
Add Nintendo 3DS port

### LINKED ISSUE #863: Integrate 3DS port
@Rinnegatamante wrote a nice Nintendo 3DS port which is mature enough for being integrated in the main codebase.

https://github.com/Rinnegatamante/easyrpg-player-3ds

Scheduling this for 0.4.2, so will happen this month. :)

Ghabry:
Finally! Thanks to @zegeri for the initial port and @Rinnegatamante for bringing the prototype in a usable state over the last weeks :).
Together with my performance improvments we went up from 16 FPS to ~56-59 FPS on old 3DS for the TestGame world map.

What makes this port special is that it does not depend on SDL and that it has a dedicated maintainer giving this port much more love.
This currently lacks MP3 and MIDI support, which will come later when mpg123 and fmmidi integration are pushed upstream.

Rinne is still adding code daily, so I suggest waiting a few more days and I will cherry pick his stuff... :)

This branch uses a slightly altered Makefile. Compile steps needed:
```make
export TOOLCHAIN_DIR=[easyrpg_3ds_toolchain-dir]
# LIBLCF is extra because this is our lib and needs recompiles
export LIBLCF_DIR=[liblcf_crosscompile-dir]

cd builds/3ds
make -j2
```

Fixes #863


Ghabry:
The commit order looks a bit wrong. Some of my commits are placed in between Rinnes o.O

Ghabry:
But a local git log shows the correct order. Probably gets confused because of the dates.

src/graphics.cpp
```diff
@@ -117,7 +117,7 @@ void Graphics::Update(bool time_left) {
 		fps = 0;
 
 		next_fps_time = current_time + 1000;
-
+
```
carstene1ns: eeek

src/main.cpp
```diff
@@ -18,13 +18,14 @@
...
+#include "output.h"
 #include <cstdlib>
 
 #ifdef USE_SDL
 #  include <SDL.h>
 #endif
 
-extern "C" int main(int argc, char* argv[]) {
+extern "C" int main(int argc, char* argv[]) {	
 	Player::Init(argc, argv);
```
carstene1ns: eeek²

Rinnegatamante:
A little thing. Release build should be compiled with -DNO_DEBUG and -DUSE_CACHE since right now NO_DEBUG will substitute bottom screen keyboard with a debug console and USE_CACHE allows usage of sounds cache (essential for a good stable framerate on a lot of games).

Ghabry:
My build crashes on the 3DS but works in Citra :(

Rinnegatamante:
My repository compiles fine, if you need, i can send you MD5 checksums for libctru/libsf2d/libkhax.

Ghabry:
Which toolchain do you currently use? Could you try with the latest I linked in the forum: https://easy-rpg.org/forums/viewtopic.php?f=4&t=625&start=200

And did you compile your repository or my pull request?

Ghabry:
Update: Found the crash reason \o/

Ghabry:
The 3DS build is ready now and works :+1: . There are some toolchain problems but they are not Player related. Something with pixman is wrong, my older lib of it works. hm

img_2817
img_2819

Rinnegatamante:
Nice! :P I think the icon i'm currently using in my repo is cooler however <.<

Ghabry:
The problem with your icon is that it doesn't follow the EasyRPG branding. <.<
But you make a valid point because the HB launcher lacks background alpha

Rinnegatamante:
It's just a zoomed version of the EasyRPG logo.

Ghabry:
Replaced the icon with a rectangular one.

Rinnegatamante:
I've noticed you have deleted the buildscript for CIA generation. What about it? It should be included in releases i think.

Ghabry:
I skipped that commit because it contained lots of Windows binaries.
I will manually readd everything except the exe/dll files.

Jenkins will also offer a elf file to make creating CIAs easier.

gnmmarechal:
Hum, you managed to actually put the whole icon in there. Although it probably will look weird on gridlauncher cuz the transparent part will be black. I'd rather have the zoomed-in one. In fact, the one I uploaded at my server for the easyRPG Updater to use if no icon is found is a zoomed-in EasyRPG icon (because pixelation occurs if I decrease the size of the icon too much). Also, I like the square.

Ghabry:
Black background is even worse. But the icon is already replaced: https://github.com/Ghabry/easyrpg-player/commit/291684737e74d0384e79e5c996d9683db378de3d#diff-1f9ccd3805917595c75fa1f51aa649bd

Ghabry:
I want to replace the sample if the CIA channel with one that we also (plan) to use for the Wii channel.
I read on gbatemp that the maximal WAV length is 3 seconds. But I don't trust gbatemp and don't see this documented anywhere. Can you confirm @Rinnegatamante

Rinnegatamante:
Not really 3 secs but something really near, like 3.4 secs. Also 3D models can be loaded so it could be cool to create a little model for the banner. (You can use Ohana 3DS to make it. A sample of 3D banner is in my CHMM2 CIA version: https://youtu.be/ZoH2PQi71yo?t=4m38s i know it can also be animated but never tried so).

Ghabry:
I ask @Rinnegatamante or @gnmmarechal to verify that the stuff in the cia_builder folder works. Especially the audio file. It is a 11025 Hz 16 PCM (Stereo). Compressed it quite heavy because I couldn't hear a difference.

Rinnegatamante:
hex_set is not a "released" tool but it's just a (really bad made) program that modifies Unique ID in workaround. Also the asked tools are not for common uses and a lot of people will not even know what PATH is.

Ghabry:
That's why I wrote "or copy them in this directory". I expect from the users to know how to get the required stuff for CIA building. 
Found the sourcecode of hex_set, this should be easy to replicate...

carstene1ns:
Well, if the tools are not for common use, the average user will not need them and will not create a custom game bundle anyway.

Rinnegatamante:
Tools are for common use for devs since normally homebrews are released in CIA format. Making a generic CIA homebrew for Easy RPG is just the worst idea you can have. That's why it should be released as ELF and tools for CIA generation should be included in the release.

carstene1ns:
Why does it not make sense?
The Player has a simple game browser included, you can add some games to the SD and are good to go.

Rinnegatamante:
If you want that no one will use EasyRPG on 3DS then you can build it so. You know 3DSX only homebrews are just ignored by users cause they're too lazy to launch hblauncher_loader?

carstene1ns:
So you say when I want to play multiple games I need to add the same application multiple times with different bundled data?

Rinnegatamante:
Oh and also you need to change the save_path detection if you want to use a single CIA build cause unique_id will be always the same.

carstene1ns:
Well, i guess we should just provide the executable then, leaving the bundle task to the average gamer?

carstene1ns:
Let's just assume I have no idea about 3ds homebrew.
How do I get it to work anyway? I got a SD card with my console, but cannot use it because I am lazy? Honestly, that seems a bit stupid to me. :bulb:

Also, what about the updater application then? How does it work? I assumed until now, that it just replaces the executable.

Rinnegatamante:
CIA homebrews are installable homebrews that will generate an icon on Home Menu, like this: https://pbs.twimg.com/media/CgSBWv1WsAE_eJF.jpg:large

3DSX homebrews are homebrews that must be launched from hblauncher. If you don't have CFW access, you must use this.

A lot of homebrews are 3DSX only cause CIA format is just uncomfortable to test.
To launch 3DSX homebrews with a CFW there is hblauncher_loader that just open hblauncher basically so you should do Home Menu -> hblauncher_loader -> 3DSX homebrew from hblauncher.
The point is, even if hblauncher_loader exists, no one cares about 3DSX homebrews cause 3DS users are too lazy to launch it and then select the 3DSX homebrew to launch.

@Ghabry : There are problems at least on Windows if i place the executables in the same path of the CIA builder ( https://rinnegatamante.it/ciab.png ). It basically says that it doesn't detect the files. Also hex_set since it's builded with cygwin (at least the one i released), it has a couple of dependecies (DLLs).

Ghabry:
I didn't say that all will use the same ID. I just said that I will replace set_hex with a powershell / sed call.

I don't see a reason to host the exe and dll in the repo. But we could provide them somewhere for download for convenience.

carstene1ns:
Ok, maybe we can get around this issue if we use some small loader from the installed cia. It would look at the same place for the real executable like the hblauncher. That way we can easily replace it when there are updates (or for testing) and the users are happy because they get a nice launcher.

Ghabry:
In that case I would suggest to hardcode a specific ID in the Player code for this "chainloader" solution and all other IDs will use romfs: + sdmc:/easyrpg/id-save.

Rinnegatamante:
CIA homebrews are not installed on a folder. (Infact as you can see from the code, to check if the launched homebrew is under hblauncher or not, you can just check for argc since CIA homebrews will not get any argument like current folder or others.)

So what's your idea?

carstene1ns:
My idea is to not have the _real_ Player in the cia, but a loader, that finds the real one on SD.

Ghabry:
carstene1ns solution is maybe a bit tricky on the 3DS.
My idea is to hardcode to folder "sdmc:/3ds/easyrpg-player/" as game and save directory when the ID is (e.g.) 0xEA51. And for all other ideas it will take the romfs: codepath you programmed.
This way they can have a easyrpg cia app installed which will show a game browser after startup. And for all others it will be "standalone mode" with a romfs game.

Rinnegatamante:
O.o if you want to launch a 3DSX file from a CIA file, it's basically not possible. hblauncher_loader launches a payload that has hardcoded value (boot.3dsx).

Rinnegatamante:
Ehm, for how i wrote the startup code, if romFs is empty, it will just launch normally using sdmc and it will not detect any game folder so it will result in the game browser appearing. (don't know where it will go to see for the games) https://github.com/Rinnegatamante/easyrpg-player-3ds/blob/master/src/main_data.cpp#L84-L91

carstene1ns:
Depends what the filesystem provides after launching, i.e. the working directory.

Ghabry:
Okay in that case we can just assume: When launched from a .cia and the romfs is empty the Player will search for games in uuuuhm. sdmc:/3ds/easyrpg-player/

carstene1ns:
That seems good to me, we can provide this 'empty' cia and the users can use it or just the also provided 3dsx to create an own cia (bundled with the tools).

Ghabry:
Without a (detected) romfs the CIA will now load games from sdmc:/3ds/easyrpg-player/ . Also updated the scripts, double clicking the batch file works now with the exe files in the same path.

Ghabry:
Here is a zip for testing: http://easy-rpg.org/downloads/easyrpg-3ds.zip

Contains 3dsx, CIA (without romfs, searches roms in sdmc:/3ds/easyrpg-player) and ELF. I would be interested in if the CIA works and if you can create an ELF/CIA with RomFS.

This was 99% Rinnes Code but I moved it to another file so I just want to be sure that I didn't break anything. :D

Ghabry:
okay I asked some people to test the CIA. It finds the games in /3ds/easyrpg-player for the non romfs case and the audio file got converted to 32k and Plays correctly in the 3ds menu now.

From my side this is merge ready.
Will need a Jenkins Job update and a Website update (download links for 3dsx, cia, elf and backlink) and a blog announcement afterwards.

Because you use adf.ly everywhere on your website feel free to provide adf.ly links on your website for downloading the 3ds port and advertise it on gbatemp or where-ever. Maybe also bundled with some games. Just for Information, the license permits this [for games it depends on the game license but this is not my problem].

Rinnegatamante:
I was going to download the various builds right now, what a timing! Lol.
Anyway, really good to know it works fine.
I'm going to try to make a 3D banner, i have little experience with Blender, maybe i can produce something quite good :P

Ghabry:
Well you could test the romfs-cia-case. The zip above still applies, the CIA has broken title audio but doesn't really matter for testing.

Ghabry:
Concerning the 3d model I can provide you with a non-polished one @Rinnegatamante . Needs some work and must be exported to a 3ds friendly format. Just gimme a few minutes.

Ghabry:
Have fun: [EasyRPG_3D_Logo.zip](https://github.com/EasyRPG/Player/files/233166/EasyRPG_3D_Logo.zip)

Rinnegatamante:
404 :P

Ghabry:
[EasyRPG_3D_Logo.zip](https://github.com/EasyRPG/Player/files/233174/EasyRPG_3D_Logo.zip)

Rinnegatamante:
Looks like the model is untextured on Blender for me. Is it normal?

Ghabry:
No that's not normal. That's blender.

Ghabry:
I tried it now. Importing the "obj" file works for me and displays correctly.

Ghabry:
unbenannt
 how it looks for me

Rinnegatamante:
Yep, i was using the Texture Mode when Solid Mode should be used. Anyway looks like Ohana3DS has some problem importing it.

ItsSchoko:
@Rinnegatamante Are there any guidelines for exporting to Ohana3DS?

Rinnegatamante:
Unfortunately not, it's kinda hard to convert a model. It's relative simple to modify textures for existent models but it's pretty hard to convert a new model.

Rinnegatamante:
If you like it, i made one by just modifying an already existent 3D banner. Here's the preview: https://goo.gl/photos/GpQ5AeyzZiLPndQ9A

Ghabry:
Well, then lets stick to this edited one when creating one from scratch is not possible yet.

Rinnegatamante:
Here you can find the bcmdl file for this banner and also the binray file. (You'll probably want to change the sound used: http://rinnegatamante.it/bfiles.rar )

Ghabry:
Thx. Could you tell me how to change the audio track? Have no idea how to use the 3DS tools ;)

Rinnegatamante:
You must download ba-GUI-nnertool and extract the .bin banner in the archive, then you must replace the bcwav file with your own audiofile and then you can re-build the binary file with the same tool.

Ghabry:
I really love btw that the bannertool command line tool always segmentation faults when a file is missing or an argument is wrong. Quality software.

Ghabry:
okay now I do how to do it from the command line. I quickly pressed pause when a command line window popped up and looked in the taskmgr to figure out the command line. :D

Ghabry:
@Rinnegatamante or @gnmmarechal 
Here is a new CIA: http://easy-rpg.org/downloads/easyrpg-player-cia.zip  Could you please test if this has the 3d banner of Rinne and a working sound effect?

carstene1ns:
@Siltala reports the CIA shows the banner correctly and plays the sound fine, however it crashes on startup.

Ghabry:
The download link in my previous post got updated and contains now 2 CIAs. Please test both.

Ghabry:
Ah crap I used the wrong toolchain directory ~.~ ... I should really clean up my stuff...

Rinnegatamante:
So does the CIA still needs a test or not since it has been merged? I have enough time for a couple of tests if needed.

gnmmarechal:
Awesome!

Ghabry:
@Rinnegatamante 
We found somebody else in our IRC channel to do the test for us and he said it works :)

The elf is now also available in a zip archive. Please update your links to point to that zip, I will remove the gigantic elf in a day or two.

gnmmarechal:
@Ghabry I may implement that, but I don't know if it really is necessary - after all, it's a really small download anyway.

Ghabry:
Well it is a useful information to the user imo. I don't care about the download, we have more then enough free traffic, I'm just suggesting some ideas if you should be bored and want to do more programming :D

gnmmarechal:
I see. I may add it, whenever I get the time. Atm I have some more ideas
(ex. direct game download & installation, there's a ton of RPGM2k/2k3
freeware that I can add to make some sort of... PS Store for RPGs? I guess)

PR #870, opened 2016-04-23
Make freetype an optional dep

Ghabry:
This makes freetype automagic on when available. Otherwise it defaults to off (I hope, copy-pasted from that gentoo article *g*)

Removed freetype from Wii, not sure if we will ever need it for this platform...


carstene1ns:
I have a small fixup to the autotools part, but otherwise looks good to me.
https://gist.github.com/carstene1ns/2220cdadaabe106f44bd5a4c1d8df104

PR #873, opened 2016-04-24
Add mpg123 and fmmidi support

### LINKED ISSUE #721: Space Funeral: Unrecognized format when playing mp3
Reported by a user on Google Play.
I wouldn't care about noise when playing MP3 but "Unrecognized format" warning is new :O.

Not tested yet.

### LINKED ISSUE #752: Embed an analog software synthesizer for MIDI playback
The [fmmidi](http://unhaut.x10host.com/fmmidi/) software is a BSD-licensed Frequency Modulation (FM) syntheziser which emulates the [Yamaha YM2608](https://en.wikipedia.org/wiki/Yamaha_YM2608) chip (the FM part). The code only costs around 100 kb.
The main advantage of adding this allows good MIDI playback with low CPU and memory footprint and not depending on large wavetable digital patches. An alternate solution instead of this would be a reasonably smaller .pat set.
[It sounds funny](https://fran.cr/owncloud/index.php/s/0jGfb6yK6buZrJf/download) with RTP's Opening2.

This would require to pass the stream to SDL, e.g. in a SDL_mixer channel (and hopefully OpenAL too) instead of using libao (which is a really good library, but _another_ library).
There are around a couple of cpp source files, which could be copied to our source tree, as this code is not designed as library and does not look to have maintenance activity.

### LINKED ISSUE #781: HOME (OFF fangame): crash while transitioning into boss battle
This game crashes when fighting in map with enemy group id 77 (Butcher Burned).

[Game download](http://rpgmaker.net/games/6361/downloads/). [Savegame](https://github.com/EasyRPG/Player/files/149452/Save01.zip).

Reported by MJ via Android bug report email.

### LINKED ISSUE #818: Support libmpg123 for MP3 playback
By using this amazing library, we would get the following:

- **Plays broken MP3** created with broken tools using libLAME incorrectly, properly skips data until finding the first valid frame (compliant decoder), which would be a fix for #721.

- **Gapless playback loops** where possible (it even supports special LAME gapless tags if they exist)

- libmpg123 is **maintained** (libmad is dead from more than a decade and needs a bunch of patches) and packaged in popular distros.

- Because we need new code for this, it would **prevent noises or crashes** due to buffer issues in existing libmad's dynamic_mp3/Mix_PlayMusic code when resampling (#781, #677).

- Supports MPEG-1/2/2.5 without issues (SMPEG crashes with some 2/2.5). Also supports I/II/III layers like libmad (layer III is enough I guess).

- LGPLv2.1, this license is compatible with code combination (and therefore static linking) with GPLv3. With libmad was also fine because it is GPLv2+.

Below there is a test program to demonstrate how it plays files properly.

Ghabry:
This is not completely ready yet, the opt-in logic fails.

I need the PR builders to test the MIDI playback (speed). :)

Fix #721
Fix #752
Fix #781
Fix #818


fdelapena:
MP3 playback works for me with Monigote Fantasy:

https://easy-rpg.org/play/index-pr.html?game=monigote_fantasy

However:

- Title music playback (mp3) is not playing (note: large mp3 file, but no async info such halt info in console):
- Map mp3 music works, however the battle music is a 48khz song here (emscripten is using 44100hz sampling freq here), is using libmad instead mpg123 maybe?

About fmmidi: no luck with emscripten yet :(

Ghabry:
My debug code says the callback fires and the buffer is filled :(

Ghabry:
I removed AudioCVT and allow ALL_CHANGES but this nothing, not even noise... Maybe this is broken with emscripten? :/

carstene1ns:
Under Linux I get `fallthrough <mp3file>` btw. for the PeuterisGrey music.

Ghabry:
I assume it has an empty frame at the beginning and does not end with .mp3? :D

carstene1ns:
Starts with ID3 tag and I renamed the files from wav to mp3.

carstene1ns:
At least fmmidi works under linux \*g\*

Ghabry:
I'm quite certain this is somehow broken in emscripten :(

Ghabry:
While the hook is enabled normal sound effects are also broken :o. The whole Player goes silent.

Ghabry:
SDL_mixaudio seems to be the problem. I got noise :+1:

fdelapena:
Now it works for me for both fmmidi and mpg123 without freq or noise issues \o/ (my browser guessed freq is 44100).

Ghabry:
The purpose of the heuristic is to distinguish between MOD and MP3 files. I will try to tweak the heuristic based on Monigote MP3s and AE MOD files later...

Ghabry:
Jenkins: Test this please

Ghabry:
Jenkins: Test this please (sorry for noise)

Ghabry:
The new heuristic filters perfectly AE_EN Mods from Monigote MP3s:

```
Fall through .\Music\absolu.mod.mp3
MP3 .\Music\amable ternura.mp3
MP3 heuristic .\Music\Ambiente helado.mp3
MP3 heuristic .\Music\Castlevania - Circle of the Moon.mp3
MP3 .\Music\Castlevania - Lago del Diablo.mp3
MP3 heuristic .\Music\Castlevania incesante.mp3
MP3 .\Music\Choujin Sentai - Jetman - Area A.mp3
MP3 .\Music\Choujin Sentai - Jetman - Area B.mp3
MP3 .\Music\Choujin Sentai - Jetman - Area C.mp3
MP3 .\Music\Choujin Sentai - Jetman - Area D.mp3
MP3 .\Music\Choujin Sentai - Jetman - Area E.mp3
MP3 .\Music\Choujin Sentai - Jetman - Area F.mp3
Fall through .\Music\dead_windmills.xm.mp3
MP3 .\Music\Digital Devil Story - Exodus.mp3
MP3 .\Music\Digital Devil Story - Nuclear Dawn.mp3
MP3 heuristic .\Music\Digital Devil Story - Sanity's Requiem.mp3
MP3 .\Music\Digital Devil Story - Uncertainty.mp3
MP3 .\Music\Don Doko Don 2 - Villa del Champinon.mp3
MP3 heuristic .\Music\Double Dragon III - Temple of Chaos.mp3
MP3 heuristic .\Music\Dracula's Curse - Barco Fantasma.mp3
MP3 heuristic .\Music\Dracula's Curse - CorazÃ³n Embravecido.mp3
MP3 heuristic .\Music\Dracula's Curse - Gruta.mp3
MP3 heuristic .\Music\Dracula's Curse - PrÃ³logo al infierno.mp3
MP3 heuristic .\Music\Dracula's Curse - Reto del CorazÃ³n.mp3
MP3 heuristic .\Music\Dracula's Curse - RetÃ³rica Lunar.mp3
MP3 heuristic .\Music\Dracula's Curse - Ruta Neblinosa.mp3
MP3 heuristic .\Music\Dragon Fighter - Fight in the Cave.mp3
MP3 heuristic .\Music\Dragon Fighter - Fight in the Dungeon.mp3
MP3 heuristic .\Music\Dragon Fighter - Fight in the water dream.mp3
MP3 .\Music\Dragon Quest III - Dungeons.mp3
MP3 heuristic .\Music\Dragon Quest IV - Aventura Fantastica.mp3
Fall through .\Music\drnkref.xm.mp3
Fall through .\Music\em2.mod.mp3
MP3 .\Music\Expo Solidaria.mp3
MP3 heuristic .\Music\Final Fantasy III - Nostalgia.mp3
MP3 heuristic .\Music\Final Fantasy III - Que monon!.mp3
MP3 .\Music\G.I. Joe - Cobra.mp3
MP3 .\Music\G.I. Joe - Sewers of Snake Eyes.mp3
MP3 .\Music\G.I. Joe - Tribute.mp3
MP3 heuristic .\Music\Gun Nac - Boss A.mp3
MP3 .\Music\Gun Nac - Boss B.mp3
MP3 .\Music\Gun Nac - Boss C.mp3
MP3 .\Music\Harmony of Dissonance Hearth.mp3
Fall through .\Music\jazznocn.xm.mp3
Fall through .\Music\knight_of_fire_-_soul_chill.it.mp3
MP3 .\Music\Linde Enjoyable.mp3
Fall through .\Music\mega_destruction.xm.mp3
MP3 heuristic .\Music\Metal Gear Ghost Babel - Tragedy.mp3
MP3 .\Music\Mighty Bomb Jack.mp3
MP3 heuristic .\Music\Mother - Descansa y Repone tu Salud.mp3
Fall through .\Music\musix-pointless.mod.mp3
MP3 heuristic .\Music\por Martin Aparicio - Batalla!.mp3
Fall through .\Music\reddream.xm.mp3
MP3 .\Music\Rockman 1 - Victory!.mp3
MP3 heuristic .\Music\Rockman 4 - Desert Eagle.mp3
MP3 heuristic .\Music\shatterhand - Conglomerado Naval.mp3
MP3 heuristic .\Music\Rockman 6 - Hard Castle.mp3
MP3 heuristic .\Music\shatterhand - Factory Shaft.mp3
MP3 .\Music\shatterhand - satellite.mp3
MP3 heuristic .\Music\shatterhand - Parque de Diversiones.mp3
MP3 heuristic .\Music\Side Pocket Bar.mp3
MP3 heuristic .\Music\Tengen Tetris.mp3
MP3 .\Music\the Chordettes - Mr. Sandman auchan.mp3
```

fdelapena:
:+1: just as a reminder, looping works but the last time I checked the inn call it didn't finish after the first loop, hanging the game there.

Ghabry:
Now everything is implemented except the correct handling of the WITH_FMMIDI macro
Ugly debug output is also gone.

Ghabry:
@Rinnegatamante  The API is stable now and besides some smaller ToDos the coding is finished.

So if you want you can checkout my branch and try to get the audio system integrated for 3DS (only BGM, SE is not used for MP3 and MIDI)

```
git checkout -b Ghabry-audio master
git pull https://github.com/Ghabry/easyrpg-player.git audio
```

See the docs in https://github.com/Ghabry/easyrpg-player/blob/audio/src/audio_decoder.h

and the usage at
https://github.com/Ghabry/easyrpg-player/blob/audio/src/audio_sdl.cpp#L235
and
https://github.com/Ghabry/easyrpg-player/blob/audio/src/audio_sdl.cpp#L52 (there is also DecodeAsMono)

You will not need SetFormat, *Pitch and *Fade because this is just to workaround SDL mixer limitations but you have direct control over a DSP.

Ghabry:
This is ready now. Two things must be tested by others:
1. openal_audio still compiles
2. MIDI fallback mode works

Rinnegatamante:
Toolchains on Jenkins lacks of fmmidi apparently. I need an updated portlibs to do my tests just to be sure to not have problems cause libraries.

Ghabry:
FmMidi is shipped with the Player directly you don't need extra libs. Just define "WANT_FMMIDI" to 1 and it will work when you call AudioDecoder::Create.

See also: https://github.com/EasyRPG/Player/pull/873/commits/06c761841e557c9ad33e3e3014b272ec0a3635aa

It is from this website http://milkpot.sakura.ne.jp/fmmidi/ and no Linux distribution ships this package so we embedded it directly. ^^

Ghabry:
ソフトウェアFM音源MIDIプレイヤー

Sofutowea FM -- MIDI Pureia - Love Katakana :D

Rinnegatamante:
During compilation i get this: c:\users\alessio\desktop\audio\easyrpg-player-audio\src\main_data.h:22:18: fatal error: data.h: No such file or directory

Ghabry:
That's because liblcf is not part of the portlibs anymore. Because it's our own lib it is extra:
https://easy-rpg.org/jenkins/view/liblcf/job/liblcf-3ds/

Just extract it into the Portlibs dir (or point LIBLCF_DIR to the folder when using my Makefile)

Rinnegatamante:
Yeah, when this error appeared, i've already placed liblcf in the same dir of portlibs.

Ghabry:
Hm then make sure that $TOOLCHAIN_DIR/include/data.h exists. Otherwise no idea oO

Rinnegatamante:
About the output audio format, is it possible to enforce a specific output with SetFormat? (3DS supports only ADPCM, PCM16 and PCM8 audiocodecs).

Rinnegatamante:
Tried a first implementation for mp3 files and this code resulted in an instant system crash with Ib when starting title.mp3: http://pastebin.com/LBeFFf4L

Log:

[2011-04-04 23:02:49] Debug: EasyRPG Player 0.4.1 started
[2011-04-04 23:02:49] Debug: ============================
[2011-04-04 23:02:49] Debug: csnd:SND has been selected as audio service.
[2011-04-04 23:02:49] Debug: CLI: 
[2011-04-04 23:02:50] Debug: Detected encoding: ibm-943_P15A-2003
[2011-04-04 23:02:50] Debug: Using . as Game and Save directory
[2011-04-04 23:02:51] Debug: Loading game Ib
[2011-04-04 23:02:51] Debug: Using RPG2k Interpreter
[2011-04-04 23:02:53] Debug: MP3 heuristic: ./Music/title.mp3

Rinnegatamante:
Added more debug checks ( http://pastebin.com/Xa5HX952 ) and got this in log file:
[2011-04-05 03:32:06] Debug: EasyRPG Player 0.4.1 started
[2011-04-05 03:32:06] Debug: ============================
[2011-04-05 03:32:06] Debug: csnd:SND has been selected as audio service.
[2011-04-05 03:32:06] Debug: CLI: 
[2011-04-05 03:32:06] Debug: Detected encoding: ibm-943_P15A-2003
[2011-04-05 03:32:06] Debug: Using . as Game and Save directory
[2011-04-05 03:32:08] Debug: Loading game Ib
[2011-04-05 03:32:08] Debug: Using RPG2k Interpreter
[2011-04-05 03:32:10] Warning: Creating AudioDecoder
[2011-04-05 03:32:10] Debug: MP3 heuristic: ./Music/title.mp3
[2011-04-05 03:32:10] Warning: Getting Format

So it looks like it crashes after doing the GetFormat / SetFormat calls.

Rinnegatamante:
Nvm, found what caused the problem, mp3 playback added. Has some slight issues but not so big problems.

Ghabry:
You probably forgot to call "Open". Hm, maybe I should change the API that Create opens the file stream directly, would make sense :/

I recommend that you pass "S16" instead of "U16" in the SetFormat call because this work for FmMidi.

Concerning the detection I recommend that you do it like in the SDL version:

Place the AudioDecoder::Create before the ```// Reading and parsing the magic``` and when "Create" returns nullptr the file format was not supported and it falls through to your WAV/OGG code (and otherwise you call OpenAudioDecoder (which is currently OpenMP3)). This way you get MP3 detection (which is a bit tricky) and MIDI support at once with the same code bash :)

carstene1ns:
btw. `src/audio_decoder.h` needs `#include <cstdio>` to compile on PSP (for `FILE *`).

Rinnegatamante:
AS @carstene1ns pointed, the 3DS port too needs cstdio header. Added it on the header file.

Ghabry:
I'm not aware of a huge stack allocation in the decoding phase.

DecodeAsMono will do a heap allocation once of size * 2 (size of both buffers) the first time it is called and needs more time because it has to do the mono conversion.

Decode writes into the passed buffer directly and allocates nothing.

The sample rate should also not matter because this doesn't influence the buffer size (assumption, you pass the buffer)

Rinnegatamante:
That's the current stage about MP3 ( https://www.youtube.com/watch?v=wJY8nX5HnIM&feature=youtu.be ). Basically it works quite good but sometimes there are slight problems in looping apparently. Gonna do more tests and see where can be the problem.

Ghabry:
Great. Sounds really good :)

Ghabry:
@Rinnegatamante 
Did you also try Midi already? Because it also uses the AudioDecoder the code would be exactly the same as for MP3. I would be especially interested if the CPU speed of the 3DS is fast enough for this because fmmidi is quite CPU heavy compared to everything else :o

You can reduce the CPU usage by reducing the samplerate because then you have more time for calculating the samples. Because FmMidi "synthesizes" the data on the fly you can pass everything and it will work.

Hm, I should add a "GetType" function so you can conditionally check for Midi.

```
// Must be S16 and Stereo, otherwise FmMidi will refuse it
SetFormat(44100 / 2, AudioDecoder::Format::S16, 2)
```

@carstene1ns 
Maybe a lower sample rate can also fix the MIDI issues on the PSP? (when it is a power of 2 of the SDL2 one [per default 44100] SDL2 will resample fine). Minus that SDL2 resample step 22050 would half the CPU usage and 11025 quarter it  :).

Rinnegatamante:
The MIDI playback looks COMPLETELY broken on 3DS even if set as pretty low samplerates like 16000 (used by default by 3ds microphone to record mono audio) or 11025.

Here's a 3dsx build (which uses 11025 even if the warning says 16000) to test it: http://rinnegatamante.it/midibroken.zip

Ghabry:
Could you measure the time in ms how long the "Decode" call takes?

Another issue I'm seeing with fmmidi is that it calls "fgetc" a lot. This is not a problem when the IO is properly buffered but as you know on 3DS this is just broken and slow.

I will try to replace this with "load whole MIDI to memory" (no big deal, MIDIs are usually 10-50 kbyte) and check if memory access-only is faster :)

Rinnegatamante:
As for the pitched problem, nvm, i'm an idiot, forgot to properly update samplerate attribute in DecodeMusic struct <.<.

Anyway, the first Decode call (from OpenAudioDecoder) required only 67 ms O.o.
And it looks like adding a check for short musics like this completely broke the audio (so i'm assuming it "fails" to decode on first call?)

`u64 tick = osGetTime();
	
	// Mono file
	if (audiotype == 1) res=audio_decoder->Decode(Sound->audiobuf, Sound->audiobuf_size);	
	
	// Stereo file
	else{
		u32 chn_size = Sound->audiobuf_size>>1;
		res=audio_decoder->DecodeAsMono(Sound->audiobuf, &Sound->audiobuf[chn_size], chn_size);
	}
	
	tick = osGetTime() - tick;
	Output::Warning("Decode required %llu ms", tick);
	
	// Music is pretty small so we don't need streaming
	if (res < Sound->audiobuf_size){
		Sound->audiobuf_size = res;
		res = 1;
	}else res = 0;`

Ghabry:
Could you paste the whole "OpenAudioDecoder" function somewhere? Maybe I spot some bug.
That it takes the ```if (res < Sound->audiobuf_size){``` codepath is also unlikely o.O

Rinnegatamante:
http://pastebin.com/m5MuaAGh

Ghabry:
Looks totally fine. What is the output of ```Output::Warning("Decode required %llu ms", tick);``` for MP3s?

Ghabry:
Could you test if "sizeof(int_least16_t)" (stdint.h) is 2?

Rinnegatamante:
Decode call for mp3 files take 1016 ms and the check for the possibility to disable streaming broke also title on Ib :/

Ghabry:
okay there is something wrong. The decode call is too fast. Maybe the buffer is too big? \*g\*
Hm, have to do some tests at the PC, this is strange.

Rinnegatamante:
Fixed the check for the streaming disabling but there still are strange behaviours in mp3 and midi is still broken.

Ghabry:
Stupid question but could you verify that WANT_FMMIDI is correctly defined? Just want to exclude trivial errors :D

In audio_decoder.cpp

```
#if WANT_FMMIDI == 1
		return std::unique_ptr<AudioDecoder>(new FmMidiDecoder());
```

Replace this with

```
#if WANT_FMMIDI == 1
		Output::Debug("MIDI"); // or Warning
		return std::unique_ptr<AudioDecoder>(new FmMidiDecoder());
```

Rinnegatamante:
This is the MP3 playback issue i have: https://www.youtube.com/watch?v=3lQtfF-bT58

If you need a MIDI ones too, just tell.

Rinnegatamante:
Adding the MIDI warning in audio_decoder showed it both (i'm currently using the DecodeAsMono since i'm on csnd:SND).

Ghabry:
The video is private :(. For the sound issues I will write again when I'm at a PC with a dev environment.

Rinnegatamante:
Link fixed.

Ghabry:
I bet this error is because "Decode" does not clear the end of the buffer when it writes less bytes then the buffer has space (at the end of the song).
I will fix this bug.

Ghabry:
Could you test if the audio_decoder->Open() call returns true? And I would be interested in the midi sample on youtube :)

Ghabry:
The buffer seems to be too big for FmMidi. Sometimes (especially at the beginning) it somehow returns super quick and writes nothing to the buffer. When you use a smaller buffer size this doesn't happen.

Could you try it with a "normal" buffer size for audio hardware like 8192?

Ghabry:
Fixed the worst memory leaks now (2 MB per MIDI) and 200 KB per minute and looping of MP3s is (almost) gapless now (tested with 10000 Hz "music"). Not possible for fmmidi because it doesn't tell you how much buffer it filled :(

Ghabry:
And for @Rinnegatamante I fixed the strange FM Midi behaviour returning quickly and the first time(s) it is called and keeping the buffer empty...

Ghabry:
Semi OT: Patch for SDL2 to support Windows Native MIDI Pause/Resume: https://bugzilla.libsdl.org/attachment.cgi?id=2251&action=edit

Ghabry:
Jenkins test this please

Ghabry:
Wii build works with MP3 :+1: 
Ready for review.

fdelapena:
jenkins: retest this please <3

Ghabry:
The WMA case is solved: A user commented that this happens when Windows asks you while copying if you want to convert the files and click "yes" (You get a prompt when copying "unknown" files to a MTP device). So I will really add WMA to the file finder and tell the user to not do this ;)

@fdelapena mentioned that the pitching of FM Midi also changes the frequency. Therefore I will disable the pitching for FM Midi and we wait until we have a proper resampler to handle all formats at once.

builds/wii/Makefile
```diff
@@ -7,6 +7,15 @@ ifeq ($(strip $(DEVKITPPC)),)
 $(error "Please set DEVKITPPC in your environment. export DEVKITPPC=<path to>devkitPPC")
 endif
 
+ifeq ($(strip $(TOOLCHAIN_DIR)),)
+$(error "Please point TOOLCHAIN_DIR to your 3ds portlibs directory")
```
carstene1ns: Syncing too much, eh? :grinning:

builds/wii/Makefile
```diff
@@ -27,16 +36,25 @@ INCLUDES	:=	src
...
+				-Wl,--gc-sections
 
 #---------------------------------------------------------------------------------
 # any extra libraries we wish to link with the project
 #---------------------------------------------------------------------------------
 
-LIBS	:=	 -llcf -lexpat -lSDL_mixer -lSDL -lpixman-1 -lpng -lmad \
-			-liconv -lvorbisidec -lz -lfat -lwiiuse -lbte -logc -lm -lwiikeyboard
+LIBS	:=	-llcf -liconv -lpixman-1 -lpng -lz \
+			-lSDL_mixer -lSDL -lvorbisidec -logg -lmpg123 \
```
carstene1ns: ogg is not needed.

src/decoder_fmmidi.cpp
```diff
@@ -0,0 +1,141 @@
...
+	size_t samples = (size_t)length / sizeof(int_least16_t) / 2;
+
+	float delta = (float)samples / (frequency * pitch);
+
+	// FM Midi somehow returns immediately at the beginning when mtime is too small
+	// This increments mtime until FM Midi is happy
+	int notes = 0;
+	do {
+		seq->play(mtime, this);
+		notes = synthesize(reinterpret_cast<int_least16_t*>(buffer), samples, frequency * pitch);
```
fdelapena: Removing the ` * pitch` here makes behave as RPG_RT:

```cpp
		notes = synthesize(reinterpret_cast<int_least16_t*>(buffer), samples, frequency);
```

builds/wii/Makefile
Ghabry: Maybe should replace 3ds with ${PWD##*/} :D

carstene1ns:
jenkins: test this please :+1:

carstene1ns:
jenkins: test this please (now with batteries included)

carstene1ns:
jenkins: test this please 4 REAL

carstene1ns:
jenkins: test this please https://youtu.be/FGBhQbmPwH8

builds/wii/Makefile
```diff
@@ -91,16 +106,19 @@ export INCLUDE	:=	$(foreach dir,$(INCLUDES), -I$(TOPDIR)/$(dir)) \
...
 
 #---------------------------------------------------------------------------------
 # build a list of library paths
 #---------------------------------------------------------------------------------
 export LIBPATHS	:=	$(foreach dir,$(LIBDIRS),-L$(dir)/lib) \
 					-L$(LIBOGC_LIB) \
-					-L$(PORTLIBS)/lib
+					-L$(PORTLIBS)/lib \
+					-L$(LIBLCF_DIR)/lib \
+					-L$(TOOLCHAIN_DIR)/lib
```
carstene1ns: You can just add them to `$LIBDIRS` above (L63).

Rinnegatamante:
Sorry for taking to much time for 3DS adaptation but i've currently started my exams session on university so i'm slowly doing my tests to find what causes the strange behaviours i reported before.

Is updating my local repository worth it for MIDI files? I'm currently on a branch outdated by 7-8 days.

Ghabry:
Hello. I think I solved the strange FmMidi issue in the meanwhile. I suggest that you also try a smaller buffer size (a few kbyte max).

So when you rebase your branch on upstream it is worth a try. But take your time, when you are busy with University this can wait.

Rinnegatamante:
Here's a video showing MIDI issues on 3DS with latest version of the sourcecode: https://www.youtube.com/watch?v=UznVTMPnS4g

(Sorry for the noise but a train decided to pass just when i was recording <.< )

Ghabry:
I'm also busy the next weeks, so communication will be a bit slower ^^. Do you have the sourcecode hosted somewhere? Could you push it in a branch of your fork?

Rinnegatamante:
Got really little time to work on MIDI and i've still not found a solution. What about making a pull for MP3 support (at least) since i'll be super busy till July? I can just "lock" the MIDI support and add a temporary preprocessor flag to debug it.

Ghabry:
Sounds fine, lets do it this way.

PR #877, opened 2016-04-27
3DS: Removed hex_set reference in README.md

Rinnegatamante:
Removed hex_set reference in README.md since it was replaced by a powershell/bash script.


Ghabry:
hm, minor oversight...
Could you "git commit --amend" that commit and replace the last "," with an "and"? (and then push -f)

PR #879, opened 2016-04-28
crash fix in stop_count + utf8 on Windows

Ghabry:
VS2015 wants a BOM now otherwise it will compile utf8 strings incorrectly :(.
This worked in older Version of VS :/

Also fix that div/0 that carstene1ns encountered in UiD

screenshot_0

### DIFF
--- a/src/game_event.cpp
+++ b/src/game_event.cpp
@@ -557,7 +557,10 @@ void Game_Event::MoveTypeRandom() {
 	int last_direction = GetDirection();
 	switch (rand() % 6) {
 	case 0:
-		stop_count -= rand() % stop_count;
+		stop_count -= rand() % (stop_count + 1);
+		if (stop_count < 0) {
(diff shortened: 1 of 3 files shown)


src/game_event.cpp
```diff
@@ -557,7 +557,12 @@ void Game_Event::MoveTypeRandom() {
...
 	switch (rand() % 6) {
 	case 0:
-		stop_count -= rand() % stop_count;
+		if (stop_count > 0) {
+			stop_count -= rand() % stop_count;
+			if (stop_count < 0) {
+				stop_count = 0;
+			}
+		}
 		return;
```
carstene1ns: Just leaving this here in case you did miss it:
`2016-04-24 21:49:05 <zegeri> Ghabry: It should be "stop_count -= rand() % (stop_count + 1);"`

PR #881, opened 2016-04-30
Add built-in FMMIDI support to Autotools

fdelapena:
As requested by Ghabry (#752), an opted-out by default FMMIDI option for Autotools.

The `--enable-*` option differs from `--with-*` because this is a built-in feature, not an external package. That's why I suggest to name it `--enable-fmmidi` without the "builtin", as it feels redundant.

This will need `#ifdef HAVE_FMMIDI` guards (don't forget `#include "system.h"`) where used.

CMake users: Just declare `-DHAVE_FMMIDI` as usual.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -23,6 +23,10 @@ AC_ARG_VAR([EM_GAME_URL], [Game URL/directory (only used for the Emscripten port
 AS_IF([test "x$EM_GAME_URL" != "x"],[
 	AC_DEFINE_UNQUOTED([EM_GAME_URL], ["$EM_GAME_URL"], [Game URL (Emscripten)])
 ])
+AC_ARG_ENABLE([fmmidi], AS_HELP_STRING([--enable-fmmidi], [Enable built-in FM MIDI synthesizer]))
+AS_IF([test "x$enable_fmmidi" = "xyes"], [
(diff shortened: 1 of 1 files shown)


Ghabry:
Iirc someone also suggested a way to fallback to use sdl midi by default and fallback to fmmidi when it fails. But that needs another audio-sdl code change. Not sure if its currently worth it.

carstene1ns:
I also had some commit for it prepared: https://github.com/carstene1ns/easyrpg-player/commit/94fa946432a5a90cc961fe89f630b09ea13accff
We talked about allowing fallback mode for fmmidi if sdl_mixer does not play midi. `WANT_FMMIDI` can be 1 or 2.
Also i really dislike having empty objects compiled and linked in, so i disabled them if unneeded.

carstene1ns:
And a friendly reminder to **guard** the fmmidi sources properly.
Currently it is bloat and will stay bloat, even if someone `-disable`[s]`-fmmidi`.
 (I know you also wrote it above, but this is important IMHO)

Ghabry:
:+1: for your fmmidi want (fallback) and guarding.

fdelapena:
carstene1ns' autotools branch is definitely better, could I discard this PR?

Ghabry:
Okay. I will cherry pick carstene1ns commit into my branch instead.

PR #883, opened 2016-05-02
Android: Add the direct edit of the buttons layout after creation

FaithFeather:
So, that may be just a little mess from me ('cause, i've got some problems when things are not really logical), but for me, it's really impractical to not be able to modify the button layout directly after having create it. Until now, when the users want to create a layout, they need to click on "Add an input layout", create the layout and click again on an other button to edit the layout. With just adding the edit after the creation, you save time and it's like... More practical ? 

(For now, i've only test it with the android emulator 'cause my phone hate me, but i plan to test it with an other phone, though that should work).

### DIFF
--- a/builds/android/app/src/main/java/org/easyrpg/player/SettingsActivity.java
+++ b/builds/android/app/src/main/java/org/easyrpg/player/SettingsActivity.java
@@ -334,7 +334,7 @@ public class SettingsActivity extends Activity {
 					layout.setButton_list(
 							ButtonMappingModel.InputLayout.getDefaultButtonList(getApplicationContext()));
 					mapping_model.add(layout);
-
+					editInputLayout(layout) ;
 					refreshAndSaveLayoutList();
 				}
 			}


Ghabry:
At first I wondered if this is a PR for the editor but you are talking about Android.
This change makes sense to me.

Btw do you have an idea how to make it more obvious that you can change the button layout? Because lots of 1 ★ ratings are because people can't find the Shift button -.-

FaithFeather:
Erf, forgot to precise that was the Android port. My bad.
I don't really understand the problem with that... Is the problem that they don't find the reaaally big icon for edit the name/layout ?

Ghabry:
The users are too stupid to open the settings and scroll down.
Some user still complain that they can't add games from the external storage though we even tell them this when no game is found how to add new directories. m(

FaithFeather:
Mhh... Maybe make a quick tutorial on the first launch, with several steps ? 
Without too much text ('cause if it's too long to read, people just don't read xD)

Ghabry:
Maybe support Google Now. "Hey Google where the fuck is the shift button"

FaithFeather:
Or a different settings menu. One for each : like one for the buttons, one for the sound... Though that may be useless when there's only one thing that can be change.

Ghabry:
Jenkins: Test this please 💗

fdelapena:
To make the button mapping feature easier to reach instead the gear wheel icon we could display a game specific options after tapping every game title in the game list and containing some actions such Play, Add/edit buttons, Change encoding, etc.

Issue #705, opened 2015-12-30
OS X : Double click on the Player file doesn't find game


Medhoc:
When we double click on the easyrpg-player file, it launch the game browser, which doesn't find any game.
We have to execute it with a terminal to launch the game or the game browser (which is abble to find games in this case).

Ghabry:
The game browser finding only works when the browser is one directory above the games currently

BlisterBoy:
There is a misunderstood, if the player is in a game folder, it launches the game browser (not the game) which found no game.
If the player is in a folder containing games, it launch the game browser which found no game.

In all cases, the Player seems to not detect games if we execute it via a double click.
But everything is fine when we execute it with a terminal.

Ghabry:
Could you patch the player to output getcwd() on startup? And then give me the output for both cases (double click and console)

carstene1ns:
The launcher (how is it called again, Finder?) will likely start in the home directory.
I actually had this problem the other way around when packaging mkxp.
It switches to the directory of the executable on startup to read configuration files and assets:
```c++
	/* set working directory */
	char *dataDir = SDL_GetBasePath();
	if (dataDir)
	{
		int result = chdir(dataDir);
		(void)result;
		SDL_free(dataDir);
	}
```
So when installed in `/usr/bin` this makes no sense :grinning:. However, this code might be useful here.

BlisterBoy:
Ho nice ^^.

I've patched the Player as you asked me Ghabry. This is the result.
Execution with terminal :
```/Users/mehdi/Desktop/Dyhortfight2```

Execution by double click : 
```/Users/mehdi```

fdelapena:
An OS X specific patch could compare user homedir with the getcwd, this hack probably will be reliable enough (it will fail to load /home/user/RPG_RT.ldb game directly).
The purpose of this condition is to prevent chdir when launched from command line from a different path.

Ghabry:
Just got reminded of this. I don't like the SDL dependency but is a better short-term solution then trying to get the Objective C compiler working which is required for the MacOS API >.<

Medhoc:
This works like a charm :).

PR #897, opened 2016-05-15
Fix bugs encountered (mostly in battles) in Ara Fell

Ghabry:
This fixes lots of bugs I found while playing through Ara Fell and defeating the first boss. The biggest change was moving most Interpreter commands to the Interpreter (from map) because this breaks battle otherwise.
Many visual Problems were fixed in battle.
If I didn't fail in cut-paste the Interpreter should work as good as before.

The map is here but for some reason when starting the battle it crashes (works locally)
https://easy-rpg.org/play/?game=af_demo&start-map-id=223&start-party=1&test-play

I also fixed a buffer Overflow in vsnprintf which occured in Spanish Dragonball Z because the map tree was too deep.
And double clicking in the Finder on MacOS should work now...

Not ready yet. I need somebody to test it on a Mac ;)

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -47,6 +47,7 @@
 #include "player.h"
 #include "util_macro.h"
 #include "reader_util.h"
+#include "game_battle.h"
 
 Game_Interpreter::Game_Interpreter(int _depth, bool _main_flag) {
 	depth = _depth;
@@ -1343,6 +1344,10 @@ bool Game_Interpreter::CommandChangeHP(RPG::EventCommand const& com) { // Code 1
 			amount += hp * (-1) + 1;
 		}
 		actor->ChangeHp(amount);
+
+		if (actor->IsDead()) {
+			Game_Battle::SetNeedRefresh(true);
+		}
 	}
 
 	if (lethal) {
@@ -1379,6 +1384,7 @@ bool Game_Interpreter::CommandChangeCondition(RPG::EventCommand const& com) { //
 		} else {
 			if(state_id == 1) {
 				actor->ChangeHp(-actor->GetHp());
+				Game_Battle::SetNeedRefresh(true);
 			}
 			actor->AddState(state_id);
 			CheckGameOver();
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -255,7 +255,7 @@ void Game_Interpreter::SetupStartingEvent(Game_CommonEvent* ev) {
 }
 
 void Game_Interpreter::CheckGameOver() {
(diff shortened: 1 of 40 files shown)


src/game_interpreter.cpp
```diff
@@ -907,6 +908,46 @@ bool Game_Interpreter::CommandControlVariables(RPG::EventCommand const& com) { /
...
+				break;
+			case 6:
+				// Enemy Spirit
+				value = (*Main_Data::game_enemyparty)[com.parameters[5]].GetSpi();
+				break;
+			case 7:
+				// Enemy Agility
+				value = (*Main_Data::game_enemyparty)[com.parameters[5]].GetAgi();
+				break;
+			}
```
Ghabry: Look what we forgot :o

Ghabry:
Regression in EHR is fixed.

The only problem I'm aware of in Ara Fell is that the "Scout" battle command is behaving totally wrong :/

src/graphics.cpp
```diff
@@ -227,6 +227,11 @@ void Graphics::Freeze() {
 }
 
 void Graphics::Transition(TransitionType type, int duration, bool erase) {
+	if (erase == screen_erased) {
+		// Don't allow another erase or non-erase
+		return;
+	}
+
```
Zegeri: This doesn't reproduce the behaviour of RPG_RT when using Show Screen in a non-erased screen. For example, try using Show Screen with the transition type Scroll Right.

Ghabry:
Enough fixes for this PR. The next commits will fix regressions.

The first boss can be defeated now. That boss absorbs other enemies to generate a "shield". The Interpreter checked if all enemy HP was 0, then the real boss can be attacked. Due to different bugs the sum was never 0.

ForceFlee is not needed, see it as a bonus ;)

fdelapena:
First regression found: battler sprites are the same for all allied in EHR:
The allied character is the same for all

Force flee will unbreak multiple games, nice bonus :+1:.

Ghabry:
RPG2k3 supports many ways to do animations.

1. The "right hand" animation fades out the battler and then it stays at an empty frame (old behaviour: it was looping or stuck at the last non-empty frame)
2. The battle animation plays
3. Not implemented: A "fade-out" animation is played (is an extra per-actor setting), not used by many games but Ara Fell does. This makes Lita's attacks look a bit strange.
4. The idle (or state) animation plays for the battler

Ghabry:
Got confirmation that #705 is fixed

Ghabry:
Fixed the reported regressions and solved more (easy to resolve) issues I encoutered.

Ghabry:
\*pfew\* this is ready now :+1:

Battle should be a bit better now, also in other games ;)

PR #899, opened 2016-05-24
Some minor changes on layout and a function (v2) 

sushi24:
Hi everybody, 
here's a new pull request with my previous correction. I hope this time everything is alright! ^^


Ghabry:
I disabled the "squash and merge" Option a while ago to prevent doing it by accident but here it would make sense...

PR #907, opened 2016-06-07
3DS: Added MP3 support and debug code for MIDI tracks.

Rinnegatamante:
3DS: Added MP3 support and debug code for MIDI tracks.

### DIFF
--- a/src/3ds_decoder.cpp
+++ b/src/3ds_decoder.cpp
@@ -33,7 +33,7 @@
 #include "3ds_decoder.h"
 #endif
 extern DecodedMusic* BGM;
-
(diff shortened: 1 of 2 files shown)


Ghabry:
Jenkins: Test this please

Sorry for the delay. I want to test it quickly on hardware but had no time yet during the workweek >.< but tomorrow is weekend

Rinnegatamante:
Apparently there are problems in MP3 code: http://rinnegatamante.it/board/showthread.php?tid=89&pid=289#pid289

On CIA build it will cause a crash, 3DSX will cause a freeze.
It means something weird happens (like an out of ram or some other unexpected behaviours).

PR #908, opened 2016-06-11
Added optional dependency to libsndfile and libspeexdsp

ChristianBreitwieser:
Hello,

As explained before in the [forum](https://easy-rpg.org/forums/viewtopic.php?f=4&t=633)  these are two additions to the audio layer of the player:
 
- decoder_libsndfile to enable wav playback without sdl (depends on libsndfile)
- audio_resampler to enable arbitrary resampling (depends on libspeexdsp, or libsamplerate)

The AudioResampler is meant to wrap any AudioDecoder and intersects if SetFormat or SetPitch are called.

I haven't changed anything in the sdl audio files so these changes only affect background music as sound effects don't use the AudioDecoder class.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -31,6 +31,8 @@ libeasyrpg_player_la_SOURCES = \
 	src/cache.h \
 	src/color.cpp \
 	src/color.h \
+	src/decoder_libsndfile.cpp \
+	src/decoder_libsndfile.h \
 	src/decoder_mpg123.cpp \
 	src/decoder_mpg123.h \
 	src/decoder_fmmidi.cpp \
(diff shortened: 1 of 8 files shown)


Ghabry:
Jenkins: Test this please

For Sound effects it's fine. SDL is able to play most of them. We only get resampler issues when streaming data because sounds are played in one go but no music.

Thanks, looks really great. Much better then the code I have to review at my company :+1: Going to test it :)

This will need adjustments for Windows (Visual Studio) but I will do this with another pull request after merging this one because it's ugly ;)

Ghabry:
Works perfect in SDL, even with nonsense rates like 10000 or 23456

ChristianBreitwieser:
Thanks for the kind reply!
The block seperating comments are removed.

carstene1ns:
btw. is there a reason you guarded only the internal stuff from the libraries in `audio_resampler.h` and `decoder_libsndfile.h`? It will not link without the libs (if using the functions) anyway, so it would make sense to empty the files when the preprocessor macros are not defined imo.

ChristianBreitwieser:
No there is no "real" reason - I just tried to follow the style of the two existing decoders (fmmidi and mp123).

In fact the header files aren't used anyway if the macros aren't defined. The inclusion in `audio_decoder.cpp` is macro dependent as well.

Should i change it?

Ghabry:
Finnally found time to test the tempo. Works great :+1:

Ghabry:
Jenkins: Test this please

PR #912, opened 2016-06-19
Don't half fmmidi frequency twice

fdelapena:
The FMMIDI frequency was already being halved from 44100 to 22050. The line removed was halving MIDI again (22050 to 11025 Hz), doing FMMIDI playback too much phone-like.

The half frequency in FFMIDI is for performance purposes on low end ports (used for synthesizing, saves a lot of CPU), not needing to be resampled later (which spends additional CPU).


Ghabry:
@ChristianBreitwieser

We remove this line because SDL code and 3ds code already handles this and this line makes it more complecated because the behaviour depends on the existance of a resampler.

ChristianBreitwieser:
Sorry for putting it in in the first place.

PR #913, opened 2016-06-19
Fix #703, #884

### LINKED ISSUE #703: Pom Gets Wi-Fi: Caterpillar system not working properly
[Game](http://rpgmaker.net/games/5441/). [Savegame](https://fran.cr/owncloud/index.php/s/RJ3WQ6sD8jz2g2M/download).
This game has a caterpillar system (events following hero steps) not behaving as in RPG_RT. This makes the player movement hard to do because events are blocking the hero when trying to walk in horizontal axis. Also, the hero character is moving some frames faster than events following it.

### LINKED ISSUE #884: Order of updates of events that walk into each other is incorrect
Here's to document some really crazy behaviour in RPG_RT.
Imagine you have some events lined up. Each of them has a move route that makes them walk to the position of the event right next to it, like here:
captura de pantalla_2016-05-02_20-23-35
In RPG_RT, they all walk together nicely **without colliding into each other**. While in Player some of them will collide with the one right next to it for a frame and create enought space between them to walk. If the events have IDs 1, 2, 3 and 4 and they're in the map lined up horizontally in that order from left to right. Player will update first the event 1, that will try to walk right and collide with 2. 2 will collide with 3 and 3 with 4. 4 can walk, so it'll be slightly ahead of the rest of them. In the next frame, 3 gets slightly ahead of 1 and 2. And so on.
captura de pantalla_2016-05-02_20-38-29
(Image from Player. Up: From left to right, events 1, 2, 3, 4 collide as described. Down: From left to right, events 4, 3, 2, 1 don't collide)
Here's the interesting part, in RPG_RT it doesn't matter the order of events, they all walk always together. Ghabry's eventtracer is again useful to tell what's happening. Let's use the followind ordering of events: 2, 1, 4, 3. The tracer reports that the order of execution the move commands has been 1, 4, 3, 2. Weird, we expected 1, 2, 3, 4. After trying other permutations it can be deduced that **when an event checks the passability of a tile, it executes the move route of the events that are in that tile to see if they move out of the way**. So in that last example, 1 is causing 4 to update because he's in the way, which is causing 3 to update because he's in the way. 2 isn't blocking anyone, so he gets updated as soon the tree of updates that 1 leads ends.
de
Crazy stuff.

[Download a test sample](https://dl.dropboxusercontent.com/u/30277047/TestEventsOrder.7z). This bug will be fixed when the two lines of events walk with the same spacing between them.

scurest:
Also fixes [gadesx's issue with character movement](https://community.easyrpg.org/t/problem-moving-various-events-at-time/260).

out

It does _not_ fix #884 in the case that one of the characters is a vehicle. I didn't even test whether a vehicle will trigger early updates (but it seems likely). Will wait for a testcase for that.

### DIFF
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -271,7 +271,14 @@ void Game_Character::MoveTypeCustom() {
 	if (IsStopping()) {
 		move_failed = false;
 
-		for (; (size_t)active_route_index < active_route->move_commands.size(); ++active_route_index) {
+		int original_index = active_route_index;
+		bool looped_around = false;
+		while (true) {
+			if (active_route_index == original_index && looped_around) {
+				// We've gone around a full loop; stop here
+				break;
+			}
+
 			if (!IsStopping() || wait_count > 0 || stop_count < max_stop_count)
 				break;
 
@@ -406,10 +413,15 @@ void Game_Character::MoveTypeCustom() {
 			if (move_failed) {
 				if (active_route->skippable) {
 					last_move_failed = false;
-					continue;
+				} else {
+					break;
 				}
+			}
 
-				break;
+			++active_route_index;
+			if ((size_t)active_route_index >= active_route->move_commands.size()) {
+				looped_around = true;
+				active_route_index = 0;
 			}
 		}
 
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -254,8 +254,18 @@ void Game_Character::MoveTypeCustom() {
 		int original_index = active_route_index;
 		bool looped_around = false;
 		while (true) {
+			if ((size_t)active_route_index >= active_route->move_commands.size()) {
+				if (active_route->repeat) {
+					looped_around = true;
+					active_route_index = 0;
+					SetMoveRouteRepeated(true);
+				} else {
+					break;
+				}
+			}
+
 			if (active_route_index == original_index && looped_around) {
-				// We've gone around a full loop; stop here
+				// We've gone around a full loop; stop here for now
 				break;
 			}
 
@@ -399,18 +409,10 @@ void Game_Character::MoveTypeCustom() {
 			}
 
 			++active_route_index;
-			if ((size_t)active_route_index >= active_route->move_commands.size()) {
-				looped_around = true;
-				active_route_index = 0;
-			}
 		}
 
 		if ((size_t)active_route_index >= active_route->move_commands.size() && IsStopping()) {
-			// End of Move list
-			if (active_route->repeat) {
-				active_route_index = 0;
-				SetMoveRouteRepeated(true);
-			} else if (IsMoveRouteOverwritten()) {
+			if (IsMoveRouteOverwritten()) {
 				CancelMoveRoute();
 				Game_Map::RemovePendingMove(this);
 				stop_count = 0;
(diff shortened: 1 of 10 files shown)


carstene1ns:
random rant: I am in for using c++11 where it makes sense and is possible.
For example using `auto` for temporary variables when checking our own liblcf types that have long names is okay.
However, when the `Player::GetFrames()` function returns an `int` and your `frame_count_at_last_update_parallel` is also an `int`, using `auto` here is a bit too much.

scurest:
I can definitely change it, but the reason I used it there is that the code is quite agnostic to what type exactly `GetFrames()` returns, so long as it's a fresh value every frame.

Should I change [`auto tile_id = other.GetTileId();`](https://github.com/EasyRPG/Player/pull/913/files#diff-488e769e6277b870601b99f538eadcdaR488) too?

Ghabry:
imo, it's an unwritten rule in the codebase to only use "auto" when it's some work to write (like iterator or some range-based for). For primitive types we always use the type and not auto.
Would be nice to have this adjusted.

scurest:
Got it. Fixed.

Ghabry:
@zegeri, you are the map interpreter expert. Could you take a look at this when you have some time? :)

Zegeri:
After b3026d2, a Wait move command won't do anything.

scurest:
Ah. I see what you mean. But `Wait` commands are only generated by custom move routes, and [the check here](https://github.com/scurest/Player/blob/4b5295adaf0f9cc85e5ac42407955e43a441f2b3/src/game_character.cpp#L272) actually prevents the next move command from running until the wait is over, so they still seem to work. I see your point though. Should I

    -    if (stop_count >= max_stop_count) {
    +    if (wait_count == 0 && stop_count >= max_stop_count) {

?

(I notice, now, I've carelessly left a useless `return` in that function too.)

Zegeri:
You're right, the wait command will work, but in the very specific case where a repeating move route ends with a Wait Command, the Move Route Repeated flag will be set to true 20 frames earlier because it's [before that check](https://github.com/scurest/Player/blob/4b5295adaf0f9cc85e5ac42407955e43a441f2b3/src/game_character.cpp#L261). That can have consequences if there's a Proceed With Movement running anywhere. I'm afraid I can't check this properly for a few days. 

Other than that, this branch looks great to me.

scurest:
It's exactly like you said :) Current `master` doesn't pause for the wait duration at the end of a repeating move route during a Proceed With Movement.

Zegeri:
Looks good to me.

fdelapena:
It seems like there is a regression, the main thread locks up in possibly related stuff. Found when loading a savegame done in previous master and even reproduceable on a new fresh game after returning from a save scene:

http://ftp.vector.co.jp/62/45/3309/sou_1.03.zip

Japanese game, unpack with `unar` is recommended.

It will hang after saving, the first sequence is mostly automatic to reach this hang. It was not hanging before.

scurest:
Yeah, my bad, guys :( I would have caught this if I'd tested that last commit properly.

The problem is I swapped

    wrap the index to the beginning of the list if we're past the end
    check if we've come back to the original index

for the case where the original index was one-past-the-end. But the other order 

    check if we've come back to the original index
    wrap the index to the beginning of the list if we're past the end

breaks it in the very common case that the original index is 0. So I fixed it with a rather ugly duplicated condition check.

Um, I think I need to open a new PR, right?

PR #921, opened 2016-07-07
Class handling

Ghabry:
Classes are handled silly in RPG Maker 2003. The class settings don't matter at all on startup (it only shows the class name in the status scene), everything else is only applied when "Change Class" event is used...

The naming in liblcf is really inconsistent what do you think about adjusting this?

mighty_guard <-> super_guard
lock_equipment <-> fix_equipment
two_weapon <-> two_swords_style

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -119,7 +119,7 @@ bool Game_Actor::IsSkillUsable(int skill_id) const {
 
 	// Actor must have all attributes of the skill equipped as weapons
 	const RPG::Item* item = GetEquipment(0);
-	const RPG::Item* item2 = GetTwoSwordsStyle() ? GetEquipment(1) : nullptr;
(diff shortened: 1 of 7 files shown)


carstene1ns:
Is sword the only weapon that can be used two-handed? I guess no, so weapon should be the right term here (and not caring, whatever rpg maker uses here, their translation is not good in some other cases too).

Ghabry:
The problem here is that liblcf has two different names for the same things depending on if you look in RPG::Actor, RPG::Class or RPG::SaveActor.

two_weapon is obviously better then two_swords_style, bad translation.

Ghabry:
test

carstene1ns:
jenkins: test this please

carstene1ns:
Do you want to rename `bool Game_Actor::GetTwoSwordsStyle()` too or should it stay this way?

Ghabry:
good point, gimme a moment

carstene1ns:
looks good to me now, merging after jenkins is ready.

PR #923, opened 2016-07-09
Additional Audio decoders

carstene1ns:
What this does in detail:
- adds another decoder for midi (based on libwildmidi)
  - sounds mostly like timidity from `sdl_mixer`, but supports pitch changes through our `audio_resampler`
  - disabled by default (needs additional magic to find configuration file)
- adds an ogg vorbis decoder (based on either libtremor or libogg, libvorbis and libvorbisfile)
- can use fallback decoders, when initializing a library fails
  - currently only works for mpg123 and wildmidi
  - ogg can be played by `libsndfile` if the oggvorbis decoder is unavailable
  - mp3 will be played by `sdl_mixer` (if smpeg/libmad is available), if mpg123 fails
- reworked the other `audio_decoder`s a bit to fix some coding style and hide all internal structures.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -39,6 +39,8 @@ libeasyrpg_player_la_SOURCES = \
 	src/decoder_mpg123.h \
 	src/decoder_fmmidi.cpp \
 	src/decoder_fmmidi.h \
+	src/decoder_wildmidi.cpp \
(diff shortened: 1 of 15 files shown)


Ghabry:
Looks good to me :+1:

PR #926, opened 2016-07-10
Fix regression when original_index = 0

scurest:
Fixes the regression in 葬 (and probably many others :) from #913.

(Sorry!)

### DIFF
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -266,6 +266,9 @@ void Game_Character::MoveTypeCustom() {
 					looped_around = true;
 					active_route_index = 0;
 					SetMoveRouteRepeated(true);
+					if (original_index == 0) {
+						break;
+					}
 				} else {
 					break;
 				}


src/game_character.cpp
```diff
@@ -266,6 +266,9 @@ void Game_Character::MoveTypeCustom() {
 					looped_around = true;
 					active_route_index = 0;
 					SetMoveRouteRepeated(true);
+					if (original_index == 0 && looped_around) {
```
carstene1ns: You set `looped_around` to true directly above. Seems like this check is not needed then.

scurest: Oops.

Ghabry:
Jenkins: Test this please

Ghabry:
Will update the compiler to "Update 3", maybe that helps o.O

Ghabry:
Jenkins: Test this please

PR #931, opened 2016-07-13
Fix #928

### LINKED ISSUE #928: Airship vehicle shouldn't trigger battle encounters and other glitches
According to multiple reports, battle encounters shouldn't be triggered when boarded. Only with the airship vehicle, other vehicles (boat and ship) are OK.

Also, it seems like airship vehicle shouldn't have sprite animation when not boarded.

For all vehicles: the sprite does some weird movement when moving it on map, it looks like some incorrect graphics update.

scurest:
* The flickering motion was caused by a double update. `remaining_steps` got updated by the player, the player synced the airship, and then the airship updated `remaining_steps` again. When the vehicle moves fast enough, this made a visible difference during motion. Fixed with a second sync.
* In  	6768eba, I'm not sure if the brackets should go around `pattern != original_pattern` or not. When does that case occur?

### DIFF
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -186,7 +186,7 @@ void Game_Character::Update() {
 			anime_count++;
 	} else {
 		stop_count++;
-		if (IsSpinning() || IsContinuous() || pattern != original_pattern)
+		if ((walk_animation && (IsSpinning() || IsContinuous())) || pattern != original_pattern)
 			anime_count++;
 	}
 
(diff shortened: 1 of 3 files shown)


Zegeri:
>I'm not sure if the brackets should go around pattern != original_pattern or not. When does that case occur?

When a movement has ended, but the pattern isn't the middle frame. In RPG_RT, stopping the animation while on a step doesn't freeze the pattern until the pattern is again the middle frame, so the bracket shouldn't go around ```pattern != original_pattern```.

PR #932, opened 2016-07-13
Add missing #include <cassert>

scurest:
...which fixes the CMake build.


Ghabry:
The CMake build really needs some love and maybe a PR builder @fdelapena & @carstene1ns ?

PR #942, opened 2016-07-19
Fix regression introduced in 81c910a (SP cost calculation)

Ghabry:
see commit description

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -638,8 +638,11 @@ void Game_BattleAlgorithm::Normal::Apply() {
 	AlgorithmBase::Apply();
 
 	source->SetCharged(false);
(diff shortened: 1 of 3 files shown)


src/game_battlealgorithm.cpp
```diff
@@ -639,7 +639,8 @@ void Game_BattleAlgorithm::Normal::Apply() {
 
 	source->SetCharged(false);
 	if (source->GetType() == Game_Battler::Type_Ally && static_cast<Game_Actor*>(source)->GetWeaponId() != 0) {
-		source->ChangeSp(-source->CalculateSkillCost(static_cast<Game_Actor*>(source)->GetWeaponId()));
+		Game_Actor* src = static_cast<Game_Actor*>(source);
```
carstene1ns: I think you can move the cast two lines up an use `src` already in the condition.

PR #943, opened 2016-07-21
misc fixes

carstene1ns:


src/sdl_ui.cpp
```diff
@@ -171,12 +171,17 @@ SdlUi::SdlUi(long width, long height, bool fs_flag) :
...
-	audio_.reset(new EmptyAudio());
+	if (!Player::no_audio_flag) {
+		audio_.reset(new ALAudio());
+		return;
+	}
 #endif
+	audio_.reset(new EmptyAudio());
 }
 
 SdlUi::~SdlUi() {
```
Ghabry: Will this also work for the case that no audio hardware is present but no disable-audio argument is passed?

carstene1ns: No, to check for that you would need to initialize SDL with audio and fallback to no-audio, if it fails.

src/registry.cpp
```diff
@@ -35,15 +35,27 @@
...
 	char value[1024];
 	DWORD size = 1024;
 	DWORD type = REG_SZ;
 	HKEY key_handle;
+	REGSAM desired_access = KEY_QUERY_VALUE;
+
+	switch (view) {
+		case KEY32:
+			desired_access |= KEY_WOW64_32KEY;
+			break;
```
Ghabry: I think it wold be enough to only use KEY_WOW64_32KEY.

Because the RTP is always installed in the 32bit path.

carstene1ns: Well, this is to be future proof, who knows what kind of registry keys we want to access in the future.

Ghabry: Yeah sorry I didn't check the other code which always invokes with KEY32. Nevermind.

src/game_map.cpp
```diff
@@ -1329,17 +1329,13 @@ void Game_Map::UpdateParallax() {
 		return;
 
 	if (map_info.parallax_horz && map_info.parallax_horz_auto) {
-		parallax_x +=
-			map_info.parallax_horz_speed > 0 ? 1 << (map_info.parallax_horz_speed - 1) :
-			map_info.parallax_horz_speed < 0 ? -1 << (-map_info.parallax_horz_speed - 1) :
-			0;
+		int acc = 1 << (std::abs(map_info.parallax_horz_speed) - 1);
```
scurest: If `map_info.parallax_horz_speed` is zero, the RHS of the left shift is negative. I don't know whether this case occurs or not, but the old code handles it.

carstene1ns: While testing this never happened, I need to find the exact code, but remember removing the check for `0`, because it is not needed.

carstene1ns: I guess `parallax_horz` and/or `parallax_horz_auto` was always false and so a speed of `0` does not exist.

Well, i will just add the check back in, in the condition above, as it does not hurt to have.

Ghabry: I think a value of 0 is possible, at least you can enable auto speed in the settings and provide a speed of 0.

carstene1ns:
@scurest: Fixed your remark, thanks!

Ghabry:
Actually, as documented on msdn, HKCU\Software is really not redirected at all. So this doesn't alter anything on Windows, always worked, even on 64bit (just tested it). But Wine does redirect? o.O

carstene1ns:
Yeah, at least older versions did. However, this will not fix it when using them.

PR #952, opened 2016-07-25
Update CMake File

Ghabry:
This synchronizes the CMake file with the features of the autoconf file and adds a shiny summary table :)

I also tried to simplify the code (400 lines *cough*)　and removed some features I don't see a use for. Wine and Shinonome are gone, Wine test makes no sense and Shinonome can be manually triggered if you really need a recompile.


CMakeLists.txt
```diff
@@ -5,66 +5,26 @@ set(CMAKE_CXX_STANDARD_REQUIRED on)
...
 set(LIBRARY_OUTPUT_PATH ${CMAKE_CURRENT_BINARY_DIR}/lib)
-#link_directories(${LIBRARY_OUTPUT_PATH})
-#cmake_policy(CMP0015 NEW)
 
 list(APPEND CMAKE_MODULE_PATH
-	"${CMAKE_CURRENT_SOURCE_DIR}/lib/liblcf/builds/cmake/Modules"
 	"${CMAKE_CURRENT_SOURCE_DIR}/builds/cmake/Modules")
 
+if(CMAKE_BUILD_TYPE MATCHES "Debug")
+	add_definitions(-D _DEBUG=1)
```
carstene1ns: style: remove space before definition

CMakeLists.txt
```diff
@@ -5,66 +5,26 @@ set(CMAKE_CXX_STANDARD_REQUIRED on)
...
+endif()
+	
 add_definitions(
   ${PNG_DEFINITIONS}
 
   -D USE_SDL=1)
-
+
```
carstene1ns: style: space

CMakeLists.txt
```diff
@@ -73,54 +33,175 @@ if(NOT DEFINED CMAKE_TOOLCHAIN_FILE)
...
+# Sound system to use
+set(PLAYER_AUDIO_BACKEND "SDL2_mixer" CACHE STRING "Audio system to use. The SDL2_mixer audio system provides advanced features required by RPG Maker. Options: SDL2_mixer OpenAL OFF")
+set_property(CACHE PLAYER_AUDIO_BACKEND PROPERTY STRINGS SDL2 OpenAL)
+
+# Configure Audio backends
+
+if(${PLAYER_AUDIO_BACKEND} MATCHES "SDL2_mixer")
+	# SDL2_mixer Audio
+	find_lib_req(SDL2_mixer ON)
+	add_definitions(-D HAVE_SDL_MIXER=1)
```
carstene1ns: style: remove space before definition

CMakeLists.txt
```diff
@@ -73,54 +33,175 @@ if(NOT DEFINED CMAKE_TOOLCHAIN_FILE)
...
+		add_definitions(-DHAVE_WILDMIDI)
+	endif()
+	
+	# Provide fmmidi options
+	set(PLAYER_ENABLE_FMMIDI "Fallback" CACHE STRING "Enable internal MIDI sequencer. Fallback (default) mode will use it when the external MIDI library fails. Options: Fallback ON OFF")
+	set_property(CACHE PLAYER_ENABLE_FMMIDI PROPERTY STRINGS Fallback ON OFF)
+	if(${PLAYER_ENABLE_FMMIDI} MATCHES "Fallback")
+		add_definitions(-DWANT_FMMIDI=2)
+	elseif(${PLAYER_ENABLE_FMMIDI} MATCHES "ON")
+		add_definitions(-DWAND_FMMIDI=1)
```
carstene1ns: typo

CMakeLists.txt
```diff
@@ -73,54 +33,175 @@ if(NOT DEFINED CMAKE_TOOLCHAIN_FILE)
...
+
+	# libsndfile
+	option(PLAYER_WITH_LIBSNDFILE "Play WAV audio with libsndfile." ON)	
+	find_lib("SndFile" PLAYER_WITH_LIBSNDFILE)
+	if(${RETVAL})
+		add_definitions(-DHAVE_LIBSNDFILE)
+	endif()
+	
+	# libogg
+	option(PLAYER_WITH_LIBOGG "Play OGG audio with libogg or tremor." ON)
```
carstene1ns: This does only care for Ogg/Vorbis for now, so removing libtremor makes sense IMHO

CMakeLists.txt
```diff
@@ -73,54 +33,175 @@ if(NOT DEFINED CMAKE_TOOLCHAIN_FILE)
...
+	# libogg
+	option(PLAYER_WITH_LIBOGG "Play OGG audio with libogg or tremor." ON)	
+	find_lib("OggVorbis" PLAYER_WITH_SPEEXDSP)
+	if(${RETVAL})
+		add_definitions(-DHAVE_OGGVORBIS)
+	endif()
+	
+	# wildmidi (optional, OFF)
+	option(PLAYER_WITH_WILDMIDI "Play MIDI audio with wildmidi. (optional)" OFF)
+	find_lib("wildmidi" PLAYER_WITH_WILDMIDI)
```
carstene1ns: this will likely not work, as the library name is mixed case (tested?)

CMakeLists.txt
```diff
@@ -73,54 +33,175 @@ if(NOT DEFINED CMAKE_TOOLCHAIN_FILE)
...
+	endif()	
+elseif(${PLAYER_AUDIO_BACKEND} MATCHES "OpenAL")
+	# OpenAL Audio
+	add_definitions(-DNO_SDL_MIXER)
+	find_package(OpenAL REQUIRED)
+	find_package(SndFile REQUIRED)
+	find_package(FluidSynth REQUIRED)
+	include_directories(${OPENAL_INCLUDE_DIR} ${SNDFILE_INCLUDE_DIR} ${FLUIDSYNTH_INCLUDE_DIR})
+	list(APPEND EASYRPG_PLAYER_LIBRARIES ${OPENAL_LIBRARY} ${SNDFILE_LIBRARIES} ${FLUIDSYNTH_LIBRARY})
+	add_definitions(-D HAVE_OPENAL=1)
```
carstene1ns: style: remove space before definition

builds/cmake/Modules/FindOggVorbis.cmake
```diff
@@ -0,0 +1,10 @@
+find_path(OGGVORBIS_INCLUDE_DIR ogg/ogg.h)
+find_library(OGGVORBIS_LIBRARY NAMES ogg libogg)
```
carstene1ns: the required lib is vorbisfile (which provides the same high-level api as tremor), so this will not work correctly

builds/cmake/Modules/Findwildmidi.cmake
```diff
@@ -0,0 +1,10 @@
+find_path(WILDMIDI_INCLUDE_DIR wildmidi_lib.h)
+find_library(WILDMIDI_LIBRARY NAMES wildmidi libwildmidi WildMidi libWildMidi)
+
+include(FindPackageHandleStandardArgs)
```
carstene1ns: wildmidi has currently no pkg-config support (but i want to fix that upstream so letting it stay :D)

CMakeLists.txt
Ghabry: This works because find_library uses that argument to open "Findwildmidi.cmake" but I can rename both to be consistent with the lib name (WildMidi)

CMakeLists.txt
```diff
@@ -130,45 +211,50 @@ aux_source_directory(${CMAKE_CURRENT_SOURCE_DIR}/src SRCS)
...
+	if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/resources/${MAN_NAME})
+		install(FILES resources/${MAN_NAME} DESTINATION share/man/man6)
+	endif()
 endif()
 
 # CPack
 set(CPACK_GENERATOR "ZIP" "TGZ")
 if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
-  list(APPEND CPACK_GENERATOR "NSIS")
+	list(APPEND CPACK_GENERATOR "NSIS")
```
carstene1ns: generating a windows installer for the portable player was always a bit off IMO, so maybe we should remove that option now

CMakeLists.txt
```diff
@@ -73,54 +33,175 @@ if(NOT DEFINED CMAKE_TOOLCHAIN_FILE)
...
+	add_definitions(-DNO_SDL_MIXER)
+	find_package(OpenAL REQUIRED)
+	find_package(SndFile REQUIRED)
+	find_package(FluidSynth REQUIRED)
+	include_directories(${OPENAL_INCLUDE_DIR} ${SNDFILE_INCLUDE_DIR} ${FLUIDSYNTH_INCLUDE_DIR})
+	list(APPEND EASYRPG_PLAYER_LIBRARIES ${OPENAL_LIBRARY} ${SNDFILE_LIBRARIES} ${FLUIDSYNTH_LIBRARY})
+	add_definitions(-D HAVE_OPENAL=1)
+elseif(${PLAYER_AUDIO_BACKEND} MATCHES "OFF")
+	add_definitions(-DNO_SDL_MIXER)
+elseif()
```
carstene1ns: style: should this not be `else()` ?

Ghabry:
@carstene1ns 
This is again open for review. But before merging I think I will squash some commits together

tests/filefinder.cpp
```diff
@@ -6,33 +6,33 @@
...
+		FileFinder::SetDirectoryTree(tree);
 	}
-*/
+
 	void CheckIsDirectory() {
 		assert(FileFinder::IsDirectory("."));
 	}
 
 	void CheckEnglishFilename() {
-		assert(!FileFinder::FindImage("Backdrop", "castle").empty());
```
fdelapena: The old test was doing a "touch" command for a RTP castle filename written in Japanese, maybe TestGame-2000 should have some filename in Japanese or vice-versa (use castle in Japanese in this string), to make this function name still meaningful. Or just rename the function :stuck_out_tongue_winking_eye:.

Ghabry: Well it searches for a file that has a English name :D. But I could change it to a japanese -> english lookup

carstene1ns:
Unrelated minor visual issue: the sdl2_mixer version stuff does not work, [these definitions](https://github.com/EasyRPG/Player/blob/master/builds/cmake/Modules/FindSDL2_mixer.cmake#L111-L113) need to be changed as the header still uses the old version macros (without `2`).

CMakeLists.txt
```diff
@@ -5,122 +5,202 @@ set(CMAKE_CXX_STANDARD_REQUIRED on)
...
-  -D USE_SDL=1)
+if(CMAKE_BUILD_TYPE MATCHES "Debug")
+	add_definitions(-D_DEBUG=1)
+endif()
+	
+add_definitions(-DUSE_SDL=1)
 
 if(CMAKE_GENERATOR MATCHES "Makefile")
-  add_definitions(-fno-rtti)
+	add_definitions(-fno-rtti)
```
carstene1ns: This should actually be added to `CXXFLAGS`.

CMakeLists.txt
```diff
@@ -5,122 +5,202 @@ set(CMAKE_CXX_STANDARD_REQUIRED on)
...
 
-if(CMAKE_BUILD_TYPE MATCHES "Debug")
-  add_definitions(-D _DEBUG=1)
+# Detect all required libraries
+foreach(i Pixman ZLIB PNG SDL2)
+	find_lib_req(${i} ON)
+endforeach()
+
+# Always enable Wine registry support on non-Windows
+if(CMAKE_SYSTEM_NAME MATCHES "Windows")
```
carstene1ns: Why not `if(NOT <...>)``

carstene1ns:
IMO the "line" status messages are not needed and look bad as the summary is separated by empty lines anyway and the length of them is arbitrary, but this may be personal preference.

CMakeLists.txt
```diff
@@ -5,122 +5,202 @@ set(CMAKE_CXX_STANDARD_REQUIRED on)
...
+	set(RETVAL ${RETVAL} PARENT_SCOPE)
+endfunction()
+
+function(find_lib LIBNAME COND)
+	find_lib_int(${LIBNAME} ${COND} OFF)
+	
+	# Export _FOUND variable to parent scope which is global scope
+	set(${LIBNAME}_FOUND ${${LIBNAME}_FOUND} PARENT_SCOPE)
+	set(EASYRPG_PLAYER_LIBRARIES ${EASYRPG_PLAYER_LIBRARIES} ${NEW_LIBS} PARENT_SCOPE)
+	set(RETVAL ${RETVAL} PARENT_SCOPE)
```
carstene1ns: Why is `RETVAL` needed? Seems like just for convenience, because the status could be queried by using `${LIBNAME}_FOUND` at higher level too.

CMakeLists.txt
```diff
@@ -130,103 +210,174 @@ aux_source_directory(${CMAKE_CURRENT_SOURCE_DIR}/src SRCS)
...
-endforeach()
+option(PLAYER_ENABLE_TESTS "Execute unit tests after compilation finishes" ON)
+
+if(PLAYER_ENABLE_TESTS)
+	enable_testing()
+
+	set(TEST_GAME_REPOSITORY_PATH "${CMAKE_CURRENT_SOURCE_DIR}/lib/TestGame")
+	if(NOT EXISTS ${TEST_GAME_REPOSITORY_PATH})
+		execute_process(COMMAND ${GIT_EXECUTABLE} clone
+			"https://github.com/EasyRPG/TestGame.git"
```
carstene1ns: btw. we could just use `--depth=1` here too as the history is not needed either.

CMakeLists.txt
Ghabry: Yes is just my pseudo-return variable because looks like that you can't really return something from Cmake functions...

CMakeLists.txt
```diff
@@ -5,122 +5,202 @@ set(CMAKE_CXX_STANDARD_REQUIRED on)
...
-if (libmpg123_FOUND)
-  include_directories(${MPG123_INCLUDE_DIR})
-  list(APPEND EASYRPG_PLAYER_LIBRARIES ${MPG123_LIBRARY})
-  add_definitions(-DHAVE_MPG123)
+# freetype and harfbuzz
+option(PLAYER_WITH_FREETYPE "Support FreeType font rendering" ON)
+find_lib("Freetype" PLAYER_WITH_FREETYPE)
+if(${RETVAL})
+	add_definitions(-DHAVE_FREETYPE)
+	option(PLAYER_WITH_HARFBUZZ "Enable HarfBuzz text shaping (Requires FreeType)" ON)
```
carstene1ns: `CMAKE_DEPENDENT_OPTION` could be used instead for harfbuzz.

CMakeLists.txt
```diff
@@ -5,122 +5,202 @@ set(CMAKE_CXX_STANDARD_REQUIRED on)
...
-link_directories(${ICU_LIBRARY_DIRS})
-
 set(EXECUTABLE_OUTPUT_PATH ${CMAKE_CURRENT_BINARY_DIR}/bin)
 set(LIBRARY_OUTPUT_PATH ${CMAKE_CURRENT_BINARY_DIR}/lib)
-#link_directories(${LIBRARY_OUTPUT_PATH})
-#cmake_policy(CMP0015 NEW)
 
 list(APPEND CMAKE_MODULE_PATH
-	"${CMAKE_CURRENT_SOURCE_DIR}/lib/liblcf/builds/cmake/Modules"
 	"${CMAKE_CURRENT_SOURCE_DIR}/builds/cmake/Modules")
```
carstene1ns: Some boilerplate I would like to propose to be added here:
```cmake
# Set a default build type if none was specified
if(NOT CMAKE_BUILD_TYPE)
	message(STATUS "Setting build type to 'Debug' as none was specified.")
	set(CMAKE_BUILD_TYPE Debug CACHE STRING "Choose the type of build." FORCE)
	set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
endif()
```
This makes sure a build type is defined, so different flags can be used.
Cmake will for example use `-g` for "Debug" builds and optimize `-O3` for "Release" build by default.

Ghabry:
I hope this was the last change now :D

Ghabry:
No it wasnt. Found a problem in combination with PLAYER_BUILD_LIBLCF... trying to fix this tomorrow >.>

CMakeLists.txt
```diff
@@ -2,231 +2,560 @@ cmake_minimum_required(VERSION 3.1)
...
+# liblcf
+option(PLAYER_BUILD_LIBLCF "Instead of detecting liblcf the liblcf repository is cloned into lib/liblcf and built together with the Player. This is convenient for development" OFF)
+set(PLAYER_BUILD_LIBLCF_GIT "https://github.com/EasyRPG/liblcf.git" CACHE STRING "Git repository of liblcf to clone when building liblcf. Requires PLAYER_BUILD_LIBLCF=ON.")
+
+if(PLAYER_BUILD_LIBLCF)
+	# liblcf is built as part of this cmake file
+	set(LIBLCF_PATH "${CMAKE_CURRENT_SOURCE_DIR}/lib/liblcf")
+	find_package(Git REQUIRED)
+	if(NOT EXISTS ${LIBLCF_PATH})
+	execute_process(COMMAND ${GIT_EXECUTABLE} clone
```
carstene1ns: reason why this does not use `--depth=1` anymore?

Ghabry: Because you can't really work with a unshallow clone when you want to edit code but I figured out in the meanwhile that there is an unshallow command for me usecase :D

CMakeLists.txt
```diff
@@ -2,231 +2,560 @@ cmake_minimum_required(VERSION 3.1)
...
-  add_definitions(-D _DEBUG=1)
+# Adds a library to the liblist if COND is true and the lib is found
+# When REQ is true the lib is required
+function(find_lib_int LIBNAME COND REQ)
+	if(${COND})
+		list(APPEND EASYRPG_LIB_LIST ${LIBNAME})
+	
+		if(${REQ})
+			find_package(${LIBNAME} REQUIRED)
+		else()
```
Ghabry: Just testing the new review feature.

CMakeLists.txt
```diff
@@ -2,231 +2,560 @@ cmake_minimum_required(VERSION 3.1)
...
+			find_package(${LIBNAME})
+		endif()
+		
+		string(TOUPPER ${LIBNAME} LIBNAMEUP)
+		include_directories(${${LIBNAMEUP}_INCLUDE_DIR} ${${LIBNAMEUP}_INCLUDE_DIRS})
+		set(NEW_LIBS ${${LIBNAMEUP}_LIBRARY} ${${LIBNAMEUP}_LIBRARIES} PARENT_SCOPE)
```
Ghabry: This stuff should only be set when FOUND :-1:

builds/cmake/Modules/FindFluidSynth.cmake
```diff
@@ -3,3 +3,8 @@ find_path(FLUIDSYNTH_INCLUDE_DIR "fluidsynth.h")
 
 include(FindPackageHandleStandardArgs)
 find_package_handle_standard_args(FluidSynth DEFAULT_MSG FLUIDSYNTH_LIBRARY FLUIDSYNTH_INCLUDE_DIR)
+
+set(FLUIDSYNTH_INCLUDE_DIRS ${FLUIDSYNTH_INCLUDE_DIR})
+set(FLUIDSYNTH_LIBRARIES ${FLUIDSYNTH_LIBRARY})
+
+mark_as_advanced(FLUIDSYNTH_INCLUDE_DIR FLUIDSYNTH_LIBRARY)
```
carstene1ns: This is needed for XMP, too :dancer:.

Ghabry: hmmm

PR #965, opened 2016-08-01
Fix some audio issues

Ghabry:
Another bigger change is that now full paths are passed to the audio interface instead of the api using file finder. This is needed to support ineluki properly (will do another PR for this later).
And need to compile test OpenAL and 3ds.


carstene1ns:
This looks good for merging.

Ghabry:
I will just test it with the 3DS PR builder before

Issue #948, opened 2016-07-25
Suggestion: Provide access to a "Debug" menu option in Scene_Menu when TestPlay mode is on


Ghabry:
Because some TestPlay stuff is not reachable easily on some ports I thought about adding a "Debug" menu item in the normal menu when TestPlay is on (at least if it's not blocked by the game, maybe provide an overwrite like "Press B 5 times in one second" to still open it) which provides options like:

| `Help-Text` |
| --- |
| `Open Load Menu` |
| `Open Save Menu` |
| `Switch/Variable Editor` |
| `[  ] Walk through Walls` |
| `[  ] Instant text` |
| `[  ] Zegeri's event overlay` |

Ghabry:
Closing this. There is a debug scene by now and it already got many features. Just open new issues to request new features.

PR #966, opened 2016-08-01
Add Debug option to Scene Menu (TestPlay only)

Ghabry:
Related #948 

All applies to TestPlay mode only:

This makes accessing our debug scene easier, when in TestPlay mode a "Debug" item will be shown in the normal menu.

The debug sceen also got 2 new options (besides var and switch editing): Save and Load, they do what the name says (and ignore "save forbidden" setting).

Some games forbid menu calling but don't worry, this can be overriden with "Shift+5x B" (when SHIFT is not pressed anymore the counter is reset).

### DIFF
--- a/src/scene_debug.cpp
+++ b/src/scene_debug.cpp
@@ -27,6 +27,8 @@
 #include "game_map.h"
 #include "game_system.h"
 #include "scene_debug.h"
+#include "scene_load.h"
+#include "scene_save.h"
 #include "player.h"
 #include "window_command.h"
 #include "window_varlist.h"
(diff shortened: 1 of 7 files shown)


src/scene_debug.cpp
```diff
@@ -99,44 +115,58 @@ void Scene_Debug::Update() {
...
+
+	if (current_var_type == TypeSwitch) {
+		var_window->SetShowSwitch(true);
+	} else if (current_var_type == TypeInt) {
+		var_window->SetShowSwitch(false);
+	}
+
+	var_window->SetVisible(current_var_type != TypeGeneral);
+
+	printf("%d\n", current_var_type);
```
carstene1ns: debug leftover.

Ghabry:
Yeah that was an oversight :/

For testing I recommend Ib because it blocks the main menu: https://easyrpg.org/play/pr966/?game=ib&testplay

Ghabry:
Sorry for the noise. Somehow some local files got added when I ammended -.-

PR #969, opened 2016-08-02
Add support for Inelukis MP3 Patch

Ghabry:
Differences to the real patch because we use the BGM Api directly:
 - It uses our BGM Api directly, this means compared to the real patch it will not overaly with the normal BGM which must be stopped manually (Needs testing in games if this is a problem)
 - The Patch allows playing multiple MP3s at the same time, but who needs this. (and Disharmony fixed this problem)
 - No support for the "loop" statement on line 2, it will always loop.
 - Our code does not care about the format, it can play all formats we support (but nobody uses this obviously).

This patch is enabled for all games, I see no risk in doing this. In the unlikely case a game really uses a filename with ".link" and then fails, we will see this in an Android bug report ;)

Algorithm:
When .link found it reads the first line and converts that path into a canonical one using the game directory as a root and passes it to FindDefault which finds the file. Simple ^^

### DIFF
--- a/src/game_system.cpp
+++ b/src/game_system.cpp
@@ -27,6 +27,8 @@
 #include "output.h"
 #include "graphics.h"
 #include "main_data.h"
+#include "player.h"
+#include "reader_util.h"
 #include "scene_save.h"
 
 namespace {
@@ -326,6 +328,7 @@ void Game_System::OnBgmReady(FileRequestResult* result) {
 
 		// The first line contains the path to the actual audio file to play
 		std::string line = Utils::ReadLine(*stream.get());
+		line = ReaderUtil::Recode(line, Player::encoding);
 		
 		Output::Debug("Ineluki link file: %s -> %s", path.c_str(), line.c_str());
 
--- a/src/game_system.cpp
+++ b/src/game_system.cpp
@@ -317,8 +317,6 @@ void Game_System::OnBgmReady(FileRequestResult* result) {
 	}
 
 	if (Utils::EndsWith(result->file, ".link")) {
-		Output::Debug("Ineluki link file: %s", path.c_str());
-		
 		// Handle Ineluki's MP3 patch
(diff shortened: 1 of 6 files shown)


carstene1ns:
One thing I do not like about this is the generated noise in the log. You will get 3 lines related to one file in some cases:
```
[2016-08-02 23:10:09] Debug: Ineluki link file: ../RPG_Maker Games/CalmFalls/Music/17-Prisonic_Fairytale.link.wav
[2016-08-02 23:10:09] Debug: Linking to: ..\mp3\17-Prisonic_Fairytale.MP3
[2016-08-02 23:10:09] Debug: MP3 heuristic: ../RPG_Maker Games/CalmFalls/mp3/17-Prisonic_Fairytale.MP3
```
Can we get rid of the heuristic line already? It should be well tested enough until we get another bug reports after dropping sdl_mixer... :P Also it would be okay to join the Ineluki lines in one.

Ghabry:
Done

src/game_system.cpp
```diff
@@ -307,23 +308,54 @@ void Game_System::SetTransition(int which, int transition) {
...
+	Audio().BGM_Play(path, data.current_music.volume, data.current_music.tempo, data.current_music.fadein);
 }
 
 void Game_System::OnSeReady(FileRequestResult* result, int volume, int tempo) {
 	auto item = se_request_ids.find(result->file);
 	if (item != se_request_ids.end()) {
 		se_request_ids.erase(item);
 	}
-
+
```
carstene1ns: :scream:

carstene1ns:
If this works for games using the escape_symbol for directory traversal, then looks good to me.

Ghabry:
Now that you say this, I forgot something. The read line must be converted from game encoding to utf8. >.>

carstene1ns:
Works for me correctly with Calm Falls & Nebulus intro. Merging.

PR #970, opened 2016-08-03
3ds: Fix AudioDecoder

Ghabry:
This fixes the crash when the BGM changes. No idea what the real cause is, race condition. I just added Mutex everywhere... works but I will ask some people for a hardware test.

Pitch support for Audio Decoder added

Midi is temporary, will turn it off again when it still not works.


Ghabry:
WAV, OGG and MP3 confirmed working without crashes.
Midi stays disabled, lets resolve this someday with wildmidi.

carstene1ns:
you should remove `WANT_FMMIDI=1` from the Makefile then to gain a bit of space ;-)

Ghabry:
as you wish ;)

src/3ds_decoder.cpp
```diff
@@ -421,11 +426,11 @@ int OpenAudioDecoder(FILE* stream, DecodedMusic* Sound, std::string const& filen
 		#ifdef MIDI_DEBUG
 		samplerate = 11025;
 		Output::Warning("MIDI track detected, lowering samplerate to 11025 Hz.", filename.c_str());
```
carstene1ns: Nitpicking a bit, but the second argument is useless here, as there is no %s in the string.

Ghabry: Is not my code but good find :)

Issue #875, opened 2016-04-26
Embric of Wulfhammer's Castle: Game occasionally freezes in 993 Anti-Paladins battle


Siltala:
EasyRPG occasionally freezes specifically during the 993 Anti-Paladins battle in this game.
To be more exact, the game technically remains responsive, but the battle system itself is stuck, meaning that the ATB gauges do not refill and thus it is impossible to actually do anything. As far as I can tell, there is no way to get around the problem.

* Game download: https://aeresland.net/EmbricDELUXE_2.2.2.7z
* Savegame: https://www.heavylobster.com/files/Save07.lsd

To reproduce, walk towards the east to start the encounter after going through some dialogue (the 'Zhen die without explanation' option is fastest). Then just keep battling until the freeze occurs. It doesn't seem to be caused by any particular player actions (can be reproduced with autobattle) and it doesn't seem to occur every single time I do the battle either, so it may take several attempts to reproduce. It is still quite common, though.

(this was tested with latest git as of this typing)

carstene1ns:
As we found out today, this is a battle system deadlock caused by a skill reviving a party member.
This case is already handled for items, though.

Ghabry:
Have a fix for this but have to do more testing.

Ghabry:
I know in the meanwhile why this doesn't affect Ara Fell: As Vampires Dawn 2 this game uses battle animations for the battlers instead of normal animated battlers. Have to compare the code to find what we do different there.

Zegeri:
Now the Anti-Paladins don't disappear when they die.

Ghabry:
Uh that's bad, will fix it :(

PR #975, opened 2016-08-06
Battle: Fix problems with non-fading out enemies

Ghabry:
Now everything fades out again and Battle 2k system had a bad timing issue: Damage and Kill animation were without a pause, looked wrong, the kill animation waits now for damage.


carstene1ns:
So this addresses the remark by @Zegeri in #875... :+1:

meme

PR #978, opened 2016-08-07
Fix Yume2kki ugliness

Ghabry:
This game is stupid and has invalid inventory in the slots.
The weapon contains a switch e.g., RPG_RT filters this and the editor shows it as (None) but the database still contains the value. Way to go Enterbrain!

Furthermore this 350 MB games saves tons of space *cough* by crossreferencing files via
Panorama\\..\\Title\\Title3, nice, that path is now rewritten to Title\\Title3 by the FileFinder.
I also tested files directly in the root directory (Panorama\\..\\Title3), becomes Title3 :+1: 
That strange feature is only used when filefinder path != canonical, so only happens when "." or ".." is a path component which can't happen when using the editor without hex editing.

### DIFF
--- a/src/filefinder.cpp
+++ b/src/filefinder.cpp
@@ -99,9 +99,26 @@ namespace {
 			return em_file;
 #endif
 
-		std::string const lower_dir = Utils::LowerCase(dir);
+		std::string lower_dir = Utils::LowerCase(dir);
 		std::string const escape_symbol = Player::escape_symbol;
 		std::string corrected_name = Utils::LowerCase(name);
+
+		std::string combined_path = MakePath(lower_dir, corrected_name);
+		std::string canon = MakeCanonical(combined_path, 1);
+		if (combined_path != canon) {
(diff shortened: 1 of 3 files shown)


carstene1ns:
Will test this later. Hope we find other games fixed by this using our Android userbase... :grinning:

Ghabry:
I also want to check if i  simplify the lookup for ineluki mp3 due to these changes here

Ghabry:
I also want to check if i  simplify the lookup for ineluki mp3 due to these changes here

Ghabry:
No can't simplify ineluki, the canonical path is also used for debug log. Even if this would only save one line... ;)

Ghabry:
this is ready for testing now.

Ghabry:
The invalid equipment filtering is a gigantic clusterfuck. Happens in like 33% of my games :o

Expect complains of users losing items because I also clean up when loading saves ;)

PR #981, opened 2016-08-08
PSVITA: Various stuffs for PSVITA port

Rinnegatamante:
That fix the coredump when exiting the player.


Ghabry:
Besides this I think it's a good idea to try to fix the bugs in the audio system that users report.
Which are: Bad resampling (not sure what causes this) and SE problems.

Is it possible for you to merge psp2_decoder into audio_psp? At least all these callbacks are useless, they were useful on the 3DS because the audio system that time was a mess but here you always only have one callback, rendering them obsolete.
This would also get rid of that ```extern unique_ptr<AudioDecoder>``` in audio_psp which is not very C++-ish at all ^^
This would also make it easier to understand the code and to find what breaks the resampler, must be something simple.

According to the gbatemp post you linked the severe problems are:
"Yume Nikki, LcdDem and similar Yume-Nikki fangames: Run very well, but once you activate the bicycle/sneakers/(item for walking faster) and walk around tiles that make different noises, the app will probably crash."

Sounds like a problem in your Sound code when too many SEs are played at the same time?

"All games so far: Stuttering when sound effects play. It's very slight, but still noticeable. Gets worse if many sound effects play on sequence"
I think this is not your fault and we should fix this with a Sound cache because other slow platforms have this, too.

Ghabry:
A user on gbatemp reported that the broken Music and sound is fixed. (was the build without basic wav decoder).

May I ask you to move the "Added a basic wav decoder" commit to a seperate branch and create another pull request? Because this is of general interest for other ports and needs an extra review.

Ghabry:
You commited EasyRPG.velf in the last commit. You could delete it in another one but that would still make it part of the repository and everybody has to download 8 MB of useless data :-1: 

So please remove it: (make sure to be on the right branch (```git branch```).

Because it is the last commit you can "amend" that commit:
```
# Remove the file from the commit index
git rm --cached path/to/EasyRPG.velf

# Repeat the commit (just save and exit in the editor)
git commit --amend

# Do a forced push, because server repo mismatches now with local one
git push -f origin
```

Ghabry:
Jenkins: Test this please

PR #990, opened 2016-08-12
Implement Skills Teleport and Escape

Ghabry:
Basicly besides video the only two events that were not fully implemented ;) (okay the events were implemented already but not the skills to use them).
Probably due to lazyness, only took two hours to get this coded.

That TerrainTag commit unrelated, just observed this while testing and is easy fix.


Ghabry:
Test case:
https://easy-rpg.org/play/?start-map-id=2&new-game
3rd row from the top, last character

carstene1ns:
> GetTerrainTag must return 9 when out of bounds

hmmm, magic numbers... :dash:

PR #994, opened 2016-08-14
Android: add expat for XML parsing

carstene1ns:
Needed for EasyRPG/buildscripts#11.


carstene1ns:
Jenkins: test this please <3

That being said, expat and the new library versions are already installed.

PR #1000, opened 2016-08-18
Fix FastWAV issues and crash in AudioDecoder-loop

### LINKED ISSUE #992: PS Vita: Crash when SE channels are exhausted
This happens because SE_Play does not check if there are any free channels.

Could be e.g. fixed by initializing a Semaphore to se channel count (7?) and then aquiring the Semaphore at the beginning of SE thread-for-loop and releasing it at the end of the for-loop.

The main thread could check this semaphore and when it is 0 (= would block) issue a Output::Warning and ignore that SE (like the SDL audio version does).

Ghabry:
This fixes some flaws in the FASTWAV implementation:

 - Seek always returned false when "file_" was set, resulting directly in a crash in OpenAL, otherwise during the first rewind (assert triggered)
 - Bits per sample was read from the wrong field, was off by 6 bytes.
 - "data" chunk was not always found because the code assumed that the chunk position is 4-byte-aligned
 - Chunks after "data" would play as noise
 - Code was not big endian aware

The audio decoder will now only go 10 recursions deep when trying to refill the buffer after a Rewind. This fixes a typical stack overflow case.

Fix #992

### DIFF
--- a/src/audio_decoder.cpp
+++ b/src/audio_decoder.cpp
@@ -22,6 +22,7 @@
 #include "filefinder.h"
 #include "output.h"
 #include "system.h"
+#include "utils.h"
 
 #include "decoder_fmmidi.h"
 #include "decoder_mpg123.h"
@@ -40,6 +41,10 @@ void AudioDecoder::Resume() {
 }
 
 int AudioDecoder::Decode(uint8_t* buffer, int length) {
+	return Decode(buffer, length, 0);
+}
+
+int AudioDecoder::Decode(uint8_t* buffer, int length, int recursion_depth) {
 	if (paused) {
 		memset(buffer, '\0', length);
 		return length;
@@ -53,16 +58,21 @@ int AudioDecoder::Decode(uint8_t* buffer, int length) {
 		memset(&buffer[res], '\0', length - res);
 	}
 
-	if (IsFinished() && looping) {
+	if (IsFinished() && looping && recursion_depth < 10) {
 		++loop_count;
 		Rewind();
 		if (length - res > 0) {
-			int res2 = Decode(&buffer[res], length - res);
+			int res2 = Decode(&buffer[res], length - res, ++recursion_depth);
 			if (res2 <= 0) {
 				return res;
 			}
 			return res + res2;
 		}
+	}
+
+	if (recursion_depth == 10 && loop_count < 50) {
+		// Only report this a few times in the hope that this is only a temporary problem and to prevent log spamming
+		Output::Debug("Audio Decoder: Recursion depth exceeded. Probably stream error.");
 	}
 
 	return res;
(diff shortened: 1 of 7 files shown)


Ghabry:
Will delegate the PR builder of the Vita version to some users so they can tell me if it's fixed...

Ghabry:
One success story: https://gbatemp.net/threads/release-easyrpg-vita-rpg-maker-2000-2003-player-for-psvita.437040/page-4#post-6611111

Ghabry:
Another success story: https://gbatemp.net/threads/release-easyrpg-vita-rpg-maker-2000-2003-player-for-psvita.437040/page-4#post-6611222

Probably the last one because we only have 2 users on gbatemp ;)

Ghabry:
This 3rd user also said it works: https://gbatemp.net/threads/release-easyrpg-vita-rpg-maker-2000-2003-player-for-psvita.437040/page-4#post-6613853

So I guess this is mergeready :+1:

carstene1ns:
Did someone say merge?
sleepwalking

carstene1ns:
The fastwav_decoder now plays noise on Wii in Yume Nikki title scene, libsndfile plays it fine. Seems like some swapping is wrong/broken.

Ghabry:
Yepp, the buffer is little endian and not byteswapped. Easy fix, we can do this together with the wildmidi PR >.>

Ghabry:
```diff
diff --git a/src/decoder_wav.cpp b/src/decoder_wav.cpp
index 23086f4..b9c7c5d 100644
--- a/src/decoder_wav.cpp
+++ b/src/decoder_wav.cpp
@@ -147,7 +147,22 @@ int WavDecoder::FillBuffer(uint8_t* buffer, int length) {
        }
 
        int decoded = fread(buffer, 1, real_length, file_);
-       if (decoded < length)
+
+       if (Utils::IsBigEndian()) {
+               if (output_format == AudioDecoder::Format::S16) {
+                       uint16_t* buffer_16 = reinterpret_cast<uint16_t*>(buffer);
+                       for (int i = 0; i < real_length / 2; ++i) {
+                               Utils::SwapByteOrder(buffer_16[i]);
+                       }
+               } else if (output_format == AudioDecoder::Format::S32) {
+                       uint32_t* buffer_32 = reinterpret_cast<uint32_t*>(buffer);
+                       for (int i = 0; i < real_length / 4; ++i) {
+                               Utils::SwapByteOrder(buffer_32[i]);
+                       }
+               }
+       }
+
+       if(decoded < length)
                finished = true;
        
        return decoded;
```

This fixes the buffer swapping but some sounds still sound wrong (too high pitched). Maybe disable FASTWAV again for Wii... and we solve this later with some audio cache.

carstene1ns:
¯\_(ツ)_/¯

PR #1001, opened 2016-08-18
PSVITA: SE spamming crash bugfix.

Rinnegatamante:
As the title says, this fix the SE spamming crash bug reported here: https://github.com/EasyRPG/Player/issues/992

### DIFF
--- a/src/audio_psp2.cpp
+++ b/src/audio_psp2.cpp
@@ -26,6 +26,7 @@
 #include <stdio.h>
 #include <cstdlib>
 #include "audio_decoder.h"
+#include <atomic>
 
 // Internal stuffs
 #define AUDIO_CHANNELS 7 // PSVITA has 8 audio channels but one is used for BGM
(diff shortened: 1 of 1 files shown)


Ghabry:
Does it pass my artificial test case?

Rinnegatamante:
Yes, as reported also on IRC.

src/audio_psp2.cpp
```diff
@@ -186,6 +187,7 @@ static int sfxThread(unsigned int args, void* arg){
 		free(sfx->audiobuf2);
 		sfx_decoder[sfx->id].reset();
 		free(sfx);
+		availThreads++;
```
Ghabry: That usage of a volatile variable has a race condition. It is not guaranteed that ++ or -- are one operation, could have a register load which could result in incorrect values.
Especially because volatile disables the optimizer. The purpose of volatile was never multithreading but for accessing memory mapped hardware IO ports.

Assuming the PSVita has this (should be in newlib, is just normal ARM assembly stuff):
remove the volatile, include <atomic> and change the variable type to.
std::atomic<std::uint8_t>

Rest can stay the same, operators are overloaded.

Ghabry:
Closing because got cherry-picked into #1000 to make testing (broken WAV code) easier.

PR #1005, opened 2016-08-22
Added unstripped elf preservation, a better way to close audio threads and a vpk builder for standalone games

Rinnegatamante:
As the title says. 
I've also added the -g flag by default since arm-none-eabi-vita-strip will make the final elf looking pretty much as big as if the flag isn't used.


carstene1ns:
Btw. `$(basename file[s]...)` should get the file part sans the extension, so we could use:
```make
%.velf: %.elf
	$(PREFIX)-strip -g -o $(basename $<)-stripped.elf $<
	vita-elf-create $(basename $<)-stripped.elf $@
```
This way we do not have an ugly filename with multiple extensions. :dancers:.

Ghabry:
Looks good to me. But I would like to see a response from a user that the standalone mode doesnt crash :D

Rinnegatamante:
I can try it right now, give me a couple of minutes :P

carstene1ns:
The `make clean` target lacks `$(TARGET)-stripped.elf` now :see_no_evil:

PR #1006, opened 2016-08-22
Wii: Wildmidi, some small endianess-and-other-fixes

carstene1ns:


Ghabry:
Jenkins: Test this please

Ghabry:
I 99% sure I found the bug that crashes on the PS Vita. As pointed out by the libretro people fread reads sometimes (when the file ends) negative bytes in FillBuffer of WAVDecoder. Though instead of fixing it they only worked around it. Anyway I fail in math and here is a proper fix which also fixes yume nikki music loop for me.

I wonder why reading a huge amount of bytes (size_t) is just ignored instead of crashing on linux o.O

```cpp
	int real_length;

	// Handle case that another chunk is behind "data"
	if (cur_pos + length >= audiobuf_offset + chunk_size) {
		real_length = audiobuf_offset + chunk_size - cur_pos;
		cur_pos = audiobuf_offset + chunk_size;
	} else {
		real_length = length;
		cur_pos += length;
	}
```

[remove the curpos = curpos + length that is after "int real_length"]

src/audio_resampler.h
```diff
@@ -187,7 +187,7 @@ class AudioResampler : public AudioDecoder {
 			spx_uint32_t input_frames_used, output_frames_gen;
 			spx_uint32_t ratio_num, ratio_denom;
 		} conversion_data;
-		SpeexResamplerState * conversion_state;
+		SpeexResamplerState * conversion_state = nullptr;
 	#elif defined(HAVE_LIBSAMPLERATE)
 		SRC_DATA conversion_data;
 		SRC_STATE * conversion_state;
```
Ghabry: Could you add nullptr here, too? I bet this will crash with the same problem.

Ghabry:
gbatemp users confirmed that the lines I suggested indeed fixed ALL crashes :+1:

carstene1ns:
Will update this in approximately ~10 hours.

carstene1ns:
> I wonder why reading a huge amount of bytes (size_t) is just ignored instead of crashing on linux o.O

Likely the used C library (glibc) refuses that, but newlib lacks that check.

Issue #856, opened 2016-04-04
IDAT: invalid distance too far back


lpis:
__Name of the game__:Nepheshel

__Player platform__ (e.g. Windows, Linux, Android, ...):Android

__Attach files__ (as a .zip archive or link them):
[Nepheshel203.zip](https://github.com/EasyRPG/Player/files/201908/Nepheshel203.zip)
 - A savegame next to the problem (if it isn't at the beginning of the game)
 - The easyrpg_log.txt logfile (you can find it in the game directory)

__Describe the issue in detail and how to reproduce it__:
![screenshot_2016-03-17-12-28-16](https://cloud.githubusercontent.com/assets/18257069/14240876/6f65a36a-fa8b-11e5-92cc-5faecaf60249.png)

As soon as the file opened, this error (says "IDAT: invalid distance too far back") comes up and never work

Ghabry:
Thanks. This looks like a problem with our PNG image parser. We will investigate.

fdelapena:
I've tested the attached game on desktop and loads PNG images properly. There is a known bug at least between libpng 1.6.0 and 1.6.3 showing this message. Maybe the libpng version used in the toolchain is outdated.

Ghabry:
I can reproduce the bug under Windows (same Version as libpng on Android)

Ghabry:
And we can't really fix this because libpng calls the error handler and when our error handler returns libpng terminates the application m(

fdelapena:
It works for me with libpng 1.6.18, maybe there is a bug in the most recent version.

Ghabry:
ImagePNG needs some error handling (a bool return) so that the callee can draw a checkerboard e.g.

I just remembered how to work around the libpng Exit:
```
jmp_buf env;

static void on_png_error(png_structp, png_const_charp error_msg) {
	Output::Warning("%s", error_msg);
	longjmp(env, 1);
}

// beginning of ReadPNG
	int jmp_res = setjmp(env);
	if (jmp_res == 1) {
		return;
	}
```

\o/

carstene1ns:
@lpis: While we unfortunately cannot make the original image work with our used library, it can at least be replaced by yourself locally:
[![default](https://cloud.githubusercontent.com/assets/4691314/15621268/17af6120-2460-11e6-8c32-4db0cb5f31b3.png)](https://p.f4ke.de/JBEP2VKTPQ-nepheshelsystem.zip)
[...]

PR #1008, opened 2016-08-23
Image refactor

Ghabry:
This ends what @carstene1ns started :+1:

### DIFF
--- a/src/bitmap.cpp
+++ b/src/bitmap.cpp
@@ -71,6 +71,7 @@ Bitmap::~Bitmap() {
 }
 
 bool Bitmap::WritePNG(std::ostream& os) const {
+#ifdef SUPPORT_PNG
 	size_t const width = GetWidth(), height = GetHeight();
 	size_t const stride = width * 4;
 
@@ -83,6 +84,9 @@ bool Bitmap::WritePNG(std::ostream& os) const {
 							 0, 0, 0, 0, 0, 0, width, height);
 
 	return ImagePNG::WritePNG(os, width, height, &data.front());
+#else
+	return false;
+#endif
 }
 
 int Bitmap::GetWidth() const {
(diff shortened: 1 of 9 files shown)


src/bitmap.cpp
```diff
@@ -533,19 +551,38 @@ Bitmap::Bitmap(const std::string& filename, bool transparent, uint32_t flags) {
...
+		free(pixels);
+
+		pixels = nullptr;
+
+		return;
+	}
+
 	Init(w, h, (void *) NULL);
-	ConvertImage(w, h, pixels, transparent);
+	if (pixels)
```
carstene1ns: This is not necessary anymore, as you `return` before, in case of error.

src/bitmap.cpp
```diff
@@ -559,17 +596,32 @@ Bitmap::Bitmap(const uint8_t* data, unsigned bytes, bool transparent, uint32_t f
...
+		Output::Warning("Unsupported image");
+
+	if (!img_okay) {
+		return;
+	}
 
 	Init(w, h, (void *) NULL);
-	ConvertImage(w, h, pixels, transparent);
+	if (pixels)
+		ConvertImage(w, h, pixels, transparent);
```
carstene1ns: same.

carstene1ns:
Looks good so far.
What about using the embedded System image (as mentioned in #856), does this work without changing anything (Cache)?

Ghabry:
screenshot_0

Yeah it works I also tested invalid BMP and invalid XYZ codepaths yesterday, they behave the same.

One minor problem here is that the message is slightly misleading because it says "not found". The cache can't differentiate between not found and invalid because the return value is the same (empty image :/).  Hm but maybe I can fix this one by moving the warning to another location.

And the Output appears twice this is because our cache is a bit flush-friendly. I have plans to improve the cache but not in this PR :)

Ghabry:
Thats better (deleted the title to show a difference)


screenshot_1

src/cache.cpp
```diff
@@ -53,11 +53,18 @@ namespace {
...
 
+			BitmapRef bmp = BitmapRef();
+
 			if (path.empty()) {
-				return BitmapRef();
+				Output::Warning("Image not found: %s/%s", folder_name.c_str(), filename.c_str());
+			} else {
+				bmp = Bitmap::Create(path, transparent, flags);
+				if (!bmp) {
+					Output::Warning("Image invalid: %s/%s", folder_name.c_str(), filename.c_str());
```
fdelapena: Nitpicky comment: maybe "Invalid image" is better.

Ghabry: at least thats proper grammar

src/cache.cpp
```diff
@@ -53,11 +53,18 @@ namespace {
...
+				}
 			}
-
-			return (cache[key] = Bitmap::Create(path, transparent, flags)).lock();
+
```
carstene1ns: :o

src/bitmap.cpp
```diff
@@ -947,7 +947,7 @@ void Bitmap::FillRect(Rect const& dst_rect, const Color &color) {
 void Bitmap::Clear() {
 	memset(pixels(), '\0', height() * pitch());
 
-        RefreshCallback();
+	RefreshCallback();
```
Ghabry: before somebody complains here: This replaces 9 (!) spaces with 2 tabs :dancer:

Ghabry:
Was there anything left here?

Issue #998, opened 2016-08-16
Nobihazard: Hitting zombies from the side (or back) gives 100s of SE channel warnings


Ghabry:
When the zombies are blocked by a wall or hit from the side they will generate hundreds of "SE: Out of channel" warnings.

Download: https://nobihaza.wordpress.com/2014/02/21/doraemon-nobitas-resident-evil-english-release/

Savegames:  
[nobihaza english.zip](https://github.com/EasyRPG/Player/files/419252/nobihaza.english.zip)

Load No3 and let the zombies get stuck e.g. by walking against that lantern.

Depends on #1356

Ghabry:
What fixes it in my test case but not in the real code is this:
https://github.com/EasyRPG/Player/compare/master...Ghabry:development?expand=1

So basicly make Turning never fail, so the move doesnt fail, the frequency applies and plays the sound with enough gap.
But still failing on this map (save 3)... Because this only happens when two events collide I guess it has to do with "MakeWay". Maybe @Zegeri has an idea.

Ghabry:
Looks like this only happens when the move route is frequency 8.
I'm not sure why but all our checks use ``GetMoveFrequency() > 7 ? 0`` instead of 8 when setting ``max_stop_count``.
I can prevent the channel spamming by changing the check from 7 to 8. The SE still plays faster (which means the move executes too often) but at least the warning is gone.

This bug is exposed by parallel event:
Event with Custom moveroute (repeat, ignore illegal moves)
```
Turn toward Player
Play SE
1 Step forward
```
And event script:
```
Move [Empty route]
Wait 0.1
```

Ghabry:
I revoke this patch, that change breaks the behaviour of "Turn" for Frequency 8.

fmatthew5876:
I just tried this custom move type with freq=8, repeat, skip.

```
Play SE
Turn to Player
Move foward
```

RPG_RT plays the sound rapidly once the event gets stuck.

This makes sense to me, as this will retry every frame.

fmatthew5876:
Its highly likely #1696 is the cause of this one

fdelapena:
#1696 is fixed but the behavior still happening here. If this behavior is expected, at least player should not display any warnings and should reuse existing channels (olders first) when the same audio file is being player over again. If we can probe RPG_RT only uses 4 channels for this single sound at once, then this should be implemented as part of #1356.

Ghabry:
[...]

PR #1009, opened 2016-08-24
Fix various minor problems

Ghabry:
Collection of stuff I encountered which doesnt fit in single PRs

The TurnToHero/AwayHero thing needs a test if this is RPG Maker behaviour (turn direction) but this fixes some sound spamming in #998 

Maybe keep this open for a while maybe I find something else :D


Ghabry:
Todo list:
 - [ ] Fix broken tile in CLock of ATonement TV
 - [ ] Message boxes in UiD battle overlap the other battle windows (Regression >.<)

carstene1ns:
Apropos battle and windows: did you notice at battle end (in düsterburg) you can see the background image duplicated beyond it when it is fading out/hidden?

fdelapena:
> you can see the background image duplicated (...)

That issue seems to happen in all 2k game battles at least with these transitions, where backdrops are 320x160 and the window gets removed before the transition, which repaints the screen as pixel values change.

Ghabry:
Proper fix for this would be: Keep the Messagebox open

carstene1ns:
Also thought that, it would be the easiest at least.

carstene1ns:
The music commit (2e441add9e90) has a bit of debug output left.

Ghabry:
The last commit fixes the 1 HP problem in EHR and the 0 SP problem for Beta and Omega in OFF. ChangeClass command is still completely broken but at least the typical incorrect case (class applied on startup) is fixed.

Ghabry:
Okay, I'm done here.

PR #1016, opened 2016-08-28
Audio SE performance improvements

### LINKED ISSUE #895: Support arbitrary pitch changes for all sound types
__Name of the game__: Currently, any game using WAV and OGG for music or sounds.

__Player platform__: Any, with special mention of Wii (SDL1.2) and 3DS (non-SDL port), not using the  broken SDL mixer resampler.

__Describe the issue in detail and how to reproduce it__: SDL2 has well and long standing issues with the built-in resampler, which does noise or even may crash for some frequency conversions. Because this does not seem to be fixed anytime soon and because some ports does not use SDL or have an old version without arbitrary resampling (SDL 1.2), this issue supersedes #349.

This is needed to do proper playback for at least WAV files, quite popular in RPG Maker 200x games. For MIDI and MP3 there are no issues, as current implementations have its own resampler working properly. Also, for proper WAV support for platforms without SDL support, the libsndfile library will need a resampler (#859).

A proposal for our resampling needs:

**Speex resampler**. Features:

- Distro supported, available in the [speexdsp](http://speex.org/downloads/) package.
- Fixed and float truly arbitrary resampler with adjustable quality, designed to be fast, portable for low memory needs. It even includes some CPU specific optimizations (SSE, NEON).
- BSD licensed, which is app/game stores friendly.
- It works ;).

Ghabry:
Audio SE cache. Sound effects are cached now in a ~~5~~ 1 MB cache, only the first access does a cache miss. Afterwards it takes the decoded buffer directly. This also gives AudioDecoder for SE the first time. Should fix 24bit Pom Gets Wifi SEs and gives proper playback speed for the Wii that has a broken SDL resampler \o/.

3DS and PSVita don't use the cache yet.

Cache clearing strategy:
When the cache limit is reached all entries not being accessed in the last 10 seconds are removed.

Fixes #895.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -20,6 +20,8 @@ libeasyrpg_player_la_SOURCES = \
 	src/audio_decoder.h \
 	src/audio_resampler.cpp \
 	src/audio_resampler.h \
+	src/audio_secache.cpp \
+	src/audio_secache.h \
 	src/background.cpp \
 	src/background.h \
 	src/baseui.cpp \
--- a/builds/vs2015/PlayerLib.vcxproj
+++ b/builds/vs2015/PlayerLib.vcxproj
@@ -435,6 +435,7 @@
     <ClCompile Include="..\..\src\audio_decoder.cpp" />
     <ClCompile Include="..\..\src\audio_resampler.cpp" />
     <ClCompile Include="..\..\src\audio_sdl.cpp" />
+    <ClCompile Include="..\..\src\audio_secache.cpp" />
     <ClCompile Include="..\..\src\background.cpp" />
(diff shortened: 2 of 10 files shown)


Ghabry:
That one is also not risky. Will outsource this to the Wii communtiy to test :D. At least in Dolphin helps alot.

Ghabry:
Got feedback in the forum, they are happy ;). Concerning psvita and 3ds: Woukd prefer another PR for tjis because i need other testers.

Ghabry:
Jenkins: Test this please ... fu android

Ghabry:
otherwise this is ready now ^^

PR #1017, opened 2016-08-28
Tone: Performance improvements

Ghabry:
Tone change of the screen is now calculated per sprite. This means the tone is only calculated once and afterwards the Blit is normal speed as if no tone is changed. Special case is a "gradient change", when the tone changes per frame. In that case it uses the old behaviour because is faster for that case.

Gives a significant speed boost on slow hardware for that case obviously ;)

### DIFF
--- a/src/bitmap.cpp
+++ b/src/bitmap.cpp
@@ -1012,29 +1012,38 @@ void Bitmap::ToneBlit(int x, int y, Bitmap const& src, Rect const& src_rect, con
 		int gs = pixel_format.g.shift;
 		int bs = pixel_format.b.shift;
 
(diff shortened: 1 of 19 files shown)


Ghabry:
This could have regressions, though unlikely. Will also ask Wii people to test ;).

BlisterB:
For mute the audio : it would be nice to modify this option in game.
For pausing the game : for example, if you open the menu, the game is not paused (it's problematic on some circumstances).

Ghabry:
wrong topic >.>
github just lacks administration features

Ghabry:
Got feedback in the forum, they are happy ;)

PR #1018, opened 2016-08-28
Improve Bitmap cache

Ghabry:
Instead of using weak_ptr is now shared_ptr and gives a 10 MB cache. Otherwise behaviour is the same as before. This prevents tons of cache misses (e.g. every frame during gauge battle) we usually get.

Cache clearing strategy:
When the cache limit is reached all entries not being accessed in the last 10 seconds are removed.

**Debug output is enabled** to better showcase that the cache works. will disable this before merging.

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -34,10 +34,15 @@
 
 namespace {
 	typedef std::pair<std::string,std::string> string_pair;
-	typedef std::pair<BitmapRef, uint32_t> bitmap_pair;
+
+	struct CacheItem {
+		BitmapRef bitmap;
+		uint32_t last_access;
+	};
+
 	typedef std::pair<std::string, int> tile_pair;
 
-	typedef std::map<string_pair, bitmap_pair> cache_type;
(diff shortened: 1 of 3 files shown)


Ghabry:
That one here should be save to merge after testing (and removing that commit)

Ghabry:
Here I observed OOM for TSTT on wii. So what you think? Reduce cache size? I can even set it to 0 and just use the timeout logic... actually maybe thats worth a try :D

src/cache.cpp
```diff
@@ -34,23 +34,47 @@
 
 namespace {
 	typedef std::pair<std::string,std::string> string_pair;
+	typedef std::pair<BitmapRef, uint32_t> bitmap_pair;
```
scurest: Wouldn't it be better to define a struct here? The `.second.second` stuff is kind of hard on the eyes.

Ghabry:
PSA: This is done, struct is renamed, cache size is 0 (flush after 5 seconds of non-use)

PR #1040, opened 2016-09-18
Nintendo 3DS: Disable khax

Ghabry:
Because this crashes on many normal firmwares... :( and DSP works flawless anyway.


carstene1ns:
Compiles, and khax was run only on old firmware anyway.. so merging.

PR #1050, opened 2016-10-03
Interpreter: Fix Equipment command and Vehicle movement

Ghabry:
Vehicles with custom event routes were not moving at all, one function call missing.
This fixes "Dunkle Schatten" - The ship never reached our heros.

The equipment command has a function "unequip all", for some reason this was never implemented :o
This removes our "Item -1 is invalid" debugmsg ;). "Item 0 is invalid" is still generated when a variable is really 0, that's correct.

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -123,8 +123,8 @@ bool Game_Actor::IsSkillUsable(int skill_id) const {
 	const RPG::Skill& skill = Data::skills[skill_id - 1];
 
 	// Actor must have all attributes of the skill equipped as weapons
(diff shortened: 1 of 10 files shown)


Ghabry:
When reviewing please do a full text search and look for GetEquipment, SetEquipment, ChangeEquipment calls that I missed or fixed incorrectly.

Ghabry:
Jenkins: Test this please

Ghabry:
Jenkins: Test this please. Gradle <.<

Ghabry:
Oh, thanks :)

PR #1054, opened 2016-10-05
Fix Razas battle start (Transitions in parallel interpreters)

Ghabry:
See commit description

Also fixes the following case:
```
Teleport to Map2
HideScreen
ShowScreen
Wait 0
HideScreen
ShowScreen
```

Delete Map 2 and run that event parallel. The error "Map 2 not found" is executed at the Wait 0.0 position (before it appeared after the first HideScreen)

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -63,8 +63,8 @@ Game_Interpreter::Game_Interpreter(int _depth, bool _main_flag) {
 	Clear();
 }
 
- Game_Interpreter::~Game_Interpreter() {
- }
+Game_Interpreter::~Game_Interpreter() {
+}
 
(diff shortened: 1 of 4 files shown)


Ghabry:
semi-ot: The map not found error could be replaced with a warning. Basicly breaks the game anyway but at least reduces the amount of error calls even more.

carstene1ns:
About the map error: I have started restructuring our logging and error handling a bit,
there will be (recoverable) `error`s and `fatal`s (=> fatal errors) soon.
- `error` will close the current game (i.e. takes you back to the game browser if available)
- `fatal` will do what our current `Error()` does.

Only some of our current `Error()` calls will be upgraded to `fatal`s.

PR #1055, opened 2016-10-06
Battle: Fix Hitrate calculation for normal attacks and Fatigue

Ghabry:
Finally...

Concerning hitrate if not obvious:
Before the code evaluated as (100 - (100 - hit_chance)) * REST because was in 2 different lines.
But correct order is 100 - (100 - hit_chance) * REST, so the 100 - is the last.

The requirement for "std::ceil" can be tested via: Actor has 0/0 SP and check is for 34-100% exhaust. When HP is full is 33.3% exhaust, but RPG_RT triggers the 34% case.

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -568,9 +568,11 @@ bool Game_BattleAlgorithm::Normal::Execute() {
 			crit_chance += crit_chance * weapon.critical_hit / 100.0f;
 			multiplier = GetAttributeMultiplier(weapon.attribute_set);
 		}
-		to_hit = (int)(100 - (100 - hit_chance));
-		if(weaponID != -1 && !Data::items[weaponID].ignore_evasion) {
-			to_hit *= (1 + (1.0 * (*current_target)->GetAgi() / ally->GetAgi() - 1) / 2);
+
+		if (weaponID != -1 && !Data::items[weaponID].ignore_evasion) {
+			to_hit = (int)(100 - (100 - hit_chance) * (1 + (1.0 * (*current_target)->GetAgi() / ally->GetAgi() - 1) / 2));
+		} else {
+			to_hit = (int)(100 - (100 - hit_chance));
 		}
 	} else {
 		// Source is Enemy
--- a/src/game_party.cpp
+++ b/src/game_party.cpp
@@ -18,6 +18,7 @@
 // Headers
 #include <algorithm>
 #include <cassert>
+#include <cmath>
 #include "system.h"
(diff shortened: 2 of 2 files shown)


src/game_party.cpp
```diff
@@ -546,19 +547,28 @@ int Game_Party::GetAverageLevel() {
...
+		total_sp += a->GetMaxSp();
+	}
+
+	// SP is always 33.3% of fatigue, which means a 0 SP actor is never above 66%
+	if (total_sp == 0) {
+		total_sp = 1;
 	}
 
-	return party_exh / (int)actors.size();
+	return (int)std::ceil(100 - 100.0 * (((float)hp / total_hp * 2.0f + (float)sp / total_sp) / 3.0f));
```
fdelapena: Nitpick: not sure how the compiler will handle this, but the 100.0 does not have f, so it might go from float to double.

Corner case check: is somehow possible to exist total_hp being zero (or null) to get a div0 error?

src/game_battlealgorithm.cpp
```diff
@@ -568,9 +568,11 @@ bool Game_BattleAlgorithm::Normal::Execute() {
 			crit_chance += crit_chance * weapon.critical_hit / 100.0f;
 			multiplier = GetAttributeMultiplier(weapon.attribute_set);
 		}
-		to_hit = (int)(100 - (100 - hit_chance));
-		if(weaponID != -1 && !Data::items[weaponID].ignore_evasion) {
-			to_hit *= (1 + (1.0 * (*current_target)->GetAgi() / ally->GetAgi() - 1) / 2);
+
+		if (weaponID != -1 && !Data::items[weaponID].ignore_evasion) {
+			to_hit = (int)(100 - (100 - hit_chance) * (1 + (1.0 * (*current_target)->GetAgi() / ally->GetAgi() - 1) / 2));
```
fdelapena: Corner case check: Is possible to exist Agi = 1 to get a div0 error?

src/game_party.cpp
Ghabry: The missing f is a valid complaint. Forgot it at this location (and we somehow use double everywhere else but isn't really worth the speed penalty).

src/game_battlealgorithm.cpp
Ghabry: In theory total_hp == 0 and agi == 0 if the database is not altered... Though about AGI I'm not 100% sure

Ghabry:
AGI 0 and HP 0 are not possible without editing the database manually, so will not sanitize this because we don't sanitize this anywhere :/

PR #1056, opened 2016-10-09
RPG2k battle message fixes

kakurasan:
Some RPG2k battle messages are incorrect. Also, lots of Japanese messages are broken.

#### Fix wrong uses of Data::terms.*

Wine + RPG_RT.exe:
wine-ally-critical
Unpatched ("会心の一撃！！" is expected in this game, not "痛恨の一撃！！"):
unpatched-ally-critical
Patched:
patched-ally-critical

#### Fix broken Japanese messages if the encoding is CP932

Wine + RPG_RT.exe:
wine-ally-1mplost
Unpatched (Broken Japanese message. Also, "攻撃力 (ATK)" is incorrect):
unpatched-ally-1mplost
Patched:
patched-ally-1mplost

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -28,6 +28,11 @@
 #include "rpg_skill.h"
 #include "util_macro.h"
 
+/** Interprets char literals as utf-8 */
+#ifdef _MSC_VER
+#pragma execution_character_set("utf-8")
+#endif
+
 static int max_hp_value() {
 	return Player::IsRPG2k() ? 999 : 9999;
 }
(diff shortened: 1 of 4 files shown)


carstene1ns:
First of all: Thank you for this contribution!

*I will not do much here, because @Ghabry is the battle system expert...* 😀

Question about `actor_critical` vs. `enemy_critical` and `actor_hp_absorbed` vs. `enemy_hp_absorbed`:
Did we switch the meaning of these or did we just switch the names? Otherwise they would need to be switched in [liblcf fields](https://github.com/EasyRPG/liblcf/blob/master/generator/csv/fields.csv#L410-L411).

About the Japanese messages fixes: Maybe is better to use a variable and assign it before (set to " " or the Japanese literal) to avoid having the [exact same lines](https://github.com/kakurasan/Player/blob/5c2015414c88f3cf8ba586478bec57ae4de5b8bd/src/game_battlealgorithm.cpp#L1053-L1055) in both conditions. For the other cases I am not actually sure if non-japanese games do not need the additional spaces or [strings](https://github.com/kakurasan/Player/blob/5c2015414c88f3cf8ba586478bec57ae4de5b8bd/src/game_battlealgorithm.cpp#L274) as well, so we could simplify that a bit then.

Now let's see if our CI compiles the changes flawlessly:
Jenkins: test this please

Ghabry:
Thanks for this. Didn't knew Japanese battle system inserts hardcoded particles in between.

As carstene1ns said I'm the battle expert and will review this, unfortunately I will likely not have time before Thursday to take a look, but I will as soon as possible.

Ghabry:
Jenkins: Test this please

Ghabry:
Jenkins: Test this please
with updated liblcf ;)

Ghabry:
Do you know how to do history rewriting (rebase) with git?
In that case you could remove commits 3ccd4fc, cec9179 and 613e749

Ghabry:
Jenkins: Test this please

Ghabry:
While testing the battles I noticed that animations are broken, probably another regression from the bitmap shrinking :/

PR #1057, opened 2016-10-09
Basic Mouse/Touch support

carstene1ns:
_The touch part is not tested, as I do not have such device._
(Maybe @fdelapena tries this out with his iPhone 😼)

### DIFF
--- a/resources/easyrpg-player.6.adoc
+++ b/resources/easyrpg-player.6.adoc
@@ -51,6 +51,9 @@ for some additional binary patches.
 *--enable-mouse*::
   Use mouse click for decision and scroll wheel for lists.
 
+*--enable-touch*::
+  Use one/two finger tap for decision/cancel.
+
 *--hide-title*::
   Hide the title background image and center the command menu.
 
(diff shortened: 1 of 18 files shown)


Ghabry:
As said in the other PR no time, so only one thing:
Afaik @zegeri had some touch movement code for the map, maybe would be a worth to add his code (not online yet?) to allow MV-like navigation :D

Zegeri:
>Afaik @zegeri had some touch movement code for the map, maybe would be a worth to add his code (not online yet?) to allow MV-like navigation :D

Not worth it for now, my code had some issues with Touch Hero/Collision With Hero triggers and didn't work in looping maps. Also I deleted that branch by accident ¯\\\_(ツ)_/¯.

carstene1ns:
I think this should stay open for test under iOS (needs some toolchain work before), will try to add swipe up/down left/right and will disable mouse for now (maybe add some `--enable-mouse` parameter) so we can make sure the touch input is recognized and not due some bug is using mouse instead (this may be platform/driver dependent, at least [is disabled by a check](https://github.com/carstene1ns/easyrpg-player/blob/31842f0ce210618c21cbdc0122c3eee63a9c46a4/src/sdl_ui.cpp#L844-L846), but no idea if it works).

carstene1ns:
This one could be merged for now, as the additional features are guarded by CLI options (and can be toggled on the fly by `scene_options` later).

PR #1058, opened 2016-10-13
Add CJK Font fallback.

### LINKED ISSUE #260: Add missing Chinese glyphs
Currently, EasyRPG Player does not have proper Chinese font support .

Here is a game with texts in Traditional Chinese:
http://download.seesaa.jp/contents/win/ent/rgame/11279/

Note: .zip archives contain files with non-ascii encoding

DBLobster:
Hi, I add a WenQuanYi 9pt font to resolve missing CJK glyphs.

WenQuanYi font is GPL v2 (with font embedding exception).

I only update automake script. Can someone update other platform build script?

Thanks.

master
wqy

-
Edit by carstene1ns:
Fixes #260.


Ghabry:
I'm not a lawyer so I can't say if GPLv2 with Font embedding exception is incompatible to GPLv3+.

I was worried about another thing: We plan to relicense the Player with a GPL Linking Exception which would be incompatible to GPLv2+FontEmbedding but you putted the fonts in a different file so for such a GPL-incompatible build we could just ifdef out that part :+1:

Ghabry:
Updating the other platform files is simple. Most platforms just compile all cpp files they can find, except CMake and Windows (Visual Studio).

Where to add the files:
CMakeLists.txt, just add file to the filelist.

VIsual Studio:
builds/vs2015/PlayerLib.vcxproj
Add a "ClCompile" block

builds/vs2015/PlayerLib.vcxproj.filters
Add new "ClCompile" with Filter "Source Files\Shinonome"

Ghabry:
Jenkins: Test this please

DBLobster:
Thanks.

I updated CMake and Visual Studio files.

Ghabry:
Jenkins: Test this please

Ghabry:
Tested via the chinese wadanohara game. Works for me on Windows and Linux (though not in Wine but that's not our problem) and all glyphs are displayed (can't say if the correct ones, lack of Chinese knowledge, so I trust the dev here ^^')

Ghabry:
@carstene1ns any opinion on this?

carstene1ns:
IMO we are covered by the FE, Player renders the font unmodified and makes a (dynamically generated) document.
About the size: well, will only make problems on our constrained platform ports.

Off topic: We already have a lot included in the executable, but leaving things out and making it modular is a big task and having different builds around (for example one for Chinese, one for Western and one for Japanese with different fonts) will make everything more complicated for now.

Ghabry:
With different font files I meant the embedded ones. When a chinese game is used we could use WenQuanYi instead of shinonome for kanji lookup because WenQ has chinese kanji font.
I don't want to make custom builds, just a runtime switch :D

PR #1061, opened 2016-10-14
MSVC: Use utf8 compiler argument introduced in VS2015.2

Ghabry:
And fix a bug detected by the debug crt.


carstene1ns:
We really need to update the build instructions to say that `VS2015 Update 2` is needed by now. The wiki still says `VS2010` and `VS2012` are supported and the README generally only states `Visual Studio`.
As the wiki stuff will change soon anyway, it would be nice if you could add the requirement in the README for now.

Ghabry:
Accepted

Ghabry:
Jenkins: Test this please <3

PR #1062, opened 2016-10-15
De-magic-numbify some Z layer settings and BattleAnim-Picture-Render Order

### LINKED ISSUE #1195: Embric of Wulfhammer's Castle: Heavy flickering in intro scene
The carriage scene of this game's intro has heavy flickering, which was not an issue when I last tried quite a while ago. I bisected this and it seems the following commit causes the issue:

https://github.com/EasyRPG/Player/commit/ce4787d716690ac62a7a5aada3ad0e5f15e3d0ab

Problem is showcased in the following video: https://my.mixtape.moe/ywaqhy.mp4 
This was tested in Linux. To reproduce, simply start a new game and go through the intro until the scene appears.

Ghabry:
I hope I got all Z layer usages and didn't break something :/
The Z layers are now categorized in categories and the category marker are the upper 8 bit minus the sign bit. So 127 categories are possible (like a IPv4 Class A network ;)).

I didn't touch the Map Z layer stuff because I don't want to break the event order rendering... (but is in theory possible to figure out now because is all in PriorityMap)

The order has some limitations. See game_picture.cpp comments.

Fix #1195

### DIFF
--- a/src/game_picture.cpp
+++ b/src/game_picture.cpp
@@ -51,19 +51,17 @@ void Game_Picture::UpdateSprite() {
 
 	sprite->SetX((int)data.current_x);
 	sprite->SetY((int)data.current_y);
-	if (Player::IsRPG2k3()) { // TODO: Including new RPG2k
+	if (Player::IsMajorUpdatedVersion()) {
 		// Battle Animations are above pictures
-		// ToDo: Include RPG Maker 2000 1.5 Value! and newer (How to detect?)
-		// ToDo: Exclude RPG Maker 2003 <1.05 (How to detect?)
-		sprite->SetZ(PriorityPictureNew + data.ID);
+		sprite->SetZ(Priority_PictureNew + data.ID);
 	} else {
 		// Battle Animations are below pictures
-		sprite->SetZ(PriorityPictureOld + data.ID);
+		sprite->SetZ(Priority_PictureOld + data.ID);
 	}
 	sprite->SetZoomX(data.current_magnify / 100.0);
 	sprite->SetZoomY(data.current_magnify / 100.0);
-	sprite->SetOx((int)(sprite->GetBitmap()->GetWidth() / 2));
-	sprite->SetOy((int)(sprite->GetBitmap()->GetHeight() / 2));
+	sprite->SetOx(sprite->GetBitmap()->GetWidth() / 2);
+	sprite->SetOy(sprite->GetBitmap()->GetHeight() / 2);
 
 	sprite->SetAngle(data.effect_mode != 2 ? data.current_rotation * 360 / 256 : 0.0);
 	sprite->SetWaverPhase(data.effect_mode == 2 ? data.current_waver : 0.0);
--- a/src/game_picture.cpp
+++ b/src/game_picture.cpp
@@ -51,7 +51,15 @@ void Game_Picture::UpdateSprite() {
 
 	sprite->SetX((int)data.current_x);
 	sprite->SetY((int)data.current_y);
-	sprite->SetZ(1100 + data.ID);
+	if (Player::IsRPG2k3()) { // TODO: Including new RPG2k
+		// Battle Animations are above pictures
+		// ToDo: Include RPG Maker 2000 1.5 Value! and newer (How to detect?)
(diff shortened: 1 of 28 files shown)


src/window_message.cpp
```diff
@@ -58,7 +58,7 @@ Window_Message::Window_Message(int ix, int iy, int iwidth, int iheight) :
 	}
 
 	visible = false;
-	SetZ(10000);
+		SetZ(PriorityWindow + 10);
```
fdelapena: accidental tab press

src/game_picture.cpp
```diff
@@ -55,7 +55,15 @@ void Game_Picture::UpdateSprite() {
 
 	sprite->SetX((int)data.current_x);
 	sprite->SetY((int)data.current_y);
-	sprite->SetZ(1100 + data.ID);
+	if (Player::IsRPG2k3()) { // TODO: Including new RPG2k
+		// Battle Animations are above pictures
+		// ToDo: Include RPG Maker 2000 1.5 Value! and newer (How to detect?)
+		// ToDo: Exclude RPG Maker 2003 <1.05 (How to detect?)
```
carstene1ns: Can this not be detected using kakurasan's changes?

src/sprite_battler.cpp
```diff
@@ -259,7 +259,7 @@ void Sprite_Battler::CreateSprite() {
 
 	SetX(battler->GetBattleX());
 	SetY(battler->GetBattleY());
-	SetZ(battler->GetBattleY()); // Not a typo
+	SetZ(PriorityBattle + battler->GetBattleY()); // Not a typo
```
carstene1ns: While we are at it, why not making this comment a bit more constructive?
Right as it is, this needs thinking over the purpose. Maybe something like this:
`// overlaps from top to bottom`

Ghabry: The typo was about using GetBattleY instead of GetBattleZ (which doesn't exist). Yes these lines basicly mean that battler with a lower Y (so higher in the scene) appear behind battlers with higher Y.

src/game_picture.cpp
Ghabry: Yes the code is outdated and already handled by my other patch :)

src/sprite_battler.cpp
carstene1ns: Yeah, I understand what the comment means, but in my opinion should be reworded to make it better / constructive (and more clear for new contributors).

-
`//TODO: This is wrong` is a bad comment.
`//TODO: Needs foo and bar to work correctly` is a better comment.

Ghabry: Yeah I agree with you

Ghabry:
For this issue I would vote for a move directly behind 0.5.1 because (the not commited) Z-changes in above/below/same as hero events to be prepared for the next RPG2k3 version (pokes Cherry) will likely cause bugs. :)

src/window_message.cpp
carstene1ns: Seems like you forgot to remove this 😁

Ghabry:
Needs some ingame-testing first, though. Havn't checked the battle systems yet, was just happy that it compiled...

src/window_message.cpp
```diff
@@ -58,7 +58,7 @@ Window_Message::Window_Message(int ix, int iy, int iwidth, int iheight) :
 	}
 
 	visible = false;
-	SetZ(10000);
+		SetZ(Priority_Window + 10);
```
carstene1ns: still accidental tab press here.

src/sprite_battler.cpp
```diff
@@ -269,7 +269,8 @@ void Sprite_Battler::CreateSprite() {
 
 	SetX(battler->GetDisplayX());
 	SetY(battler->GetDisplayY());
-	SetZ(battler->GetBattleY()); // Not a typo
+	// Battlers at the bottom appear above battlers at the top
```
carstene1ns: great comment!

Ghabry:
Regression: Somehow the SetXYZ line in Sprite_Battler is duplicated and appears in Update() again and I missed that line. So the Z-value is incorrect and monster become (sometimes) invisible. I think simply removing that 3 lines is safe. Will submit a fix

PR #1066, opened 2016-10-25
Fix GCW0 port

gameblabla:
I had to add a header in utils.cpp to make it compile on the GCW0. The new GCW0 toolchain has support for SDL2 so let's use that instead. Since the old script no longer worked, i modified it so it loads RTPs and Games from the home folder instead. Let me know if there are any issues.

### DIFF
--- a/builds/opendingux/Makefile
+++ b/builds/opendingux/Makefile
@@ -21,15 +21,22 @@ ifeq ($(BUILD_TARGET), dingoo)
 	FREETYPE_CFLAGS := -I$(SYSROOT)/usr/include/freetype2
 	ICU_CFLAGS := 
 	ICU_LIBS := 
+	LIBMPG123_CFLAGS := 
+	LIBMPG123_LIBS := 
+	LIBSNDFILE_CFLAGS :=
+	LIBSNDFILE_LIBS :=
 	CXXFLAGS = 
+	SDL_CONFIG = $(SYSROOT)/usr/bin/sdl-config
+	LIBSDLMIXER_LIBS = -lSDL_mixer
 else
 ifeq ($(BUILD_TARGET), gcwzero)
 	TARGET = gcw-zero/EasyRPG
 	TOOLCHAINDIR ?= /opt/gcw0-toolchain
 	SYSROOT = $(TOOLCHAINDIR)/usr/mipsel-gcw0-linux-uclibc/sysroot
 	BINDIR = $(TOOLCHAINDIR)/usr/bin
-	CFLAGS = -O2 -fomit-frame-pointer -ffunction-sections -ffast-math -G0 -std=gnu++11 -DLCF_SUPPORT_ICU
-	LDFLAGS = -Wl,--gc-sections
+	CFLAGS = -Ofast -fomit-frame-pointer -fdata-sections -ffunction-sections -mips32r2 -mmxu -std=gnu++11 -DLCF_SUPPORT_ICU 
+	CFLAGS += -DGCW_ZERO -DWANT_FMMIDI=2 -DHAVE_MPG123 -DHAVE_LIBSNDFILE -DWANT_FASTWAV
+	LDFLAGS = -Wl,--as-needed,--gc-sections -flto
 	PKG_CONFIG := $(BINDIR)/pkg-config
 	PNG_LIBS := $(shell $(PKG_CONFIG) --libs libpng)
 	PIXMAN_CFLAGS := $(shell $(PKG_CONFIG) --static --cflags pixman-1)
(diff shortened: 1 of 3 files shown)


Ghabry:
Hu, did you read our blog post? We stated that GCW0 is out of maintanance except if somebody cares to maintain. Can't test it obviously, but looks good to me. Just waiting for the other devs if they have anything to say.

Do you plan to publish compilations on some external site manually? In that case that's all to do. Otherwise we provide a service, that is compiling everything through our own CI system but for this we need a toolchain that works on x86_64 Linux and all crosscompiled libs. These buildscripts are one-click-deployed through our buildscript repo. See e.g. here for the Wii: https://github.com/EasyRPG/buildscripts/tree/master/wii . So if you want you can also provide such a deployment script for us.
Note about the script: You won't need Wildmidi (you use FMMidi) and you have to uncomment icu (which is now 58.1, not 56.1 btw) and comment iconv (not needed, ICU used). And if the lib is already available in the toolchain you don't have to compile it again obviously.

Ghabry:
Jenkins: Test this please (silly, but I want a green check)

builds/opendingux/gcw-zero/run.sh
```diff
@@ -1,11 +1,25 @@
 #!/bin/sh
-cd `dirname $0`
+SOURCE="$1"
+DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
```
Ghabry: Logic question: How does this result in $HOME/.easyrpg when $1 was not passed

gameblabla: I don't understand your question. $1 can not be empty because you need to choose a file in the file browser. Plus, even if $1 is not passed, $HOME always redirects to /usr/local/home, it doesn't matter.

Ghabry: Ah okay, was just confused why you forced "$HOME/.easyrpg" as project dir before and now not anymore but this explains it.

Ghabry:
Jenkins: Test this please <3

builds/opendingux/Makefile
```diff
@@ -21,30 +21,42 @@ ifeq ($(BUILD_TARGET), dingoo)
...
 ifeq ($(BUILD_TARGET), gcwzero)
 	TARGET = gcw-zero/EasyRPG
 	TOOLCHAINDIR ?= /opt/gcw0-toolchain
 	SYSROOT = $(TOOLCHAINDIR)/usr/mipsel-gcw0-linux-uclibc/sysroot
 	BINDIR = $(TOOLCHAINDIR)/usr/bin
-	CFLAGS = -O2 -fomit-frame-pointer -ffunction-sections -ffast-math -G0 -std=gnu++11 -DLCF_SUPPORT_ICU
-	LDFLAGS = -Wl,--gc-sections
+	CFLAGS = -Ofast -fomit-frame-pointer -fdata-sections -ffunction-sections -mips32r2 -mmxu -std=gnu++11 -DLCF_SUPPORT_ICU 
+	CFLAGS += -DGCW_ZERO -DWANT_FMMIDI=2 -DHAVE_MPG123 -DHAVE_LIBSNDFILE -DWANT_FASTWAV
+	LDFLAGS = -Wl,--as-needed  -Wl,--gc-sections -flto
```
carstene1ns: These two linker flag lines should be simplified to one line:

```makefile
LDFLAGS = -Wl,--as-needed,--gc-sections -flto
```

builds/opendingux/Makefile
```diff
@@ -56,10 +68,10 @@ BINFILES = $(foreach dir, $(DATA), $(wildcard $(dir)/*.*))
...
 
 # Compiler flags
-CFLAGS += -Wall -DUSE_SDL -DHAVE_SDL_MIXER `$(SDL_CONFIG) --cflags` $(PIXMAN_CFLAGS) $(FREETYPE_CFLAGS) $(ICU_CFLAGS)
-CXXFLAGS += $(CFLAGS) -fexceptions -fno-rtti
-LDFLAGS += -lstdc++ -lexpat -lSDL -lSDL_mixer -lz -lfreetype -lvorbisidec -lgcc -lm -lc -lpthread \
-           -ldl `$(SDL_CONFIG) --libs` $(PIXMAN_LIBS) $(ICU_LIBS) $(PNG_LIBS)
+CFLAGS += -Wall -DUSE_SDL -DHAVE_SDL_MIXER `$(SDL_CONFIG) --cflags` $(PIXMAN_CFLAGS) $(FREETYPE_CFLAGS) $(ICU_CFLAGS) $(LIBMPG123_CFLAGS) $(LIBSNDFILE_CFLAGS)
+CXXFLAGS += $(CFLAGS) -fno-rtti -fno-exceptions
+LDFLAGS += -lstdc++ -lexpat $(LIBSDL_LIBS) -lz -lfreetype -lvorbisidec -lgcc -lm -lc -lpthread \
+           -ldl `$(SDL_CONFIG) --libs` $(PIXMAN_LIBS) $(ICU_LIBS) $(PNG_LIBS) $(LIBMPG123_LIBS) $(LIBSNDFILE_LIBS)
```
carstene1ns: The use of `sdl-config` and `LIBSDL_LIBS` makes us link here with the same lib again (Yes, we did this before, too. and yes, `--as-needed` will take care of it).

Issue #137, opened 2013-08-11
Graphic of charsets don't appear well


gadesx:
I think that I found the reason of the wrong graphics of charsets, 
if the size of the Charset is 288x256 all work fine, but if this have other size like 128x128 the graphics give problems.

project with tests (see 128x128 with rm2k3 and easy and see)
http://www.mediafire.com/?s1xsdb27sc1scp4

that was the reason why my game had missing graphics
http://img593.imageshack.us/img593/1936/5lya.png

PR #1067, opened 2016-10-29
Add large CharSet support

fdelapena:
This recovers large CharSet support removed since #137 fix,
allowing CharSet files starting with a dollar sign to be
calculated by 4x2 SpriteSets. This differs from VX Ace behavior.
The purpose of using large 4x2 spritesets instead individual
spritesets like VX Ace does is to keep reduced file access
on games wanting to use multiple large characters.
Also keeps code lesser complicated for this custom feature.

Test case to use with the PR:
[largecharset.zip](https://github.com/EasyRPG/Player/files/559675/largecharset.zip)

### DIFF
--- a/src/sprite_character.cpp
+++ b/src/sprite_character.cpp
@@ -25,8 +25,8 @@ Sprite_Character::Sprite_Character(Game_Character* character) :
 	character(character),
 	tile_id(0),
 	character_index(0),
-	chara_width(24*(TILE_SIZE/16)),
-	chara_height(32*(TILE_SIZE/16)) {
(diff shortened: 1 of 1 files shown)


Ghabry:
Works for me and looks funny ;)

screenshot_0
screenshot_1

Ghabry:
As mentiomed in chat output.h is a useless include but is probably a problem in general in the codebase ;)

carstene1ns:
Yeah, I would have run `include-what-you-use` already to fix cases like this, but this tool is soo broken... 👎

PR #1069, opened 2016-10-29
Battle 2k3: Implement "Enable Combo", skill reflection, (de)buff and minor things

Ghabry:
Enable combo simply repeats the whole action and reduces a counter.

Skill reflect is a bit more "fun" but was surprisingly easy to implement in the BattleAlgorithm (good API design): GetTarget lies to the caller and returns the source when reflected. Therefore battleAlgorithm will handle the whole damage calc and applycation automatically correct. 
The only exception is PlayAnimation because the animation is played twice: Once normal and once on the reflected target.

(De)buffing is that [x] Reduce Resistance option. When checked this buffs or debuffs the Attack Elements by 1/-1. Maximum shift is +-1 and when e.g. an actor is buffed with +1 and the enemy debuffs it becomes 0.

Related: #821

Bonus:
2k3: Damage numbers "jump" now
Multitarget attacks did not target all actors when one was dead
"Attack party" flag for weapons
Preemptive attack for 2k

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -685,6 +685,20 @@ bool Game_Actor::HasStrongDefense() const {
 	return GetData().super_guard;
 }
 
+bool Game_Actor::HasPreemptiveAttack() const {
+	const RPG::Item* item = GetEquipment(RPG::Item::Type_weapon);
+	if (item && item->preemptive) {
+		return true;
+	}
+	if (HasTwoWeapons()) {
+		item = GetEquipment(RPG::Item::Type_weapon + 1);
+		if (item && item->preemptive) {
+			return true;
+		}
+	}
+	return false;
+}
+
 const std::vector<int16_t>& Game_Actor::GetSkills() const {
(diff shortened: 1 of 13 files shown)


Ghabry:
Here some bad video (somehow offset is wrong): https://gfycat.com/ObviousBogusIberianchiffchaff

Uses "3x combo" and Poison causes reflect. As you can see the first poison attack is not reflected but then the enemy gets poison and the next 2 are reflected.
Afterwards Ice is used which is reflected 3 times.

Ghabry:
And looks like our damage calculation is still wrong

Ghabry:
Done. Will not add "Attack twice with weapon" and "Two weapon attack" in this PR because this needs a bigger BattleAlgorithm change...

**Merge warning**: When #1050 merged don't merge this until I fixed the GetEquipment calls! (Same for #1069 before #1050)

Ghabry:
Right, have to update GetEquipment. Thank you past self

PR #1074, opened 2016-11-01
Android Game Browser: Add XYZ image support

Ghabry:
If you wonder why a new lib: Because I don't want to load SDL while in the gamebrowser.
The XYZ is converted to PNG (without file-io writes in a buffer) and this is returned to Java and processed by the normal Android Image API.

Before:
screenshot_2016-11-01-17-52-58

After:
screenshot_2016-11-01-17-50-10


Ghabry:
For reference: To create a stub header invoke in "app/src/main/java":
```
javah -cp . org.easyrpg.player.java.file
```

Where "java.file" points to the Java file (without .java ending) that contains the native method.

fdelapena:
LGTM, despite it duplicates the code, keeps the JNI code separated, though maybe could be ifdef'd somehow and use some function overload or whatever, not sure if the point was to keep the code pollution free this way.

Ghabry:
Add a guard to allow loading of the GameBrowser even if library loading fails for devs that forget to run ndk-build. @BlisterB approves this :P

BlisterB:
Hahahaha, I finally used to take the .so files from the .apk. Really quicker !

It's good to see all of your nice PRs ! I'm sorry to not be really active, I don't have the motivation to code out of work (maybe Dark Souls 3 is a part of the explanation). I hope I will have the motivation to rework input layout in the next months :).

Issue #898, opened 2016-05-16
Yume Nikki misc problems


Zekario:
**Name of the game**: Yume Nikki

**Player platform**: Linux

**Describe the issue in detail and how to reproduce it**: In the latest release, Neon World causes a crash shortly after entering. Sleeping in the spaceship causes it to 'drag' part of the screen as well (https://i.sli.mg/oWGnoy.png).

I tried using the git versions of player and liblcf, but it crashes when trying to access the save screen.

Here is my log, the only thing I did was enter the Neon World:

```
[2016-05-16 15:55:24] Debug: EasyRPG Player 0.4.1 started
[2016-05-16 15:55:24] Debug: ============================
[2016-05-16 15:55:24] Debug: CLI: --encoding shift-jis 
[2016-05-16 15:55:24] Error: Couldn't initialize audio.
ALSA: Couldn't open audio device: File descriptor in bad state
[2016-05-16 15:55:29] Debug: EasyRPG Player 0.4.1 started
[2016-05-16 15:55:29] Debug: ============================
[2016-05-16 15:55:29] Debug: CLI: --encoding shift-jis 
[2016-05-16 15:55:29] Debug: Opened audio at 44100 Hz (stereo), format: S16LSB
[2016-05-16 15:55:29] Debug: Using . as Game and Save directory
[2016-05-16 15:55:29] Debug: Loading game ゆめにっき
[2016-05-16 15:55:29] Debug: Using RPG2k3 Interpreter
[2016-05-16 15:55:31] Debug: Loading Save Save01.lsd
[2016-05-16 15:55:31] Debug: Loading Map Map0003.lmu
[2016-05-16 15:55:31] Debug: Tree: 現実の部屋
[2016-05-16 15:55:39] Debug: Loading Map Map0007.lmu
[2016-05-16 15:55:39] Debug: Tree: 夢ベランダA < 夢の部屋
[2016-05-16 15:55:40] Debug: Loading Map Map0006.lmu
[2016-05-16 15:55:40] Debug: Tree: 夢の部屋
[2016-05-16 15:55:44] Debug: Loading Map Map0009.lmu
[2016-05-16 15:55:44] Debug: Tree: 扉部屋
[2016-05-16 15:55:44] Debug: Image size out of bounds: Panorama/扉部屋背景  (80x80 < 560x496 < 640x480)
[2016-05-16 15:55:50] Debug: Loading Map Map0020.lmu
[...]

PR #1077, opened 2016-11-14
Fix lack of transparency after rotating

Ghabry:
This is a regression introduced with the performance optimisations for bitmap blits.
In this case getting rid of the fast path doesn't hurt much because is only taken when a rotation happens.

Related: #898


Ghabry:
Test case: [Save06.lsd.zip](https://github.com/EasyRPG/Player/files/590098/Save06.lsd.zip)

Go next to the steam thing and press "1" multiple times. When no UFOs appear reenter and repeat. When they become stationary you will see that there is no black border anymore

screenshot_0
screenshot_1

PR #1078, opened 2016-11-16
Overclocked GPU/BUS,added shaders,disabled vsync.

Rinnegatamante:
This pull will do four different things:

- Adds filters (aka shaders) support (at the moment there are 4 shaders available: None, Sharp Bilinear, LCD 3x, xBR). These will be interchangeable during player execution by the user by swinging the touchscreen left/right.

- Increases GPU, GPU crossbar and BUS clocks (from 111 mhz to 222 mhz) allowing shaders to get the best results.

- Disables vsync. This will produce a slight frameboost without any bad conseguence.

- Fix some typo errors and updates Makefile for new vitasdk compatibility.

### DIFF
--- a/src/psp2_ui.cpp
+++ b/src/psp2_ui.cpp
@@ -67,6 +67,62 @@ char* shader_names[] = {
 	"xBR"
 };
 
+namespace {
+	vita2d_texture* gpu_texture;
+	vita2d_texture* next_texture;
+	vita2d_texture* main_texture;
+	uint8_t zoom_state;
+	int in_use_shader;
+	bool set_shader;
+	SceUID GPU_Mutex, GPU_Cleanup_Mutex;
+}
+
+static int renderThread(unsigned int args, void* arg){
+	
+	for (;;){
+	
+		sceKernelWaitSema(GPU_Mutex, 1, NULL);
+		memcpy(vita2d_texture_get_datap(gpu_texture), vita2d_texture_get_datap(next_texture), vita2d_texture_get_stride(gpu_texture)*240);
+		sceKernelSignalSema(GPU_Mutex, 1);
+		
+		sceKernelWaitSema(GPU_Cleanup_Mutex, 1, NULL);
+		
+		if (main_texture == NULL) sceKernelExitDeleteThread(0); // Exit procedure
+		
+		vita2d_start_drawing();
+   
+		if (set_shader){
+			Output::Post("Shader set to %s.",shader_names[in_use_shader]);
+			set_shader = false;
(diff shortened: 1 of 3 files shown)


Ghabry:
You messed up the rebase, needs a squash.

src/psp2_ui.h
```diff
@@ -27,13 +27,9 @@
...
-vita2d_shader* shaders[SHADERS_NUM];
-char* shader_names[] = {
-	"None",
-	"Sharp Bilinear",
-	"LCD 3x",
-	"xBR"
-};
+
+extern vita2d_shader* shaders[];
+extern char* shader_names[];
```
Ghabry: Why are you declaring the shaders as extern? Are they used anywhere outside of psp2_ui.cpp?

src/psp2_ui.cpp
```diff
@@ -47,14 +58,30 @@ AudioInterface& Psp2Ui::GetAudio() {
 
 int _newlib_heap_size_user = 192 * 1024 * 1024;
 
+vita2d_shader* shaders[SHADERS_NUM];
+char* shader_names[] = {
+	"None",
+	"Sharp Bilinear",
+	"LCD 3x",
+	"xBR"
+};
```
Ghabry: If the reason for extern are strange linker errors: Declare these variables as "static"

Ghabry:
Just noticed that the Makefile lacks a check if TOOLCHAIN_DIR is defined, but isn't your fault.

Ghabry:
I get linker errors.
Please move the global variables above class Psp2Ui from psp2_ui.h into an anonymous namespace (namespace {}) in the .cpp file. Fixes it for me

Ghabry:
A user on gbatemp reported that this made (even with shaders disabled) some games, e.g. Ara Fell, significantly slower. Theres no obvious reason why. Could be related to how the upstream-version handles audio, lets wait for the new audio code to be finished.

Ghabry:
@Rinnegatamante 
A user reports a massive slowdown (15 FPS) in "Ara Fell" since the gpu-code changes, even in my "sdl_con_audio_sin_mixer" branch. Maybe scheduling or new vitasdk problem.

Ara Fell Demo Download:
http://queenscourt.org/othergames.php

Rinnegatamante:
I doubt it's an issue related to vitasdk, maybe it could be related to the new gpu rendering thread + lots of sounds thread working.

Ghabry:
We confirmed already that even with the new audio code it runs bad, but havn't tested rolling back to the non-render-thread version yet.

PR #1087, opened 2017-01-19
Fix some issues in Wolfenhain

Ghabry:
The test for Battle Commands was completely wrong (only worked for 1st actor)

The subskill fix is the same one used in other places: The check must be >=.
The diff is a mess because of the identation change, the only relevant change is from
```
case RPG::Skill::Type_normal:
case RPG::Skill::Type_subskill:
```
to
```
if (skill.type == RPG::Skill::Type_normal || skill.type >= RPG::Skill::Type_subskill) {
```

The item_usage change will cause minor issues when loading EasyRPG savegames: Items with more than one use will have an incorrect usage count until the usage limit is reached once.
Unlimited and Single use items are not affected. Unlimited doesn't care about uses at all and Single will wrap on first use (because 2 >= 1)

### DIFF
--- a/src/game_battle.cpp
+++ b/src/game_battle.cpp
@@ -286,7 +286,7 @@ bool Game_Battle::AreConditionsMet(const RPG::TroopPageCondition& condition) {
 	}
 
 	if (condition.flags.command_actor &&
-		condition.command_id != Game_Actors::GetActor(condition.turn_actor_id)->GetLastBattleAction())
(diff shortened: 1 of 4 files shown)


carstene1ns:
You can use this GitHub secret to ignore whitespace changes: https://github.com/EasyRPG/Player/pull/1087/files?w=1 :octocat:

PR #1089, opened 2017-01-24
Fix asciidoc manpage generation when using a different working folder

fdelapena:
When trying to work with gnome-builder, this issue emerged. The change in Makefile.am seems to fix this.
Also adds the gnome-builder .buildconfig to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,6 +65,7 @@ test-driver
 /builds/wii/*.dol
 /output
 /utils
+.buildconfig
 compile
 CPackConfig.cmake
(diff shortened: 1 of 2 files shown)


carstene1ns:
Need to test if this works with `make dist-check`, because it also does a VPATH / out-of-source build.

fdelapena:
How to reproduce the issue: call, from a different working folder than the containing the `configure` script, e.g.: `cd /tmp/; /path/to/script/configure` (it might ask for a `make distclean` first, just do it).

Issue #1082, opened 2016-12-14
HOME (OFF fangame) Hang when entering level 'SH1'


Zekario:
#### Attach files
[save.zip](https://github.com/EasyRPG/Player/files/650667/save.zip)
[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/650668/easyrpg_log.txt)

#### Describe the issue in detail and how to reproduce it:
Entering the room to the upper left causes Player to hang and take up all of one CPU. Tested with the continuous build.

Ghabry:
This somehow happens when too many events are pushing each other and the event in front can't move.
HOME has 15 events that push each other. In my debug build it starts to lag with 4 events.
So it's not getting stuck, just running too slow.

Ghabry:
And every of these 15 events has 15 Move right events with attribute "skip if not possible". So this accumulates to tons of "makeway" calls when cascaded from the leftmost to the rightmost :/

scurest:
I'm still breaking things with that `MakeWay` commit - -;;;;;

The `frame_count_at_last_update_parallel` field is supposed to prevent this: it lets us know if we already tried to update on the current frame so we only visit every character once in a frame. That's what [this](https://github.com/EasyRPG/Player/blob/921a708f36db92d88189af360c7579ae74c032c5/src/game_player.cpp#L367) block does. I just failed to also place this block in `Game_Event::UpdateParallel`.

If you add it there, the Player won't lock up.

Ghabry:
Great, because I had no idea how to fix it :)

Ghabry:
I fixed this in the development branch and added some comments to make the purpose more clear.

fmatthew5876:
I retested this with #1687 

Everything works and there is no noticeable change in CPU usage (Ubuntu linux).

Issue #1072, opened 2016-10-31
2k3 scene_order redo option is broken


fdelapena:
After sorting the party in the scene order (tested with 4 actors only, worth to check with lesser), it will display a selectable to confirm and redo. After pressing cancel key or selecting Redo, the decision key to retry the sorting won't work.
![captura de pantalla de 2016-10-31 10-16-12](https://cloud.githubusercontent.com/assets/174885/19861898/e29fc596-9f53-11e6-97be-8f839d837249.png)

Ghabry:
Does any game use this useless scene btw?

fdelapena:
This is a general feature accessible from the menu on default 2k3 games, not critical in any case but just wanted to document this glitch

Ghabry:
I know, was just curious :D

Issue #962, opened 2016-07-27
Embric of Wulfhammer's Castle: Equipped items with skills not correctly counted in inventory in battles


Siltala:
This is probably the most trivial bug ever, but hey, it's still a bug! :P

If you have an item that has a skill equipped, enter combat and look at your inventory, the game reports you have 0x of said item. It is still usable, though. RPG_RT correctly reports that you do have 1x of said item.

Screenshot:
![screenshot_2](https://cloud.githubusercontent.com/assets/2873449/17193725/057df9da-545d-11e6-951b-3b60acfa2792.png)
- Game download: https://aeresland.net/EmbricDELUXE_2.2.2.7z
- Savegame: http://www.heavylobster.com/files/Save08.lsd

Ghabry:
Theres another bug here. When the item is equipped and in the inventory the item appears twice in the menu.

Issue #920, opened 2016-07-03
Freeze while resizing the variable array multiple times


carstene1ns:
**Name of the game**: thouhou alive 1.07 (東方ALIVE)

**Player platform**: Android (likely all)

**Describe the issue in detail and how to reproduce it**:

The variable array is endlessly increased by some event, making the game unresponsive.
Issue submitted by mail:

> A movie stops at a prolog by the illusion volume. Is this the bug?

Log excerpt:

```
[2016-07-03 18:17:28] Debug: Loading Map Map0557.lmu
[2016-07-03 18:17:28] Debug: Tree: VS魔理沙 < マヨヒガ城 < 幻想編Chapter1
[2016-07-03 18:17:57] Debug: Resizing variable array to 1101 elements.
[...]
[2016-07-03 18:23:46] Debug: Resizing variable array to 71587 elements.
```

fdelapena:
It happens with both original or translated version.

[Game download](https://onedrive.live.com/?cid=F87EF6D0A3619382&id=F87EF6D0A3619382%21105) (Japanese, unpack with `unar`).
[Game download](http://www.mediafire.com/?twb96m5aqax9vde) (English translation by vgperson).

How to reproduce:

`easyrpg-player --start-map-id 557`

It starts after a couple of seconds, goes black screen, event with ID 0 exceeds execution limit a few times while increasing some array size elements.

Here after 71587 elements resizing it goes to the title screen.
With the vgperson's translated version it ends after 76587 elements.

Ghabry:
Have a patch which makes the resizing more efficient (always reserve id + 1000) and the output less verbose (only 10 per run).

Zegeri:
> Have a patch which makes the resizing more efficient (always reserve id + 1000) 

I'm not sure why would that be more efficient. This reallocates the vector every time, not once the capacity has been reached. This means that the whole vector could be reallocated 76587 times.

Ghabry:
You are right, without checking ``capacity()`` the patch makes no sense at all.

PR #1092, opened 2017-01-28
Fix #716, #920, #962, #1072 and #1082

### LINKED ISSUE #716: Event sprites are not being rendered on the left when their visibility is below half tile
Video to understand the graphical glitch: https://fran.cr/nextcloud/index.php/s/vLt1uCm3H6aAQ6g
The first column shows people and doors disappearing in the left screen edge when the visible part is around lesser than 8 px (half tile). The WebM video could be reproduced at 0.5x speed for better checking. This is also easier to check in the Yume Nikki doors scene.

Ghabry:
All of them are minor stuff. See commit logs.

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -226,10 +226,28 @@ void Game_Actor::ChangeEquipment(int equip_type, int item_id) {
 	}
 }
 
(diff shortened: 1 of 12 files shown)


carstene1ns:
works for me™

Issue #646, opened 2015-11-02
Event interpreter unit testing


Ghabry:
Many times we observed now that old interpreter issues that got fixed are broken due to new bugfixes.

Because the interpreter is the most fragile part of our Player I would suggest new "unit tests" which consist of a minimal database, and a single (or 2 for teleporting) map file containing an event.

Things to do:
- [ ] Write a DynRPG plugin that traces event execution, this is the reference file.
- [ ] Write a --trace-event command that outputs event execution to a file (with same format as the DynRPG event trace file)
- [ ] Compare both files -> Test passed when equal

Ghabry:
https://gist.github.com/Ghabry/0c8557fae9cabc071e76

dynrpg code

Ghabry:
Made some progress

DynRPG code repo: https://github.com/Ghabry/dynrpg-eventtracer

Changes made to Player: https://github.com/Ghabry/easyrpg-player/tree/event_unittest

Ghabry:
I need some help from automake experts ;)

The TESTS need a different working directory:
Must be "tests/" otherwise the event tracer will assert (because it can't find the trace log file)

How to supply command line arguments to tests? At least --project-path will be required.

Ghabry:
Move Route logging is now implemented in DynRPG plugin (via custom ASM code).
Only needs the Player part now (much easier ;))

Ghabry:
What do you think? Should the traces (the log files) be stored in the event-test game or in the tests folder of player?

fdelapena:
How worth is separating event trace tests from Player tests?
IMO if there are too many event trace tests it might be worth to have them separated and keep Player tests more Player specific, however having 2 different repository tests affecting Player behavior might be confusing, if we want to run tests when doing "make check" or "make distcheck" from the tarball source itself then I think it should be in the same repository for easier generation.

About running tests from the `tests/` folder: Creating a `Makefile.am` in `tests/` and adding `tests/Makefile` to `AC_CONFIG_FILES()` in `configure.ac` (it might need to re-run `autoreconf` to regenerate configure.in) and move `sometest_SOURCES = tests/sometestsource.cpp` to the new `tests/Makefile.am` without the `tests/` part: `sometest_SOURCES = sometestsource.cpp`. About passing parameters, I've linked some days ago a way described in stackoverflow using shell scripts instead of calling the binary executable directly.

Ghabry:
[...]

PR #1096, opened 2017-02-04
Record/replay input logs

scurest:
This PR lets you record and replay a log of all the input given to Player.

This moves towards #646.

------

Record a log with

    $ EasyRPG_Player --record-input input.log

and then replay it with

    $ EasyRPG_Player --replay-input input.log

If you also use the same `--seed`, you should (AFAIK) get to watch the exact same playthrough you just recorded.

The intention is that this could eventually be used for automated regression testing, as bug reporting data, to confirming bugs still exist, etc. If Ghabry's event tracer can be rigged to give compatible output it could also be used for compatibility testing with RPG_RT. Right now, you could technically use it to look for changes between Player versions by comparing save files (ie. record a playthrough where you save, modify Player, replay the log, diff the `.lsd` files). But in order to do this fast, we'd need to add a way to run the Player "headless" (ie. don't emit graphics and run updates as fast as possible) or else you'd have to watch the whole playthrough every time you do it :p

### Input Log Format

The input log is plain-text. There is one line for each call to `Input::Update` (this is roughly, but not exactly, one line per frame) that contains the `InputButton`s that were pressed when the call happened. The buttons are formatted as with `operator<<(std::ostream&, const std::bitset<BUTTON_COUNT>&)`.

The format is really just proof-of-concept and we'll probably want something more efficient and robust. If the general idea of this PR looks good and there's no bugs, I can open an issue afterwards to talk about a good format for this.


Ghabry:
Thx for the PR, will take a look later.

@scurest 
Android ships with different stdc++ versions. A more up-to-date version of stdc++ is the clang-version which is provided by Android, too. But by now we were unable to switch the c++-library because this results in strange build errors while compiling our libraries. Android is a real can of worms.

scurest:
Thanks. Np on the Android thing; the workaround isn't really worse or anything.

Ghabry:
Works for me.
There is one potential desync issue: When savegames are available the Title scene will highlight "Load". Maybe always focus "New game" when record/replay are enabled?

Ghabry:
And the replays will not sync on different platforms because "rand()" depends on the used libc. This can be resolved by replacing rand() everywhere with Utils::GetRandomNumber() which uses std::mt19937 which generates the same output on all platforms (at least the standard says so, untested).

scurest:
Yeah, you definitely need to restore the save files to reproduce a run. Even ignoring that desync, the saves have a `save_count` field that will ruin your diffs if you don't. I want to avoid having any actual Player logic depend on the input though. Is it okay to just add that the save files need to be restored too to the docs for `--replay-input`?

-----

I have a branch with the relevant fixes for `rand`. In small experiments, the save files do reproduce perfectly on it (except for the timestamp field). They still aren't portable though, because the algorithm for `uniform_int_distribution` isn't part of the standard ;) I was going to do a separate PR for that though.

Now that I think of it, there might also be an issue with the BGM ticks. Do you know how reproducible those are? It looks like they depend on the audio implementation.

Ghabry:
For this prototype implementation it's good enough for me when you document that savegames must be part of the movie. Can be fixed later using a different movie format / movie folder which somehow stores the savegames together with the movie.

Okay, then I ignore rand for now.

The timestamp is a design flaw in liblcf. Not liblcf should set the timestamp but the caller of liblcf. This is a problem in lcf2xml, too because the timestamp gets updated when converting back and forth :/. If you want you can submit a patch to liblcf which removes the "GenerateTimestamp" call and put it at an appropriate spot in Player.

BGM_Ticks doesn't return the correct value by now. When BGM_Ticks ever gets fixed it should return the same output on any platform and the only game I know which depends on BGM_Ticks is one of the "Mother" games which has the combo system implemented in an acustic way (not working in EasyRPG yet).

scurest:
Oh, that's why my save files look the same after I run them through `lcf2xml`! :)

PR #1104, opened 2017-02-19
Make RNG portable

scurest:
The intention of this PR is to make random number generation totally portable by:

#### Removing `rand` and `srand`

Replacing `rand() % a` with `GetRandomNumber(0, a-1)` generally, but also

* I changed [this line](https://github.com/EasyRPG/Player/compare/master...scurest:rand?expand=1#diff-34f34f675ce52452da855f1469fc5421R728) which was probably a bug
* `rand() / (RAND_MAX / a + 1)` was replaced with `Utils::GetRandomNumber(0, a - 1)` (is this right?)
* I also added `ChanceOf` for the common case of getting a bool with a certain probability (this could also be used for the battle algorithms)

#### Removing `uniform_int_distribution`

The algorithm it uses isn't part of the standard.

My first idea was to copy the `uniform_int_distribution` from (say) GCC's stdlib here, but it's a _lot_ of code, and it's complicated because it has to deal with RNGs with non-power-of-two ranges. We only need to consume random u32s and generate random i32 ranges, so I wrote a rejection sampler for just that case.

Hopefully I didn't get any of the math wrong (please check!).

-----

The upshot is that the seed \*\**should*\*\* determine the RNG sequence and the RNG should now be reproducible.

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -17,7 +17,6 @@
 
 // Headers
 #include <algorithm>
-#include <cstdlib>
 #include <iostream>
 #include <sstream>
 #include "game_interpreter.h"
@@ -48,6 +47,7 @@
 #include "util_macro.h"
 #include "reader_util.h"
 #include "game_battle.h"
+#include "utils.h"
 
 Game_Interpreter::Game_Interpreter(int _depth, bool _main_flag) {
 	depth = _depth;
@@ -736,7 +736,7 @@ bool Game_Interpreter::CommandControlVariables(RPG::EventCommand const& com) { /
 			int a, b;
 			a = max(com.parameters[5], com.parameters[6]);
 			b = min(com.parameters[5], com.parameters[6]);
-			value = rand() % (a-b+1)+b;
+			value = Utils::GetRandomNumber(b, a);
 			break;
 		case 4:
 			// Items
@@ -1379,7 +1379,7 @@ bool Game_Interpreter::CommandSimulatedAttack(RPG::EventCommand const& com) { //
 		result -= (actor->GetSpi() * spi) / 800;
 		if (var != 0) {
 			int rperc = var * 5;
-			int rval = rand() % (2 * rperc) - rperc;
+			int rval = Utils::GetRandomNumber(-rperc, rperc - 1);
 			result += result * rval / 100;
 		}
 
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -24,6 +24,9 @@
 
 namespace {
 	std::mt19937 rng;
+
+	/** Gets a random number uniformly distributed in [0, U32_MAX] */
+	uint32_t GetRandomU32() { return rng(); }
 }
 
 std::string Utils::LowerCase(const std::string& str) {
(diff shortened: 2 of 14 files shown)


src/game_interpreter.cpp
```diff
@@ -736,7 +736,7 @@ bool Game_Interpreter::CommandControlVariables(RPG::EventCommand const& com) { /
 			int a, b;
 			a = max(com.parameters[5], com.parameters[6]);
 			b = min(com.parameters[5], com.parameters[6]);
-			value = rand() % (a-b+1)+b;
+			value = Utils::GetRandomNumber(b,a);
```
Ghabry: Interestingly the help of the official english version (both 2k/2k3) says the interval is exclusive ("For example, if the range “0~100” is specified, the value used will be a random number above 0 but below 100.").
RPG Advocates translation says "between 0 and 100" and Dons version doesn't explain the interval at all.

But I verified that the interval is indeed inclusive, code is correct.

src/utils.cpp
```diff
@@ -338,9 +342,44 @@ void Utils::SwapByteOrder(double& d) {
...
+
+	// Rejection sampling:
+	// 1. Divide the range of uint32 into blocks of max+1
+	//    numbers each, with rem numbers left over.
+	// 2. Generate a random u32. If it belongs to a block,
+	//    mod it into the range [0,max] and accept it.
+	// 3. If it fell into the range of rem leftover numbers,
+	//    reject it and go back to step 2.
+	uint32_t m = max + 1;
+	uint32_t rem = -m % m; // = 2^32 mod m
```
Ghabry: Is that correct? -m % m is always 0.

src/utils.h
```diff
@@ -152,6 +152,15 @@ namespace Utils {
...
+	 * Has an n/m chance of returning true. If n>m, always returns true.
+	 *
+	 * @param n number of times out of m to return true (non-negative)
+	 * @param m denominator of the probability (positive)
+	 * @return true with probability n/m, false with probability 1-n/m
+	 */
+	bool ChanceOf(int32_t n, int32_t m);
+
+	/**
 	 * Seeds the RNG used by GetRandomNumber
```
Ghabry: I request a minor documentation change here: "used by GetRandomNumber and ChanceOf"

Ghabry:
I'm not sure if ``rand() / (RAND_MAX / a + 1)`` has an off-by-one error but your code does what was intended.

Good to see GetRandomNumber/ChanceOf everywhere now to improve the randomness because our old code has a non-even randomness-distribution because most of the code only uses the incorrect rand() % number code :/.

src/utils.cpp
scurest: Because `m` is unsigned, `-m` is equal to 2^32 - m here ([cite](http://eel.is/c++draft/expr.unary.op#8)) and 2^32 - m mod m = 2^32 mod m.

Ghabry: Ah now I get it. Didn't knew that it's promoted to unsigned beforehand, interesting trick :D

Ghabry:
Theres a problem with the random-ness.

Test code:
```cpp
	std::array<int, 6> arr;

	for (int i = 0; i < 10000000; ++i) {
		arr[Utils::GetRandomNumber(0, 5)]++;
	}

	for (int i : arr) {
		printf("%d\n", i);
	}
```

Output: Multiple runs give a similiar bad distribution.
```
11801336
1668455
19106669
1665291
18570228
1666413
```

scurest:
@carstene1ns Oh yeah, I do do that, don't I (when the arguments are short). Sorry! I amended the spaces in.

PR #1107, opened 2017-02-27
Screen shake fixes

### LINKED ISSUE #1101: Screen shake animation should draw black on left/right sides if there is no map information
[Video of the issue](https://www.youtube.com/watch?v=5nnSXs0gzhU). This was made using the elf file of [build 120](https://ci.easyrpg.org/job/player-3ds/120/).

When the screen shakes on a map that's only 20 tiles wide, there is no pixel information on the left and right sides. In the old player, the left and right sides simply draw black pixels in this case. (If the map is wider than 20 tiles, we see the actual tiles instead of black pixels.)

In EasyRPG 3DS, we instead see duplicated pixels on the left and right sides.

scurest:
More fixes to follow up #1098

* Black should now be drawn out of bounds of the map (fixes #1101)
* Parallax backgrounds now shake too (before they didn't). The behavior is different for looping maps and non-looping maps. For non-looping maps, basically any area of the screen that is drawn black because it is out of the map bounds is still drawn black (regardless of if the panorama loops). The panorama covers everything else.

    out

### DIFF
--- a/src/plane.cpp
+++ b/src/plane.cpp
@@ -20,6 +20,8 @@
 #include "graphics.h"
 #include "player.h"
 #include "bitmap.h"
+#include "main_data.h"
+#include "game_map.h"
 
 Plane::Plane() :
 	type(TypePlane),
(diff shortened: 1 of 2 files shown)


Ghabry:
Thanks for the gif, perfect for our 0.5.1 blog post :)

scurest:
P-please use a real game and not my ugly test map for the blog post >_>;;;

PR #1116, opened 2017-03-05
Update translations for 0.5.1

carstene1ns:
(let this stay open until release, maybe we get some new strings until then)


carstene1ns:
Nothing new on transifex 🙁, looks like this could be merged then.

PR #1117, opened 2017-03-06
Add U+2212 and U+301C glyphs

### LINKED ISSUE #514: Missing U+2212 and U+301C glyphs
Seems like SHIFT_JIS uses U+2212 glyph for the minus sign. It is currently missing from Shinonome.
Also, misses U+301C (wave dash) used in the change hero name scene).
Additionally, having a graphical representation of 0xFFFD (replacement character "�") would be nice to have, used to display wrong encoding conversion sometimes.

Seems to fail when rendering some Japanese titles in the game browser, which are already Unicode, at least for minus sign and wave dash.

## 
- [x] U+2212 (minus sign)
- [x] U+301C (wave dash)
- [x] 0xFFFD (replacement character "�")

kakurasan:
* Write SHINONOME_WQY into shinonome.h
* Add new file for additional full-width characters
* Add U+2212 glyph (copied from shinonome/kanjic/font_src.bit)

Fixes #514.

Note: It seems we need to add the glyph for *full-width* minus sign ("0x81 0x7C").


Ghabry:
Can you provide a game that uses U+2212 and fails currently?
I added SHIFT-JIS 0x817C and it renders as a "-" with your patch and without.

kakurasan:
> I added SHIFT-JIS 0x817C and it renders as a "-" with your patch and without.

You're right. I modified some texts (to display "81 7C") in TestGame-2000 but "-" is rendered instead of "�(U+FFFD)" without the patches.

RPG_RT(Win):
Windows

EasyRPG Player (unpatched):
EasyRPG Player

The patches work in game browser when there are directories that use U+2212.
And I noticed that U+301C is not rendered in game browser, so I added the glyph.

unpatched -> patched

> Can you provide a game that uses U+2212 and fails currently?

Unfortunately, I don't know any game that outputs "glyph not found: 0x2212".

Ghabry:
I didn't think about the filebrowser, good idea!

Ghabry:
Due to the Shift-JIS-to-unicode mapping it's probably not possible to get these glyphs ingame. But can confirm your Game Browser screens

PR #1122, opened 2017-03-10
Add fast forward option

### LINKED ISSUE #1010: Implement Option: Skip Frames
I would like to see a option in the debug menu, where you can disable the fps limit, so the game runs faster (like some snes/gba/... emulator can do). It's not about increasing the fps to something higher, so the game runs smoother, but to make it faster. I hope you get the idea :)

- [x] Visual indication
- [x] Android-support

Ghabry:
This adds speedup (x3) via F and x10 via F+PLUS.

Fix #1010

### DIFF
--- a/src/graphics.cpp
+++ b/src/graphics.cpp
@@ -35,7 +35,6 @@
 namespace Graphics {
 	void UpdateTitle();
 	void DrawFrame();
-	void DrawOverlay();
 
 	int framerate;
 
@@ -45,7 +44,6 @@ namespace Graphics {
 	BitmapRef black_screen;
 	BitmapRef screen1;
 	BitmapRef screen2;
-	bool frozen;
 	TransitionType transition_type;
 	int transition_duration;
 	int transition_frames_left;
@@ -192,8 +190,6 @@ void Graphics::DrawFrame() {
 	for (Drawable* drawable : global_state->drawable_list) {
 		drawable->Draw();
 	}
-
-	//DrawOverlay();
 
 	DisplayUi->UpdateDisplay();
 }
--- a/src/graphics.cpp
+++ b/src/graphics.cpp
@@ -26,6 +26,7 @@
 #include "cache.h"
 #include "baseui.h"
 #include "drawable.h"
+#include "input.h"
 #include "util_macro.h"
 #include "output.h"
 #include "player.h"
@@ -34,8 +35,12 @@ namespace Graphics {
 	void UpdateTitle();
 	void DrawFrame();
 	void DrawOverlay();
+	std::string GetFpsString();
 
+	/** Frames per second */
 	int fps;
+	/** Logic updates per second */
+	int ups;
 	int framerate;
 
 	void UpdateTransition();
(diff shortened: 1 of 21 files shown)


Ghabry:
What I can observe: The "⏩" is not rendered on my 4.0.3 system...
The speedup is bit buggy, at the beginning the framerate increases (this isn't supposed to happen, must stay at 60) and when you release the key while this happens the game still runs too fast and doesn't return to 60.

Ghabry:
Looks like with 4.1 it shows something, so I only need a fallback ">" for lower apis.

Ghabry:
Theres an unexpected problem here: When Vsync is enabled the fast forward doesn't work because the SDL UpdateDisplay calls block until the VSync happens. To solve this I could spawn a std::thread and do the SDL call there.

Ghabry:
Okay we have a problem here: Video output doesn't work in Android when I do the rendering from an extra thread because the EGL context is on the wrong thread :(

BlisterB:
Arf, android...
If you are afraid that the character ">>" doesn't show on every devices, you can create a class inheriting from the button class, and override the method that draw the button (by drawing it manually). Or you can just add a condition on the drawing method from the button class.

Ghabry:
I already fixed this, by rendering ">" in that case. GOod enough for old devices. The Android stuff won't be part of this PR anymore, because I have to fix it first >.>

src/graphics.cpp
```diff
@@ -189,25 +191,11 @@ void Graphics::DrawFrame() {
 		drawable->Draw();
 	}
 
-	DrawOverlay();
+	//DrawOverlay();
```
carstene1ns: Debug leftover

src/input_buttons.h
```diff
@@ -60,7 +60,7 @@ namespace Input {
 		PAGE_DOWN,
 		SCROLL_UP,
 		SCROLL_DOWN,
-		BUTTON_COUNT
+		FAST_FORWARD,		BUTTON_COUNT
```
carstene1ns: missing `\n`

src/player.h
```diff
@@ -238,6 +243,13 @@ namespace Player {
 	/** Game title. */
 	extern std::string game_title;
 
+	/**
+	 * The default speed modifier applied when the speed up button is pressed
+	 *  Only used for configuring the speedup, don't read this var directly use
+	 *  GetSpeedModifier() instead.
+	 */
+	extern int speed_modifier;
```
carstene1ns: Why is this variable needed anyway?

Ghabry: The idea was to have an exported variable that a potential setting dialog could alter to get more/less speedup. But currently its useless :D

carstene1ns: okay, granted.

Ghabry:
Just found a much smarter way to handle the updating.
Has two disadvantages compared to the old code but on the other hand keeps Player::update almost untouched (before Update contained tons of code):

Input is handled when speedup is on as-if the key was held for 3 (10) frames. Makes navigating while speedup is on a bit tricky (menus are unusable), just turn off the speedup before using the cursor ;).
All audio played in the 3 (10) frames starts simultaniously. Probably not noticable because the difference is a few milliseconds.

carstene1ns:
I get the `Framerate is 0 FPS` message when the gamebrowser launches, seems like the fps counter is not initialized yet or something.
Also, currently sounds are played at full length, this leads to `No free channels available` warnings more often.

PR #1126, opened 2017-03-15
Refactor/Bug fixes for Pictures

### LINKED ISSUE #1046: MovePicture does not disable Sine Effect set by ShowPicture
This affects some games, for example [_Rosy Despair_](http://s000.tinyupload.com/index.php?file_id=38911809574183228004) (and likely the countdown in the _Super Smash Brothers_ clone)

The ShowPicture has Sine with strength 9 set, the first MovePicture has it disabled, the second sets it again:
[event](https://picload.org/image/rddiawcp/20161001-101854.png)

To reproduce in Rosy Despair use `--start-map-id 40 --new-game`
Here is also a dedicated Testcase: [issue1046.zip](https://github.com/EasyRPG/Player/files/504208/issue1046.zip)

| Player | RPG_RT |
| --- | --- |
| player | rpg_rt |

### LINKED ISSUE #1063: Picture should have transparency and rotation angle should reset
There are 2 bugs reported by Brassman in the following test game:
https://www.sendspace.com/file/biiqrd
(Requires English 2003 RTP).

- [ ] There is a switch to start picture rotation. When switched off and switched on again, it continues from the previous rotation angle instead starting from the proper angle.
- [X] There is also a transparency issue when rotating the picture, which might be a regression.

scurest:
Fixes #1046, fixes #1063. There are also some other small fixes to picture behavior.

`Game_Picture` was refactored substantially. `Command{Show,Move}Picture` now have much less responsibility, and the public surface area of `Game_Picture` was reduced. Some tiny changes to PicPointerPatch handling were also made.

### DIFF
--- a/src/game_interpreter_map.cpp
+++ b/src/game_interpreter_map.cpp
@@ -581,6 +581,82 @@ bool Game_Interpreter_Map::CommandPanScreen(RPG::EventCommand const& com) { // c
 	return true;
 }
 
+// PicPointerPatch handling.
+// See http://cherrytree.at/cms/download/?did=19
+namespace PicPointerPatch {
+
+static void AdjustId(int& pic_id) {
+	if (pic_id > 10000) {
+		int new_id;
+		if (pic_id > 50000) {
+			new_id = Game_Variables[pic_id - 50000];
+		} else {
+			new_id = Game_Variables[pic_id - 10000];
+		}
+
+		if (new_id > 0) {
+			Output::Debug("PicPointer: ID %d replaced with ID %d", pic_id, new_id);
+			pic_id = new_id;
+		}
+	}
+}
+
+static void AdjustParams(Game_Picture::Params& params) {
+	if (params.magnify > 10000) {
+		int new_magnify = Game_Variables[params.magnify - 10000];
+		Output::Debug("PicPointer: Zoom %d replaced with %d", params.magnify, new_magnify);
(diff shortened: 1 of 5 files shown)


Ghabry:
All my picture test cases pass:
https://easyrpg.org/play/pr1126/?game=issue_1046&engine=rpg2k3
https://easyrpg.org/play/pr1126/?game=issue_1046  (No effect for the "No Rotate/Wave -> Then Rotate/Wave" cases due to a RPG2k bug that is emulated)

Ghabry:
Download link of the PicPointer Patch Demo game: http://cherrytree.at/cms/lang/en/download/?did=19

src/game_interpreter_map.cpp
```diff
@@ -581,179 +581,171 @@ bool Game_Interpreter_Map::CommandPanScreen(RPG::EventCommand const& com) { // c
...
+		params.bottom_trans = new_bottom_trans;
+	}
+}
+
+static void AdjustShowParams(int& pic_id, Game_Picture::ShowParams& params) {
+	AdjustId(pic_id);
+	AdjustParams(params);
+
+	// Adjust name
 	if (pic_id >= 50000) {
```
Ghabry: Bug is probably here because AdjustId overwrites the pic_id to the real number, so this check fails.

scurest: Just as you said! Thanks.

PR #1130, opened 2017-03-22
Fix BG bug introduced in fc546f70

scurest:
I introduced this bug

screenshot_0

in fc546f70. The rendering problem is caused by the bug and the interaction between a vertically looping map, transparent tiles, and a parallax BG. Details are in the commit message.

Quick PR to fix it.


Ghabry:
This is probably the same bug as in this Space Funeral video at 8:05
https://www.youtube.com/watch?v=a81cRXlMHfI

scurest:
Yeah, that's definitely it ^^; I can confirm this fixes that scene in Space Funeral.

carstene1ns:
Tested, works for me.

Issue #1129, opened 2017-03-20
Battle interpreter must halt execution when any battle end condition (Win/Lose/Escape) is fulfilled [The Huntress of the Hollow]


Ghabry:
Breaks: The Huntress of the Hollow

[Save03.lsd.zip](https://github.com/EasyRPG/Player/files/856177/Save03.lsd.zip)

This breaks the 1st ending in that savegame. (Select beheading the wolf when asked in mid-battle and use the behead skill).
The event script will set the wolf HP to -100% to kill it and then calls a Common Event which hangs the game (bug ^^)
This common event is never executed in RPG_RT because killing the Wolf is a victory condition and the interpreter stops.

Issue #1127, opened 2017-03-20
Abort Battle should not execute the Victory handler [The Huntress of the Hollow]


Ghabry:
Breaks "The Huntress of the Hollow"

[Save03.lsd.zip](https://github.com/EasyRPG/Player/files/856168/Save03.lsd.zip)

This breaks the 2nd ending in that savegame. (Select sparing the wolf when asked in mid-battle) because the Victory handler runs and sets variables to the "killed the wolf" case.

PR #1131, opened 2017-03-22
Fix game breaking bugs in "The Huntress of the Hollow" and "ThisEvent" savegame loading issue

Ghabry:
See issues #1127 and #1129 
Additionally this fixes the tutorial. Bug in "CHange Class" command (thx @carstene1ns)

Fixed the Event 10005 is invalid problem after loading a savegame. See https://community.easyrpg.org/t/some-little-player-bugs/352/8

Also removed some unused battle functions (just some cleanup)

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -187,6 +187,11 @@ bool Game_Actor::UnlearnSkill(int skill_id) {
 	return false;
 }
 
+void Game_Actor::UnlearnAllSkills() {
+	GetData().skills.clear();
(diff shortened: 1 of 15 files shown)


carstene1ns:
Did you test the "CommandChangeClass" change? I must admit, I only cared for HOTH, which uses case 2 (learn additionally).

Ghabry:
No havnt't tested case 1 yet, assumed you did it yes :D. will do

Ghabry:
Thanks for the patch, applied

PR #1136, opened 2017-03-30
Fix regression when "Abort Battle" command is used and the Encounter-command has no handler

Ghabry:
Reported by mail, fighting the woodchopper in eternal legends repeats endlessly.

This basicly means that the abort handling code in the interpreter was broken since years but was dead code until I fixed the abort case for huntress of the hollow :D

When "Abort Battle" is used and there is no other handler (Victory, Escape or Defeat) for the battle moving the interpreter to "End Battle" failed and the battle repeated. The interpreter is now advanced by one line when this happens.

Also fixed this for the defeat and the escape handler but here the bug is only possible through a malformed database because the "Execute handler" settings are part of the Enemy encounter dialog, so they are always generated when these options are on.

This fixes a regression introduced in commit 11246f2a

### DIFF
--- a/src/game_interpreter_map.cpp
+++ b/src/game_interpreter_map.cpp
@@ -287,8 +287,10 @@ bool Game_Interpreter_Map::ContinuationEnemyEncounter(RPG::EventCommand const& c
 		case 1:
 			return CommandEndEventProcessing(com);
 		case 2:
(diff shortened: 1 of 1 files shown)


Ghabry:
Jenkins: Test this please

Issue #747, opened 2016-01-31
Make the README file better and distribute it along the binary.


carstene1ns:
Currently, the README file is mainly targeted at building the Player executable.

It should:
- [ ] be polished and also contain some information only available in the wiki currently
- [ ] use the Markdown format to allow advanced rendering
- [ ] be distributed along the binary in the downloads on https://easy-rpg.org (for some platforms it could be renamed to `.txt` to avoid confusion ~~Windows~~)

PR #1139, opened 2017-04-03
RTP stuff

carstene1ns:
- Add support for our RTP
- Supports multiple RTPs to be set from environment
- Adds official English 2k RTP names


src/filefinder.cpp
```diff
@@ -462,36 +462,52 @@ void FileFinder::InitRtpPaths(bool warn_no_rtp_found) {
...
+	// Our RTP is for all engines
+	read_rtp_registry("EasyRPG", "RTP", "path");
+
 	add_rtp_path("/data/rtp/" + version_str + "/");
 #endif
+	std::vector<std::string> env_paths;
+
+	// Windows paths are split by semicolon, Unix paths by colon
+	std::function<bool(char32_t)> f = [](char32_t t) {
+		return t == ';' || t == ':';
```
Ghabry: This must be if-defed because Splitting at ":" will damage the path on Windows due to the drive letters being C:\

resources/easyrpg-player.6.adoc
```diff
@@ -132,15 +132,18 @@ arguments are supported:
...
-  Run-Time-Package(RTP).
+  Full path to a directory containing an extracted RPG Maker 2003 RTP.
 
 'RPG_RTP_PATH'::
-  Full path to a folder containing both RTPs.
+  Full path to a directory containing a combined RTP.
+
+NOTE: All '*_RTP_PATH' variables support directory lists, using colon (':') or
+semicolon (';') as separator. Useful when you have multiple translated RTP
+versions or directories with extra files.
```
Ghabry: Remove "or semicolon(';')" because no manpage support unter Windows

carstene1ns:
Our manual page is not really limited by the platform, everthing documented there should work under Windows as well. We could even ship a html version with the `player.exe` (see also #747).

Ghabry:
Jenkins: Test this please
Makes sense, I revoke my review.

PR #1143, opened 2017-04-15
Different bitmap fonts

rohkea:
This pull request:

* Adds separate ‘official’ fonts and ‘unofficial’ fonts:
    * Official fonts are Shinonome Mincho and Shinonome Gothic (they are only Japanese characters differ in those fonts). They are compatible with MS Mincho/Gothic (`this' is displayed like ‘this’). They are used when an engine is used with the default encoding (Japanese engine + Japanese encoding = official fonts).
    * Unofficial fonts are RM2000 and RMG2000; they are compatible with Don Miguel’s fonts. They are used when an engine is used with a non-standard encoding (Japanese engine + Cyrillic encoding = unofficial fonts). RMG2000 is based on Shinonome’s existing font, RM2000 is edited to match Don Miguel’s RM2000.
* Adds a possibility to change the font in the System graphics.

Here is a comparison of how this looks like:

* The first row is Don Miguel’s translation, for comparison;
* the second row is the new ‘unofficial’ fonts RM2000 and RMG2000;
* the third row is the ‘official’ fonts Mincho and Gothic; unfortunately, the difference is not seen as only Japanese characters are different. Mincho can be edited to add serifs to Latin letters. 

font-comparison

I’ve removed the ``font`` field from the ``Bitmap``, and now pass font to text-drawing routines. This is neccessary because otherwise, the ``Font::Default()`` is only called once for every bitmap, when the bitmap is instantiated, and then it’s difficult to switch from Mincho to Gothic (from RMG2000 to RM2000).


rohkea:
I’ve replaced the RM2000-compatible font with [ttyp0](http://www.typophile.com/node/92340), as suggested in chat. The left screenshot now looks like this (the right one hasn’t changed, attaching it for completeness):

screenshot_2screenshot_3

The best thing about ttyp0 is its wide coverage of scripts. We just got support for Georgian, Armenian and Greek without doing anything. Here’s the first stanza from Sappho’s Hymn to Aphrodite:

screenshot_0

Ghabry:
Please check your whitespace, is partly space instead of tab.

About TextDraw: Because you pass Font::Default() everywhere anyway how about simply hardcoding the font to Font::Default() in TextDraw instead of adding a FontRef argument?

rohkea:
> Please check your whitespace, is partly space instead of tab.

Sorry! I’ve changed by vim settings so this shouldn’t happen again.

The whitespace was in the function for checking if Engine is an official release. I’ve think this function is not needed. I’ve noticed that the official English release doesn’t use the same font as Japanese release (e.g. … is European, not Japanese in RPG2k3E), so I think it makes sense to use ttyp0/RMG2000 for RPG2k3E too, in the same way as for Don’s translation.

> About TextDraw: Because you pass Font::Default() everywhere anyway how about simply hardcoding the font to Font::Default() in TextDraw instead of adding a FontRef argument?

Yes, I can do this. I don’t think anyone relies on being able to change the font.

resources/wenquanyi/README
```diff
@@ -20,20 +20,22 @@ from wqy-bitmapsong-bdf-1.0.0-RC1.tar.gz
 wenquanyi_cjk_basic_9pt.bdf
 GPLv2+FE
 
+is generated by the following coommand:
```
rohkea: I've added this change to make the English text correspond to the Chinese text below.

Ghabry:
Font comparison: And picture 2 has a problem with missing glyphs.

resources/wenquanyi/README
carstene1ns: typoo

rohkea:
Thanks for noticing the problems! I’ve fixed the problem Shinonome Mincho not falling back on Shinonome Gothic:

screenshot_0

kakurasan:
The "displayed in Mincho" behavior seems unexpected.

histoire / RPG_RT on Japanese Windows (Gothic):
histoire / RPG_RT on Japanese Windows
histoire / Player 0.5.1 (Gothic):
histoire / Player 0.5.1
histoire / patched (Mincho):
histoire / patched

Yume Nikki / RPG_RT on Japanese Windows (Gothic):
Yume Nikki / RPG_RT on Japanese Windows
Yume Nikki / Player 0.5.1 (Gothic):
Yume Nikki / Player 0.5.1

Also note that lots of "glyph not found:" messages are displayed in Japanese games.

rohkea:
@kakurasan Thank you very much for the testing, I wouldn’t have found this myself!

> The "displayed in Mincho" behavior seems unexpected.

Look like I’ve swapped Mincho and Gothic by accident! I’ll fix this in less than an hour.

> Also note that lots of "glyph not found:" messages are displayed in Japanese games.

Thanks for noticing! Shinomome Mincho doesn’t contain all the characters and is expected to fallback on Shinonome Gothic, so those messages are not useful. I’ll change the code so that the messages are only shown if a glyph is missing in all the fonts, not just in the first font.

I’ll update my pull request in less than an hour.

rohkea:
I’ve swapped Gothic with Mincho for Japanese games, the codes are now like this:

* Gothic=ID 0 in Japanese games (might need to add it for Chinese and Korean, since they also use … in the middle of the line, but I don’t know how to check for Chinese and Korean),
* Mincho=ID 1 in Japanese games (might need to add it for Chinese and Korean),
* ttyp0=ID 0 in non-Japanese games,
* RMG2000 (Latin for Shinonome with minor edits)= ID 1 in non-Japanese games.

And I’ve changed the error reporting: now the error is only displayed when all the fonts lack a glyph, not when just one font lacks a glyph.

Ghabry:
Somehow the font switching doesn't work for me. I always get that new, small font. I assumed that RMG2000 should still give Shinonome? And isn't RMG2000 ID=0?

Based on RPG2k:
- RMG2000
- RM2000 (by Don)
my diff says:

```diff
⋊> ~/Spiel/RPG M/OFF diff RPG_RT_RM2000.xml RPG_RT_RMG2000.xml
133947c133947
<     <font_id>1</font_id>
---
>     <font_id>0</font_id>
```


About detecting other languages, see Player::IsCP932.

The checks for korean would be: "windows-949-2000" (or "949") and for chinese: "windows-936-2000" and "936". For traditional chinese I don't have a game floating around. Simply run it and check what "Detected encoding:" shows.

Ghabry:
Further tests: Setting the font in the database doesn't work at all. I always get that new small font (RM2000?).

And in tests I get this: Shouldnt it be inverted?
screenshot_20170424_162629

rohkea:
> And isn't RMG2000 ID=0?

I think there is an error in Don’s RM2k translation. The Editor says that RMG2000 is 0, RM2000 is 1. HOWEVER, the RPG_RT displays RM2000 for 0, RMG2000 for 1. This pull request matches the behaviour of the Player, so it gives ttyp0 as ID 0, RMG2000 for ID 1.

This is also consistent with the official RPG Maker 2000 release, which uses RPG2000 (serif font) for ID 0, RPG2000G (sans-serif font) for ID 1.

I can swap them if you want, as there are little consistency between the versions: some RPG_RT translations assign IDs to RM2000 and RMG2000 differently.

> And in tests I get this: Shouldnt it be inverted?

Maybe? I’m following Don’s RPG_RT (not editor) and the official English RPG Maker 2000 release. Some other releases seem to assign font IDs differently. It depends on what we should consider a reference.

> Further tests: Setting the font in the database doesn't work at all.

Indeed, I can confirm this bug. Thanks for noticing! I’ll fix it.

Ghabry:
Jenkins: Test this please ♥

rohkea:
Just as a showcase of Armenian support, here’s an extract from a poem [Like Wild Lilies](https://pionline.wordpress.com/2016/03/13/marine-petrossian-translated-from-armenian-with-the-author-and-interviewed-by-arthur-kayzakian/) by Hovhannes Shiraz:

screenshot_0

I’ve used lcf2xml to add Armenian text because EasyRPG editor is not finished yet, and RPG Maker 2000/2003 doesn’t support Armenian.

carstene1ns:
Btw. this increases our executable size by ~100 kb.

rohkea:
This pull request adds 470 characters in BITMAPFONT_RMG2000 and 3074 characters in BITMAPFONT_TTYP0.

If the size is important, we can reduce size by:

* In BITMAPFONT_WQY, remove characters that are covered by SHINONOME_GOTHIC. Since WenQuanYi is only used as a fallback, those characters are never displayed. This can decrease the character count by ~7000.
* In BITMAPFONT_RMG2000, only keep the characters that are different from SHINONOME_GOTHIC (i.e. Cyrillic and punctuation). Since BITMAPFONT_RMG2000 was partly copied from SHINONOME_GOTHIC, it has overlaps. This can reduce the character count by ~200.

The following changes could be visible to the user:

* Remove fullwidth Cyrillic from SHINONOME_GOTHIC and use narrow Cyrillic BITMAPFONT_RMG2000 instead. This is not compatible with MS Mincho/Gothic, but I doubt any Japanese game depends on Cyrillic being fullwidth. In fact, I doubt any Japanese game uses Cyrillic.
* Remove some scripts from ttyp0. The obvious candidates are Thai and Hebrew, because those scripts are not currently supported (no RTL support for Hebrew, no diacritic placement for Thai).

I think removing unused characters from WenQuanYi seems the best way to reduce the font size. It could also be nice to merge shinonome/generate_cxx_font.rb and wenquanyi/gen_wqy_font_cpp.py into one script, since they basically do the same thing.

Ghabry:
Finally found some time to test games under Windows for visual comparison. That ttyp0 font matches quite closely to the RPG_RT font but the RPG_RT font is waaaay uglier (completely forgot about this, too much used to EasyRPG fonts) :D

carstene1ns:
Yeah, same here. I do not really like the `r` in ttyp0, but I guess that is just nitpicking.

PR #1146, opened 2017-04-18
Android: Fix fast forward button and text centered in buttons

### LINKED ISSUE #1124: [Android] Fast Forward Button not rendered
Hi everyone :).

The Fast Forward option works like a charmed, really cool!

But the button is not rendered on my Device (Sony Xperia Z3 Compact, API 23), it just displays a rectangle.

@Ghabry if you are bored about this I can fix that with an hard coded shape, I can also add the option to customize the FF button : "press and release" and "press 1 time to active, press 1 time to desactivate".
As you wish, we can also split the two subject :).

BlisterB:
I've changed the character for fast forwarding by "»", the ancient one was too problematic.
I also fixed the text which wasn't centered all time on buttons (Android is a mess for those kinds of things)

Fixes #1124.


BlisterB:
I agree with you Ghabry, a rectangle is a better shape for this kind of buttons.
This code was the first step of a refactoring, I should have mentioned it.

I just added "press and release" behavior, the next step is to put back the rectangle shape.

BlisterB:
The rectangle shape has returned :).

Ghabry:
I'm not really sure how to label that checkbox for locking the Fast Forward button.

Don't really think "Enable the press-and-release behavior for Fast Forwarding button" is obvious.

Does the Android compat toolkit provide a toggle button?
I think something like.
```
Fast Forward Button: Hold (o ) Toggle
```
is better interface design.

https://cdn-images-1.medium.com/max/800/0*Yj3LanbQyOH9cvfW.jpg

BlisterB:
For what I understood, toggle buttons are for Apple design.
The Android guideline doesn't recommend such buttons for a 2 options choice (we have the checkbox for "on/off" choice). Toggle buttons are recommended when used in clusters:
https://material.io/guidelines/components/buttons.html#buttons-dropdown-buttons

But the sentence I used is probably not the best. Maybe something like is better: 
**Fast Forward Button:** 
[x] Toggle behavior

What do you think?

Ghabry:
The problm is that it's not clear that the opossite of toggle is hold.

With toggle button I didn't mean what Android calls togglebutton but what is in my ascii art.

Google calls this "Switch": https://material.io/guidelines/components/selection-controls.html#

carstene1ns:
There is a backward compatible version: https://developer.android.com/reference/android/support/v7/widget/SwitchCompat.html

carstene1ns:
layout

carstene1ns:
The remaining question is here, should the slider and drop-down change size on bigger screens? Currently there is a gap.

Ghabry:
Imo the slider should match the other sliders and take the full space.

For the combobox I think the current width is fine but I guess it uses the same Ui group as the slider, so just expand both.

carstene1ns:
We should really redesign the option screens then. They look really bad on tablets.

carstene1ns:
Will redesign this later (#1207), ready to merge.

PR #1152, opened 2017-04-23
3DS: cleanup and bug fixing

carstene1ns:
Video of the only new feature:
[video](https://my.mixtape.moe/aiegxa.webm)

---

3DS PR builder passed: https://ci.easyrpg.org/view/Pull%20Requests/job/player-3ds-pr/27/console

### DIFF
--- a/src/3ds_ui.cpp
+++ b/src/3ds_ui.cpp
@@ -173,17 +173,17 @@ bool CtrUi::IsFullscreen() {
 void CtrUi::ProcessEvents() {
 	hidScanInput();
 	u32 input = hidKeysHeld();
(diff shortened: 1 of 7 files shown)


carstene1ns:
The DSP missing case will just stay silent and add the message to the log, tested with `.cia` version here.
Unfortunately on 3DS audio is initialized too early to display the warning on screen. :/
When refactoring this some day, would be a good idea to call "early platform init code" before `Player::Init();` in `main.cpp`.

Ghabry:
Maybe Output could buffer such messages until the screen is ready.

carstene1ns:
About performance loss when using the pen too much (results in drawing on bottom screen): Not noticable in yume nikki on n3ds at least.

PR #1154, opened 2017-04-27
Use hiragana and katakana only for Japanese games in hero name scene and support Hangul

### LINKED ISSUE #364: Name scene should show upper 7 bit characters for current game encoding only
Setting japanese glyphs for a non-japanese game will show mojibake after saving and loading due to encoding conversion.

Currently it lacks support for non-japanese encodings. Glyph list proposal for missing encodings are welcome.

fdelapena:
This is a workaround which fixes #364.

The proposed approach displays hiragana and katakana for CP932 games and basic latin and symbols for the rest of encodings. Japanese games lose access to the symbol table, but original RPG_RT also does not support this.

This also adds limited Hangul script to the scene.

### DIFF
--- a/src/scene_name.cpp
+++ b/src/scene_name.cpp
@@ -53,16 +53,16 @@ void Scene_Name::Update() {
 		if (name_window->Get().size() > 0) {
 			Game_System::SePlay(Game_System::GetSystemSE(Game_System::SFX_Cancel));
 			name_window->Erase();
-		}
-		else
+		} else {
 			Game_System::SePlay(Game_System::GetSystemSE(Game_System::SFX_Buzzer));
+		}
 	} else if (Input::IsTriggered(Input::DECISION)) {
 		Game_System::SePlay(Game_System::GetSystemSE(Game_System::SFX_Decision));
 		std::string const& s = kbd_window->GetSelected();
 
 		assert(!s.empty());
 
-		if(s == Window_Keyboard::DONE || s == Window_Keyboard::DONE_JP) {
+		if (s == Window_Keyboard::DONE || s == Window_Keyboard::DONE_JP) {
 			Game_Temp::hero_name = name_window->Get();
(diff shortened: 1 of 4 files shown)


Ghabry:
Use function IsCP932 pls

fdelapena:
> Use function IsCP932 pls

Fixed 👍

carstene1ns:
Looks much better now.

src/scene_name.cpp
```diff
@@ -40,7 +40,13 @@ void Scene_Name::Start() {
 	face_window->Refresh();
 
 	kbd_window.reset(new Window_Keyboard(32, 72, 256, (SCREEN_TARGET_WIDTH/2)));
-	kbd_window->SetMode(Window_Keyboard::Mode(Game_Temp::hero_name_charset));
+	if (Player::IsCP932()) {
```
carstene1ns: A comment on how this affects the language would be nice here, e.g.:
```c++
// Japanese pages
// Korean pages
// ASCII pages
```

src/window_keyboard.cpp
```diff
@@ -33,57 +32,85 @@ const char* const Window_Keyboard::TO_KATAKANA = "<カナ>";
 const char* const Window_Keyboard::TO_HIRAGANA = "<かな>";
 const char* const Window_Keyboard::DONE_JP = "<決定>";
 
+const char* const Window_Keyboard::TO_HANGUL_1 = "<앞P>";
+const char* const Window_Keyboard::TO_HANGUL_2 = "<뒤P>";
+const char* const Window_Keyboard::DONE_KO = "<결정>";
+
 /*
- * hiragana -> katakana -> letter -> symbol -> hiragana -> ...
+ * hiragana <-> katakana; letter <-> symbol
```
carstene1ns: missing hangul here.

carstene1ns:
Great. Will merge when jenkins is ready.

PR #1161, opened 2017-05-09
Fix #1160; regression from 7f83a3bf

### LINKED ISSUE #1160: Eternal Legends: panorama movement is wrong
The panorma in Eternal Legends’s Peak of Prophecy is wrong in the current master branch and in the 0.5.1 release (but it was correct in 0.5.0):

screenshot_30

Here’s a test project with just this map: [TestPanoramaScroll.tar.gz](https://github.com/EasyRPG/Player/files/983896/TestPanoramaScroll.tar.gz) (~500 kB).

If someone needs to test it in the real game, here’s the save game: 
[Save02.lsd.gz](https://github.com/EasyRPG/Player/files/983900/Save02.lsd.gz). The game itself (~25 MB) can be downloaded here: <https://www.file-upload.net/download-7313515/EL-V2.27a.exe.html> or here <http://rmarchiv.de/games/595> (requires registration).

The problem did not exist in the Player 0.5.0.

scurest:
This was caused by a bad refactor. Essentially, when I refactored to handle scrolling up as scrolling down by a negative distance, I neglected to replace `min` by `max`. See the commit message for details.

7f83a3bf has caused a bit of trouble, huh? ><;;

### DIFF
--- a/src/game_map.cpp
+++ b/src/game_map.cpp
@@ -1405,6 +1405,11 @@ void Game_Map::Parallax::Update() {
 	}
 }
 
+/** Return the argument that is closer to zero. */
+static int closer_to_zero(int x, int y) {
+	return (std::abs(x) < std::abs(y)) ? x : y;
+}
+
 void Game_Map::Parallax::Scroll(int distance_right, int distance_down) {
 	Params params = GetParallaxParams();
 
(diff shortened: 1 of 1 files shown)


carstene1ns:
refactor

Yeah, it is the same for almost all refactoring work, forgetting one or two corner cases and only testing the basic stuff. We are partly at fault too for not proper testing with all directions.

Ghabry:
And the Panorama code was (or still is, just to a lesser extend ^^) quite awful, so I'm not too surprised about regressions. And we have no proper test-suite for regression testing which would have found this.

PR #1162, opened 2017-05-12
Add opus support

### LINKED ISSUE #1159: Opus audio support
The Opus codec is currently the state of the art on audio compression, royalty free, and sounds really well on low bitrates, such 48 kbps. This makes this codec perfect for game audio, both for sound effects and music. There are games with large file sizes due to the amount of music and effects bundled with it, and an efficient codec looks a good way to make them significantly smaller. It also has a fixed point implementation.

fdelapena:
This tries to add Opus decoder support to Player (fix #1159).

The FillBuffer code is probably wrong currently. There are no crashes with current settings here so far.

It currently does some minor buffer related noise, I guess it gets unhappy with the resampler buffering mechanism, somewhere the buffer may be too tight or some delay involved in underruns.

Amendments to this PR are totally welcome.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,7 @@ set(PLAYER_SRCS
 	src/decoder_libsndfile.cpp
 	src/decoder_mpg123.cpp
 	src/decoder_oggvorbis.cpp
+	src/decoder_opus.cpp
 	src/decoder_wav.cpp
 	src/decoder_wildmidi.cpp
 	src/decoder_xmp.cpp
@@ -227,7 +228,7 @@ if(PLAYER_BUILD_LIBLCF)
 	set(LIBLCF_PATH "${CMAKE_CURRENT_SOURCE_DIR}/lib/liblcf")
 	find_package(Git REQUIRED)
 	if(NOT EXISTS ${LIBLCF_PATH})
-	execute_process(COMMAND ${GIT_EXECUTABLE} clone "--depth=1" 
+	execute_process(COMMAND ${GIT_EXECUTABLE} clone "--depth=1"
 		"${PLAYER_BUILD_LIBLCF_GIT}"
 		"${LIBLCF_PATH}")
 	endif()
@@ -357,13 +358,20 @@ if(${PLAYER_AUDIO_BACKEND} MATCHES "^SDL.*$")
 		add_definitions(-DHAVE_LIBSNDFILE)
 	endif()
 	
-	# libogg
-	option(PLAYER_WITH_OGGVORBIS "Play OGG audio with libogg." ON)	
+	# libvorbis
+	option(PLAYER_WITH_OGGVORBIS "Play Ogg Vorbis audio with libvorbis." ON)
 	find_lib("OggVorbis" PLAYER_WITH_OGGVORBIS)
 	if(OggVorbis_FOUND)
 		add_definitions(-DHAVE_OGGVORBIS)
 	endif()
-	
+
+	# opusfile
+	option(PLAYER_WITH_OPUS "Play Opus audio with opusfile." ON)
+	find_lib("Opus" PLAYER_WITH_OPUS)
+	if(Opus_FOUND)
+		add_definitions(-DHAVE_OPUS)
+	endif()
+
 	# xmp (lite)
 	option(PLAYER_WITH_XMP "Play MOD audio with libxmp." ON)
 	find_lib("XMP" PLAYER_WITH_XMP)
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -28,6 +28,7 @@ Documentation is available at the documentation wiki: https://wiki.easyrpg.org
 - mpg123 for better MP3 audio support
 - WildMIDI for better MIDI audio support
 - Ogg+Vorbis/Tremor for OGG audio support
+- Opus for Opus audio support
```
carstene1ns: `opusfile` needs `libogg`, too.

src/decoder_opus.cpp
```diff
@@ -0,0 +1,98 @@
...
+		return false;
+
+	return true;
+}
+
+int OpusDecoder::FillBuffer(uint8_t* buffer, int length) {
+	if (!oof)
+		return -1;
+
+	return op_read_stereo(oof, reinterpret_cast<opus_int16*>(buffer), length / 2) * 4;
```
carstene1ns: I have to think about the `/2` and `* 4` here.

---
The doc says:
`int op_read_stereo (OggOpusFile *_of, opus_int16 *_pcm, int _buf_size)`

Note: `_buf_size` indicates the total number of values that can be stored in `_pcm`, while the return value is the number of samples per channel, even though the channel count is known, for consistency with `op_read()`.

- `_pcm`: A buffer in which to store the output PCM samples, as signed native-endian 16-bit values at 48 kHz with a nominal range of [-32768,32767). The left and right channels are interleaved in the buffer. This must have room for at least `_buf_size` values.
- `_buf_size`: The number of values that can be stored in `_pcm`. It is recommended that this be large enough for at least 120 ms of data at 48 kHz per channel (11520 values total). Smaller buffers will simply return less data, possibly consuming more memory to buffer the data internally.
- Returns the number of samples read per channel on success, or a negative value on failure.
---

To me this means it will not fill more samples than `length/2` anyway, because it is stereo.
This means the buffer could be half-empty here.
The return value is the number of bytes read.
This should be the returned value (number of samples per channel) `*2` (for stereo) `*2` (because samples are 16 bit), so `*4` seems right here.
However, in case of error, the negative error value is multiplied as well. I do not know how the error constants are defined, but this could possibly lead to overflows. (`#define ERROR_CODE INT_MIN`)

carstene1ns:
jenkins: test this please <3

src/decoder_opus.h
```diff
@@ -0,0 +1,54 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef EASYRPG_AUDIO_DECODER_OPUS_H
+#define EASYRPG_AUDIO_DECODER_OPUS_H
+
+#include <opus/opusfile.h>
```
carstene1ns: This needs to be guarded.

CMakeLists.txt
```diff
@@ -590,6 +598,12 @@ endif()
 
 message(STATUS "")
 
+if(Opus_FOUND)
+	message(STATUS "Opus playback: opusfile")
+endif()
```
Ghabry: Imo this should be moved in the audio section (after MOD)

```cmake
	if(Opus_FOUND)
		message(STATUS "Opus playback: opusfile")
	else()
		message(STATUS "Opus playback: None")
	endif()
```

Ghabry:
Working CMake file based on Pixman file:
```cmake
find_path(OPUS_INCLUDE_DIR_INTERNAL opus/opusfile.h)
find_library(OPUS_LIBRARY NAMES opusfile libopusfile)
if(EXISTS "${OPUS_INCLUDE_DIR_INTERNAL}")
  set(OPUS_INCLUDE_DIR "${OPUS_INCLUDE_DIR_INTERNAL}/opus")
endif()

include(FindPackageHandleStandardArgs)
find_package_handle_standard_args(Opus DEFAULT_MSG OPUS_INCLUDE_DIR OPUS_LIBRARY)

set(OPUS_INCLUDE_DIRS ${OPUS_INCLUDE_DIR})
set(OPUS_LIBRARIES ${OPUS_LIBRARY})

mark_as_advanced(OPUS_INCLUDE_DIR OPUS_LIBRARY)
```

README.md
```diff
@@ -27,14 +27,17 @@ Documentation is available at the documentation wiki: https://wiki.easyrpg.org
 - FreeType2 for external font support (+ HarfBuzz for Unicode text shaping)
 - mpg123 for better MP3 audio support
 - WildMIDI for better MIDI audio support
-- Ogg+Vorbis/Tremor for OGG audio support
+- Libvorbis / Tremor for Ogg Vorbis audio support
```
carstene1ns: Actually it is vorbisfile... 🙄

README.md
```diff
@@ -27,14 +27,17 @@ Documentation is available at the documentation wiki: https://wiki.easyrpg.org
...
+- opusfile for Opus audio support
 - libsndfile for better WAVE audio support
+- libxmp for better tracker music support
```
carstene1ns: 👍

Issue #985, opened 2016-08-10
HH3: some sounds, animations, actions (events?) not appearing or appearing incorrectly


vicusdanielson:
Greetings! Just as I wrote in the title, some sounds, animations, actions/events? (I meant e.g. the boss's attack animation, boosting himself) do not appear or they appear but incorrectly.
Game name: [Heaven and Hell Episode 3](http://www.anonelbe.net/hh3-rm2k/)
I uploaded 2 videos of my playing (I re-encoded the videos to lower quality for smaller size):
[Smartphone one](https://www.youtube.com/watch?v=bh-O0whvLJo) (where the issues occur) / [Laptop one](https://www.youtube.com/watch?v=XlYl84nxHgk) (which is the correct/how it should be)
Smartphone I played on: [Xiaomi Redmi Note 2](http://www.gsmarena.com/xiaomi_redmi_note_2-6992.php)
Android version: 5.0.2 LRX22G, MIUI version: 7.4 by xiaomi.eu (6.5.19 Developer)

So the things which I noticed to be incorrect on the smartphone one:
- [X] When battles start that beeping sound doesn't get slowed (for some reason it's made like that fast and normally it plays slower - the "pitch" in the rm2k is set to lower)
- [x] On the battle screen the names, etc appear with a "%S" at their end (e.g. Martyr%S attacks the enemy, Kharan 7%S %V damage sustaned)
- [x] The "Animal Summoning III" spell of the common enemies doesn't do its grrr\* sound
- [X] When I "Run away" from battles, it doesn't do its sound and running away is much more successful (maybe I just have better luck)
- [x] "AutoBattle" always does the basic attack, normally it does the strongest attack
- [x] The basic attack does damage one time, normally it does two times **(since 0.6)**
- [x] After I used a spell the next time I attack it doesn't select it automatically when I select Fight/NirvanaSwd (e.g. I used "Sun Flare" spell at 5:10 then next time I attacked at 5:20 it didn't auto-select that spell but always the first one instead) **(since 0.6)**
- [x] When a spell misses on an enemy it doesn't show like "it missed" or "he avoided" (e.g. when the Templar avoids Kharan's "Dead Sword" spell)
- [x] The boss, Abel always just boosts himself (the values of the boosts are incorrect too, e.g. "Attack Pwr -1, ArmorClass 25"), never attacks (I even killed him XD, got the "Debugger" then that battle restarted) **(since 0.6)**
[...]

Issue #797, opened 2016-03-05
Multilanguage game support


Ghabry:
The Player should be able to load game translations on-the-fly without embedding them in the database or maps.

See also LcfTrans in the Tools repo.

Probably will be also useful for @pureexe and @whateverzone when Thai is fully supported.

Work plan (should be discussed):
- Place translations in "translation/[Translation-Name]", file hierarchy is mirrored (RPG_RT.po, Map0001.po).
- The Translation folder can also contain asset folders (ChipSet, Pictures, ...). They are prefered over the original ones when available
- The Translation files will be UTF-8, no encoding issues \o/
- Language selection on game startup, language name extracted from "Language:" in RPG_RT.po

pureexe:
WOW!!
it's great.
but i have a question.
How can i create translation file? (.po extension)

whateverzone:
That's great news! Thank you a lot!
All the plans sound convenient though I have the same question as @pureexe and I also need to hear more details about these.

Ghabry:
Don't be too excited yet. It is still in the planning phase.

You can create ".po" files with the tool "LcfTrans" which will be in our tool repository soon. Simply invoke it in the game directory and .po files are created for Database and Maps.

Po is a popular file format used to create translations, especially open source projects use it. The content is basicly

```
msgid "Original text"
msgstr "Translated text"
```

About the loading in EasyRPG Player: No work was invested into this yet. I plan this for the 0.5 release (coming this year)

pureexe:
Thank you for describe.
I hope this feature will release soon.

Ghabry:
Another open problem is how to handle text that is too long (more then 4 lines of text). Probably just insert more lines and the Player will handle it gracefully...

The other direction is a different issue (from 2 message commands to 1 because the translated message is shorter)... Probably some in-band-signalling "\0"-char... Will document it when I have an idea.

sorlok:
Good evening,

This Issues looked really cool, so I implemented a hacky test branch to help with the discussion:
  https://github.com/EasyRPG/Player/compare/master...sorlok:snh_translate?expand=1

Most .po text works, including Terms and Message Boxes. Image swapping also works, using the directory structure discussed earlier. Screenshots:
![translate_multi](https://user-images.githubusercontent.com/513512/88684452-e0aeb400-d0c2-11ea-9323-c4a53ca2b38c.png)
[...]

PR #1163, opened 2017-05-13
Support RM2k v.1.61 (placeholders and word-wrapping)

### LINKED ISSUE #588: Support RPG Maker 2000 v1.60 and v1.61
I don't think it's worth to differentiate between 1.60 and 1.61
No idea how to detect them properly.

https://docs.google.com/document/d/1n-CHoQ0hwHWEnkoiLo7fiRUqS4hY4mhOkv_jkmz2cg8/edit?pli=1#heading=h.wcb2puxulvzi

v1.61 (2015-09-15)

Placeholders in system messages.
- [x] Battle messages can now contain placeholders (%S, %O, %V, and %U for subject, object, value and unit) instead of having a fixed order and position of those values.
- [x] Also applies to skill usage messages, state messages and inn messages.
- [x] additional values can be used in battle strings, such as the attacker’s name for evasion messages, the target name for skill messages (only for single-target skills), etc.
- [x] Battle messages which become too long for a line will automatically be word-wrapped.

v1.60
- [x] Picture operations (Show, Move, Erase) are not blocked by messages anymore.

### LINKED ISSUE #1138: Engine naming for RM2k v1.61
The English RM2k v1.61 is easy to detect, but how can it be named? 

Options for RM2k v1.61:

1. Rename older RM2k as 'Legacy' for older engines. Possible downside: Japanese RPG Maker 2000 v 1.50 is [still sold](https://tkool.jp/products/rpg2000/index), can we call it legacy if it’s still the only Japanese version available?
2. Name RM2k v1.61 as 'English', probably rename old RM2k as 'Japanese'. Possible downside: the English version v1.60 would be considered 'Japanese'.
3. Use something like 'EngineSecondMajorUpdate'. Possible downside: doesn’t seem intuitive.
4. Use something like 'EngineMessagePlaceholders'. Possible downside: doesn’t seem intuitive, not in line with other names.
5. Something else?

rohkea:
This pull request adds support for placeholders (%S, %O, %V, %U) and word-wrapping introduced in the English release of RPG Maker 2000 v. 1.61.

Word-wrapping is currently **still measuring** invisible codes like ``\C[x]``, ``\S[x]``, ``\<``, so ``\C[2]\S[5]Hello`` is longer than ``Hello``. This is what RPG Maker 2000 is doing, so I’m not introducing new incompatibility here. The current code for displaying messages (e.g. for parsing parameters) has too many side effects, and changing it was a bit too complex for me, so I’ve decided not to do it right now.

Word-wrapping is currently only applied to the battle messages, like in RPG Maker 2000 v. 1.61. However, in future it can be used in other cases:

* When game translatability with LcfTrans is finished (#797), it would make sense to enable this setting for translated games.
* And when the Editor is finished, it would probably make sense to turn it on for the games made with the editor.

The word-wrapping is enabled by setting ``Game_Message::is_word_wrapped`` to true.

This pull request is related to issue #985 HH3: some sounds, animations, actions (events?) not appearing or appearing incorrectly: this should fix "Martyr%S attacks the enemy" (but I havenʼt tested this very game).

Fix #588 - Support RPG Maker 2000 v1.60 and v1.61: this adds all the features from v1.61, but not from v.1.60
Fix #1138 - Engine naming for RM2k v1.61: I've used ``EngineEnglish`` and ``rpg2ke``

### DIFF
--- a/src/game_battlealgorithm.h
+++ b/src/game_battlealgorithm.h
@@ -311,6 +311,7 @@ protected:
 	std::string GetHpSpAbsorbedMessage(int value, const std::string& points) const;
 	std::string GetDamagedMessage() const;
 	std::string GetParameterChangeMessage(bool is_positive, int value, const std::string& points) const;
+	std::string GetStateMessage(const std::string& message) const;
 
 	float GetAttributeMultiplier(const std::vector<bool>& attributes_set) const;
 
--- a/src/game_battlealgorithm.h
+++ b/src/game_battlealgorithm.h
@@ -304,6 +304,9 @@ protected:
 	AlgorithmBase(Game_Battler* source, Game_Battler* target);
 	AlgorithmBase(Game_Battler* source, Game_Party_Base* target);
 
+	std::string GetDodgeMessage() const;
+	std::string GetHPMPRecoveredMessage(int value, const std::string& points) const;
+
 	float GetAttributeMultiplier(const std::vector<bool>& attributes_set) const;
 
 	void Reset();
--- a/src/game_battlealgorithm.h
+++ b/src/game_battlealgorithm.h
@@ -310,6 +310,7 @@ protected:
 	std::string GetCriticalHitMessage() const;
 	std::string GetHpSpAbsorbedMessage(int value, const std::string& points) const;
 	std::string GetDamagedMessage() const;
+	std::string GetParameterChangeMessage(bool is_positive, int value, const std::string& points) const;
 
 	float GetAttributeMultiplier(const std::vector<bool>& attributes_set) const;
 
--- a/src/game_battlealgorithm.h
+++ b/src/game_battlealgorithm.h
@@ -309,6 +309,7 @@ protected:
 	std::string GetUndamagedMessage() const;
 	std::string GetCriticalHitMessage() const;
 	std::string GetHpSpAbsorbedMessage(int value, const std::string& points) const;
+	std::string GetDamagedMessage() const;
 
 	float GetAttributeMultiplier(const std::vector<bool>& attributes_set) const;
 
--- a/src/game_battlealgorithm.h
+++ b/src/game_battlealgorithm.h
@@ -304,7 +304,7 @@ protected:
 	AlgorithmBase(Game_Battler* source, Game_Battler* target);
 	AlgorithmBase(Game_Battler* source, Game_Party_Base* target);
 
-	std::string GetDodgeMessage() const;
+	std::string GetAttackFailureMessage(const std::string& points) const;
 	std::string GetHpSpRecoveredMessage(int value, const std::string& points) const;
 	std::string GetUndamagedMessage() const;
 	std::string GetCriticalHitMessage() const;
--- a/src/game_battlealgorithm.h
+++ b/src/game_battlealgorithm.h
@@ -308,6 +308,7 @@ protected:
 	std::string GetHPMPRecoveredMessage(int value, const std::string& points) const;
 	std::string GetUndamagedMessage() const;
 	std::string GetCriticalHitMessage() const;
+	std::string GetHPMPAbsorbedMessage(int value, const std::string& points) const;
 
 	float GetAttributeMultiplier(const std::vector<bool>& attributes_set) const;
 
--- a/src/game_battlealgorithm.h
+++ b/src/game_battlealgorithm.h
@@ -305,10 +305,10 @@ protected:
 	AlgorithmBase(Game_Battler* source, Game_Party_Base* target);
 
 	std::string GetDodgeMessage() const;
-	std::string GetHPMPRecoveredMessage(int value, const std::string& points) const;
+	std::string GetHpSpRecoveredMessage(int value, const std::string& points) const;
 	std::string GetUndamagedMessage() const;
 	std::string GetCriticalHitMessage() const;
-	std::string GetHPMPAbsorbedMessage(int value, const std::string& points) const;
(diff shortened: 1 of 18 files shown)


Ghabry:
Thx for this addition!
"When game translatability with LcfTrans is finished (#797), it would make sense to enable this setting for translated games."
Yepp that was my plan :) (it even emits the 2k-codes)

src/window_message.cpp
```diff
@@ -126,7 +118,41 @@ void Window_Message::StartMessageProcessing() {
...
+	else {
+		text.clear();
+		for (const std::string& line : Game_Message::texts) {
+			text.append(Utils::DecodeUTF32(line)).append(1, U'\n');
+		}
+	}
+	Game_Message::texts.clear();
+	item_max = min(4, Game_Message::choice_max);
 
+	ApplyTextInseringCommands();
```
Ghabry: InserTing?

src/window_battlemessage.cpp
```diff
@@ -35,12 +38,42 @@ Window_BattleMessage::Window_BattleMessage(int ix, int iy, int iwidth, int iheig
...
+			{subject}
+		));
+	}
+	else {
+		Push(subject + message);
+	}
+}
+
+
+
```
Ghabry: lots of empty lines

src/game_battlealgorithm.h
```diff
@@ -304,6 +304,15 @@ class AlgorithmBase {
...
 	AlgorithmBase(Game_Battler* source, Game_Party_Base* target);
 
+	std::string GetAttackFailureMessage(const std::string& points) const;
+	std::string GetHpSpRecoveredMessage(int value, const std::string& points) const;
+	std::string GetUndamagedMessage() const;
+	std::string GetCriticalHitMessage() const;
+	std::string GetHpSpAbsorbedMessage(int value, const std::string& points) const;
+	std::string GetDamagedMessage() const;
+	std::string GetParameterChangeMessage(bool is_positive, int value, const std::string& points) const;
+	std::string GetStateMessage(const std::string& message) const;
```
Ghabry: Really nice refactor

src/window_message.cpp
```diff
@@ -126,7 +118,41 @@ void Window_Message::StartMessageProcessing() {
 			}
 		}
 	}
+}
+
+void Window_Message::StartMessageProcessing() {
+	contents->Clear();
+
+	if (Game_Message::is_word_wrapped) {
```
Ghabry: Is this variable somehow reset when a new game is started / savegame loaded? Because otherwise the wordwrap logic would be applied to the first displayed msg :D

rohkea: [Copying from chat is someone is going to read the history in the future] "Player::ResetGameObjects already calls Game_Message::Init() which results in SemiClear being called."

Ghabry:
Jenkins: Test this please . Need exe for review ^^'

src/utils.cpp
```diff
@@ -438,3 +439,33 @@ std::vector<std::string> Utils::Tokenize(const std::string &str_to_tokenize, con
 
 	return tokens;
 }
+
+
+std::string Utils::ReplacePlaceholders(const std::string& text_template, std::vector<char> types, std::vector<std::string> values) {
+	std::string str = text_template;
```
Ghabry: Testcase:
%S appeared! %A %A %s%%s%

(Used %A because %O is substituted to ??? >.>)

2k result: Slime appeared! %A %A Slime%Slime%
EasyRPG: Slime appeared! %A %A Slime%%s%

Ghabry:
The substitution behaviour for group attacks is different. 2kE shows ???, EasyRPG the first enemy. Though this is a design limitation in BattleAlgorithm and can be fixed another time because is just the error case nobody should see :D (and that we don't substitute %O & co. when not applicable is a bonus)

src/utils.cpp
rohkea: Thanks! I’ll fix this in less than an hour. Looks like I’m increasing the variable by 2 (skipping %%) when I should be increasing in by 1 (skipping %).

rohkea:
> 2kE shows ???, EasyRPG the first enemy.

Actually, EasyRPG shows the current enemy. I.e. when attacking a slime and a bat, it will show:

* Daisy casts attack all on Slime.
* Slime loses 10 HP.
* Daisy casts attack all on Bat.
* Bat loses 10 HP.

The RM2k behaviour:

* Daisy casts attack all on ???.
* Slime loses 10 HP.
* Daisy casts attack all on ???.
* Bat loses 10 HP.

It’s trivial to change the code to RM2k behaviour (replace ``GetTarget()->GetName()`` with ``(skill.scope == RPG::Skill::Scope_party || skill.scope == RPG::Skill::Scope_enemies) ? "???" : GetTarget()->GetName()`` in ``Game_BattleAlgorithm::Skill::GetStartMessage()``), but I really don’t see any reason to do this. The current behaviour seems logical, and people can easily imitate the RM2k behaviour by writing ``???`` instead of ``%O`` for multi-object skills.

src/utils.cpp
rohkea: Done.

Ghabry:
I have a enemy party with "Slime 1234567890ABCD" and "Bat".
The "enemy appear message" is: "%S%S%S%S%S%S%S%S%S%S%S%S%S%S%S%S%S%S%S%S". Before "BatBatBatBat" the Player will hang.

I also observe that the first two lines of the Slime name appear at once (at page 3+) instead of one-by-one.

Ghabry:
Okay works as expected now

Issue #1115, opened 2017-03-04
Horizontal/Vertical movement ignores facing direction


Ghabry:
#### Name of the game:
Space Funeral, similiar issue found in Deep8

#### Attach files (as a .zip archive or link them)
[Space Funeral issue.zip](https://github.com/EasyRPG/Player/files/678906/Space.Funeral.issue.zip) Contains savegame/logs.

#### Describe the issue in detail and how to reproduce it:
A quick way to reproduce this:
The facing direction is ignored when the move type is vertical/horizontal.
So e.g. When the actor faces up and the movement is vertical the actor moves down instead of up.

Original comment by scurest:

First, this needs the other change in #1108 or else the events (which use chip graphics) will pass right through you if you don't move into them (which is convenient, but wrong :). Other than that

- [x] The flash color doesn't go fully opaque in EasyRPG.

    ![final](https://cloud.githubusercontent.com/assets/11024420/23575627/649f5086-0055-11e7-90ef-9d067f5265dc.png) 

- [x] The direction of motion for events that cycle left/right is determined by their current facing direction in RPG_RT. We determine it with a flag `Game_Character::cycle_stat`. For example, if you stand in the bottom row and wait for the... thing... to hit you from the right, it will retreat two steps to the left, then (1) in RPG_RT continue its cycle moving right, (2) in EasyRPG, continue its cycle going left

    ![out](https://cloud.githubusercontent.com/assets/11024420/23575754/27a9cd70-0058-11e7-8a24-0023764cc304.png)

- [x] If you back that same thing into the bottom-right corner, in RPG_RT, it will stay facing left, hitting you. In EasyRPG, the direction it faces switches every frame. You can't tell with the weird face thing but if you change the event graphic, you can. **since 0.6**

    ![switch](https://cloud.githubusercontent.com/assets/11024420/23575812/49884484-0059-11e7-98f4-b484f999f0c9.gif)

    Also, if you hold right in this position, the event will constantly face right. In RPG_RT, holding right doesn't change anything (it stays facing left).

- [x] When you are in the same corner the character is red the whole time, but not in EasyRPG Player where it is always red-pause-red-pause. **patch available for 0.6.1**

That's what I noticed.

scurest:
(btw I accidentally switched the words "right" and "left" in (1) and (2) in the second bullet. In RPG_RT it continues to the right, and in EasyRPG it continues to the left. The picture is correct.)

fmatthew5876:
[...]

PR #1167, opened 2017-05-17
Improve Movement Type Cycle

scurest:
This fixes (or at least improves) the second bullet point (and only the second bullet point) in #1115. Events with movement type cycle now prefer moving forward, if possible. Basically, the algorithm is now

1. When we're ready to move, try moving forward (or in the default direction if we're facing the wrong way: Right for Left-Right, Down for Up-Down)
2. If that failed for 20 frames, keep trying to move forward, but if it fails, now try moving the opposite way too.
3. If that fails for 40 more frames, try going forward and if it fails, now move (don't just try to move) in the opposite way, and reset

By "try to move" I mean to do the same thing as doing a `Move` in a move route with the "Ignore if can't be moved" option toggled on. It turned out, though, that this also behaved differently in Player; even if you didn't move, the command wasn't totally ignored; you'd still turn to face the move direction. In RPG_RT, this doesn't happen. So I fixed this too.

The (probably totally useless in actual games) case of an event between two obstacles on either side that can't move should now also turn around with the same timing as in RPG_RT.

Yeah. Um. Do we know where in Deep 8 this was a problem? Probably be good to test it.

### DIFF
--- a/src/game_character.h
+++ b/src/game_character.h
@@ -411,17 +411,23 @@ public:
 	void MoveTypeCustom();
 
 	void Turn(int dir);
-	void Move(int dir);
+
+	enum class MoveOption { Normal, IgnoreIfCantMove };
+
+	/**
+	 * Move in the direction dir.
+	 */
+	void Move(int dir, MoveOption option = MoveOption::Normal);
 
 	/**
 	 * Moves the character forward.
 	 */
-	void MoveForward();
+	void MoveForward(MoveOption option = MoveOption::Normal);
 
 	/**
 	 * Does a random movement.
 	 */
-	void MoveRandom();
+	void MoveRandom(MoveOption option = MoveOption::Normal);
 
 	/**
 	 * Does a move to the player hero.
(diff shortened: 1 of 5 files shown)


Ghabry:
Looks good :D

deep8-movement-0 5 2

src/game_event.h
```diff
@@ -195,14 +195,20 @@ class Game_Event : public Game_Character {
 	void MoveTypeRandom();
 
 	/**
-	 * Moves left to right and switches direction if the
-	 * move failed.
+	 * Cycles between moving in default_dir and its opposite.
+	 * Tries to move the the event's movement direction, or
```
carstene1ns: This comment seems wrong here.

src/game_character.h
```diff
@@ -677,6 +683,9 @@ class Game_Character {
 		UpLeft
 	};
 
+	/** Reverses a direction, ex: RevereseDir(Up) == Down. */
```
carstene1ns: typo.

Issue #1113, opened 2017-03-04
Wrong facing direction during jump


Ghabry:
Movement:
```
Jump
Right
Right
Down
Land
```

#### Attach files (as a .zip archive or link them)

[issue1113.zip](https://github.com/EasyRPG/Player/files/818739/issue1113.zip)

#### Describe the issue in detail and how to reproduce it:
Move on the stairs, press return. The Player faces down during the jump. Expected: Faces Right.

Albeleon:
#1401 The Commit "BeginJump calculating the right direction; remove Turn commands inside Jump" solves this and explains the real RPG_RT logic (sum up: Turn doesn't work inside a Jump; the direction is calculated once it knows where it's going to land).

carstene1ns:
These things have high potential to break games if priority or order is wrong. Needs some playtest with problematic games.
Please also have a look here: #1171 → #1172.

Albeleon:
^ The new update has made sure to respect #1171 and #1172 while still solving this issue.

Issue #1112, opened 2017-03-04
Deep 8: Birds shot in "Cannon game" minigame are sometimes not playing the dead animation


Ghabry:
#### Name of the game: Deep 8

I included a minimal testcase that has the same issue.

#### Attach files (as a .zip archive or link them)
[issue1112.zip](https://github.com/EasyRPG/Player/files/818705/issue1112.zip)

#### Describe the issue in detail and how to reproduce it:

In front of you are 4 bridges, pressing ENTER on one of these bridges "kills" a bird. This is some jump-event combined with a rotation. In EasyRPG that jump-kill is sometimes not executed and the bird dissapears directly. In RPG_RT the anim is always played.

Ghabry:
The animaton only works when the movement pattern is on frame 1 (middle frame) while it happens. On frame 0, 2 (left, right) the bird just dissapears. (actually I think the wrong frame is shown which is empty in this charset)

Ghabry:
This is caused by the following code in Game_Event::Setup and is fixed by removing the if-check.
```cpp
	if (original_pattern != page->character_pattern) {
		pattern = page->character_pattern;
		original_pattern = pattern;
	}
```
git blame says the code was added by me 6 years ago :o

This needs regression tests but imo that if-check makes no sense because it prevents the events current pattern to change when the old and the new page used the same pattern setting. (and the current pattern can be different due to movements)

PR #1172, opened 2017-05-18
Fix #1171: Respect Turn commands when calculating the Jump direction.

### LINKED ISSUE #1171: Super Mario RPG The Seven Sages: Hero jumps in the wrong direction after touching a !-switch
#### Name of the game: Super Mario RPG The Seven Sages

Simple test case:
```
MoveEvent: Jump, Turn 180°, 1 Step forward, Land
```
Expected: Player turns and jumps one tile back
What happend: Player jumps one tile forward

The problem is that we ignore all kinds of direction changes, we only consider movements in StartJump

Ghabry:
This is basicly a copy-paste of all the Face-related stuff in MoveTypeCustom into BeginJump. (When a Jump is performed most commands are ignored, except Move and Face it seems)

index

### DIFF
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -629,6 +629,40 @@ void Game_Character::BeginJump(const RPG::MoveRoute* current_route, int* current
 			case RPG::MoveCommand::Code::move_forward:
 				MoveForward();
 				break;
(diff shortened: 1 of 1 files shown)


Ghabry:
The other jump related issues #1112 and #1113 are not fixed by this :(

Issue #562, opened 2015-09-03
Emscripten: support async asset preloading on startup


fdelapena:
Before having an async file implementation, static data preloading was a big file. The problem of this was not just a huge loading time: browser is not caching large files and a page refress needed to download the whole game data again.

Current async implementation is loading files on demand in the moment they are called. This allows to cache files apart of fast loading but makes the game display/hear the file a bit more later than it should, most noticeable on slow connections, degrading the game user experience.
To prevent this, files should be loaded enough time before by preloading them in some way.

carstene1ns:
If done correctly, this should use some emscripten magic:
e.g. use [`FS.createPreloadedFile()`][1] for files that are needed before anything else (MapTree, ini... later also Title screen Music/Pictures).
A drawback using this method is, that they have to be known before actual runtime, either when compiling a special version for one game only or when [packaging gamedata][2] before bundling.

For the files needed when changing the Map, [main loop blockers][3] can be used. However, this needs parsing the entire Map for appropriate files (maybe just the large ones...for example Music?)

This should not affect the async stuff much, as the loading functions may look for files in the virtual filesystem first, before actually downloading (`emscripten_*wget*`) anything.

[1]: http://kripken.github.io/emscripten-site/docs/api_reference/Filesystem-API.html#FS.createPreloadedFile
[2]: https://kripken.github.io/emscripten-site/docs/porting/files/packaging_files.html#packaging-using-the-file-packager-tool
[3]: http://kripken.github.io/emscripten-site/docs/porting/emscripten-runtime-environment.html#execution-lifecycle

Ghabry:
Are you still unhappy about the async loading strategy or can that be closed?

fdelapena:
This one could be deferred IMO. As carstene1ns pointed, critical files could be preloaded, e.g. for "The Sacred Tears: TRUE" picture based cutscene or the "+13" animation in "Desolate City".

Ghabry:
The only direct improvement I see for map changes is to declare all files directly required when the map loads as important so that the transition looks better (without stuff popping in).

fdelapena:
Another hacky but simple approach to achieve faster asset loading without messing with Player code:

- Generate via JavaScript a list of `<link rel=prefetch>` on load, by reading the JSON generated with gencache.

Most browsers have prefetching enabled by default, except those not wanting to prefetch content (e.g. for privacy reasons or bandwidth savings).

elsemieni:
[...]

PR #1178, opened 2017-05-23
Emscripten: Improve UX for map loads

Ghabry:
Don't finish the teleport until all assets that will be required by the map are downloaded.

The purpose is to make the fade-in look correct. Currently only the ChipSet
is "important". Which means other assets are missing and will "pop-in" after
the fade-in is finished which is wrong and looks ugly.

Disadvantage is that map loading will take slightly longer because the Player
waits for all file transfers. Still faster then MV, though ;)

### DIFF
--- a/src/game_picture.cpp
+++ b/src/game_picture.cpp
@@ -22,6 +22,7 @@
 #include "game_picture.h"
 #include "player.h"
 #include "main_data.h"
+#include "game_player.h"
 
 Game_Picture::Game_Picture(int ID) :
 	id(ID),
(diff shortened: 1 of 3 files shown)


fdelapena:
Related: #562

Ghabry:
For some reason under "real world" conditions the code doesn't appear to work and the panorama still comes in delayed o.O. In Chromes "limit bandwidth" function it worked :(

fdelapena:
Firefox (with e10s enabled) there is a limited bandwidth option in the mobile view if you want to re-check with it:
captura de pantalla de 2017-06-10 17-35-44

Ghabry:
I think we have to delay this because I wont have time before vacation anymore to fix this.

carstene1ns:
Okay, changing the milestone then.

carstene1ns:
I am going to close this for now, please re-open when you decide to revamp this... ;-)

Issue #1169, opened 2017-05-17
 ChipSet and Battle positioning issues


zell180:
#### Attach files (as a .zip archive or link them)
[TestEasyRPG.zip](https://github.com/EasyRPG/Player/files/1009210/TestEasyRPG.zip)

#### Describe the issue in detail and how to reproduce it:
Hi, our main coder prepared a mini-demo of two problems we are experiencing in Easy RPG only. The issues are:

1) ChipSet: in the map there is a column with a hole inside. The column appears withuot holes in RPG Maker. The hero of the mini-demo will show you the bugged column (which is part of a ChipSet) while you play it;

2) Battle positioning: in our game we implemented a FaceSet Animation System. We don't use static FaceSets but we do use Battle animations to see them moving (mouth, eyes, etc). In the mini-demo the hero will talk for the first time and the animated FaceSet will appear in the right position (in its box of our text message picture). Then if you walk to the right, after the column, the game will take control of the hero, change the camera displacement and then move the hero upwards. He will talk again and this time the animated Battle of the FaceSet will not appear in the box. THIS DOES NOT HAPPEN IN RPG MAKER. To show the animated Battle of the FaceSet, throughout our game we use Event 0001 of every map as its target. Of course this event need to be repositioned every time we open a new message box, expecially in big maps, in order to see the animation played inside the box. This repositioning is performed by a parallel process (Common Event number 894, you find it in the test case). In RPG Maker this algorith always work, in any map and in any condition. In Easy RPG the result of the calculation is wrong because the Battle appears in different, wrong, positions (not in the message box). We are experiencing this problem in big maps, the ones which left the DEBUG message "Image size out of the boundary" in the Easy RPG log.

Please check the mini-demo and let us know where the differences with RPG Maker are and if you can fix them.
Thank you!

scurest:
Really nice testcase :)

1 is caused because the tile for the event is the very first one in the tileset (`tile_id` is 0) and we assume that the first tile is always fully transparent. A simple patch to `Sprite_Character` should fix this.

Ghabry:
I second this. Great testcase. Usually we don't get such detailed reports :)

Ghabry:
Looks like in the 3rd case where the image is not visible the problem is that the reposition event isn't even executed, no idea why yet o.O

Ghabry:
After the panning EasyRPG populated the variables that receive ScreenXY and HeroXY incorrectly.

RPG_RT:
```
SX184
SY176
[...]

PR #1179, opened 2017-05-23
Always draw event sprites, even for tile_id = 0

scurest:
This fixes the first part of #1169, with the invisible event. I'm PRing it separately because this one should be much easier to review than the fix for the second part.

### DIFF
--- a/src/sprite_character.cpp
+++ b/src/sprite_character.cpp
@@ -23,7 +23,7 @@
 
 Sprite_Character::Sprite_Character(Game_Character* character) :
 	character(character),
-	tile_id(0),
+	tile_id(-1),
 	character_index(0),
 	chara_width(0),
 	chara_height(0) {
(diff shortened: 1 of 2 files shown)


Ghabry:
Made some tests with the editor to see how actor_id and character_name are populated and I think a condition on character_name is the correct behaviour. Just funny that it took so long to find a game that breaks due to this :D. Everybody using 0 as transparent tile it seems.

Ghabry:
@scurest 
I found a regression: When the event has no runnable page (no page that fulfills the start condition) then the event is really invisible instead of showing tile 0 :)

scurest:
ლ(` Д ´  )ლ

We should probably just not draw those at all. How about something like this?

    diff --git a/src/game_event.cpp b/src/game_event.cpp
    index cad9e7e5..5bed38ac 100644
    --- a/src/game_event.cpp
    +++ b/src/game_event.cpp
    @@ -345,10 +345,12 @@ void Game_Event::Refresh() {
     		// Loop in reverse order to see whether any page meets conditions...
     		if (AreConditionsMet(*i)) {
     			new_page = &(*i);
    +			SetVisible(true);
     			// Stop looking for more...
     			break;
     		}
     	}
    +	if (!new_page) SetVisible(false);
     
     	// Only update the page pointer when game is loaded,
     	// don't setup event, already done

Ghabry:
Can confirm that this fixes the regression. Will append this to my PR

PR #1187, opened 2017-05-25
Some little cosmetics

carstene1ns:


src/decoder_fmmidi.cpp
```diff
@@ -57,9 +57,9 @@ bool FmMidiDecoder::Open(FILE* file) {
 	fseek(file, 0, SEEK_END);
 	file_buffer.resize(ftell(file) - old_pos);
 	fseek(file, old_pos, SEEK_SET);
-	fread(file_buffer.data(), file_buffer.size(), 1, file);
+	size_t bytes_read = fread(file_buffer.data(), file_buffer.size(), 1, file);
 
-	if (!seq->load(this, read_func)) {
+	if ((bytes_read =! file_buffer.size()) || (!seq->load(this, read_func))) {
```
Ghabry: =!

carstene1ns: ?

Ghabry: the operator should be != . I guess this assigns 0 or 1 (negation of file_buffer.size() to bytes_read.

src/image_png.cpp
```diff
@@ -187,16 +187,19 @@ static void ReadGrayData(
...
+			uint8_t a = (*p & 0xff000000) >> 24;
+			uint8_t b = (*p & 0x00ff0000) >> 16;
+			uint8_t g = (*p & 0x0000ff00) >> 8;
+			uint8_t r = (*p & 0x000000ff);
+
+			// Black pixels are transparent
+			if (b == 0 && g == 0 && r == 0 && a == 255)
+				*p = 0;
+		}
 	}
```
Ghabry: Finally tested it under the Wii: Does indeed not work. Colors are completely wrong when transparency is used. All other graphics look fine.

carstene1ns:
I expected this. Going to fix this later.

Ghabry:
Do you plan to fix the png reader for the Wii before 0.5.2? Otherwise that commit could be removed and fixed for after 0.5.2 because the current code doesn't break anything.

carstene1ns:
Let's move this after 0.5.2 then. Will do a final rebase in a bit.

carstene1ns:
Please note that there are changes since your review:
- commit 5b52266 is new.
- commit 0c7cbac has been changed significantly in `audio_generic` and `utils`.

src/sdl_ui.cpp
```diff
@@ -786,6 +786,7 @@ void SdlUi::ProcessKeyDownEvent(SDL_Event &evnt) {
 			return;
 		}
 
+		[[gnu::fallthrough]];
```
fdelapena: in GCC6 says:

`warning: attributes at the beginning of statement are ignored [-Wattributes]`

carstene1ns: I had experimented with some "fall-through comments" as mentioned in the link I provided in the commit description, but it did not work.
Also MSVC warns about not recognizing `gnu` attributes (of course^^).
If we do not figure out a better solution, I will move the check for <kbd>Enter</kbd> into the `default` case instead.

carstene1ns: gnulib (mentioned in that bug report) does this:
```c++
#ifndef FALLTHROUGH
# if __GNUC__ < 7
#  define FALLTHROUGH ((void) 0)
# else
#  define FALLTHROUGH __attribute__ ((__fallthrough__))
# endif
#endif
```
and then
```c++
                  case TYPE_ULONGLONGINT:
                    // [...]
                    FALLTHROUGH;
                  case TYPE_LONGINT:
```
m(

carstene1ns:
jenkins: test this please

PR #1193, opened 2017-05-30
Replace "を" and "ヲ" with "ヴ" in the "Enter Hero name" scene

kakurasan:
On Windows (+ RPG_RT) there is not "を" or "ヲ" in the "Enter Hero name" scene and there is "ヴ" instead.

Hiragana:
RPG_RT on Windows (hiragana)
Katakana:
RPG_RT on Windows (katakana)

EasyRPG Player (unpatched)
EasyRPG Player (unpatched)

This patch replaces "を" and "ヲ" with "ヴ", and allows users to input "ヴ".

EasyRPG Player (patched)
EasyRPG Player (patched)


Ghabry:
Besides being consistent with RPG_RT:
Imo one of the two "wo" can be kept because why should be ヴ displayed twice.

Which one is more useful in names: を or ヲ? Pick one :D

kakurasan:
In general, both "を" and "ヲ" are *rarely used for names* in Japan and I guess that's why RPG_RT users can't input "を" or "ヲ".

> Which one is more useful in names: を or ヲ? Pick one :D

The only difference is hiragana or katakana. I think almost all Japanese users (including me) don't need "を" or "ヲ".

PR #1197, opened 2017-06-05
Fix off-screen panning

### LINKED ISSUE #1169: ChipSet and Battle positioning issues

scurest:
I think this is ready....

Keep track of how much we've actually panned so we know not to go past where the pan should be.

Should fix #1169. It don't think it fixes _any_ other panning problems though. Matching RPG_RT turned out to be a nightmare -_-;;

I needed a non-negative mod function. Ghabry added a proper one in 59b4b54, so the `mod` function I added should be changed to use that whenever both of them make in into master.

### DIFF
--- a/src/game_map.cpp
+++ b/src/game_map.cpp
@@ -60,9 +60,6 @@ namespace {
 	std::vector<Game_CommonEvent> common_events;
 
 	std::unique_ptr<RPG::Map> map;
-	int scroll_direction;
-	int scroll_rest;
-	int scroll_speed;
 
 	std::unique_ptr<Game_Interpreter_Map> interpreter;
 	std::vector<std::shared_ptr<Game_Interpreter> > free_interpreters;
@@ -88,9 +85,6 @@ void Game_Map::Init() {
 	refresh_type = Refresh_All;
 
 	location.map_id = 0;
-	scroll_direction = 0;
-	scroll_rest = 0;
-	scroll_speed = 0;
 	interpreter.reset(new Game_Interpreter_Map(0, true));
 	map_info.encounter_rate = 0;
 
@@ -237,10 +231,6 @@ void Game_Map::SetupCommon(int _id) {
 	}
 
 	refresh_type = Refresh_All;
-
-	scroll_direction = 2;
-	scroll_rest = 0;
-	scroll_speed = 4;
 
 	int current_index = GetMapIndex(location.map_id);
 	map_info.encounter_rate = Data::treemap.maps[current_index].encounter_steps;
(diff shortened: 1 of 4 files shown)


Ghabry:
The incorrect screen pan offset in the Razas intro is fixed, too 👍

carstene1ns:
Sooo, any possible candidates for regressions? (´▽`)

scurest:
Heh :) I wanted to check I didn't regress 342351d4, but I don't think I understand the bug it fixed, since characters seem to die when they hit the bottom of the screen whether the `+2`s are there or not?

Ghabry:
That +2 was added because in easyrpg the screen position was wrong by 2 pixels after walking around compared to RPGRT.
When GetScreenX and GetScreenY yield the same result as RPG_RT without the +2 and the viewport loooks correct it should be safe to remove it.

carstene1ns:
Well, I wanted to wait until the the `+2` and `mod()` stuff is changed before merging this.
We need to remember this before preparing 0.5.2.

Ghabry:
oh, right. Sorry. :/

Ghabry:
mod() is done in my PR. +2 not

PR #1199, opened 2017-06-07
Improve the Output system

Ghabry:
console-0 5 2

^----- Proposed screenshot for the blogpost

New features:
 - The message overlay can now collect all messages since startup (Graphics::Init was moved to the beginning of Player::Init) and shows them when the Ui is ready (e.g. the NDSP error message on the 3DS)
 - For file-log and on-screen similiar messages increment now a counter [x10] which is appended to the line. For the file-log a message is always written twice: The 1st time it happens and a 2nd time how often it was duplicated (this is to ensure that something is logged in case of crashes), (see screenshot)
- The FileFinder got a new function to strip path components outside of the game path (see screenshot, look at the filenames reported). When the path is outside the game path it does nothing, so RTP paths are not touched.
- Instead of hardcoding "\n" in the messages I use the linebreak function of @rohkea , simplified the code a bit (see screenshot)

Bugfixes:
 - The timeout of the console lines was coupled to the framerate, the timeout is now correctly handled in Update()
 - The linebreak function did not break the last word when the last word exceeded the limit

### DIFF
--- a/src/audio_generic.cpp
+++ b/src/audio_generic.cpp
@@ -154,7 +154,7 @@ void GenericAudio::SE_Play(std::string const &file, int volume, int pitch) {
 		}
 	}
 
-	Output::Warning("Couldn't play %s SE.\nNo free channel available", FileFinder::GetPathInsideGamePath(file).c_str());
+	Output::Warning("Couldn't play %s SE. No free channel available", FileFinder::GetPathInsideGamePath(file).c_str());
 }
 
 void GenericAudio::SE_Stop() {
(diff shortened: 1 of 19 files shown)


Ghabry:
Jenkins: Test this please :)

carstene1ns:
nope

carstene1ns:
jenkins: test this please.

carstene1ns:
jenkins: test this please (again).

Ghabry:
Rebased. But havnt cleaned up message spam yet.

carstene1ns:
Great! Maybe removing all output of PicPointer messages is not a good idea, if this regresses later. So having at least one message indicating the patch is used would be fine.

carstene1ns:
Merging this... We can further reduce message spam (i.e. picpointer) later.

PR #1200, opened 2017-06-11
opus: add seek, tell and close callbacks

### LINKED ISSUE #1159: Opus audio support

fdelapena:
Fixes #1159

### DIFF
--- a/src/decoder_opus.cpp
+++ b/src/decoder_opus.cpp
@@ -24,9 +24,24 @@
 #include "audio_decoder.h"
 #include "decoder_opus.h"
 
-static int custom_read(void *stream, unsigned char *ptr, int nbytes) {
+static int custom_read(void* stream, unsigned char* ptr, int nbytes) {
 	FILE* f = reinterpret_cast<FILE*>(stream);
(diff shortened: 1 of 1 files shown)


src/decoder_opus.cpp
```diff
@@ -29,6 +29,21 @@ static int custom_read(void *stream, unsigned char *ptr, int nbytes) {
...
+	FILE* f = reinterpret_cast<FILE*>(stream);
+	return fseek(f, offset, whence);
+}
+
+static opus_int64 custom_tell(void *stream) {
+	FILE* f = reinterpret_cast<FILE*>(stream);
+	return ftell(f);
+}
+
+static int custom_close(void* stream) {
```
carstene1ns: minor style it: can you decide whether you use `void* var`, `void * var` and `void *var`?

Issue #1191, opened 2017-05-28
RM2k battle: Change Enemy HP with Allow Knockout should trigger victory


rohkea:
The Change Enemy HP battle command can trigger victory in RPG Maker 2000, but not in EasyRPG.

Here is a test project (Degica’s RPG_RT, not Don’s, but Don’s should be same): 
[BattleBug.tar.gz](https://github.com/EasyRPG/Player/files/1034543/BattleBug.tar.gz)

As can be seen in this project, the EasyRPG ends up with a battle scene with no monsters in it. Trying to do a usual attack on a monsterless battle scene results in an infinite loop.

Ghabry:
Not even downloaded but my educated guess is that your check is on the last event line and the actual bug is here: https://github.com/EasyRPG/Player/blob/master/src/game_interpreter_battle.cpp#L37
This check won't happen for the last executed event line.
(And that behaviour is wrong anyway as stated in https://github.com/EasyRPG/Player/issues/1180)

rohkea:
No, this effect doesn’t depend on the line in which the event command is placed. I’ve just tested with adding more event commands, the battle still doesn’t end.

Ghabry:
Is this problem exclusive to events that run at the beginning of the battle or does your real game use something different?
Because I can only reproduce it for the "beginning of battle" case.

rohkea:
@Ghabry No, it’s not at the beginning of the battle, it’s when hero’s HP ≤30%. Here are the settings I use:
![The event I use](https://user-images.githubusercontent.com/3201243/27253501-8145ec5a-537e-11e7-8e5d-365f80b02802.png)

(I’ve tried it without ‘Abort Battle’, too. Didn’t work either way.)

Issue #1186, opened 2017-05-25
Battle encounters don't consider the Terrain setting of the monster groups


Ghabry:
For monster groups theres an option to set on which terrain these monsters can appear.

This must be handled in https://github.com/EasyRPG/Player/blob/master/src/game_map.cpp#L956

![screenshot_20170525_172026](https://cloud.githubusercontent.com/assets/1331889/26456874/b18a4d7e-416e-11e7-897b-be686899b2dc.png)

Issue #1180, opened 2017-05-24
7thJojo: Battle against the nurse is unwinnable


Ghabry:
The patch that fixed "Huntress of the Hollow" #1129 does not fully replicate RPG_RT (of course -_-). Turns out the behaviour is different depending on engine and Ally or enemy.

7thJojo does: Remove Actor 1; Add Actor 2. But the party only has 1 actor so after removing the lose check is True and a GameOver happens.

TestCase: "--battle-test 85 --start-party-id 1". Use "Defend" 3 times.

Expected behaviour:
2k: The victory check happens at the end of the event, so you can freely remove party members and enemies as long as at the end anybody is alive.

2k3: The victory check happens after each line but **only** for the enemy party. Allies can be removed and added without triggering a lose BUT executing event commands that yield the interpreter (Message, Wait...) while the party is empty crashes the engine with an access violation.

Issue #1176, opened 2017-05-21
Schmetterlinkskuh-Flugsimulator-Simulation 2k: Collision system doesn't work in Level 2


Ghabry:
Name: Schmetterlinkskuh-Flugsimulator-Simulation 2k

Download: http://rmarchiv.de/games/704

Savegame:
[Save01.zip](https://github.com/EasyRPG/Player/files/1017440/Save01.zip)

After loading the save level 2 starts. (Epilepsy warning!). You can't collide with the scenery and just pass through making the game unwinnable because the "End game" tiles don't work either.

![screenshot_20170522_014051](https://cloud.githubusercontent.com/assets/1331889/26288316/b8fc32b6-3e8f-11e7-96a9-7a895ba2be42.png)

Ghabry:
The problem is that the map is looping so GetTerrainID actually *wraps* (so when the map is 20x20 and you request "20" then it wraps and takes the 1st one...). Also works in negative direction.

Somehow doesn't affect other event commands. E.g. GetEventID doesn't wrap.

Thx to scurest for the idea.

CC @CherryDT . More obscure Maker behaviour.

Ghabry:
```diff
diff --git a/src/game_map.cpp b/src/game_map.cpp
index 2551b972..b4a086ca 100644
--- a/src/game_map.cpp
+++ b/src/game_map.cpp
@@ -665,7 +665,10 @@ bool Game_Map::IsCounter(int x, int y) {
 	return !!(passages_up[index] & Passable::Counter);
 }
 
-int Game_Map::GetTerrainTag(int const x, int const y) {
+int Game_Map::GetTerrainTag(int const xpos, int const ypos) {
+	int x = Game_Map::RoundX(xpos);
+	int y = Game_Map::RoundY(ypos);
+
 	if (!Game_Map::IsValid(x, y)) return 9;
 
 	unsigned const chipID = map->lower_layer[x + y * GetWidth()];
```

scurest:
To get this to match RPG_RT in all cases, we'd also have to change `RoundX` (and `RoundY`) to handle all values. We only handle negative numbers as far as `-width`, because once you go below that, the LHS in `(x + GetWidth()) % GetWidth()` becomes negative. We could replace that with `mod(x, GetWidth())` where

    // Non-negative modulus
    int mod(int x, int m) {
        return (x < 0) ? (m + x % m) : (x % m);
    }

or something.

PR #1204, opened 2017-06-17
Bug squashing (#987, #1111, #1112, #1145, #1176, #1180, #1186, #1191)

Ghabry:
Additionally fixes some compiler warnings (ReplacePlaceholders had a bug)

And some ASAN error in battle of Wolfenhain.

Monster in 2k that can't act don't act anymore. ^^

Fixed a regression introduced in #1062

### DIFF
--- a/src/game_enemyparty.cpp
+++ b/src/game_enemyparty.cpp
@@ -40,12 +40,32 @@ int Game_EnemyParty::GetBattlerCount() const {
 void Game_EnemyParty::Setup(int battle_troop_id) {
 	enemies.clear();
 	troop = &Data::troops[battle_troop_id - 1];
-	std::vector<RPG::TroopMember>::const_iterator ei;
-	for (ei = troop->members.begin(); ei != troop->members.end(); ++ei)	{
-		std::shared_ptr<Game_Enemy> enemy = std::make_shared<Game_Enemy>(ei->enemy_id);
-		enemy->SetBattleX(ei->x);
-		enemy->SetBattleY(ei->y);
-		enemy->SetHidden(ei->invisible);
+
+	int non_hidden = 0;
+	for (const RPG::TroopMember& mem : troop->members) {
+		non_hidden += (!mem.invisible ? 1 : 0);
+	}
+
+	for (const RPG::TroopMember& mem : troop->members) {
+		std::shared_ptr<Game_Enemy> enemy = std::make_shared<Game_Enemy>(mem.enemy_id);
+		enemy->SetBattleX(mem.x);
+		enemy->SetBattleY(mem.y);
+
+		if (!mem.invisible) {
+			if (troop->appear_randomly) {
+				// At least one party member must be visible
+				if (non_hidden > 1) {
+					bool hide = Utils::ChanceOf(1, 2);
+					enemy->SetHidden(hide);
+					non_hidden -= (hide ? 1 : 0);
+				}
+			} else {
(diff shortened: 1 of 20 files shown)


src/game_enemyparty.cpp
```diff
@@ -40,12 +40,32 @@ int Game_EnemyParty::GetBattlerCount() const {
...
+	for (const RPG::TroopMember& mem : troop->members) {
+		std::shared_ptr<Game_Enemy> enemy = std::make_shared<Game_Enemy>(mem.enemy_id);
+		enemy->SetBattleX(mem.x);
+		enemy->SetBattleY(mem.y);
+
+		if (!mem.invisible) {
+			if (troop->appear_randomly) {
+				// At least one party member must be visible
+				if (non_hidden > 1) {
+					bool hide = Utils::ChanceOf(1,2);
```
carstene1ns: Missing space between arguments.

src/utils.h
```diff
@@ -112,6 +112,18 @@ namespace Utils {
 	}
 
 	/**
+	 * Calculates the modulo of number i ensuring the result is positive for
```
scurest: Technically, it's only non-negative (i=0).

Even more techincally, m has to be >0, or else eg. `PositiveMod(1,-2)` is -1.

src/scene_battle_rpg2k.cpp
```diff
@@ -801,14 +801,17 @@ void Scene_Battle_Rpg2k::CreateExecutionOrder() {
 }
 
 void Scene_Battle_Rpg2k::CreateEnemyActions() {
-	std::vector<Game_Battler*> active_enemies;
-	Main_Data::game_enemyparty->GetActiveBattlers(active_enemies);
+	std::vector<Game_Battler*> enemies;
+	Main_Data::game_enemyparty->GetBattlers(enemies);
+
+	for (Game_Battler* battler : enemies) {
+		if (!battler->CanAct()) {
```
Ghabry: I forgot something here: When the enemy can't Act the action must be a "NoMove" because otherwise auto-healing of states doesn't work.

Correct code should be:
```cpp
battler->SetBattleAlgorithm(std::make_shared<Game_BattleAlgorithm::NoMove>(battler));
ActionSelectedCallback(battler);
continue;
```

Maybe @carstene1ns can amend this to his "little cosmetics" PR because thats the only one still open :D

src/sprite_battler.cpp
```diff
@@ -144,10 +144,6 @@ void Sprite_Battler::Update() {
 			}
 		}
 	}
-
-	SetX(battler->GetDisplayX());
-	SetY(battler->GetDisplayY());
-	SetZ(battler->GetBattleY());
```
Ghabry: I think I found a use for these lines: The purpose was probably (y u no comment) to support RPG2k3 "Animation 2" stuff that contains e.g. "Move to enemy". But we don't implement this by now so no damage done.

carstene1ns: I am currently working on something in sprite_battler, history says these were added for screen shaking (891631fc4f29).

scurest: They've been there since the file was first added in c6f41c7. 891631f just changed `GetBattleX` to `GetDisplayX` to account for shaking.

carstene1ns: We were talking about the added ones in `Update()` ;-)

PR #1231, opened 2017-07-29
Fix #1229

### LINKED ISSUE #1229: Acessing Actor Name from Message Box does not work
Start the game, Leave the House, head to the right.
Give the "Tiger Boy" a name, he will be happy, but using his name does not work on first try:
screenshot_0 screenshot_1

**Editor**: Town → West (Map 20)
Nothing interesting here, Actor 2 gets a name, displaying it with color afterwards:

rohkea:
Behaves strangely for cases when many names are juxtaposed \N[4]\N[4]\N[4]\N[4] (I used names N3="N4 is{\N[4]}", N4="N3 is{\N[3]}", and only the second two names were replaced), but those are pretty rare. Normal use cases should work.

### DIFF
--- a/src/window_message.cpp
+++ b/src/window_message.cpp
@@ -18,6 +18,7 @@
 // Headers
 #include <cctype>
 #include <sstream>
+#include <iterator>
 
 #include "window_message.h"
 #include "game_actors.h"
@@ -84,6 +85,7 @@ void Window_Message::ApplyTextInsertingCommands() {
 
 	// Contains already substitued \N actors to prevent endless recursion
 	std::vector<int> replaced_actors;
+	int actor_replacement_start = std::distance(text.begin(), end);
 
 	if (!text.empty()) {
 		// Move on first valid char
@@ -105,6 +107,10 @@ void Window_Message::ApplyTextInsertingCommands() {
 				bool success;
 				int parsed_num;
 				std::u32string command_result = Utils::DecodeUTF32(ParseCommandCode(success, parsed_num));
+				if (start_code < text.begin() + actor_replacement_start) {
+					replaced_actors.clear();
+				}
+
 				if (!success || std::find(replaced_actors.begin(), replaced_actors.end(), parsed_num) != replaced_actors.end()) {
 					text_index = start_code - 2;
 					continue;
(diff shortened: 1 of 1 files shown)


Ghabry:
I get random replacements. The first message box replaces more \N[X] then the next ones.
Not sure what the reason is but I think one of the problems is that the "actor_replacement_start" iterator becomes invalid when text.replace is called because the string the iterator belongs to is changed.

src/window_message.cpp
```diff
@@ -105,13 +106,20 @@ void Window_Message::ApplyTextInsertingCommands() {
...
+				}
+
 				if (!success || std::find(replaced_actors.begin(), replaced_actors.end(), parsed_num) != replaced_actors.end()) {
 					text_index = start_code - 2;
 					continue;
 				}
 
 				if (ch == 'n') {
 					replaced_actors.push_back(parsed_num);
+					if (actor_replacement_start >= text_index - 1) {
```
Ghabry: I think the invalid iterator is really the problem.
Changing actor_replace_start = end; to =start;

and the code to:
```cpp
				long actor_replacement_dif = 0;
				if (ch == 'n') {
					replaced_actors.push_back(parsed_num);
					if (actor_replacement_start <= text_index) {
						actor_replacement_dif = std::distance(text.begin(), text_index);
					}
				}

				text.replace(start_code, text_index + 1, command_result);
				// Start from the beginning, the inserted text might add new commands
				text_index = text.end();
				end = text.end();

				actor_replacement_start = text.begin();
				actor_replacement_start += actor_replacement_dif;
```
Basicly remembering the iterator distance in the old code and moving the iterator this distance in the new.

fixes the problem I mentioned for me. Could you retest?

But the following is broken now: When an actor is called \N[5] and 4 is called \N[4] :/

carstene1ns:
I would really like to have this one fixed before 0.5.3… 🤕

rohkea:
I have changed actor_replacement_start, now it’s an index and not an iterator, so it doesn’t become invalid when the string is changed. It looks somewhat ugly because of the iterator↔index conversions, but seems to work for me.

Please test.

Feedback on how to make this cleaner is welcome.

Ghabry:
Jenkins: Test this please

Issue #1132, opened 2017-03-25
VH1 heroine hangs when you go through the East Gate


johnflux:
#### Name of the game:

VH1:  VH_170318_translated.7z

Download: https://mega.nz/#F!dssjSJ4R!EAwCogg9ZEkJvowDIg319A
Password: nanako

#### Player platform:

Linux, Kubuntu 16.10

#### Attach files (as a .zip archive or link them)

[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/870454/easyrpg_log.txt)

[save.zip](https://github.com/EasyRPG/Player/files/870456/save.zip)

#### Describe the issue in detail and how to reproduce it:

Load the saved game, then travel  left a few squares.  What should happen is that you pass through the gate.

What happens instead is that it soft locks up.

johnflux:
gdb only says:

#3  0x000055555564b558 in SdlUi::ProcessActiveEvent (this=0x5555559ec4e0, evnt=...) at src/sdl_ui.cpp:713
#0  0x00007ffff5d9471d in nanosleep () at ../sysdeps/unix/syscall-template.S:84
#1  0x00007ffff72b7b45 in ?? () from /usr/lib/x86_64-linux-gnu/libSDL2-2.0.so.0
#2  0x00007ffff72324df in ?? () from /usr/lib/x86_64-linux-gnu/libSDL2-2.0.so.0
#3  0x000055555564b558 in SdlUi::ProcessActiveEvent (this=0x5555559ec4e0, evnt=...) at src/sdl_ui.cpp:713
#4  0x000055555564c2ab in SdlUi::ProcessEvents (this=0x5555559ec4e0) at src/sdl_ui.cpp:568
#5  0x0000555555574c9c in Player::Update (update_scene=update_scene@entry=true) at src/player.cpp:325
#6  0x00005555555820a2 in Scene::MainFunction (this=0x55555c4ea4b0) at src/scene.cpp:100
#7  0x0000555555573200 in Player::MainLoop () at src/player.cpp:264
#8  0x0000555555575675 in Player::Run () at src/player.cpp:259
#9  0x000055555556c79e in main (argc=<optimized out>, argv=<optimized out>) at src/main.cpp:28

johnflux:
This is against the latest git version, as of 2017/3/26  (version ed0e6432fc15e3d71defc0aac93cd6250e988a18 )

carstene1ns:
[...]

PR #1233, opened 2017-07-30
Fix walking through the east gate (#1132) in VH1

Ghabry:
See commit message


Ghabry:
When the ID of the autostart event is higher than the ID of the parallel event which enables the autostart event it will still hang. Which is correct behaviour. (The VH1 devs were lucky ;))

carstene1ns:
Wow, this looks like having potential to (un)break a lot of games 😀

PR #1238, opened 2017-08-06
lcf vNext Step one: Rewrite some data access on liblcf fields

Ghabry:
Replace +=, -=, ++, -- operations on and referencing of liblcf data fields.

This is required for the liblcf vNext rewrite and my automatic converter can't handle these special cases.

The big commit (https://github.com/Ghabry/easyrpg-player/commit/8d21a02b91397a47280fd6ce67820e1fd8bcfbde) follows in a month or two when the JSON code is more ready.

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -1030,12 +1030,16 @@ std::string Game_Actor::GetClassName() const {
 }
 
 void Game_Actor::SetBaseMaxHp(int maxhp) {
-	GetData().hp_mod += maxhp - GetBaseMaxHp();
+	int new_hp_mod = GetData().hp_mod + (maxhp - GetBaseMaxHp());
+	GetData().hp_mod = new_hp_mod;
+
 	SetHp(GetData().current_hp);
 }
 
 void Game_Actor::SetBaseMaxSp(int maxsp) {
-	GetData().sp_mod += maxsp - GetBaseMaxSp();
+	int new_sp_mod = GetData().sp_mod + (maxsp - GetBaseMaxSp());
+	GetData().sp_mod = new_sp_mod;
+
 	SetSp(GetData().current_sp);
 }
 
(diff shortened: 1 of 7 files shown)


src/game_picture.cpp
```diff
@@ -204,30 +204,30 @@ void Game_Picture::Update() {
 	if (data.time_left == 0) {
 		SyncCurrentToFinish();
 	} else {
-		auto interpolate = [=](double& current, double finish) {
+		auto interpolate = [=](double current, double finish) {
 			double d = data.time_left;
-			current = (current * (d - 1) + finish) / d;
+			return (current * (d - 1) + finish) / d;
```
carstene1ns: Why return here?

Ghabry: Because before _interpolate_ was overwriting _current_ with the new value instead of returning it.

src/game_picture.cpp
```diff
@@ -204,30 +204,30 @@ void Game_Picture::Update() {
...
-		interpolate(data.current_bot_trans, data.finish_bot_trans);
+		data.current_x = interpolate(data.current_x, data.finish_x);
+		data.current_y = interpolate(data.current_y, data.finish_y);
+		data.current_red = interpolate(data.current_red, data.finish_red);
+		data.current_green = interpolate(data.current_green, data.finish_green);
+		data.current_blue = interpolate(data.current_blue, data.finish_blue);
+		data.current_sat = interpolate(data.current_sat, data.finish_sat);
+		data.current_magnify = interpolate(data.current_magnify, data.finish_magnify);
+		data.current_top_trans = interpolate(data.current_top_trans, data.finish_top_trans);
+		data.current_bot_trans = interpolate(data.current_bot_trans, data.finish_bot_trans);
```
Ghabry: And the change to _interpolate_ was made because data.[fields] can't be taken by reference anymore with vNext.

src/game_actor.cpp
```diff
@@ -1068,19 +1072,23 @@ void Game_Actor::SetSp(int sp) {
...
 void Game_Actor::SetBaseDef(int def) {
-	GetData().defense_mod += def - GetBaseDef();
+	int new_defense_mod = GetData().defense_mod + (def - GetBaseDef());
+	GetData().defense_mod = new_defense_mod;
 }
 
 void Game_Actor::SetBaseSpi(int spi) {
-	GetData().spirit_mod += spi - GetBaseSpi();
+	int new_spirit_mod = GetData().spirit_mod + (spi - GetBaseSpi());
+	GetData().defense_mod = new_spirit_mod;
```
fdelapena: Shouldn't be `GetData().spirit_mod`?

Ghabry: Yes

PR #1239, opened 2017-08-08
Add Simplified Chinese support in hero name input

fdelapena:
Based on a table used in a forum fork.

~~~~Untested yet. Please review ;)~~~~

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -150,7 +150,8 @@ namespace {
 		int min_height, max_height;
 		std::function<BitmapRef()> dummy_renderer;
 	} const spec[] = {
-		{ "Backdrop", false, 320, 320, 160, 160, backdrop_dummy_func},
+		/* FIXME: 240px max height is 2k3 specific, for 2k is 160px */
+		{ "Backdrop", false, 320, 320, 160, 240, backdrop_dummy_func},
 		{ "Battle", true, 480, 480, 96, 480, battle_dummy_func },
 		{ "CharSet", true, 288, 288, 256, 256, charset_dummy_func },
 		{ "ChipSet", true, 480, 480, 256, 256, chipset_dummy_func },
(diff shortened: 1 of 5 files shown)


src/cache.cpp
```diff
@@ -150,7 +150,7 @@ namespace {
 		int min_height, max_height;
 		std::function<BitmapRef()> dummy_renderer;
 	} const spec[] = {
-		{ "Backdrop", false, 320, 320, 160, 160, backdrop_dummy_func},
+		{ "Backdrop", false, 320, 320, 160, Player::IsRPG2k() ? 160 : 240, backdrop_dummy_func},
```
Ghabry: I still think that this will not work because the specs are const

src/window_keyboard.cpp
```diff
@@ -89,6 +93,30 @@ std::string Window_Keyboard::items[Window_Keyboard::MODE_END][9][10] = {
...
+		{"玛", "迈", "曼", "梅", "美", "门", "米", "密", "明", "缪"},
+		{"摩", "莫", "姆", "穆", "那", "娜", "纳", "奈", "南", "尼"},
+		{"宁", "纽", "奴", "诺", "欧", "帕", "派", "佩", "皮", "普"},
+		{"奇", "琪", "琼", "丘", "萨", "撒", "赛", "桑", "瑟", "森"},
+		{"沙", "山", "珊", "史", "世", "斯", "丝", "司", "苏", "所"},
+		{"索", "塔", "泰", "坦", "汤", "特", "提", "汀", "统", "瓦"},
+		{"威", "维", "韦", "卫", "温", "沃", "乌", "西", "希", "夏"},
+		{"辛", "修", "休", "雅", "亚", "林", "琳", "留", "隆", "鲁"},
+		{"伊", "英", "尤", "则", "扎", "", Window_Keyboard::TO_ZH_CN_1, "", Window_Keyboard::DONE_ZH_CN}
+	},
```
Ghabry: I wonder how the "alphabet" for chinese is picked. Probably phonetic?

src/cache.cpp
fdelapena: Specs are const but while defining expect `int` and not `const int`.

Ghabry: The code is wrong because the const spec[] is initialized on startup. Which means the result will be either 160 or undefined behaviour (depending on if the static initializer for the engine version runs before Cache).

Ghabry: I would agree on changing the value in general to "240" until the Cache system is refactored. This is wrong for 2k but gets rid of 2k3 debug messages.

PR #1246, opened 2017-08-12
Fix #1002 (Aurora Wing). Limit the variable range to [-999999,999999] for RPG Maker 2000 games.

### LINKED ISSUE #1002: Aurora Wing: Intro has a skippable hang and first battle gets stuck after enemy turn
Aurora Wing is some Fire Emblem clone.

Download: https://rpgmaker.net/games/42/downloads/27/

Web testcase: https://easy-rpg.org/play/master/?game=aurora_wing
- [X] Intro hangs directly after the options menu shown at the beginning (Intro can be skipped with CANCEL)
- [x] after the (super slow) turns of the enemy the player does not get a turn.

Looks like "MakeWay" can't make way and therefore all movements fail.

Probably a problem with diagonal movements.

Ghabry:
-

### DIFF
--- a/src/game_interpreter.h
+++ b/src/game_interpreter.h
@@ -63,11 +63,6 @@ public:
 
 	virtual bool ExecuteCommand();
 
-	enum Sizes {
-		MaxSize = 9999999,
-		MinSize = -9999999
-	};
-
 protected:
 	friend class Game_Interpreter_Map;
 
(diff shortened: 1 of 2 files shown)


src/game_interpreter.h
```diff
@@ -64,8 +64,10 @@ class Game_Interpreter
 	virtual bool ExecuteCommand();
 
 	enum Sizes {
```
carstene1ns: C++11 allows typed enums, this might help not needing the cast to `int` later.

PR #1252, opened 2017-08-19
Fix some SDL annoyances under Linux when using F4 & F5

Ghabry:
Not tested under Windows yet.


Ghabry:
ALT+Enter is still broken for entering fullscreen (leaving works) under Linux but fixing this would need exposing the ALT-modifier to our Input code which is not worth it.

Ghabry:
No regressions under Windows

Ghabry:
> the second commit does not really work here, but this might be another bug
You clicked the maximize
Only tested under KDE.
The problem for me was: When you click the maximize button in the titlebar and press F5 afterwards the window content is rendered in the bottom left corner.

carstene1ns:
I read that commit wrong then, thought when I press <kbd>F4</kbd> and <kbd>F5</kbd> afterwards, it does the restore (which not happens). However, I cannot really maximise windows with my tiling wm, but when I use the fullscreen hotkey it does the same you noticed (rendering in bottom left corner).

PR #1254, opened 2017-08-23
Redirect Teleport and Escape skills that target areas to the owning map

### LINKED ISSUE #1253: Teleport and Escape skills don't handle Areas correctly
#### Name of the game:
DQ2　悪霊の勇者・製品版ver1.50  (and DQ but DQ2 has more such problems)

Download: http://www1.axfc.net/uploader/so/1768371 (NFSW. full of hentai ads!)

Savegame: 
[Save01.zip](https://github.com/EasyRPG/Player/files/1246942/Save01.zip)

Load the save, go to the item list and use the キメラ item. Select the only option and you get an error that map 35 is missing.

The map is indeed missing. It is an Area of map 1.

Looks like the Skills "Teleport" (teleports added through event "Add Teleport Place") and "Escape" (added through event "Add Escape Place") detect when the target is an area and change the target to the owner of the area.
The normal "Teleport" event command doesn't do this.

Ghabry:
The only games affected by this are the two japanese Dragon Quest fangames DQ列伝 and DQ2悪霊の勇者・製品版

Fix #1253

The two new functions in Game_Map could be used in some other places, too but I'm too afraid to break something...

### DIFF
--- a/src/game_player.cpp
+++ b/src/game_player.cpp
@@ -223,7 +223,14 @@ void Game_Player::ReserveTeleport(int map_id, int x, int y, int direction) {
 }
 
 void Game_Player::ReserveTeleport(const RPG::SaveTarget& target) {
-	ReserveTeleport(target.map_id, target.map_x, target.map_y, Down);
+	int map_id = target.map_id;
+
+	if (Game_Map::GetMapType(target.map_id) == RPG::TreeMap::MapType_area) {
+		// Area: Obtain the map the area belongs to
+		map_id = Game_Map::GetParentId(target.map_id);
+	}
+
+	ReserveTeleport(map_id, target.map_x, target.map_y, Down);
 
 	if (target.switch_on) {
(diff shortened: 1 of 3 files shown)


src/game_player.cpp
```diff
@@ -223,7 +223,14 @@ void Game_Player::ReserveTeleport(int map_id, int x, int y, int direction) {
 }
 
 void Game_Player::ReserveTeleport(const RPG::SaveTarget& target) {
-	ReserveTeleport(target.map_id, target.map_x, target.map_y, Down);
+	int map_id = target.map_id;
+
+	if (Game_Map::GetMapType(target.map_id) == 2) {
```
carstene1ns: We really need some enums/constants for such things.

Ghabry: I agree on this, but the build will fail now ;)

carstene1ns:
jenkins: test this please

PR #1256, opened 2017-09-05
Android translations

carstene1ns:
btw. I have an additional patch that moves the library flags to another location and such does not spew the warnings we get currently. However, this one should be simple for faster review and merge, so not included: https://paste.xinu.at/oYe/


Ghabry:
Jenkins: Test this please ♥

PR #1258, opened 2017-09-09
Migrate Android build to clang & libc++

### LINKED ISSUE #1185: Android: game screen sometimes hops into offscreen area after turning device or toggling app
happening on Android 7 (Samsung tablet):
[offscreen](https://picload.org/image/riaririi/screenshot_20170521-192612.png)
(click on the screnshot to display in native resolution)

The game itself is playable and display can be fixed by turning the screen again.

Ghabry:
Also update all libs (fixes #1185).


builds/android/app/src/main/java/org/libsdl/app/SDLActivity.java
```diff
@@ -974,8 +1045,6 @@ public void run() {
     in order to do anything useful.
 
     Because of this, that's where we set up the SDL thread
-
-    EasyRPG modification: disable touch input
```
carstene1ns: this comment should stay as it marks the beginning of our change

src/system.h
```diff
@@ -90,7 +90,7 @@
 #    define SUPPORT_KEYBOARD
 #  endif
 
-#  ifdef USE_SDL_MIXER
+#  ifdef HAVE_SDL_MIXER
 #    define SUPPORT_WAV
```
carstene1ns: These defines are not used anyway, guess they were meant for the filefinder at some point.

builds/android/app/src/main/jni/src/Android.mk
```diff
@@ -33,21 +19,21 @@ LOCAL_SRC_FILES := SDL_android_main.c \
 
 LOCAL_SHARED_LIBRARIES := SDL2
 
-LOCAL_STATIC_LIBRARIES := cpufeatures
+LOCAL_STATIC_LIBRARIES := \
+		SDL2_mixer vorbisfile vorbis ogg \
+		xmp-lite mpg123 speexdsp sndfile \
+		freetype pixman-1 png \
+		lcf expat icui18n icuuc icudata \
+		cpufeatures
```
carstene1ns: I was not sure here if it still uses our patched lib, if we move cpufeatures.

builds/android/app/src/main/jni/src/Android.mk
```diff
@@ -33,21 +19,21 @@ LOCAL_SRC_FILES := SDL_android_main.c \
 
 LOCAL_SHARED_LIBRARIES := SDL2
 
-LOCAL_STATIC_LIBRARIES := cpufeatures
+LOCAL_STATIC_LIBRARIES := \
+		vorbisfile vorbis ogg WildMidi \
+		xmp-lite mpg123 speexdsp sndfile \
+		freetype pixman-1 png opusfile opus \
+		lcf expat icui18n icuuc icudata \
+		cpufeatures
```
carstene1ns: I was not sure here if it still uses our patched lib, if we move cpufeatures.

Ghabry: Me neither, the problem with cpufeatures was only on some 4.0.3 devices. So we can just wait what happens.

src/system.h
Ghabry: yeah, most of the defines are useless

Ghabry:
Please review my SDLActivity.java 2.0.6 update if I missed some EasyRPG changes again.

carstene1ns:
Just a small observation: We had a command to set the platform/SDK to 23 in our build scripts.
This leads to the following warning now, question is, if we support(ed) the older SDKs at all.
IMO we should cleanup every legacy stuff in the app and only support the oldest device we have.

```
Android NDK: Found platform level in ./project.properties. Setting APP_PLATFORM to android-23.    
Android NDK: WARNING: APP_PLATFORM android-23 is higher than android:minSdkVersion 10 in ./AndroidManifest.xml. NDK binaries will *not* be comptible with devices older than android-23. See https://android.googlesource.com/platform/ndk/+/master/docs/user/common_problems.md for more information.
```

Ghabry:
This warning is nothing new, this warnings was even generated with the old NDK we used. But I havn't tried a hardware test yet with the updated APK, can't say if they changed the behaviour now.
Well, the oldest SDK we can test easily is SDK 15 (my phone).

I have nothing against deprecating API < 15 but there should be some reason like "refactor that removes part of the compat library" and the last version for 2.3.3 shouldn't have any regressions 👍.

carstene1ns:
My point is just: when we cannot test it we should not support it.

carstene1ns:
jenkins: test this please (green checkmark pls.)

carstene1ns:
hmm, a clean build says, minimum SDK is 14… 🍇
> Android NDK: APP_PLATFORM not set. Defaulting to minimum supported version android-14.    
Android NDK: WARNING: APP_PLATFORM android-14 is higher than android:minSdkVersion 10 in ./AndroidManifest.xml. NDK binaries will *not* be comptible with devices older than android-14. See https://android.googlesource.com/platform/ndk/+/master/docs/user/common_problems.md for more information.

Ghabry:
Well, the make-standalone-sdk.py already disliked that I wanted API9. So will need a quick test in an emulator (if you can still find a 2.3.3 emu) before 0.5.3.
Maybe we are lucky because we don't call much Android API from native code, except SDL which probably has conditions.

Issue #1248, opened 2017-08-13
Investigate minimum Actor Level and other off-by-one's


carstene1ns:
In *COLORS: Lost Memories* the initial actor level is 1 and this gets decreased in the intro by 1, so is effectively 0, RPG_RT allows that. We cap this to 1 in `Game_Actor::SetLevel()`.
This can be seen when looking at savegames created by RPG_RT and Player for this game.

We might have similar problems with attributes such as Atk, Def, Agi, etc.

Ghabry:
The number shown in the savegame scene is wrong in RPGRT (for easyrpg saves) in general. The Deep8 dev told me that all values are shown as 1 and when loading e.g. a OFF save in RPGRT the SP are 1.

carstene1ns:
Upon further investigation I do not know what this game is doing.
I created a testcase in 2k and 2k3, decreased the level and HP as this game does and both do not allow 0.
However, in this game it is possible (even when I replace the RPG_RT).

![20170815-040002](https://user-images.githubusercontent.com/4691314/29299353-4a9983f6-816e-11e7-9f37-9ffd6cb762c4.png)

carstene1ns:
Your sanity check branch yields:

```
Warning: Actor 1: Invalid level 0
Debug: Actor 2: Removing invalid item 18 (of type 0) from equipment slot 1 (needs type 1)
Debug: Actor 3: Removing invalid item 18 (of type 0) from equipment slot 1 (needs type 1)
Debug: Actor 4: Removing invalid item 18 (of type 0) from equipment slot 1 (needs type 1)
Warning: Actor 5: Invalid level 0
Warning: Actor 7: Invalid level 0
```

Ghabry:
This is caused by a manual database edit. The initial_level is set to 0.
Seems that RPG_RT accepts any level for this field and uses stats of 0 and EXP of 0 when the level is invalid.

Adding EXP works as expected (when you have 1 EXP you become LV1) and Level ups, too. But when the new level is below 1 showing the "level up message" will hang RPG_RT. Without the message everything works fine.

Ghabry:
Are you sure about the "Level 1 to 0 decrease"? Even when I set the initial level to 0 I can't decrease again to this level with RPG_RT.

PR #1259, opened 2017-09-09
Add sanity checks to all LCF data structure access

Ghabry:
to reduce crashes and exploit surface.

### DIFF
--- a/src/background.cpp
+++ b/src/background.cpp
@@ -25,6 +25,8 @@
 #include "background.h"
 #include "bitmap.h"
 #include "main_data.h"
+#include "reader_util.h"
+#include "output.h"
 
 Background::Background(const std::string& name) :
 	visible(true),
@@ -47,33 +49,38 @@ Background::Background(int terrain_id) :
 
 	Graphics::RegisterDrawable(this);
 
-	const RPG::Terrain& terrain = Data::terrains[terrain_id - 1];
+	const RPG::Terrain* terrain = ReaderUtil::GetElement(Data::terrains, terrain_id);
+
+	if (!terrain) {
+		Output::Warning("Invalid terrain ID %d", terrain_id);
+		return;
+	}
 
 	// Either background or frame
-	if (terrain.background_type == RPG::Terrain::BGAssociation_background && !terrain.background_name.empty()) {
-		FileRequestAsync* request = AsyncHandler::RequestFile("Backdrop", terrain.background_name);
+	if (terrain->background_type == RPG::Terrain::BGAssociation_background && !terrain->background_name.empty()) {
+		FileRequestAsync* request = AsyncHandler::RequestFile("Backdrop", terrain->background_name);
 		request_id = request->Bind(&Background::OnBackgroundGraphicReady, this);
 		request->Start();
 		return;
 	}
 
 	// Frame
-	if (!terrain.background_a_name.empty()) {
-		FileRequestAsync* request = AsyncHandler::RequestFile("Frame", terrain.background_a_name);
+	if (!terrain->background_a_name.empty()) {
+		FileRequestAsync* request = AsyncHandler::RequestFile("Frame", terrain->background_a_name);
 		request_id = request->Bind(&Background::OnBackgroundGraphicReady, this);
 		request->Start();
 
-		bg_hscroll = terrain.background_a_scrollh ? terrain.background_a_scrollh_speed : 0;
-		bg_vscroll = terrain.background_a_scrollv ? terrain.background_a_scrollv_speed : 0;
+		bg_hscroll = terrain->background_a_scrollh ? terrain->background_a_scrollh_speed : 0;
+		bg_vscroll = terrain->background_a_scrollv ? terrain->background_a_scrollv_speed : 0;
 	}
 
-	if (terrain.background_b && !terrain.background_b_name.empty()) {
-		FileRequestAsync* request = AsyncHandler::RequestFile("Frame", terrain.background_b_name);
+	if (terrain->background_b && !terrain->background_b_name.empty()) {
+		FileRequestAsync* request = AsyncHandler::RequestFile("Frame", terrain->background_b_name);
 		request_id = request->Bind(&Background::OnForegroundFrameGraphicReady, this);
 		request->Start();
 
-		fg_hscroll = terrain.background_b_scrollh ? terrain.background_b_scrollh_speed : 0;
-		fg_vscroll = terrain.background_b_scrollv ? terrain.background_b_scrollv_speed : 0;
+		fg_hscroll = terrain->background_b_scrollh ? terrain->background_b_scrollh_speed : 0;
+		fg_vscroll = terrain->background_b_scrollv ? terrain->background_b_scrollv_speed : 0;
 	}
 }
 
--- a/src/background.cpp
+++ b/src/background.cpp
@@ -52,7 +52,7 @@ Background::Background(int terrain_id) :
 	const RPG::Terrain* terrain = ReaderUtil::GetElement(Data::terrains, terrain_id);
 
 	if (!terrain) {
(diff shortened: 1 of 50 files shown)


src/game_actor.cpp
```diff
@@ -1133,18 +1213,36 @@ int Game_Actor::GetHitChance() const {
...
+	 * The following actor data is trusted after the initial verify:
+	 * - Class, SetClass filters invalid data
+	 * - Level is between 1 and 99
+	 * - Equipment
+	 * - Learned Skills
+	 * - ...
+	 *
+	 * Data::items (but not skills of items!)
+	 *
+	 */
```
Ghabry: have to update this comment

src/game_actor.cpp
```diff
@@ -459,9 +484,8 @@ int Game_Actor::CalculateExp(int level) const
 }
 
 void Game_Actor::MakeExpList() {
-	int final_level = Data::actors[actor_id - 1].final_level;
-	exp_list.resize(final_level, 0);;
-	for (int i = 1; i < final_level; ++i) {
+	exp_list.resize(99);
```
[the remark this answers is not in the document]
Ghabry: Yes, this looks safe because GetLevel() is sanitized.

Ghabry:
Fixed the reported problems and found more problems in the code.
Did a new commit to make the review easier.

src/window_battlestatus.cpp
```diff
@@ -102,6 +103,7 @@ void Window_BattleStatus::RefreshGauge() {
 		}
 
 		for (int i = 0; i < item_max; ++i) {
+			// The always only contains valid battlers
```
carstene1ns: who?

carstene1ns:
Looks good to me now apart from a few language issues in the comments, but this does not matter.
Not going to merge this, until at least a second review is available :grin:.

Ghabry:
And that review should take at least 30 minutes ;)

src/scene_order.cpp
```diff
@@ -19,6 +19,7 @@
 #include <algorithm>
 #include <string>
 #include <vector>
+#include <reader_util.h>
```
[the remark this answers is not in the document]
Ghabry: Is really unused. 👍

src/background.cpp
```diff
@@ -25,6 +25,8 @@
 #include "background.h"
 #include "bitmap.h"
 #include "main_data.h"
+#include "reader_util.h"
```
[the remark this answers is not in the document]
Ghabry: Fixed all your notes except this one because it already uses quotes

Ghabry:
Don't merge this yet. Map data (Passages and Upper/Lower) is not verified...

Ghabry:
Adjusted my sanity check code to the latest Battle CMD changes of 20kdc. Ready for merge now.

src/game_battler.cpp
```diff
@@ -524,8 +562,11 @@ int Game_Battler::GetSpi() const {
 	int n = min(max(base_spi, 1), 999);
 
 	for (int16_t i : GetInflictedStates()) {
-		if(Data::states[i - 1].affect_spirit) {
-			n = AffectParameter(Data::states[i - 1].affect_type, base_spi);
+		// States are guaranteed to be valid
+		const RPG::State& state = *ReaderUtil::GetElement(Data::states, i);
+
```
carstene1ns: Stray newline or missing in others.

Try to keep `GetAtk`, `GetDef()`, `GetSpi()` and `GetAgi()` the same.

src/game_interpreter.cpp
```diff
@@ -2811,12 +2869,18 @@ bool Game_Interpreter::CommandReturnToTitleScreen(RPG::EventCommand const& /* co
...
 bool Game_Interpreter::CommandChangeClass(RPG::EventCommand const& com) { // code 1008
-	int class_id = com.parameters[2];
+	int class_id = com.parameters[2]; // 0: No class, 1+: Specific class
 	bool level1 = com.parameters[3] > 0;
 	int skill_mode = com.parameters[4]; // no change, replace, add
 	int stats_mode = com.parameters[5]; // no change, halve, level 1, current level
 	bool show = com.parameters[6] > 0;
 
+	const RPG::Class* cls = ReaderUtil::GetElement(Data::classes, class_id);
+	if (class_id > 0) {
```
carstene1ns: I do not really understand this. What happens here?

src/scene_battle.cpp
```diff
@@ -314,16 +342,42 @@ void Scene_Battle::AssignSkill(const RPG::Skill* skill) {
...
+					Output::Warning("AssignSkill: Item %d references invalid skill %d", skill_item->ID, skill_item->skill_id);
+					return;
+				}
+			}
+		}
+		default:
+			break;
+	}
+
+
```
carstene1ns: Stray newline.

carstene1ns:
Additionally, I do not really like the additional braces you introduced in some `switch` `case`s.
IMO proper indentation is sufficient there.

We will need some code style document in the future to address such things.

Ghabry:
Thx for the feedback. Do you mean braces like here:
https://github.com/Ghabry/easyrpg-player/blob/5fdc7d78e07f6c8196a6f3405cfbda25c7f3c7e9/src/game_interpreter.cpp#L2823-L2830

Unfortunately the braces are required here because you can't declare variables in a case without introducing a scope :/

src/game_actor.cpp
```diff
@@ -421,17 +447,17 @@ int Game_Actor::GetBaseAgi() const {
 	return GetBaseAgi(true, true);
 }
 
-int Game_Actor::CalculateExp(int level) const
-{
+int Game_Actor::CalculateExp(int level) const {
+	const RPG::Class* klass = ReaderUtil::GetElement(Data::classes, GetData().class_id);
+
 	double base, inflation, correction;
 	if (GetData().class_id > 0) {
```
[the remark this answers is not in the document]
Ghabry: Indeed. Looks like a bad rebase of the class_changed rename.

carstene1ns:
Okay, did not know they were needed then.

carstene1ns:
There seems to be a problem when returning to the gamebrowser from some games (maybe data is not cleared or cleared in the wrong order?): `Warning: Removing invalid item # from party`

Ghabry:
Can you telll me a game where thus can be reproduced?

fdelapena:
I guess it's the mentioned in EasyRPG/Player#1248

carstene1ns:
Example games: Embric, Jojo, Lavender, Mondschein
To reproduce load a save or begin the game, press <kbd>F12</kbd> and choose Exit in the Menu.

Ghabry:
Had to add special handling for level 0 everywhere to prevent nullptr-dereferences due to new sanity checks (guess before it read out of bounds and somehow survived).

The last commit was a random find while testing Ara Fell.

Ghabry:
Sorry for the rebase, was to prevent a future conflict. Had to remove a sanity check in "SetupBattle" because the original code has a bug and can be completely deleted.  Will open an issue for this.

The last unreviewed commit was:
"Add exception in the sanity check and special handling for Level 0"

carstene1ns:
Well, have further investigated the reason why the message pops up:
When returning from game to title scene via <kbd>F12</kbd>, `Data` is not cleared when [`Player::ResetGameObjects()`](https://github.com/Ghabry/easyrpg-player/blob/sanitycheck/src/scene_title.cpp#L58) is called. Thus, the items are still valid. However, when returning to the gamebrowser, they are still there, but [`Data::Clear()`](
https://github.com/Ghabry/easyrpg-player/blob/sanitycheck/src/scene_gamebrowser.cpp#L60-L61) has been called, so they are invalid.

As a fix, i think clearing the party items is necessary when leaving the game, either in `ResetGameObjects()` or where the party members are removed.

Ghabry:
Looking forward to the bug reports. ;)

PR #1260, opened 2017-09-09
Emterpretify to support frameskip in the web player

Ghabry:
(already updated the jenkins PR job)

To allow sleep the following functions must run through the emterpreter (slow!):
main -> Player::Run -> Player::MainLoop -> Scene::MainFunction -> Player::Update -> SDLUi::Sleep

The only "complex" function here is Player::Update but even this function contains mostly only function calls which execute through asm.js (except SDLUi::Sleep) so the performance penalty is probably not noticable (and if it really runs slower this is frameskipped now :P)

### DIFF
--- a/src/async_handler.cpp
+++ b/src/async_handler.cpp
@@ -59,13 +59,18 @@ namespace {
 	void download_success(unsigned, void* userData, const char*) {
 		FileRequestAsync* req = static_cast<FileRequestAsync*>(userData);
 		//Output::Debug("DL Success: %s", req->GetPath().c_str());
+
 		req->DownloadDone(true);
+
+		emscripten_sleep_with_yield(1);
 	}
 
 	void download_failure(unsigned, void* userData, int) {
 		FileRequestAsync* req = static_cast<FileRequestAsync*>(userData);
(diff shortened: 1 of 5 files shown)


Ghabry:
Jenkins: Test this please

hmmmm, my local build works with
```
  -s EMTERPRETIFY=1 -s EMTERPRETIFY_ASYNC=1 \
  -s EMTERPRETIFY_WHITELIST='["_main", "_ZN6Player3RunEv", "_ZN6Player8MainLoopEv", "_ZN5Scene12MainFunctionEv", "__ZN6Player6UpdateEb", "_ZN5SdlUi5SleepEj"]'
```

Ghabry:
The problem is the following. Because we have assertions enabled the runtime will cancel when the emterpreter runs but at the same time javascript executes. This is the case for our audio "thread". The emterpreter things this is not emterpreted (which is correct) code and aborts.
So, @fdelapena and @carstene1ns. Your opinion. Two options:

* Disable assertions and use the emterpreter to support frameskips
* Keep them on and reject this PR

Ghabry:
closing this one in favor of a new sleep-free frame limiting approach that doesn't depend on the Emterpreter

PR #1273, opened 2017-10-06
Begin Rpg2k3 1.12 support

Ghabry:
This implements many RPG2k3 1.12 picture features:

 - Spritesheets
 - Z layer
 - Picture ID, opacity & co. through variable
 - Name substitution (as known from PPP)
 - Survive map change

Not part of this PR:
 - Battle pictures
 - Affected by Tone/Shake/Flash

Test cases: https://github.com/EasyRPG/TestGame/pull/36


carstene1ns:
jenkins: test this please 😘

Ghabry:
Jenkins: Test this please

Ghabry:
Besides the TestGame-2000 map I only know one "game" that uses it. I will try to get this Testgame by Archeia that is shipped with RPG2k3 1.12a.

Ghabry:
[ArcheiaPictureTutorial.zip](https://github.com/EasyRPG/Player/files/1380629/ArcheiaPictureTutorial.zip)

Ghabry:
I found two bugs in the testgame: THe "Animated picture" option shows the full image for one frame.
WHen leaving the basement the magic circle suddenly appears again after the fadeout. Though not sure if this is just one of our many bugs with transitions.

Ghabry:
Fixed by moving "data.frames = data.frames + 1;" after the if-block. Rest remains unchanged.
Ready from my side.

Ghabry:
-

src/game_picture.cpp
```diff
@@ -167,13 +223,28 @@ void Game_Picture::RequestPictureSprite() {
...
+	if (data.spritesheet_rows < 1 || data.spritesheet_cols < 1) {
+		return false;
+	}
+
+	if (data.spritesheet_rows == 1 && data.spritesheet_cols == 1) {
+		return false;
+	}
+
+	return true;
+}
```
carstene1ns: Why not simpler?

```cpp
if (data.spritesheet_rows > 1 || data.spritesheet_cols > 1)
	return true;

return false;
```

Ghabry: Good find 👍

Ghabry: Simplified the check a bit.
But have to keep the first if-check because it's possible that only one of the values is > 1 due to uninitialized arguments.

carstene1ns:
This branch is broken when there are pictures and you use fast-forward. They stay, seems like `erase_on_map_change` needs to be set somewhere or maybe `Game_Screen::Reset()` needs `true`?

Ghabry:
hu? How can fast-forward affect this. Will check.

carstene1ns:
forget "fast forward", was a flawed test.

carstene1ns:
Works now, thanks!

PR #1278, opened 2017-10-09
Add Russian name input

rohkea:
Not very elegant (RuLatn and Letter are almost identical), but useful for many games.

screenshot_0

Can also be used for the following languages with no known games: Belarusian, Bulgarian, Crimean Tatar, Ukrainian (although those will feel second-class citizens as the extra letters are below Russian and not in their alphabetic places).

**NOT** useful for Serbian and Macedonian (too many extra letters). There are no known games in Serbian and Macedonian, but we’d eventually need to devise an encoding-independent way of specifying a language for such things. (Also needed for UTF-8 support.)

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -1034,6 +1034,11 @@ bool Player::IsCJK() {
 			IsBig5() || IsCP936());
 }
 
+bool Player::IsCP1251() {
+	return (encoding == "ibm-5347_P100-1998" ||
+			encoding == "windows-1251" || encoding == "1251");
+}
+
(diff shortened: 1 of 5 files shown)


Ghabry:
When fdela refactored the name scene for his Hacktober T-Shirt the duplicate letter array should be gone ;)

rohkea:
OK, I forgot the letter Ш ^^' Will add in another PR the evening

Ghabry:
Hm, I even reviewed the alphabet if its complete ^^', missed it too

Issue #1228, opened 2017-07-29
Yume 2kki: Invisible events shown on maps with empty tileset graphics


carstene1ns:
In **Yume 2kki** there are some tilesets without file name in the database (not really possible with the editor, likely manually done). They are used on some maps like the music room and the events using them should be invisible. Instead we draw a black box currently.

current (0.5.2/master) | expected (0.5.1/RPG_RT)
-|-
![s2](https://picload.org/image/rwrlclgr/screenshot_2.png) | ![s0](https://picload.org/image/rwrlcowi/screenshot_0.png)
![s3](https://picload.org/image/rwrlclga/screenshot_3.png) | ![s1](https://picload.org/image/rwrlcoww/screenshot_1.png)

Editor:
![editor](https://picload.org/image/rwrlcowl/20170729-163633.png)

This bug was introduced by #1179 (and not fixed by 8c8ab5303da). Before they were not visible, because `tile_id` was `0` in this case. btw. we actually check for this case in `Sprite_Character::OnTileSpriteReady()` and draw that black box:
```c++
	if (!Game_Map::GetChipsetName().empty())
		tile = Cache::Tile(Game_Map::GetChipsetName(), tile_id);
	else
		tile = Bitmap::Create(16, 16, Color()); // black box
```

Changing that to be a transparent box does not work (likely because we optimize that away somewhere). This patch seems to work, but I am not happy with it because everything is skipped:
```diff
diff --git a/src/sprite_character.cpp b/src/sprite_character.cpp
index fadfb09b..28702a09 100644
--- a/src/sprite_character.cpp
+++ b/src/sprite_character.cpp
@@ -45,7 +45,8 @@ void Sprite_Character::Update() {
 			FileRequestAsync* char_request = AsyncHandler::RequestFile("CharSet", character_name);
 			request_id = char_request->Bind(&Sprite_Character::OnCharSpriteReady, this);
 			char_request->Start();
-		} else {
+		// Yume 2kki has cleared tilesets (empty tileset_name) for some maps, making events transparent.
+		} else if (!Game_Map::GetChipsetName().empty()) {
 			FileRequestAsync* tile_request = AsyncHandler::RequestFile("ChipSet", Game_Map::GetChipsetName());
 			request_id = tile_request->Bind(&Sprite_Character::OnTileSpriteReady, this);
 			tile_request->Start();
@@ -92,16 +93,7 @@ bool Sprite_Character::UsesCharset() const {
 }
 
[...]

PR #1280, opened 2017-10-09
Small Bugfixes

carstene1ns:


Ghabry:
Based on the milestone list this fixes all issues except #1228 for which we havn't decided the best way to fix this yet.

PR #1284, opened 2017-10-12
Allow re-using name-input layouts for different languages

### LINKED ISSUE #1272: Ib fan translation uses the Japanese text entry field instead of the English one
#### Name of the game:
Ib (Most recent English translation by vgperson)

#### Player platform:
Windows

[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/1361426/easyrpg_log.txt)
[Save01.zip](https://github.com/EasyRPG/Player/files/1361432/Save01.zip)

In the "Secret Room" of The Dungeon in Ib, there's a room where you have to enter a name in order to proceed. With the standard RTP, the name entry uses English characters, allowing you to enter it, but 
with EasyRPG, it uses the Japanese name entry instead.

To reproduce:
- Load the attached save
- Enter the northwestern room
- Answer "Sure" at the dialogue prompt, then speak with Garry (the man) three times
- Wait for the objects in the room to start shaking, examine the newly uncovered canvas in the northeast part of the room, then examine the northwesternmost bookshelf
- Choose "Read how to enter"
- Examine the back of the leftmost bookcase in the southeast part of the room
- Stand between the bookcase and Garry, face west, then examine the gap
- Examine the canvas, and choose "Close eyes, count to three"
- Pull the cord that appears
- Examine the rightmost orb and choose "Enter a word"

rohkea:
Seems to work for me, please test.

Feedback is welcome. I’m particularly not happy with the forward declaration of ``struct Keyboard_Layout;``.

---
EDIT by @carstene1ns: Fixes #1272.

### DIFF
--- a/src/window_keyboard.cpp
+++ b/src/window_keyboard.cpp
@@ -23,25 +23,12 @@
 #include "bitmap.h"
 #include "font.h"
 
-const char* const Window_Keyboard::TO_SYMBOL = "Symbol";
-const char* const Window_Keyboard::TO_LETTER = "Letter";
-const char* const Window_Keyboard::DONE = "Done";
+const char* const Window_Keyboard::DONE = "<Done>";
 const char* const Window_Keyboard::SPACE = "SPACE";
-
-const char* const Window_Keyboard::TO_KATAKANA = "<カナ>";
-const char* const Window_Keyboard::TO_HIRAGANA = "<かな>";
+const char* const Window_Keyboard::NEXT_PAGE = "<Page>";
 const char* const Window_Keyboard::DONE_JP = "<決定>";
-
-const char* const Window_Keyboard::TO_CYRILLIC_RU = "<Абвг>";
-const char* const Window_Keyboard::TO_LATIN_RU = "<Abcd>";
 const char* const Window_Keyboard::DONE_RU = "<OK>";
-
-const char* const Window_Keyboard::TO_HANGUL_1 = "<앞P>";
-const char* const Window_Keyboard::TO_HANGUL_2 = "<뒤P>";
 const char* const Window_Keyboard::DONE_KO = "<결정>";
-
-const char* const Window_Keyboard::TO_ZH_CN_1 = "<翻页>";
-const char* const Window_Keyboard::TO_ZH_CN_2 = "<前页>";
 const char* const Window_Keyboard::DONE_ZH_CN = "<确定>";
 
 /*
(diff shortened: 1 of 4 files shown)


Ghabry:
Test case:

https://easyrpg.org/play/pr1284/?start-map-id=36&start-position=6%2011&new-game

To test: via &encoding= : 932, 949, 936, 1251, 1252

Ghabry:
When you use encoding 1252 and talk to the right-most event (which launches name scene on katakana page/ page 2) the continue button is not rendered.

rohkea:
Thanks for noticing! It should be OK now.

rohkea:
Thanks! I’ve changed int’s to size_t’s as suggested in the IRC, and reordered the initialisations.

carstene1ns:
btw. one thing that has annoyed me since ever (mostly noticeable in the symbol page):
When you press down and there is no symbol under it, it will jump one to the left, but only when there is one, otherwise it jumps to the top or skips the empty row.

Demonstration: Press <kbd>Down</kbd> here.

screenshot

Ghabry:
Which behaviour would you prefer? That it always jumps at ⇖ in the next row or that it goes downwards and jumps to the top?

carstene1ns:
IMO it should never jump left anyway.

carstene1ns:
Added a commit that fixes this behaviour.
It was actually needed to support moving down on the wide fields.
Achieving this differently now (*hack alert*).

src/window_keyboard.cpp
```diff
@@ -16,20 +16,21 @@
...
-const char* const Window_Keyboard::DONE_RU = "<OK>";
-const char* const Window_Keyboard::DONE_KO = "<결정>";
-const char* const Window_Keyboard::DONE_ZH_CN = "<确定>";
+const char* Window_Keyboard::DONE = "<Done>";
+const char* Window_Keyboard::SPACE = "SPACE";
+const char* Window_Keyboard::NEXT_PAGE = "<Page>";
+const char* Window_Keyboard::DONE_JP = "<決定>";
+const char* Window_Keyboard::DONE_RU = "<OK>";
+const char* Window_Keyboard::DONE_KO = "<결정>";
+const char* Window_Keyboard::DONE_ZH_CN = "<确定>";
```
Ghabry: How about "constexpr char Window_Keyboard::DONE[] =" ;)?

src/window_keyboard.cpp
```diff
@@ -237,42 +238,37 @@ void Window_Keyboard::Refresh() {
...
-			}
+			col = (col + col_max - 1) % col_max;
 		}
+	}
 
+	// Special handling for wide fields
+	if (col > 0 && col < col_max) {
+		std::string k = GetKey(row, col - 1);
+		if (k[0] == '<' || k == SPACE)
+			col = std::min(col + skip_dir, col_max - 1);
```
Ghabry: Is it possible check for SYMBOL and DONE based on the row-col position instead?
Imho checking for "<" is a bit ugly.

carstene1ns: Yeah, my intention was here to not have to compare with all the different DONEs. But you are right, as long as they stay on their position your idea is good.

src/window_keyboard.cpp
carstene1ns: This does not work. Either I get `-Wwriteable-strings` warnings or the build errors out with
`error: non-const static data member must be initialized out of line` or
`error: declaration of constexpr static data member 'SPACE' requires an initializer` or
`error: redefinition of 'SPACE' with a different type: 'char *const' vs 'const char *'`, depending on how i modif/move them. 🤕

Ghabry: o.O okay leave it as is.

Issue #1268, opened 2017-09-21
Airship shadow already moves independend of Airship when a moveroute is active while boarding


Ghabry:
Basicly the airship shadow moves synchronized with the (invisible) Player character while the airship still plays the ascend animation. Afterwards the airship jumps directly on the Player tile.

No idea why the Player is allowed to move while the animation plays.

Issue #1198, opened 2017-06-05
Razas: Incorrect movement frame "left" used instead of "center"


Ghabry:
Not fixed by my fix for #1112

#### Name of the game: Razas
You can buy it here for 99ct: https://razas.itch.io/razas

The dev gave us 5 keys a while ago, I'm not sure if all are used yet. I can give you one.

Not tested if the scene is in the public demo

#### Describe the issue in detail and how to reproduce it:

Start the game with command line "--start-map-id 70 --start-position 20 0"

Just fast-forward enter-smash through the cutscene until the right hero changes from his "star" animation to a standing animation.

In RPG_RT the foot position is centred, in EasyRPG the right food is to the front (not sure if this is left or right stepping through, at least not centered)

Ghabry:
Razas has one event page set to "Step left" but because the Animation type is "standing animation" the step is reset to Middle in RPG_RT but not in Player.

PR #1288, opened 2017-10-18
Various fixes (#1268, #1198, two undocumented GetActors uses, Bitmap::Create(w, h, col). Split from other PR)

20kdc:
This includes a fix for #1268, a fix for #1198, and adds support for two undocumented(?) GetActors uses.
This was split from #1287, so the testcase for this (which proves said GetActors calls exist) is there.

The future suggestion-aka-nitpick referring to this part is:
"`IsStopping` in Player ought to be something like `IsReadyToMove`?"

### DIFF
--- a/src/game_player.cpp
+++ b/src/game_player.cpp
@@ -673,6 +673,13 @@ bool Game_Player::GetOffVehicle() {
 	return true;
 }
 
+bool Game_Player::IsStopping() const {
+	// Prevent MoveRoute execution while the airship is ascending/descending (Issue #1268)
+	if (InAirship() && !GetVehicle()->IsMovable())
+		return false;
+	return Game_Character::IsStopping();
+}
+
 bool Game_Player::IsMovable() const {
 	if (IsMoving() || IsJumping())
 		return false;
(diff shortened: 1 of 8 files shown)


Ghabry:
68fe9c3 becomes easier to review by [removing the whitespace](https://github.com/EasyRPG/Player/pull/1288/commits/68fe9c3176e52e829e841aa3ae80982d99d55e75?w=1) and an interesting find. @CherryDT this could be exported via the Editor GUI ^^.

Looks good to me. Because we still have some time until 0.5.4 this PR is a good oportunity to talk about renaming IsMoving and IsStopping as 20kdc already proposed. Just thinking about good names :)

carstene1ns:
Jenkins: ok to test

btw. Jenkins spam disabled successfully 🥇

src/game_player.cpp
```diff
@@ -673,6 +673,13 @@ bool Game_Player::GetOffVehicle() {
 	return true;
 }
 
+bool Game_Player::IsStopping() const {
+	// Issue #1268 (player moving too early on a moveroute when using airship) was because MoveRoutes use this.
```
Ghabry: How about "// Prevent MoveRoute execution while the airship is ascending"

Ghabry:
Jenkins: Test this please

src/game_player.cpp
20kdc: *nods* Will add that now.

20kdc:
Whoops, forgot to rebase to latest while dealing with the comment. Will do that now.

PR #1290, opened 2017-10-22
EasyRPG Player 0.5.3

carstene1ns:


Ghabry:
Last chance to ensure that 0.5.3 is ready. When we need to retag we'll make OpenBSD-senpai angry again ^^'.

carstene1ns:
Well, if anything happens, we just push 0.5.4 in a week or so (is easy, because we can just rename the milestone instead of moving all issues around again^^')

PR #1297, opened 2017-10-28
Fix building with sdl2_mixer 2.0.2

carstene1ns:
Quick hack, just to get it building again.

### DIFF
--- a/src/audio_sdl_mixer.cpp
+++ b/src/audio_sdl_mixer.cpp
@@ -242,9 +242,9 @@ void SdlMixerAudio::BGM_OnPlayedOnce() {
 
 	if (!bgm_stop) {
 		played_once = true;
+
 		// Play indefinitely without fade-in
-		if (Mix_GetMusicType(bgm.get()) != MUS_MP3_MAD)
-			Mix_PlayMusic(bgm.get(), -1);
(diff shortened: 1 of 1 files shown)


src/audio_sdl_mixer.cpp
```diff
@@ -242,8 +242,12 @@ void SdlMixerAudio::BGM_OnPlayedOnce() {
 
 	if (!bgm_stop) {
 		played_once = true;
-		// Play indefinitely without fade-in
+
+// introduced in SDL2_mixer 2.0.2, which deprecated libmad
+#ifndef SDL_MIXER_COMPILEDVERSION
 		if (Mix_GetMusicType(bgm.get()) != MUS_MP3_MAD)
+#endif
+			// Play indefinitely without fade-in
```
fdelapena: Multiline if without brackets.

Ghabry: Imo the whole MUS_MP3_MAD if-check can be removed because for the next cleanup all the other ugly hacks for <= 2.0.1 will be removed, too.

carstene1ns: I am fine with this. 👍

carstene1ns:
done.

PR #1257, opened 2017-09-08
Implement obscure SetVehicle behaviour (and 2 minor fixes)

Ghabry:
See the comments in the sourcecode.

The "Panorama" problem was observed in the game reported by Penguin "Oech" Dude directly at the beginning.

And the MessageOverlay problem was observed while testing the SetVehicle warnings I added :D.

PR #1303, opened 2017-10-30
Implement even more obscure SetVehicleLocation behaviour

Ghabry:
SetVehicleLocation actually supports changing the map without doing a transition (not as thought in #1257) while the party is boarded but only when the target map has an event with the same event ID.

Because this sounds more like a bug than intended behaviour this was implemented using a normal teleport and without a check if the target map has a valid event. Doesn't match the RPG_RT behaviour due to the transition but this bug is strange and only used by a few VIPRPG games, not worth implementing perfectly.

The patch is good enough to fix the test case. I hope not even worse offenders appear. I reordered the code a bit so it matches other teleport code. Moved the vehicle reposition code after the "return false" in the if-block to ensure that the vehicle isn't moved before a teleport can start.

Test case: http://velvel.net/Cynthia/List/Search/%E5%91%AA%E3%81%84%E3%81%AE%E3%81%93%E3%81%86%E3%82%82%E3%82%8A%E3%81%95%E3%82%93/true/1

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -1612,19 +1612,40 @@ bool Game_Interpreter::CommandSetVehicleLocation(RPG::EventCommand const& com) {
 	int x = ValueOrVariable(com.parameters[1], com.parameters[3]);
 	int y = ValueOrVariable(com.parameters[1], com.parameters[4]);
 
-	// Check if the party is in the current vehicle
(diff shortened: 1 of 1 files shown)


Ghabry:
Hopefully not because the "teleport" case I added usually crashes the engine :D.

carstene1ns:
allrighty then.

PR #1307, opened 2017-11-07
Traditional Chinese support for name input scene

fdelapena:
Thanks to @kane159 for the glyph table 👍


carstene1ns:
jenkins: test this please

PR #1318, opened 2017-12-13
Fixes for GCW0 port

gameblabla:
This fixes the GCW0 port, again.
I switched back to using SDL 1.2 because it runs faster on the GCW0 than the SDL2 version.
(etnaviv, the 3D driver, still sucks on it and will probably stay that way. SDL2 version works fine tho)
Let me know if that's an issue.

I also added some headers to some files (i think they got removed ? not sure) because these are required for compiling. 

Switching to using -Wl,--no-as-needed because it does not compile with -as-needed.

I tried it on my GCW0 and it works okay. It seems smoother too.

Btw, it also requires this pull request too : [here](https://github.com/EasyRPG/liblcf/pull/220).

### DIFF
--- a/builds/opendingux/gcw-zero/readme.txt
+++ b/builds/opendingux/gcw-zero/readme.txt
@@ -4,8 +4,8 @@ Installation Instructions
 RTP
 ---
 * Install the RTP files on your computer
-* RPG2K games: Copy the RTP files, not the RTP folder iteself, to /media/data/local/share/easyrpg/rtp2k
-* RPG2K3 games: Copy the RTP files, not the RTP folder itself, to /media/data/local/share/easyrpg/rtp2k3
+* RPG2K games: Copy the RTP files, not the RTP folder iteself, to /usr/local/home/.easyrpg/rtp2k
+* RPG2K3 games: Copy the RTP files, not the RTP folder itself, to /usr/local/home/.easyrpg/rtp2k3
 
 Games
 -----
--- a/src/system.h
+++ b/src/system.h
@@ -58,6 +58,10 @@
 #  define WORDS_BIGENDIAN
 #endif
 
+#ifdef OPENDINGUX
+#  include <sys/types.h>
+#endif
+
 #define SUPPORT_BMP
 #define SUPPORT_PNG
 #define SUPPORT_XYZ
--- a/builds/opendingux/Makefile
+++ b/builds/opendingux/Makefile
@@ -34,9 +34,9 @@ ifeq ($(BUILD_TARGET), gcwzero)
 	TOOLCHAINDIR ?= /opt/gcw0-toolchain
 	SYSROOT = $(TOOLCHAINDIR)/usr/mipsel-gcw0-linux-uclibc/sysroot
 	BINDIR = $(TOOLCHAINDIR)/usr/bin
-	CFLAGS = -Ofast -fomit-frame-pointer -fdata-sections -ffunction-sections -mips32r2 -mmxu -std=gnu++11 -DLCF_SUPPORT_ICU 
+	CFLAGS = -Ofast -fomit-frame-pointer -fdata-sections -ffunction-sections -mips32r2 -std=gnu++11 -DLCF_SUPPORT_ICU 
 	CFLAGS += -DGCW_ZERO -DWANT_FMMIDI=2 -DHAVE_MPG123 -DHAVE_LIBSNDFILE -DWANT_FASTWAV
-	LDFLAGS = -Wl,--as-needed,--gc-sections -flto
+	LDFLAGS = -Wl,--no-as-needed,--gc-sections -flto
 	PKG_CONFIG := $(BINDIR)/pkg-config
 	PNG_LIBS := $(shell $(PKG_CONFIG) --libs libpng)
 	PIXMAN_CFLAGS := $(shell $(PKG_CONFIG) --static --cflags pixman-1)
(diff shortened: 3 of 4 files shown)


Ghabry:
Jenkins: Test this please

src/filefinder.h
```diff
@@ -21,6 +21,8 @@
 // Headers
 #include "system.h"
 
+#include <stdio.h>
```
Ghabry: Is this really needed? 3 lines below cstdio is included.

Ghabry:
Thx for the PR. Please use \<cstdio\> instead of \<stdio.h\> because this is C++ code (just for style, has no relevant semantic change)

I'm a bit confused that adding sys/types.h didn't break the Windows build. Thought this is POSIX only and not part of the C standard o.O.

gameblabla:
My bad, i did a mistake. :< I fixed the issues.

Ghabry:
Well we have a similiar ugly include in system.h for GEKKO, so should be fine this way imo.

Ghabry:
Jenkins: Test this please

gameblabla:
Had to add NDEBUG to the Makefile due to [liblcf#220](https://github.com/EasyRPG/liblcf/pull/220). I think there are no more issues.

carstene1ns:
Hoi, thanks for the pull request!
I am able to build without the added headers myself, however, I have built the toolchain myself (from git), so might be because it is more recent. Adding them is not wrong.

Can you please test this version: [EasyRPG-PR-SDL2.zip](https://github.com/EasyRPG/Player/files/1557427/EasyRPG-PR-SDL2.zip) It has the other parts of your pull request applied, I would like to know if it makes a difference with our new libs, but still using SDL2.

Reason is, SDL1.2 is not really supported anymore and we will likely not find/fix bugs with it in time, as we do not test builds regularly.

builds/opendingux/gcw-zero/readme.txt
```diff
@@ -4,8 +4,8 @@ Installation Instructions
 RTP
 ---
 * Install the RTP files on your computer
-* RPG2K games: Copy the RTP files, not the RTP folder iteself, to /media/data/local/share/easyrpg/rtp2k
-* RPG2K3 games: Copy the RTP files, not the RTP folder itself, to /media/data/local/share/easyrpg/rtp2k3
+* RPG2K games: Copy the RTP files, not the RTP folder iteself, to /usr/local/home/.easyrpg/rtp2k
+* RPG2K3 games: Copy the RTP files, not the RTP folder itself, to /usr/local/home/.easyrpg/rtp2k3
```
carstene1ns: Maybe we should simply mention "home directory" here. Opinions?

src/system.h
```diff
@@ -58,6 +58,10 @@
 #  define WORDS_BIGENDIAN
 #endif
 
+#ifdef OPENDINGUX
+#	include <sys/types.h>
```
carstene1ns: We use two spaces for indentation instead.

Ghabry:
Maybe it would be also interesting to test with "SDL_CreateSoftwareRenderer".

builds/opendingux/gcw-zero/readme.txt
fdelapena: If this one is the only valid folder (or at least an always valid one), it prevents unnecessary user questions because they didn't understand the documentation properly IMHO.

carstene1ns: @fdelapena: I suggest looking here, it might answer your question: https://github.com/EasyRPG/Player/blob/master/builds/opendingux/gcw-zero/run.sh#L16-L19

gameblabla: Most GCW0 users are not very fluent with Linux and don't know what i'm talking about when i say Home folder.
I think it's just less trouble to just post the full path instead.

gameblabla:
@carstene1ns I tried your build and it does not work on my GCW0. (Both from the OPK file and the terminal)
In the terminal, it says this
./EasyRPG: can't load library 'S1_SaIS1_EER9LcfWriter'
Unless built statically, i don't think binaries built with your toolchain will work.

I don't feel like using SDL2 on the GCW0 because the drop in speed is noticeable compared to SDL1.2, like 5~10 FPS. (just tried the SDL2 version right now)
Maybe it would be wise to separate the two SDL versions ? Might be better than using ifdefs everywhere.
This could change in the future when the new firmware comes out though.

Ghabry:
semi-ot: The GCW0 still gets new firmware? I thought the device is dead.

Ghabry:
Jenkins: Test this please

carstene1ns:
>  ./EasyRPG: can't load library 'S1_SaIS1_EER9LcfWriter'

liblcf is build as part of Player. this only makes sense, if garbage collection or something removes it.

gameblabla:
@Ghabry Well yeah, it's been in development since forever... And no more GCW0s are made. 
You can draw your own conclusions.

I'll try to provide some OPK files [https://github.com/gameblabla/Player/releases](here). Too lazy as f to make it work with Jenkins. (because official binaries are only 32-bits :/)

carstene1ns:
Where does this development take place? At least the buildroot repository has no commits since years.

gameblabla:
Development does not happen in buildroot but in linux.
[https://github.com/gcwnow/linux/tree/jz-4.9-wip]()
He made a summary of his progress here :
[https://github.com/gcwnow/linux/wiki/Upstream-status]()

He has not worked on it since early October and pcercui is pretty much the only guy working on this.
It's still missing a lot of things so RIP.

carstene1ns:
Here is another build that should have fixed the library issue: [EasyRPG.opk](https://ci.easyrpg.org/view/Pull%20Requests/job/player-gcw0-pr/17/artifact/builds/opendingux/gcw-zero/EasyRPG.opk)
Can you please test this again? Thank you!

gameblabla:
@carstene1ns Your build works now. It's just as slow as my SDL2 version.

carstene1ns:
Yay! Then my toolchain works now, so only issue left is using SDL1.2 and then the port can bitrot again… 😀

gameblabla:
Ok so, after thinking about it for a while, i edited the Makefile to just use SDL2 instead.
I'll just provide my own builds for SDL 1.2. I don't think you'll get strange bug reports anyway...
I also removed the mmxu flag because EasyRPG does not use the MXU instructions and it could cause issues with newer GCC versions.

carstene1ns:
jenkins: ok to test.

PR #1321, opened 2017-12-22
Small fixes for the emscripten port

carstene1ns:
The braces change is only to support @rohkea's game... 😀

It seems, that currently the emscripten port is broken, as it fails inside SDL2 for me.
Note: buildscript needs now: ` -s 'EXPORTED_RUNTIME_METHODS=["Pointer_stringify"]' ` for SDL2.

### DIFF
--- a/src/async_handler.cpp
+++ b/src/async_handler.cpp
@@ -176,9 +176,11 @@ void FileRequestAsync::Start() {
 		request_path += path;
 	}
 
-	// URL encode % and #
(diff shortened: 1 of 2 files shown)


carstene1ns:
Closing this for now.
Both changes are not actually needed (currently).

- The square braces are correctly interpreted by apache and other web servers (i.e. `python3 -mhttp.server`), so no need.
- The `EM_ASM` double braces are needed as of `1.37.25`, while the current buildscript uses `1.37.0`.

PR #1326, opened 2018-01-01
Change types from int to int32_t where necessary for liblcf compat

Ghabry:
Required for https://github.com/EasyRPG/liblcf/pull/222

Only required on Wii, Vita and 3DS.

Didn't break Color, tone change still works.

### DIFF
--- a/src/game_party.cpp
+++ b/src/game_party.cpp
@@ -102,12 +102,12 @@ int Game_Party::GetItemCount(int item_id, bool get_equipped) {
 
 void Game_Party::GainGold(int n) {
 	data.gold = data.gold + n;
-	data.gold = std::min(std::max(data.gold, 0), 999999);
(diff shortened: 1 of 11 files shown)


src/game_party.cpp
```diff
@@ -108,12 +108,12 @@ int Game_Party::GetItemCount(int item_id, bool get_equipped) {
 
 void Game_Party::GainGold(int n) {
 	data.gold = data.gold + n;
-	data.gold = std::min(std::max(data.gold, 0), 999999);
+	data.gold = std::min(std::max(data.gold, (int32_t)0), (int32_t)999999);
```
carstene1ns: These casts look ugly. `min/max` can use the common type of the arguments, but needs explicit type then, i.e.
`max<int32_t>(some, value);` Maybe is better.

Ghabry: Thanks, changed this.

PR #1331, opened 2018-02-11
Fix Frozen triggers

Ghabry:
https://my.mixtape.moe/gswfhe.webm


src/game_picture.cpp
```diff
@@ -146,6 +146,7 @@ void Game_Picture::Show(const ShowParams& params) {
 	data.flags.affected_by_flash = (params.flags & 32) == 32;
 	data.flags.affected_by_shake = (params.flags & 64) == 64;
 	last_spritesheet_frame = 0;
+	sheet_bitmap.reset();
```
carstene1ns: good find!

PR #1334, opened 2018-02-13
Improve portability for "dirent->d_type"

Ghabry:
Checks now for "_DIRENT_HAVE_D_TYPE" (defined by the dirent_win.h header, checked this) which is the documented way to test for this extension.

DT_UNKNOWN is handled now correctly for filesystem that don't provide the file type without stat'ing.

### DIFF
--- a/src/filefinder.cpp
+++ b/src/filefinder.cpp
@@ -770,8 +770,6 @@ FileFinder::Directory FileFinder::GetDirectoryMembers(const std::string& path, F
 		return result;
 	}
 
-	static bool has_fast_dir_stat = true;
-
 #ifdef PSP2
 	struct dirent ent;
 	while (readdir(dir, &ent) > 0) {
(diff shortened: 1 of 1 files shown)


src/filefinder.cpp
```diff
@@ -788,24 +786,28 @@ FileFinder::Directory FileFinder::GetDirectoryMembers(const std::string& path, F
...
-		bool is_directory;
+
+		static bool has_fast_dir_stat = true;
+		bool is_directory = false;
 		if (has_fast_dir_stat) {
 			#ifdef PSP2
 			is_directory = S_ISDIR(ent.d_stat.st_mode);
+			#elif defined(_DIRENT_HAVE_D_TYPE)
+			if (ent->d_type == DT_UNKNOWN) {
+				has_fast_dir_stat = false;
```
Ghabry: I could move the deleted output::debug lines here but I question there usefulness

carstene1ns: At least they are an indicator when porting to a new platform... but could be left out, right.

PR #1350, opened 2018-04-28
Battle 2k: Always allow escape escaping from random encounters

Ghabry:
Finally something simple to review ;)

Due to an oversight GetAllowEscape was used to check if random encouter escapes are allowed but this flag is only for escape spells, Escaping is always possible in random encouters >.>

Iirc this was reported via mail as a bug in Vampires Dawn, but can't find it anymore.


carstene1ns:
Also mentioned on [gbatemp](https://gbatemp.net/threads/release-easyrpg-player-rpg-maker-2k-2k3-emulator-for-switch.497328/page-3#post-7906584).

Ghabry:
Thanks! Then this was not via mail but via gbatemp.

Issue #1349, opened 2018-04-15
Witch's Heart: When throwing the broom it always lands at the top-left corner


Ghabry:
#### Name of the game: Witch Heart

http://vgperson.com/games/witchheart.htm

Kimik0:
More details:
The magic broom is an item obtained at the start of the fourth scenario. It is needed to complete that scenario, so if it doesn't work, the game can't be played beyond that point.
Flying around on the broom works, but when landing, the player is always teleported to the top-left corner of the map, whether there is anything there or not. If it is part of the room, you can continue playing. If it isn't, the only action that does anything at that point is activating the broom again, but as soon as you try to land..
In other words, activating the broom is an effective way to get stuck on maps where the top-left corner isn't part of the room, and in any case the broom can't be used in the intended way (to travel to otherwise unreachable areas).

Since this game was translated only recently, it's pretty popular at the moment (as GameMaker 2k3 RPGs go). It'd be really cool if you could look into this issue soonish ^_^

Ghabry:
Thanks for that information! I became aware of Witch's Heart due to multiple Android bug reports. Usually we only try to fix the bug but in case of Witch's Heart I liked the game so much that I started playing it instead ;) (and wanted to fix it as soon as I encounter the bug).

By now I only played Scenario 1 and started with 2, this explains why I havn't encountered it yet :)

Kimik0:
You can use the save file I attached if you just want to check on the bug (I forgot to equip the broom, but it's in the Esc menu). I guess it would be a little spoilerish if you're playing the game yourself though ^_^;

In case it helps to know, I tried running EasyRPG with all the different engine types and that didn't change anything.

vgperson:
I got reports of this a few times (though at first I was like "I dunno, probably can't do anything about that on my end"), so I looked into this myself.

The way the Magic Broom works is that it's really the Airship, which via Common Event 64 gets moved to the player's position when you press the button under the right conditions, so you can then board it.

This part works fine in EasyRPG, but the problem is in the "descending" part. Common 65 checks if the "riding broom" switch is on (Common 66 sets it when "riding Airship" is true), but the "riding Airship" condition is now false. In other words, checking if "you've fully gotten off the Airship." When the condition is met, it resets the Airship's position to 1,1 (an unreachable position) to keep it from being boarded until 64 puts it under the player again.

[...]

Issue #605, opened 2015-09-28
Allow to load EXFONT from file or RPG_RT.exe resource


fdelapena:
EXFONT is an embedded bitmap resource containing symbols displayed with $a-$z and $A-$Z. Some games have this asset customized by patching the RPG_RT.exe file.

Because EasyRPG Player executable does not contain a windows resource based asset easy to change and needs to rebuild the source, several users asked for an alternate way to customize these icons.

carstene1ns:
In a [forum thread](https://easy-rpg.org/forums/viewtopic.php?t=593) it was suggested to read the `RPG_RT.exe` to check for applied patches. This is a major task, but it would make exfont detection and use simple, as we could strip the exfont from `Player` and just use whatever is present in `RPG_RT.exe`.
What I really do not like about this approach is that we need to rely on that binary too much, so it would be great to allow enabling everything the "exereader" does without having the `RPG_RT.exe` at all.
This comes to mind: bundling a rpg2k game with free assets together with the `Player` executable.
Some compatible patches could be enabled in `RPG_RT.ini`, for example the simple ones like disabling/skipping Title scene…

That said, reading the `EXFONT.bmp` from file if present is a good thing!

I am going to add an extractor script to Tools repository soon.

Ghabry:
For patches (for old games at least) I was thinking about some type of database that was distributed with the Player and installed into the program directory. Which contains patches for all games and is regularly updated. To make this configuration-free for the user.

carstene1ns:
This would need some way to differentiate between different versions of games, i.e. compare the checksum of the `RPG_RT.exe`.
This would have the additional benefit that all games using the same patches and no custom resources would work ootb.
However, if  we would like the Player to be a drop-in replacement for RPG_RT, adding them to ini is still an option, as you do not need to distribute anything in some shared directory.

Ghabry:
Yeah the database and the local (in game directory) option format should be the same (and game config wins over db config).

I would use a super simple heuristic for now: Take the game name from the RPG_RT.ini file. And then wait until people complain about strange bugs :P

carstene1ns:
Heh, good thing is we support a lot of patches without modifying anything yet. (MP3 and other music support?)
And there are some patches that have the same effect, but are implemented differently.
We can support them in one go. :grinning:.

fdelapena:
> (MP3 and other music support?)

[...]

PR #1352, opened 2018-05-05
Allow loading of custom ExFont file if available in the game directory

Ghabry:
Sorry for using FindImage(".", "ExFont") instead of adding a new FindImage("ExFont") and that strange check in FileFinder for the root directory, but I'm currently heavily modifying filefinder.cpp and don't want to reapply gigantic changes ^^.

Related: #1349


carstene1ns:
Also related: #605

PR #1357, opened 2018-05-06
Fix BGM Volume after fadeout and add SE cancelling

Ghabry:


carstene1ns:
Does this cancelling of fade effects also work with SDL_mixer / OpenAL?

Ghabry:
SDL_Mixer removes the song completely after fadeout, which means SetVolume just won't work.
Need to check how RPG_RT behaves, if it restarts the song or continues.

src/scene_gamebrowser.cpp
```diff
@@ -50,7 +50,6 @@ void Scene_GameBrowser::Continue() {
 #ifdef _WIN32
 	SetCurrentDirectory(L"..");
 #endif
-	Audio().BGM_Fade(800);
```
Ghabry: Removed this line because BGM_Stop is called few lines later.

Ghabry:
Havn't tested the unloved OpenAL but it works now in the generic decoder and SDL2 Mixer.
Should work everywhere because it simply uses BGM_Play now.

PR #1365, opened 2018-05-15
Update CMake scripts

Ghabry:
Requires the liblcf cmake PR to work.

This updates all the CMake stuff to use targets.

Works for Windows, Linux and MacOSX, including static builds.

Under Windows I have a vcpkg issue: The opus vcxproj file compiles a sample c++ file in which means the main is overwritten when linking static m(. Must be fixed in vcpkg repo.

Simplified the code a bit by making Wildmidi always ON, doesn't matter if SDL2 or SDL2_Mixer is used. Still no love for OpenAL yet ^^'.

Added a new magical function player_find_package to reduce code duplication.


Ghabry:
sorry, no OSX for now... the server migration removed the VMX flag >.>, contacted support.

CMakeLists.txt
```diff
@@ -1,14 +1,9 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.7)
 
-set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_CXX_STANDARD_REQUIRED on)
-include(CMakeDependentOption)
-
-project(EasyRPG_Player CXX C)
-set(PACKAGE_VERSION "0.5.3")
+project(EasyRPG_Player VERSION 0.5.3 LANGUAGES CXX C)
```
carstene1ns: Should be save to remove `C` here.

CMakeLists.txt
```diff
@@ -1,14 +1,9 @@
...
 
 # Source Files
-set(PLAYER_SRCS
+add_library(${PROJECT_NAME}_Static
```
carstene1ns: Does cmake still not support having different kinds of things (executable, library) with the same name? :-1: Maybe change the name to `player_aux` or something.
Also, this might be a shared library without explicit setting `STATIC`, no?

CMakeLists.txt
```diff
@@ -579,19 +547,19 @@ if(${PLAYER_AUDIO_BACKEND} MATCHES "^SDL2.*$")
...
+	if(SPEEXDSP_FOUND)
+		message(STATUS "Resampler:	 speexdsp")
 	elseif(SDL2_mixer_FOUND)
 		set(SDL_MIXER_USED ON)
-		message(STATUS "Resampler:     SDL2")
+		message(STATUS "Resampler:	 SDL2")
 	else()
-		message(STATUS "Resampler:     None")
+		message(STATUS "Resampler:	 None")
 	endif()
```
carstene1ns: Maybe align everything with spaces (like the rest)?

Ghabry: This was all really well aligned until fdela came and changed the Ogg playback string to Ogg vorbis playback :P.
Already played around with some alignments (and no alignments). The problem with no spacing at all is that it's harder to read IMHO.

CMakeLists.txt
Ghabry: is a static library by default because BUILD_SHARED_LIBS is OFF by default (strange default). But you are right, must force STATIC here.

Is not possible to have two targets with the same name but you can cheat:
https://gist.github.com/jlgerber/eafc4ee2b9954e27dd2bb009496b1b03#file-cmakelists-txt-L13
Will do this to get rid of _Static.

carstene1ns: nice, you are already using half of the cheat currently :+1:

Ghabry:
good news: Even without vcpkg we can already switch to CMake completely.
I just need to rename pixman to pixman-1, then it works with my hand-rolled buildsystem.
Well and fix some linker errors ;).

ras0219-msft:
The opus issue should be fixed in vcpkg master!

Ghabry:
@ras0219-msft 
yepp this is what I planned to do :+1:

carstene1ns:
Good, waiting for the (hopefully final) rebase.

Ghabry:
oooh, I understood this wrong. The opus issue was already solved in the vcpkg package repo. Awesome :D

CMakeLists.txt
carstene1ns: ping

CMakeLists.txt
carstene1ns: pong

Ghabry:
Okay, that's enough (annoying) CMake work for now.

Issue #1125, opened 2017-03-15
Pictures: Update Order and Positions


scurest:
1. In RPG_RT, pictures are updated before events.

    You can check this with an event like

        <> Show Picture: ..., Zoom: 100%
        <> Move Picture: ..., Zoom: 75%, Wait 0.0s

    For one frame (the same frame these commands run) the picture will be shown at 100% zoom. (You can also check this in Ghabry's event tracer by printing out the picture's member variables during event processing and checking that they have already updated for the frame.)

    In EasyRPG, events are updated before pictures. This can be solved by transposing [these lines](https://github.com/EasyRPG/Player/blob/afc98cc7da239eea2c3f72de2ceb96bbfa132e95/src/scene_map.cpp#L139-L140). I'm reluctant to do it though because

    1. It also places all other screen effects' updates before the events', and I don't know if that's right. Could be solved pretty easily by breaking picture updating specifically out of updating `Game_Screen`.
    2. Part of picture updating also looks at the current map position and uses it to move pictures that are "fixed to the map". So if the screen moves during event processing, currently since the pictures are updated later, they will observe the change, but if they happen before, they won't. I can't seem to actually trigger any graphical bug with this though (maybe the screen never moves far enough in one frame?), so I don't know if this is just paranoia on my part >_>

2. For pictures that are fixed to the map on a looping map, in EasyRPG they currently jump when you cross the map edges. Compare [this issue with battle animations](https://github.com/EasyRPG/Player/pull/574#issuecomment-138691723).  If you think of a wrapping map as a plane consisting of repeating copies of the map, the picture should be fixed to the _overall plane_. This means for example that if you show a picture and cross the right edge of the map, you won't see the picture on this "new copy" of the map. Once you cross back over the left-edge, you'll see it again.

    The underlying reason for the bug (and the battle animation one) is that we don't track a "global position" in this plane, only a "local" position within the map itself.

-----

Here's an untested idea to fix both 1.ii and 2: split updating into two phases.

1. The first phase updates pictures' zoom/transparency/effects etc. but not their position.
2. Events are updated. We keep track of how much the map scrolls in each direction.
3. The second phase looks at the scroll amounts and use them to update pictures/parallax/battle animation positions.

This would also let us get rid of the `old_map_{x,y}` fields in our `Game_Picture` class that don't have an analogue in `RPG::SavePicture`.

Of course the best idea is to figure out whatever RPG_RT does and do that.

fmatthew5876:
[...]

PR #1369, opened 2018-05-30
 Turn Transition to a Drawable; Implement missing transitions (RandomBlockUp/Down, WaveIn/Out, ZoomIn/Out, MosaicIn/Out)

### LINKED ISSUE #900: Some transition effects not implemented
- [x] TransitionRandomBlocks
- [x] TransitionRandomBlocksUp
- [x] TransitionRandomBlocksDown
- [x] TransitionZoomIn
- [x] TransitionZoomOut
- [x] TransitionMosaicIn
- [x] TransitionMosaicOut
- [x] TransitionWaveIn
- [x] TransitionWaveOut

### LINKED ISSUE #1094: Transition code is not a drawable and doesn't support frameskips
Noticable on slow systems like the 3DS where transitions take significantly longer then elsewhere, due to the way they are implemented.
They are a hardcoded part of Graphics::Draw instead of being a Drawable.

Albeleon:
Implemented Transitions (RandomBlockUp/Down, WaverIn/Out, ZoomIn/Out, MosaicIn/Out)

Implementation of the transitions left:

-RandomBlockUp/Down: This is made like Random, defining the order of blocks when the transition is initiated, and then Update the same way, so they share that code. The inicialization however is more complex to create an effect both incremental and random:

       · First we need to reorder our array incrementally (index 0 is 0, index 1 is 1, etc.), so we have an order established. In RandomBlocksUp we reverse the order.
 
      · The progress in the array is marked by a new vertical line to print, and it prints point up until 10 vertical lines later. So, the best way to print with this semi-random incremental order is:
                1 > We shuffle the array with the first 10 lines (600 in this case).
                2 > The first 80 (number of block per vertical line) will be fixed.
                3 > We shuffle the array from the position after the first 80, and we shuffle another 10 position lines.
                4 > Continue until there are no next 10 lines to print. This way, we've given a priority to the lines from one direction and the ones from the other direction won't have the opportunity to be printed until later.

       · This effect by itself replicates fine the effect, but there are some weird details that need a little more code to solve:
                - When it starts, the first 10 lines have the same probability of having random pixels, so the first 3 or so lines don't have more than them. To solve it, during the first iterations the number of lines where it is allowed to be printed is not the next 10, but 1, 3, 5, etc. That way it starts smoother.
                - When it ends, some blocks left can still be hanging on to the end in the beginning of the transition, which makes it jarring when they suddenly disappear. To reduce those ones, we use right after the shuffle a partial_sort function to put in each iteration a couple of the lowest blocks first in the order to print, so progressively the ones at the bottom get eliminated enough to mitigate the jarring effect. There are some times it doesn't print the lowest of all of them or it skips some test to not feel like there's an automatic drawer working behind.

       · The complexity of the Transition inicialization is (w = width, h = height, b = block size (4)): for [h] * (shuffle [w * h / b^2] + partial_sort [[w * h / b^2] * (log(cte) ~ 1)]) => 2 * w * h^2 / b^4 (with our data => 144.000) . The Transition Update is: for [w * h / b^2] (with our data => 4800).

       · POSSIBLE ALTERNATIVE TO RandomBlock Update: Instead of printing in each Update the number of blocks, we can use a Bitmap Mask where each frame we insert the corresponding blocks, and then mask Screen1 with this Mask. We would only use a few blits to insert new blocks in the Mask, and a BlitMask, instead of doing O(w * h / b^2) blits.

-ZoomIn/Out: This transition zooms in and out of the hero's position, and needs some adjustments to properly reflect the zoom of the RPG Maker:

       · First, we create a zoom_position vector parameter that is set when the transition is initialized. If the current scene is a map (there is a hero character), it returns their screen position. If it's not, it returns the center of the screen (160, 120).

       · In the update, we use StretchBlit using the percentage as the progress of the position and the extension of the zoom.

       · Because ZoomIn and Out are similar but with the percentage in reverse and a different screen pointer, to save code we make them one and the same, and when it's ZoomOut we reverse the percentage and change the screen pointer from one to the other.

       · This zoom is not enough to replicate the effect. The way the previous code would work, it would zoom on the character, but its relative position on the screen remains the same (if the character is 90% right of the screen, even with zoom it will remain there). In RM2000, a character too far off the center will have a zoom that will be next to the closest border until the character is closed enough (for our cases, when it's in between 25% and 50% of the screen), and then it continues the normal zoom. So all those variables are to take in consideration this behaviour.

       · Related to the one above, X and Y act independently this time, and therefore to simplify code and variables we can use vectors to reflect their positions, and use a for to make the same operations for both.

       · The complexity is one StretchBlit per Update.

- MosaicIn/Out: This transition blurs the screen. It is blurred incrementally (the size of each block/mosaic grows linearly according to the percentage, starting from size 1 and ending in size 40). It doesn't need initialization, only Update:

       · We cannot just scale down, and then up the screen to have the blurred effect because:
                1 > When the width and height size isn't divisible by the current mosaic size ( w|h % size != 0 ), because it's an integer, pixels are lost, so when it's scaled up again it doesn't cover the whole screen.
                2 > When the mosaic size starts getting higher and higher, given the low number of width and height (320x240), some adjacent sizes give the same result when you divide the parameters with them (320 / 31 = 10 + m; 320 / 32 = 10), so the screen will not change the colors of each pixel and it will look like the transition froze for a moment.

       · Even though it's not very efficient, so far we have to print each individual block. For each size, we print a block that takes the first pixel inside (upper-left) and colors that part with FillRect (StretchBlit has some issue that I will post in the Issues section). I hope FillRect is more efficient than StretchBlit too. We access the color of that pixel position with a static_cast and the int32_t to rgba function.

       · To not give the feeling that the mosaics are too disorienting, nor too obvious they grow from a pivot point in the upper-left corner, we move their offset with the operation (w / 2) % mosaic_size (and same with the y position and height) so they are more or less focused on the center and their alteration is a right balance between not too jumpy and not too mechanic.

       · Just like ZoomIn/Out, both versions of the Mosaic Transition are the same except the Screen pointer and the reverse percentage, so we do the exact same to save code.

       · The complexity of the code is the highest so far (w = width, h = height, m = mosaic size): for [w / m] * for [h / m] = w * h / m^2 FillRects (with our data, for the lowest size (m = 2) => 19.200. It would be good to find a more efficient code than this.

-WaverIn/Out: This transition dissolved the screen like a wave. There is already a perfectly workable function to create Wave effects (WaverBlit), so we just use it with different parameters until we find the one that is the closest to the Wave effect in RM2000. The complexity is one WaverBlit (which inside it uses one Blit per line (h = 240)) each Update.

Fix #900 
Fix #1094


Ghabry:
When the FPS layer is active it is included in the transition effects.
This is a problem with all effects that move stuff around and should be fixed as part of #1094

AUTHORS.md
```diff
@@ -17,3 +17,4 @@ EasyRPG Player authors
 * scurest
 * Shin-NiL
 * Takeshi Watanabe (takecheeze)
+* Alberto León Meaños (albeleon)
```
Ghabry: Please respect the alphabetical order.

You will be the first in the list, congratulation

Ghabry:
About your question in the chat:

"old_instances" is only filled when "Pop" is used to remove scenes from the scene stack.

There is "Scene::Find" for finding a specific scene but this won't help you here because the scene will be also on the stack while IN a battle -> incorrect zoom.

One possible option is to replace
``void Graphics::Transition(TransitionType type, int duration, bool erase = false)``
with
``void Graphics::Transition(TransitionType type, int duration, bool erase = false, int x_pos = -1, int y_pos = -1)``
and then modify the invocation of the Transition functions everywhere when appropriate.

This is not ideal because is only used for zoom but I don't know a better way.

builds/vs2015/PlayerLib.vcxproj
```diff
@@ -757,4 +759,4 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
+</Project>
```
carstene1ns: need to update the `.filters` as well. missing newline.

src/transition.cpp
```diff
@@ -0,0 +1,344 @@
...
+	return true;
+}
+
+bool Transition::IsActive() {
+	return current_frame <= total_frames;
+}
+
+bool Transition::IsErased() {
+	return screen_erased && !IsActive();
+}
```
carstene1ns: missing newline.

src/transition.h
```diff
@@ -0,0 +1,122 @@
...
+	int total_frames;
+	bool screen_erased;
+
+	std::vector<int> zoom_position;
+	std::vector<uint32_t> random_blocks;
+
+	void SetAttributesTransitions();
+};
+
+#endif
```
carstene1ns: missing newline.

src/utils.h
```diff
@@ -174,10 +174,10 @@ namespace Utils {
...
 	/**
-	 * Gets the seeded Random Number Generator (RNG).
-	 *
-	 * @return the random number generator
-	 */
+	* Gets the Random Number Generator (RNG) from Utils.
+	*
+	* @return the random number generator
+	*/
 	std::mt19937 &GetRNG();
```
carstene1ns: this file changed upstream. you need to update the branch you are rebasing on.

Albeleon:
I forgot to say it, but now the transitions are also compatible with TransitionNone (when in Default Teleport you put "Not erased" as an option). Here the result: https://i.imgur.com/F2KJBdI.mp4

Albeleon:
Small update fixing a few bugs in the Zoom transition and TransitionNone + inmediate transition. Now the black flash on the video I linked before in the last transition doesn't show up.

src/player.cpp
```diff
@@ -251,42 +252,38 @@ void Player::Update(bool update_scene) {
...
-	}
+		}
 	if (Input::IsTriggered(Input::TOGGLE_ZOOM)) {
 		DisplayUi->ToggleZoom();
-	}
+		}
 	if (Input::IsTriggered(Input::TOGGLE_FULLSCREEN)) {
 		DisplayUi->ToggleFullscreen();
-	}
+		}
```
Ghabry: Not sure if github renders this incorrectly but the new indentation looks wrong.

src/transition.cpp
```diff
@@ -0,0 +1,348 @@
...
+* (at your option) any later version.
+*
+* EasyRPG Player is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
+*/
```
Ghabry: As usual the stars * are incorrect. They need one extra whitespace to align correctly.

src/transition.h
```diff
@@ -0,0 +1,122 @@
...
+// Headers
+#include <vector>
+#include <string>
+#include "drawable.h"
+#include "system.h"
+#include "scene.h"
+
+
+/*
+* Renders the frame overlay.
```
Ghabry: bogus documentation comment

src/transition.h
```diff
@@ -0,0 +1,122 @@
...
+	int GetZ() const override;
+	DrawableType GetType() const override;
+	
+	/**
+	* Defines a screen transition.
+	*
+	* @param type transition type.
+	* @param linked_scene scene transitioning, it should be either the scene summoning (this) or the current instance (Scene::instance.get())
+	* @param duration transition duration.
+	* @param erase erase screen flag.
```
Ghabry: again the comment whitespace in the whole file

src/graphics.cpp
```diff
@@ -16,130 +16,88 @@
...
 	Cache::Clear();
 }
 
-void Graphics::Update(bool time_left) {
-	if (next_fps_time == 0) {
-		next_fps_time = DisplayUi->GetTicks() + 1000;
-	}
+void Graphics::Update() {
+	//FPS:
+	if (next_fps_time == 0) { next_fps_time = DisplayUi->GetTicks() + 1000; }
```
Ghabry: use proper formating: 
if () {
 // code
}

src/transition.h
```diff
@@ -0,0 +1,123 @@
...
+#include <string>
+#include "drawable.h"
+#include "system.h"
+#include "scene.h"
+
+
+/**
+* Transition class.
+* Shows the transition between two screens.
+*/
```
carstene1ns: ^ forgot these.

Albeleon:
I link #1125 because maybe with the refactor this has been affected. Right now I can't test, later I'll try to.

src/graphics.cpp
```diff
@@ -161,291 +121,54 @@ void Graphics::UpdateTitle() {
...
 	}
 
-	if (global_state->zlist_dirty) {
-		global_state->drawable_list.sort(SortDrawableList);
-		global_state->zlist_dirty = false;
-	}
-
-	current_scene->DrawBackground();
-
+	if (state.drawable_list.size())
```
Ghabry: !drawable_list.empty()

src/graphics.cpp
```diff
@@ -495,3 +217,7 @@ int Graphics::GetDefaultFps() {
 MessageOverlay& Graphics::GetMessageOverlay() {
 	return *message_overlay;
 }
+
+Transition *Graphics::GetTransition() {
```
Ghabry: Is it possible that Transition is nullptr? If not you can do it like MessageOverlay above and return by & reference.

src/player.cpp
```diff
@@ -198,6 +198,7 @@ void Player::Init(int argc, char *argv[]) {
 
 void Player::Run() {
 	Scene::Push(std::shared_ptr<Scene>(static_cast<Scene*>(new Scene_Logo())));
+	Graphics::UpdateSceneCallback();
```
Ghabry: is this a bugfix or unnecessary?

Ghabry: yeah is a bugfix, segfaults without this line

Ghabry:
When choosing Exit in the Title Screen I get a segfault because transision is null. (was already deleted in Graphics::Quit)

```
Transition::IsActive transition.cpp:343
Graphics::IsTransitionPending graphics.cpp:171
Player::Run player.cpp:221
main main.cpp:28
__libc_start_main 0x00007ffff111806b
_start 0x00005555556ec33a
```

Ghabry:
Could you make some changes to the commit list? (rebase and set all commits to "edit")

- Transition change: …  **Change:** Sum up the whole change in the first line, e.g. "Make Transition a Drawable and implement missing transitions (RandomBlockUp/Down, WaverIn/Out, ZoomIn/Out, MosaicIn/Out)"
- ~Style/Formatting errors corrected (Edit: carstene1ns' also corrected)~ Just style fixes **Squash into previous one**
- Fix Issue ( #1371 ): "4.2: RM2000 + RM2003: There should be white fla… … 
- Format fixes and Checking in Graphics::UpdateSceneCallback whether Tr… …  **Changes:** You probably mean IsTransitionPending not UpdateSceneCallback
- Updating "4.2: RM2000 + RM2003: There should be white flashes when a … … **Changes**: State in the first line what was changed e.g. you have this sentence later "When the Transition to a Battle was TransitionNone (appending those flashes), there was some weird behavior."
- Update: When the Transition to a Battle was TransitionNone (appending those flashes), there was some weird behavior. These changes have been made to make it compatible and correct. **Changes:** Remove the "Update:". In General please don't prefix stuff like "Update:" or "PR:" and similiar in front of the messages.
- New Update: RandomBlock Draw modification. … **Change:** Reword it e.g. to "Improve RandomBlock Draw performance"

src/player.cpp
```diff
@@ -251,24 +252,20 @@ void Player::Update(bool update_scene) {
...
 		if (cur_time < next_frame) {
 			DisplayUi->Sleep((uint32_t)(next_frame - cur_time));
 		}
-	} else {
-		Graphics::Update(false);
 	}
 #endif
 
-	// Normal logic update
+	// Input:
```
Ghabry: ping

src/player.cpp
```diff
@@ -307,17 +304,17 @@ void Player::Update(bool update_scene) {
...
-			Graphics::Update(false);
+	for (int i = 0; i < speed_modifier; ++i) {
+		Graphics::Update();
+		if (update_scene) {
 			Scene::instance->Update();
 			++frames;
-			// Scene changed or webplayer waits for files.
-			// Not save to Update again, setup code must run
+
+			// Scene changed or webplayer waits for files. Not save to Update again, setup code must run:
```
Ghabry: pong

PR #1370, opened 2018-06-05
Initial Enemy appearance message: Only visible enemies.

Albeleon:
This PR intends for, in the initial messages about enemy appearances, not including the ones that are hidden just like RM2000. For this:

        - A "Game_EnemyParty::GetVisibleEnemies" function is created.
        - A variable with the visible enemies is initialized right after "enemies" and saved in .cpp. Here I'm not sure whether it should be a private attribute in the header instead, please tell me which one is better.
        - Change in Scene_Battle_Rpg2k the references to GetEnemies for GetVisibleEnemies.

### DIFF
--- a/src/scene_battle_rpg2k.cpp
+++ b/src/scene_battle_rpg2k.cpp
@@ -849,7 +849,7 @@ int Scene_Battle_Rpg2k::GetDelayForLine() {
 }
 
 void Scene_Battle_Rpg2k::SetWaitForEnemyAppearanceMessages() {
-	if ((enemy_iterator == Main_Data::game_enemyparty->GetEnemies().end() &&
+	if ((enemy_iterator == visible_enemies.end() &&
 			!battle_message_window->GetHiddenLineCount()) ||
 			battle_message_window->IsPageFilled()) {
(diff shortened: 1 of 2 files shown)


Ghabry:
I don't think this new function is necessary. Instead you could use the function GetActiveBattlers of Game_Party_Base which handles Hidden & Dead.

Albeleon:
It is, because the new function GetVisibleEnemies, just like GetEnemies, doesn't create a new vector like GetActiveBattlers each time it's invoked; it just gives the address of one that is already defined everytime. This also means that whenever the iterators begin() and end() are invoked, they always refer to the same data and can be equivalent. Since GetEnemies is used three times during two functions, it's important then that the address is the same, or else sooner or later we find this error when trying to do the equivalence:
Code: https://i.imgur.com/rY1iccZ.png
Error: (it gives error in line 916 -> 855) https://i.imgur.com/hdMAr40.png
A possible alternative solution is then, right after the Scene_Battle2k is initialized and the enemies have been loaded, create a protected global variable "visible_enemies" that the two functions use, so that way they refer to the same object without creating a new function in Game_Enemyparty. Please tell me if you'd prefer that.

Issue #1371, opened 2018-06-10
Battle RM2000/2003 - Graphic / Music / Sound Message Display issues: Remaining


Albeleon:
Issues related with Graphics, Music, Sound and Message Display mainly focused for RPG Maker 2000. A few of them also apply to RM2003 just by extension:

1. Battle Events:
- [x] 1.1: RM2000: When a Battle Event is executing (whether by "Wait seconds" or an animation loading), the window battle screen should be empty and you shouldn't be able to access commands.
- [x] 1.2: RM2000: Animations executed in a battle event should not wait until the animation is over when the option "wait to end" is selected.
- [x] 1.3: RM2000 + RM2003: When an enemy dies in a event, it has different effects depending of whether it dies by HP damage or status condition = Death. In the first one, it gradually disappears with monster dying sound. In the second one, it just disappears inmediately.
- [x] 1.4: RM2000: If an event page makes all their actions before the message display window is needed to display, the status window should be able to refresh inmediately without skipping a frame (for example, if "Event page: (Turn 0x + 0); DecreaseHP(Player1, 999, Dead)".

2. Message Display:
- [x] 2.1: RM2000: If no hero party is ready to attack and they can't choose actions, the command options shouldn't flash for a few frames.
- [x] 2.2: RM2000: If a character recovers themself from a negative condition, the message should be printed and not empty.
- [x] 2.3: RM2000 + RM2003: The battle command window (both left [manual, automatic, escape] and right [attack, skill, defend, objects]) should move smoothly of position from one another instead of inmediately (in RM2003, type of Window Battle Screen B and C).
- [x] 2.4: RM2000 (not tested in RM2003): If you have the "special skill name" (to give your character a different name to the skill command) checkbox unmarked, it should show the default skill name instead of what is inside.
- [x] 2.5: RM2000: If the second line of a Skill is not empty: 1. The first line should have a pause before the second line is printed, and 2. The animation should only start when the second line is printed.
- [x] 2.6: RM2000 (issue #1102 ): When an enemy dies, a message should be displayed right after the attack inside the same message.
- [x] 2.7: RM2000 + RM2003: The Victory message should have pauses after each line. The first line with the victory message has a longer pause than the others (in RM2003 only the first line should have a pause).
- [x] 2.8: RM2000: If an enemy causes a condition (poison, sleep, etc.) and fails, the message should be it failed/missed.
[...]

PR #1380, opened 2018-06-24
ToneBlit performance and Method Implementation, Sprite Flashes, other fixes

### LINKED ISSUE #1225: Colored "Boxes" in Battle

---
#### Player platform:
Windows64

[Save05.zip](https://github.com/EasyRPG/Player/files/1157550/Save05.zip)

unbenannt

Colored boxes, if multiple "highlights" are on one battler at a time. These don't appear in RPG_RT.

Reproduction steps: In the Mesdor Sewers have the slime blobs poison you. The character starts to flash green periodically. The box visible in the screenshot appears once a second highlight (active battler, target of healing spell) becomes active. This also appears to happen with other status effects, for example a status effect like "bleeding" (Red flashing of the character) causes red "squares" to appear.

Albeleon:
The issues: ToneBlit is not very efficient, Battle tones and Gradual TintScreen are tinted differently, and Sprite flashes.

ToneBlit is very heavy performance-wise, which forces us to use two different ways to implement tintscreen in our screen:
1. We use ToneBlit for any individual tile, background and sprite when they're drawn.
2. After all the drawables are drawn, we use ToneBlit to the screen.

 - The 1st method, with a little modification, allows us to generate Sprite flashes that ignore the tintscreen (for example, a white flash with a dark-red tintscreen, it should appear completely white instead of red-ish). The 2nd doesn't allow that.
 - The 1st method is more efficient when the tintscreen remains static, but slower when the tone screen is gradually changing.
 - The 2nd is used for Gradual Tintscreen and Battle scenes. The 1st one is used anywhere else.

The ideal solution would be using the first to fix the flash bug. For that, we try to make the ToneBlit function more efficient. It sacrifices legibility in order to gain a little more of performance. These changes (cache variables, combining checks, etc.) improve the performance so there's not much difference for the 1st and 2nd method when a gradual tone screen happens, and therefore the 1st method is viable for everything.

To implement this:
 - Bitmap: We modify the ToneBlit to make it more efficient.
 - Sprite: We invert the positions of ToneBlit and BlendBlit. This way the flash color will have priority and not be tinted.
 - Spriteset_Map: Now in all the cases it uses the 1st method, so the code that allows the 2nd method is deleted.
 - Spriteset_Battle: We change from the 2nd method to the 1st one. We set the tone for the background and the battler sprites.
 - Background: We include the option of setting and getting a tone, and then draw it.
 - Screen: We eliminate the tone functions since no other class uses them.

Also, another fixes:
 - Game_Character flashes were limited to 128 no matter what the actual value was. An attribute flash_alpha has been added to account for this, and when the flash is set and invoked in its subclasses (Event, Vehicle, Player) we use that attribute.
 - Battle_Animation "flash_length" has been changed from 5 to 12, because the way it was before it was almost unnoticeable in comparison with RPG_RT.
 - Battle_Animation and Game_Interpreter_Map: The value of the flashes before with the << 3 operation meant the flash value for each part would be [0, 248]. With "x * 255 / 31", the values is really [0, 255].

Fix #1225

### DIFF
--- a/src/background.cpp
+++ b/src/background.cpp
@@ -29,7 +29,7 @@
 #include "output.h"
 
 Background::Background(const std::string& name) :
-	visible(true),
+	visible(true), tone_effect(Tone()),
 	bg_hscroll(0), bg_vscroll(0), bg_x(0), bg_y(0),
 	fg_hscroll(0), fg_vscroll(0), fg_x(0), fg_y(0) {
 
@@ -43,7 +43,7 @@ Background::Background(const std::string& name) :
 }
 
 Background::Background(int terrain_id) :
-	visible(true),
+	visible(true), tone_effect(Tone()),
 	bg_hscroll(0), bg_vscroll(0), bg_x(0), bg_y(0),
 	fg_hscroll(0), fg_vscroll(0), fg_x(0), fg_y(0) {
 
@@ -109,6 +109,15 @@ DrawableType Background::GetType() const {
 	return type;
 }
 
+Tone Background::GetTone() const {
+	return tone_effect;
+}
+
+void Background::SetTone(Tone tone) {
+	if (tone_effect != tone) {
+		tone_effect = tone;
+	}
+}
 void Background::Update(int& rate, int& value) {
 	int step =
 		(rate > 0) ? 1 << rate :
@@ -143,4 +152,8 @@ void Background::Draw() {
 
 	if (fg_bitmap)
 		dst->TiledBlit(-Scale(fg_x), -Scale(fg_y), fg_bitmap->GetRect(), *fg_bitmap, dst_rect, 255);
+
+	if (tone_effect != Tone()) {
+		dst->ToneBlit(0, 0, *dst, dst->GetRect(), tone_effect, Opacity::opaque);
+	}
 }
--- a/src/background.h
+++ b/src/background.h
@@ -23,6 +23,7 @@
 #include "system.h"
 #include "drawable.h"
 #include "async_handler.h"
+#include "tone.h"
 
 class Background : public Drawable {
 public:
@@ -32,6 +33,8 @@ public:
 
 	void Draw() override;
 	void Update();
+	Tone GetTone() const;
+	void SetTone(Tone tone);
 
 	int GetZ() const override;
 	DrawableType GetType() const override;
@@ -48,6 +51,8 @@ private:
 
 	bool visible;
 
+	Tone tone_effect;
+	Tone current_tone;
 	BitmapRef bg_bitmap;
 	int bg_hscroll;
 	int bg_vscroll;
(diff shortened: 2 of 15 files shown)


Albeleon:
#1371 The original reason why I checked this:
"4.12: RM2000 (not tested in RM2003): Maybe it's my impression but when an attack animation shows up to an enemy, shouldn't the enemy flashes inside the animation (not afterwards) be clearer? In the video, sometimes they don't appear."

Ghabry:
Could you split the single commit in two?
The bitmap.cpp change is completely unrelated to the rest.

The easiest way to do this hear would be doing "amend commit" and unchecking "bitmap.cpp". Then commiting "bitmap.cpp" using an additional commit.

Ghabry:
For more context: The Tone/Saturation code had a optimisation that decided if the change is a gradient or static.
See #1017

But this change had visual issues e.g. for weather effects (can't find the issue number) and needs testing if @Albeleon fixed it.

In general the assumption that there is a "screen layer" and everything below it is affected by tone doesn't hold true anymore since Cherrys RPG2k3E release because each picture can be individually configured if it is affected by tone.

This PR needs some benchmarking.

Ghabry:
This also fixes #1225

Ghabry:
This also needs a rebase on upstrema to fix the Android build failure.

Ghabry:
For benchmarking you can use this patch which unlocks the FPS but still keeps the updates at 60:

```diff
diff --git a/src/player.cpp b/src/player.cpp
index f646bc9c..d39cac16 100644
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -248,7 +248,7 @@ void Player::Update(bool update_scene) {
 	static const double framerate_interval = 1000.0 / Graphics::GetDefaultFps();
 	next_frame = start_time + framerate_interval;
 
-#ifdef EMSCRIPTEN
+#if 1
 	// Ticks in emscripten are unreliable due to how the main loop works:
 	// This function is only called 60 times per second instead of theoretical
 	// 1000s of times.
@@ -303,12 +303,17 @@ void Player::Update(bool update_scene) {
 	}
 
 	Audio().Update();
-	Input::Update();
 
 	std::shared_ptr<Scene> old_instance = Scene::instance;
 
 	int speed_modifier = GetSpeedModifier();
 
+	double cur_time = (double)DisplayUi->GetTicks();
+	if (cur_time < next_frame) {
+		return;
+	}
+	Input::Update();
+
 	for (int i = 0; i < speed_modifier; ++i) {
 		Graphics::Update();
 		if (update_scene) {
```

Ghabry:
With my benchmark I get the following FPS:

 Test  | Upstream  | This PR
---|---|---
No effect |  1050 | 1050
Tone gradient  | 650  | 970  
Saturation gradient | 623 | 970
Tone+Sat grad. | 500 | 960

The FPS measurement is not fully precise, error is ±50 FPS but you can see that this PR is significantly faster in all operations.

PR #1381, opened 2018-06-24
EXE ExFont extraction.

20kdc:
While this code probably isn't perfect, it's hopefully simple and hopefully works (It seems to so far).
If no ExFont file is present as a known image, one will be written to the save directory.
NOTE: This occurs even if one is already in the save directory that is not recognized as a game directory image.
This quirk is so Emscripten-based game updates work properly.
Also, adds support for loading ExFont.bmp from the save directory.

So far, only tested with "Mother: Cognitive Dissonance", but EXFONT format does not appear to vary.

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -257,6 +257,8 @@ namespace {
 	}
 }
 
+std::vector<uint8_t> Cache::exfont_custom;
+
 #define cache(elem) \
 	BitmapRef Cache::elem(const std::string& f) { \
 		bool trans = spec[Material::elem].transparent; \
@@ -292,18 +294,11 @@ BitmapRef Cache::Exfont() {
 	cache_type::iterator const it = cache.find(hash);
 
 	if (it == cache.end() || !it->second.bitmap) {
-		std::string exfont_file = FileFinder::FindImage(".", "ExFont");
-		if (exfont_file.empty()) {
-			// Check for automatically extracted ExFont in the save directory
-			std::shared_ptr<FileFinder::DirectoryTree> save_tree = FileFinder::CreateSaveDirectoryTree();
-			// Due to API limitations only check for ExFont.bmp
-			exfont_file = FileFinder::FindImage(*save_tree, "ExFont.bmp");
-		}
-		BitmapRef exfont_img;
 		// Allow overwriting of built-in exfont with a custom ExFont image file
-		// Probe before using LoadBitmap because this generates a warning when the file is missing
-		if (!exfont_file.empty()) {
-			exfont_img = Bitmap::Create(exfont_file, true);
+		// exfont_custom is filled by Player::CreateGameObjects
+		BitmapRef exfont_img;
+		if (!exfont_custom.empty()) {
+			exfont_img = Bitmap::Create(exfont_custom.data(), exfont_custom.size(), true);
 		} else {
 			exfont_img = Bitmap::Create(exfont_h, sizeof(exfont_h), true);
 		}
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -294,10 +294,10 @@ BitmapRef Cache::Exfont() {
 	if (it == cache.end() || !it->second.bitmap) {
 		std::string exfont_file = FileFinder::FindImage(".", "ExFont");
 		if (exfont_file.empty()) {
-			// Extracted EXFONTs have a specific name and are only in the save directory tree.
-            // To clarify - CreateSaveDirectoryTree *must* be called to ensure the resulting directory tree is up to date.
-            std::shared_ptr<FileFinder::DirectoryTree> save_tree = FileFinder::CreateSaveDirectoryTree();
-			exfont_file = FileFinder::FindDefault(*save_tree, "ExFont.bmp");
+			// Check for automatically extracted ExFont in the save directory
+			std::shared_ptr<FileFinder::DirectoryTree> save_tree = FileFinder::CreateSaveDirectoryTree();
+			// Due to API limitations only check for ExFont.bmp
+			exfont_file = FileFinder::FindImage(*save_tree, "ExFont.bmp");
 		}
 		BitmapRef exfont_img;
 		// Allow overwriting of built-in exfont with a custom ExFont image file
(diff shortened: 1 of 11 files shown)


src/exe_reader.cpp
```diff
@@ -0,0 +1,189 @@
...
+bool EXEReader::ResNameCheck(uint32_t i, const char * p) {
+	if (GetU16(i) != strlen(p))
+		return false;
+	while (*p) {
+		i += 2;
+		if (GetU16(i) != *p)
+			return false;
+		p++;
+	}
+	return true;
```
Ghabry: Does the code support big endian?

20kdc: Big-endian platforms or big-endian EXEs?

Ghabry: Platforms of course, there is no RPG_RT for PowerPC (luckily)

20kdc: Yes. See GetU32/GetU16/GetU8. For simplicity's sake, GetU32 is based off of GetU16 which is based off of GetU8. Thus, no endian-specific code to worry about.

fdelapena:
Some thoughts in favour to get this merged:

This patch is not doing reverse engineering to the executable, this means, no executable code is parsed or reversed to reach them. In fact, the PE format is a documented format which contains some blocks with static data which gets parsed even by the Windows explorer to extract the icon. There are also open source programs like 7-zip which show these resources without getting any complain.

To sum up, I think there is no license violation here, as this is done in a fair and standard file format handling way.

20kdc:
Should I rebase & squash, or just rebase?

src/player.cpp
```diff
@@ -695,26 +695,29 @@ void Player::CreateGameObjects() {
...
+			EXEReader exe_reader = EXEReader(exe);
+			exe_reader.GetExfont(filename);
+			exe.close();
 #ifdef EMSCRIPTEN
-		// From scene_save : Save changed file system
-		EM_ASM(
-			FS.syncfs(function(err) {
-			});
-		);
+			// From scene_save : Save changed file system
```
Ghabry: this emscripten code block looks at the wrong position, does not belong in the exe-else block

Ghabry:
normal rebase is enough

20kdc:
Rebased, but it's been a while so I should double-check stuff.

src/player.cpp
20kdc: It's part of the block immediately after saving a new ExFont and before loading it.

20kdc: ...Wait, no, that's a removal...

Ghabry: oh I didn't see that you added this code, thought this was our savegame code, nvermind

20kdc:
Things seem fine, but test for explosions

Ghabry:
Okay devs, please give your feedback. To sum up the controversial around this PR:
Many of us are against parsing RPG_RT.exe to configure the Player. Especially auto-detecting patches via RPG_RT-parsing is highly questionable and I'm against it.
Resource parsing is a corner-case. This feels less terrible than patch detection, therefore I'm not completely against it but it still feels wrong.
Therefore: Tell me your current opinion.


@20kdc would you allow relicensing the code to MIT license? Then we could also use it in the EasyRPG/Tools repo as a standalone resource ripper (for title and exfont).

20kdc:
Feel free to use my PR's modifications and additions under any license you wish.
As for my opinion, it is that resource parsing is merely an extension of the project's structure - if informally - while patch detection is closer to Wine-like activity (but in a hackier manner) and by nature is going to be an incoherent mess.

fmatthew5876:
Our objective is to emulate the game experience as faithfully as possible. If game developers are putting game content into the exe then we have no other choice but to rip it out.

So I am not against this for game content.

carstene1ns:
The general idea of this PR is fine. However, the emscripten stuff should be removed.
Games need preprocessing anyway, so extraction of exfont/logos should be done before.
This also allows removing `ForceLoadExfont()`, which I consider a hack anyways.

Ghabry:
@20kdc 
To sum this up: Remove the emscripten handling, just search for "RPG_RT.exe" as usual, it will silently fail.

Which means  the workflow should be:

unnamed0

20kdc:
...I also rebased.

src/cache.cpp
```diff
@@ -293,6 +293,12 @@ BitmapRef Cache::Exfont() {
 
 	if (it == cache.end() || !it->second.bitmap) {
 		std::string exfont_file = FileFinder::FindImage(".", "ExFont");
+		if (exfont_file.empty()) {
+			// Extracted EXFONTs have a specific name and are only in the save directory tree.
+            // To clarify - CreateSaveDirectoryTree *must* be called to ensure the resulting directory tree is up to date.
```
carstene1ns: whitespaceeeeeeee.

src/options.h
```diff
@@ -57,6 +57,21 @@
...
+ * So in case of Emscripten, instead the filename used is "exfont.dll",
+ *  which makes it clear the file is only being downloaded for the EXFONT resource.
+ * The best way to handle this is of course to just have people extracting the resource themselves,
+ *  but we can't really go around expecting people to do that.
+ * However much protection that gives... debate in PR comments.
+ */
+#ifndef EMSCRIPTEN
+#define EXE_NAME "RPG_RT.exe"
+#else
+#define EXE_NAME "exfont.dll"
```
carstene1ns: remove.

src/scene_logo.cpp
```diff
@@ -120,11 +120,14 @@ void Scene_Logo::OnIndexReady(FileRequestResult*) {
 	tree->SetImportantFile(true);
 	FileRequestAsync* ini = AsyncHandler::RequestFile(INI_NAME);
 	ini->SetImportantFile(true);
+	FileRequestAsync* exe = AsyncHandler::RequestFile(EXE_NAME);
+	exe->SetImportantFile(true);
```
carstene1ns: This should be reordered I think.

Ghabry: can be removed, as emscripten won't parse RPG_RT.exe and exfont loading is already below

20kdc:
I should probably squash my commits at some point.

Ghabry:
Didn't want to overwrite this branch because I made so many changes that I consider doing this bad style unless discussed.

https://github.com/Ghabry/easyrpg-player/commits/exexfont

What I did: Bit of squashing and rewording.
I changed the code to only check if:
  - ExFont in the game directory and when not found
  - reads the RPG_RT.exe ExFont
In both cases the ExFont is directly read in a memory buffer, allowed simpler logic in Cache::ExFont

Used djb2_hash to detect if the ExFont is custom. Just for informational purposes.

carstene1ns:
Well, I am ok with your changes.

PR #1384, opened 2018-06-28
ZIP file support for standalone mode on Android.

Brianum:
Add support for reading the game from a ZIP file (game.zip) in standa…lone mode instead of a plain directory.

This addresses issues with filenames containing umlauts or other special characters not being allowed into AssetManager's asset bundle.
Also allows for compression to save space.


Ghabry:
AssetManager does not allow non-ASCII characters? Never tested this but sounds strange.

Looks okay to me, fortunately this copy/extract nonsense can be skipped soon (tm) when the virtual filesystem support is ready.

PR #1389, opened 2018-06-29
Generate quick light red translucid flash when step on damaged terrain.

Albeleon:
Problem: When stepping in damaged terrain and there's more than 0 damage, there's no flash indicating it.

Solution: Put that flash after trial and error getting the exact color, power and time. Also, no need to invoke the damage function to the party when the damage is 0.

### DIFF
--- a/src/game_party.cpp
+++ b/src/game_party.cpp
@@ -411,7 +411,7 @@ int Game_Party::GetRunCount() {
 	return data.escapes;
 }
 
-void Game_Party::ApplyDamage(int damage) {
+void Game_Party::ApplyDamage(int damage, bool lethal) {
 	if (damage <= 0) {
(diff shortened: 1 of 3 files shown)


Ghabry:
I just tested one corner case:
Terrain can not kill party members, the HP only reaches "1".

The easiest way to fix this would be adding a "bool lethal" (like in Game_Interpreter ChangeHP command e.g.) as a new argument to ApplyDamage and when false it is capped to 1 HP.

I also found a bug in ApplyDamage:
https://github.com/EasyRPG/Player/blob/master/src/game_party.cpp#L414-L425

The SetHp() call must be changed to ChangeHp(-damage) because only ChangeHp handles the 0 HP case correctly...

PR #1390, opened 2018-06-29
[Regression fix] Stop sounds (OFF) / (...) only when CommandPlaySound is called.

Albeleon:
Problem: An option where, if an (OFF), Empty or filename in parenthesis is called, makes all the sound effects stop (SE_Stop). This also happened in instances outside "CommandPlaySound" (where the creator puts it intentionally), for example in Battle sound animations for effects that include anything but a flash, and generated sound cuts everywhere.

Solution: Add in SePlay and OnSeReady an optional input bool "stop_sounds" (by default "false") that only stop all sounds if it's true, with the OnSeReady input being dependent of the SePlay input. We also put SePlay in CommandPlaySound with stop_sounds = true, the only case where we want it to.

### DIFF
--- a/src/game_system.cpp
+++ b/src/game_system.cpp
@@ -139,13 +139,15 @@ void Game_System::BgmFade(int duration) {
 	force_bgm_play = true;
 }
 
-void Game_System::SePlay(RPG::Sound const& se) {
+void Game_System::SePlay(RPG::Sound const& se, bool stop_sounds) {
 	static bool ineluki_warning_shown = false;
 
 	if (se.name.empty()) {
 		return;
 	} else if (se.name == "(OFF)") {
-		Audio().SE_Stop();
+		if (stop_sounds) {
+			Audio().SE_Stop();
+		}
 		return;
 	}
 
@@ -181,7 +183,7 @@ void Game_System::SePlay(RPG::Sound const& se) {
 	}
 
 	FileRequestAsync* request = AsyncHandler::RequestFile("Sound", se.name);
-	se_request_ids[se.name] = request->Bind(std::bind(&Game_System::OnSeReady, std::placeholders::_1, volume, tempo));
+	se_request_ids[se.name] = request->Bind(std::bind(&Game_System::OnSeReady, std::placeholders::_1, volume, tempo, stop_sounds));
 	request->Start();
 }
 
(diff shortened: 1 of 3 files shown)


Albeleon:
This also solves in #1371 the issue "X.2: Every sound should correctly play without interruptions, whether in animations or the cursors."

src/game_system.h
```diff
@@ -244,7 +244,7 @@ namespace Game_System {
 	void PlayMemorizedBGM();
 
 	void OnBgmReady(FileRequestResult* result);
-	void OnSeReady(FileRequestResult* result, int volume, int tempo);
+	void OnSeReady(FileRequestResult* result, int volume, int tempo, bool stop_sounds = false);
```
Ghabry: you can remove the default argument here

Ghabry:
Minor style note: Please add {} around if one-liners. Yes I know that the code is full of them but we slowly try to have a consistent style.

src/game_system.h
```diff
@@ -118,7 +118,7 @@ namespace Game_System {
 	 *
 	 * @param se sound data.
```
Ghabry: Needs doc-comment:
```
@param stop_sounds If true stops all SEs when playing (OFF). Only used by the interpreter.
```

src/game_system.cpp
```diff
@@ -139,13 +139,15 @@ void Game_System::BgmFade(int duration) {
...
 
-void Game_System::SePlay(RPG::Sound const& se) {
+void Game_System::SePlay(RPG::Sound const& se, bool stop_sounds) {
 	static bool ineluki_warning_shown = false;
 
 	if (se.name.empty()) {
 		return;
 	} else if (se.name == "(OFF)") {
-		Audio().SE_Stop();
+		if (stop_sounds) {
```
carstene1ns: This one could be moved upwards, e.g.: `if ((se.name == "(OFF)" && stop_sounds) {`

src/game_system.cpp
```diff
@@ -432,15 +434,17 @@ void Game_System::OnBgmReady(FileRequestResult* result) {
...
+void Game_System::OnSeReady(FileRequestResult* result, int volume, int tempo, bool stop_sounds) {
 	auto item = se_request_ids.find(result->file);
 	if (item != se_request_ids.end()) {
 		se_request_ids.erase(item);
 	}
 
 	std::string path;
 	if (isStopFilename(result->file, FileFinder::FindSound, path)) {
-		Audio().SE_Stop();
+		if (stop_sounds) {
```
carstene1ns: same here.

PR #1393, opened 2018-07-07
Step State Damage: "Implement HP and SP decrease by a state on map" and "Prevent Terrain Damage"

Albeleon:
Depends on EasyRPG/liblcf#228

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -1344,3 +1344,15 @@ void Game_Actor::RemoveInvalidData() {
 		SetLevel(1);
 	}
 }
+
+bool Game_Actor::PreventsTerrainDamage() {
+	for (auto object_id : GetWholeEquipment()) {
+		RPG::Item *object = ReaderUtil::GetElement(Data::items, object_id);
+		if (object != nullptr && (object->type == RPG::Item::Type_shield || object->type == RPG::Item::Type_armor
+			|| object->type == RPG::Item::Type_helmet || object->type == RPG::Item::Type_accessory) && object->no_terrain_damage) {
+			return true;
+		}
+	}
+
+	return false;
+}
--- a/src/game_party.h
+++ b/src/game_party.h
@@ -306,6 +306,19 @@ public:
 	 * Removes invalid actors and items from the party
 	 */
 	void RemoveInvalidData();
+
+	std::vector<int16_t> GetInflictedStates() const;
+
+	/**
+	* Applies damage to the game party based on their stats.
+	*
+	* @return Whether the actor suffered some damage.
+	*/
+	bool ApplyStateDamage();
+
+private:
+	std::vector<int> state_steps_hp;
+	std::vector<int> state_steps_sp;
 };
 
 #endif
--- a/src/game_player.cpp
+++ b/src/game_player.cpp
@@ -719,8 +719,12 @@ void Game_Player::BeginMove() {
 			Game_System::SePlay(terrain->footstep);
 		}
 		if (terrain->damage > 0) {
-			red_flash = true;
-			Main_Data::game_party->ApplyDamage(terrain->damage, false);
+			for (auto hero : Main_Data::game_party->GetActors()) {
+				if (!hero->PreventsTerrainDamage()) {
+					red_flash = true;
+					hero->ChangeHp(-std::max<int>(0, std::min<int>(terrain->damage, hero->GetHp() - 1)));
+				}
+			}
 		}
 	} else {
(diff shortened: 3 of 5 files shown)


src/game_player.cpp
```diff
@@ -712,17 +712,65 @@ bool Game_Player::CanWalk(int x, int y) {
...
+				if ((++state_steps_sp[state_id - 1]) >= state->sp_change_map_steps) {
+					state_steps_sp[state_id - 1] -= state->sp_change_map_steps;
+					hero->ChangeSp(-state->sp_change_map_val);
+					red_flash = true;
+				}
+			}
+		}
+	}
+
+	//Terrain:
```
Ghabry: whitespace after // and don't use these colons :
The comment isn't very useful in general

``
 	int terrain_id = Game_Map::GetTerrainTag(GetX(), GetY()); 
 	const RPG::Terrain* terrain = ReaderUtil::GetElement(Data::terrains, terrain_id);
``
these 2 lines belong to the terrain and should be moved down

src/game_player.cpp
```diff
@@ -712,17 +712,65 @@ bool Game_Player::CanWalk(int x, int y) {
...
+					red_flash = true;
+					hero->ChangeHp(-std::max<int>(0, std::min<int>(terrain->damage, hero->GetHp() - 1)));
+				}
+			}
 		}
 	} else {
 		Output::Warning("Player BeginMove: Invalid terrain ID %d at (%d, %d)", terrain_id, GetX(), GetY());
 	}
+
+	//Red Flash:
```
Ghabry: this one is also of low value

src/game_player.cpp
```diff
@@ -712,17 +712,65 @@ bool Game_Player::CanWalk(int x, int y) {
...
 	if (terrain) {
 		if (!terrain->on_damage_se || (terrain->on_damage_se && (terrain->damage > 0))) {
 			Game_System::SePlay(terrain->footstep);
 		}
 		if (terrain->damage > 0) {
-			Main_Data::game_screen->FlashOnce(31, 10, 10, 19, 1);
-			Main_Data::game_party->ApplyDamage(terrain->damage, false);
+			for (auto hero : Main_Data::game_party->GetActors()) {
+				bool receive_damage = true;
+				for (auto object_id : hero->GetWholeEquipment()) {
```
Ghabry: this complecated loop could be moved in a Game_Actor::IsImmuneToTerrainDamage function

src/game_player.cpp
```diff
@@ -712,17 +712,65 @@ bool Game_Player::CanWalk(int x, int y) {
 void Game_Player::BeginMove() {
 	int terrain_id = Game_Map::GetTerrainTag(GetX(), GetY());
 	const RPG::Terrain* terrain = ReaderUtil::GetElement(Data::terrains, terrain_id);
+	bool red_flash = false;
+
+	// Step Damage for States:
+	for (auto hero : Main_Data::game_party->GetActors()) {
+		for (auto state_id : hero->GetInflictedStates()) {
+			RPG::State *state = ReaderUtil::GetElement(Data::states, state_id);
```
Ghabry: This whole loop logic could be in a Game_Actor::ApplyStateDamage function.
About the state steps counter: Is this really global for all actors? I mean when one hero is poisoned and another is poisoned one step later, do they receive damage at the same time?

src/game_actor.cpp
```diff
@@ -1344,3 +1344,45 @@ void Game_Actor::RemoveInvalidData() {
...
+		RPG::State *state = ReaderUtil::GetElement(Data::states, state_id);
+		if (state_steps_hp.size() < state_id) {
+			state_steps_hp.resize(state_id);
+		}
+		if (state_steps_sp.size() < state_id) {
+			state_steps_sp.resize(state_id);
+		}
+		if (state->hp_change_map_val > 0) {
+			if ((++state_steps_hp[state_id - 1]) >= state->hp_change_map_steps) {
+				state_steps_hp[state_id - 1] -= state->hp_change_map_steps;
```
Ghabry: can't you simply set it to 0 instead of subtracting steps?

Ghabry:
Depends on https://github.com/EasyRPG/liblcf/pull/228

src/game_party.h
```diff
@@ -306,6 +306,19 @@ class Game_Party : public Game_Party_Base {
...
 	 */
 	void RemoveInvalidData();
+
+	std::vector<int16_t> GetInflictedStates() const;
+
+	/**
+	* Applies damage to the game party based on their stats.
+	*
+	* @return Whether the actor suffered some damage.
+	*/
```
Ghabry: You messed up the * again. This automatic formating is really annoying Visual Studio behaviour ^^'
Can be fixed as part of some other PR:

Issue #1306, opened 2017-11-05
"Change Params" not properly capped when going below 1


Ghabry:
Assuming you have 10 HP and you do "Change Params" with "-100 HP" the HP value becomes internally "-90 HP" in EasyRPG but is adjusted to 1 HP when accessed.

This behaviour appears to be consistent with the English release of 2k3 but because this is 1.09 based this was probably fixed in this version and not by Cherry?
In 1.08 the "-100" would result in "1 HP" so doing another +10 would be "11 HP" and not "-80 HP"

Also needs a test in 2k :/
Not aware of any game except Deep 8 that relies on this < 1.09 behaviour. Found by the Deep8 author who told me he will fix it in his event code so it is 1.09 compatible.

fmatthew5876:
This is likely fixed now in master. Needs a retest.

Ghabry:
The main problem here was that it is depending on the version of the RPG\_RT engine being used.

Not worth emulating until any game breaks because of this, just need to agree on one behaviour.

fmatthew5876:
I tested this event code in master:

```
Change HP -999 (no death)
Change HP +10
```

Result: 11HP

```
Change HP -999, Allow Death
Change HP +10
```
Result: 0 HP, Dead

matches latest steam RPG_RT.

In my opinion this is the only sensible thing to do. We have to take a stand and I vote we stand with the sensible thing and be compatible with latest versions of RPG_RT.

So I vote we close this issue.

PR #1394, opened 2018-07-07
ChangeHP, AddState and RemoveState changes.

### LINKED ISSUE #1128: A bunch of Vampires Dawn bugs
1) [x] Asgar "Life gift" doesn't scale from intelligence, just heals for 40. On RPG_RT it heals for ~50-60.
2) [x] Escape button always greyed out.
3) [x] Seems like equipment from "other" slot with status effect immunities not working. Like desert giants in desert still apply bleeding on heroes with item(protective bloodring on Asgar and Alaine) that gives bleed immunity. Same happened with ring giving silence immunity(protective ring).
- [x] Sunblock cloak that should block damage on world map in desert also not working(equipped on Valnar).
4) [x] Paralyze not working on enemies. Message "Enemy was paralyzed" is displayed but they keep to act normally. Can be reproduced by casting ice magic on enemies until "Enemy is paralyzed" appears.

https://drive.google.com/file/d/0B5F-baO8aHjrUTVXdFIxdWpBNDg - Game with savefile(currently in desert so everything can be reproduced easily) and log

Albeleon:
The way RPG_RT works:
 - You cannot change a battler's HP when they're dead.
 - If ChangeHP kills a character, it adds the Death state to that battler.
 - When Death state is cured, even if alone, it always sets the HP to 1, instead of other functions.
 - When a character is caused Death, all the other states are erased; the defense, charge, gauge and stat modifiers are reset, and the current attribute shifts erased.

We implement all of this, with these particular changes:
 - Now there is no difference between Game_Actor::ChangeHP and Game_Enemy::ChangeHP, so we delete this function from both and add the ChangeHP characteristics to the superclass Game_Battler.
 - In Game_Battler::UseItem, the function now doesn't have to set the HP. However, it needs to have an int "revived" that checks if the character was revived, so in case the function heals HP too after dead, it substracts 1 since that value is automatically put after the Death state is removed. We also move the state cure before the HP change, so the Death State is cured before ChangeHP is executed (if this function was executed before Death state was cured, due to our new conditions, it wouldn't do anything).

---

Fixes #1128.

### DIFF
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -212,7 +212,7 @@ bool Game_Battler::UseItem(int item_id) {
 
 	if (item->type == RPG::Item::Type_medicine) {
 		bool was_used = false;
-
+		int revived = 0;
 		int hp_change = item->recover_hp_rate * GetMaxHp() / 100 + item->recover_hp;
 		int sp_change = item->recover_sp_rate * GetMaxSp() / 100 + item->recover_sp;
 
@@ -221,32 +221,28 @@ bool Game_Battler::UseItem(int item_id) {
 			if (item->state_set.empty() || !item->state_set[0]) {
 				return false;
 			}
-
-			// Revive gives at least 1 Hp
-			if (hp_change == 0) {
-				ChangeHp(1);
-				was_used = true;
-			}
 		} else if (item->ko_only) {
 			// Must be dead
 			return false;
 		}
 
+		for (int i = 0; i < (int)item->state_set.size(); i++) {
+			if (item->state_set[i]) {
+				was_used |= HasState(Data::states[i].ID);
+				if (i == 0 && HasState(i + 1))
+					revived = 1;
+				RemoveState(Data::states[i].ID);
+			}
+		}
+
 		if (hp_change > 0 && !HasFullHp()) {
-			ChangeHp(hp_change);
+			ChangeHp(hp_change - revived);
 			was_used = true;
 		}
 
 		if (sp_change > 0 && !HasFullSp()) {
 			ChangeSp(sp_change);
 			was_used = true;
-		}
-
-		for (int i = 0; i < (int)item->state_set.size(); i++) {
-			if (item->state_set[i]) {
-				was_used |= HasState(Data::states[i].ID);
-				RemoveState(Data::states[i].ID);
-			}
 		}
 
 		return was_used;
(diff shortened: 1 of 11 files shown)


Ghabry:
To everybody reviewing this:
There are multiple ways this breaks code, don't approve this without checking at least:

- Using of healing items in the Item scene
- Using of healing skills in the Skill scene
- Using of specific event commands that alter the HP (need to make a list first)

src/game_party.h
```diff
@@ -311,6 +311,13 @@ class Game_Party : public Game_Party_Base {
 	 */
 	bool ApplyStateDamage();
 
+	/**
+	* Gets the actor with the highest level. If there are many, choose the one with the earliest position in the group.
+	*
```
carstene1ns: indentation.

Albeleon:
#1306 The first commit was done under a few assumptions that differ from this issue. I have to check this and both issues with more detail.

src/game_battler.cpp
```diff
@@ -261,21 +261,22 @@ bool Game_Battler::UseItem(int item_id) {
 		case RPG::Item::Type_armor:
 		case RPG::Item::Type_helmet:
 		case RPG::Item::Type_accessory:
-			return item->use_skill && UseSkill(item->skill_id);
+			return item->use_skill && UseSkill(item->skill_id, Main_Data::game_party->GetHighestLeveledActor());
 		case RPG::Item::Type_special:
-			return UseSkill(item->skill_id);
+			return UseSkill(item->skill_id, Main_Data::game_party->GetHighestLeveledActor());
```
Ghabry: The observation that using of items that invoke skills uses the HighestLeveledActor instead of the current Actor is wrong.

Proof:
1. Create a "Special" Item that Invokes the Skill "Heal".
2. Set "Heal" to Fire element
3. Configure Fire as: A 999%, C 0%
4. Give the 1st actor Fire A and the 2nd actor Fire C
5. Try to use the item ingame. Result: 2nd actor can't use the item on itself.

Ghabry:
Made some further tests:
Your observation is partly correct.

It depends on the scope of the skill.

- When "The User" is the scope you get a glitched cursor in the menu (always on the first member) but the one who gets the item applied is the one with the highest skill
- When "One Ally" the attribute of the Ally applies
- When "All Allies/Party": I have no idea what happens here o.O

Ghabry:
I take over this one...

fmatthew5876:
I tested this branch and "When Death state is cured... " works. It fixes a bug in HH3's battle cleanup events.

Ghabry:
For getting this mergable for now I replaced the ``Main_Data::game_party->GetHighestLeveledActor()`` with "this" for Skills of items for now and added a FIXME that the code doesn't work properly for all cases of scope.

src/game_battler.cpp
```diff
@@ -803,3 +863,21 @@ void Game_Battler::ShiftAttributeRate(int attribute_id, int shift) {
 		--old_shift;
 	}
 }
+
+float Game_Battler::GetAttributeMultiplier(const std::vector<bool>& attributes_set) const {
+	float physical = -1001, magical = -1001;
```
Ghabry: This is confusing because we have then GetAttributeMultiplier in BattleAlgorithm (commit "3.16: RM2000: Attribute modifier is the maximum of the percentage da" of #1448 and in Game_Battler... :/   (and with the same -1001 issue)

Ghabry: see https://github.com/EasyRPG/Player/pull/1448/commits/8ef8f2e9d95b391373e211b49f57d6a11a74d229

fmatthew5876:
The first commit has been merged. The rest of the commits are now in #1529 and #1526. So nothing is left outstanding in this branch.

carstene1ns:
Closing this then.

PR #1395, opened 2018-07-07
Restrict access to Skill Menu if character has "Do Nothing" restriction.

Albeleon:
Solution: If the character can't act when they're selected for the skill menu, reproduce a Buzzer sound and do nothing else.


Ghabry:
as mentioned in the chat the Row command has no SE anymore, otherwise :ok:

PR #1397, opened 2018-07-08
Misc fixes

Albeleon:


src/game_actor.h
```diff
@@ -423,6 +423,13 @@ class Game_Actor : public Game_Battler {
 	 */
 	const std::vector<int16_t>& GetWholeEquipment() const;
 
+	/**
+	* Checks whether it has an item equipped.
```
carstene1ns: Suggestion: `Checks if the actor has a specific item equipped.`

Also, below: `Whether the item is equipped`

src/game_system.cpp
```diff
@@ -453,3 +453,19 @@ void Game_System::OnSeReady(FileRequestResult* result, int volume, int tempo, bo
 
 	Audio().SE_Play(path, volume, tempo);
 }
+
+bool Game_System::IsSoundStopFilename(RPG::Sound sound) {
```
carstene1ns: This is IMO not really needed, as the file list is cached on startup.

src/scene_file.cpp
```diff
@@ -111,7 +111,9 @@ void Scene_File::Start() {
 
 	index = latest_slot;
 
+	refresh = true;
 	Update();
+	refresh = false;
 	Refresh();
```
carstene1ns: I do not really understand why we need the double refresh here? Can this not be shuffled around?

src/scene_map.cpp
```diff
@@ -250,6 +253,7 @@ void Scene_Map::FinishTeleportPlayer() {
 void Scene_Map::CallBattle() {
 	Main_Data::game_data.system.before_battle_music = Game_System::GetCurrentBGM();
 	Game_System::SePlay(Game_System::GetSystemSE(Game_System::SFX_BeginBattle));
+	Game_System::BgmPlay(Game_System::GetSystemBGM(Game_System::BGM_Battle));
```
carstene1ns: Why is this needed? `Scene_Battle::Start()` will start the music too.

src/game_system.h
```diff
@@ -246,6 +247,9 @@ namespace Game_System {
 
 	void OnBgmReady(FileRequestResult* result);
 	void OnSeReady(FileRequestResult* result, int volume, int tempo, bool stop_sounds);
+
+	bool IsSoundStopFilename(RPG::Sound sound);
+	const RPG::Sound* GetFirstSound(RPG::Animation *animation);
 }
```
Ghabry: Imo this whole commit is overcomplecating the problem.
The only thing you want to do is: Play the first SE of the animation, otherwise nothing.

So why don't you add a new function void ``SePlay(const RPG::Animation& animation)`` which simply delegates the work to the other existing SePlay. Then you can also remove IsSoundStopFilename

```
/**
 * Plays the first sound effect of an animation if it has one.
```

src/scene_actortarget.cpp
```diff
@@ -121,7 +121,9 @@ void Scene_ActorTarget::UpdateSkill() {
 			return;
 		}
 		if (Main_Data::game_party->UseSkill(id, actor, target_window->GetActor())) {
-			Game_System::SePlay(Game_System::GetSystemSE(Game_System::SFX_UseItem));
+			const RPG::Sound* sound = Game_System::GetFirstSound(ReaderUtil::GetElement(Data::animations, ReaderUtil::GetElement(Data::skills, id)->animation_id));
+			if (sound != nullptr)
+				Game_System::SePlay(*sound);
```
Ghabry: This needs sanity checks because this is external data which is untrusted.
Look at other places where ReaderUtil::GetElement is used and add similiar Output::Warning handling.
At least for the animation because the skill is already sanitized in UseSkill()

src/scene_title.h
```diff
@@ -103,6 +102,9 @@ class Scene_Title : public Scene {
 	/** Contains the state of continue button. */
 	bool continue_enabled;
 
+	/** If it returns to the scene from Scene_Load, check this to do a quick transition */
+	bool quick_transition;
+
```
Ghabry: I really don't like this commit.
It is very invasive considering the only purpose is a minor problem which is restarting the BGM when going from load to title or did I miss something else?

src/scene_file.cpp
Ghabry: I also think this looks more like a hack then a solution

src/scene_map.cpp
Ghabry: Because the battle music already starts when the screen flash appears on the map.
The call in Start() is still needed for the BattleTest.

src/scene_title.h
Albeleon: It's not only the BGM, it's not cleaning the cache of all the data and restart the Scene_Title from the beginning, and also, make sure that the TransitionIn from Title->Load to Title is a simple quick one, instead of the slow FadeIn. The only way I found for Scene_Title to detect this case was using a new "bool" (we cannot reuse "restart_title_cache").

src/scene_actortarget.cpp
Albeleon: GetFirstSound() already checks if the animation is "nullptr" and returns a "nullptr" sound in that case. But if you still prefer that I make a sanity check for animation, then I'll do it.

src/game_interpreter.cpp
```diff
@@ -1331,6 +1331,25 @@ bool Game_Interpreter::CommandChangeEquipment(RPG::EventCommand const& com) { //
...
+				Main_Data::game_party->AddItem(item_id, 1);
+			}
+
+			if (actor->HasTwoWeapons() && slot == RPG::Item::Type_weapon) {
+				RPG::Item* new_equipment = ReaderUtil::GetElement(Data::items, item_id);
+				RPG::Item* equipment1 = ReaderUtil::GetElement(Data::items, actor->GetWeaponId());
+				RPG::Item* equipment2 = ReaderUtil::GetElement(Data::items, actor->GetShieldId());
+
+				if (equipment1 && !equipment2 && !equipment1->two_handed && !new_equipment->two_handed) {
+					actor->ChangeEquipment(slot + 1, item_id);
```
Ghabry: Is there a way to do all this logic in actor->ChangeEquipment? In the previous code the only thing the interpreter did was delegating the work to the actor class.
Also these "return true" are wrong because you are in a for-loop, use "continue".

src/game_system.h
Albeleon: This doesn't solve the problem, which is "if it's a sound that is not (OFF) or similars, stop looking at more sound effect animations", and there's no way to know by executing SePlay if the sound was (OFF) or not. Remember that every effect executes the sound "(OFF)". That's why I need some kind of check. And I didn't want to use the default IsStopFilename because that searches for the files and all that, and that could be not efficient, specially in a function that is looping until it find a valid sound.

src/scene_map.cpp
Albeleon: That was a mistake on my part, I forgot to take it away when putting the commit, my bad. I'll update it and take it away.

src/game_interpreter.cpp
Albeleon: Oops, the "return -> continue" is right, I'll correct it

If we do this in actor->ChangeEquipment, then it will affect too if we change weapons from the menu.
i.e. I'm in the Equipment menu, my character with two weapons have only the first one equipped, and I just want to change that one, that's it. If I put that inside ChangeEquipment, then instead of replacing the old weapon for the new one, it will just insert the new one in the second slot and let the first one as it is.
This is a feature only the Event Command does.

src/game_system.h
Ghabry: The filenames are cached so calling the FileFinder hundreds of times is fine but I didn't consider how finding the first SE works, so have to rethink this.

Ghabry: well I would prefer a SePlay(const RPG::Animation& animation) which uses the normal "IsStopFilename" func that already exists and then delegates the work to SePlay when it's not a StopFilename.
This will fail for a small corner case in emscripten (SE start/ends with ( / ) ) but I don't see a good way to solve this for this.

src/scene_actortarget.cpp
Ghabry: yes pls do a check for a valid animation beforehand imo functions where NULL is just an invalid argument the argument type should be reference & and not pointer \*.

src/game_system.cpp
```diff
@@ -187,6 +187,17 @@ void Game_System::SePlay(RPG::Sound const& se, bool stop_sounds) {
 	request->Start();
 }
 
+void Game_System::SePlay(const RPG::Animation &animation) {
+	std::string path;
+	std::vector<RPG::AnimationTiming>::const_iterator it = animation.timings.begin();
+	for (; it != animation.timings.end(); ++it) {
+		if (!isStopFilename(it->se.name, FileFinder::FindSound, path) && !path.empty()) {
```
Ghabry: this can be replaced with a for(const auto& anim)-loop

The path.empty check is not necessary, isStopFilename already ensures non-empty paths when returning true

src/window_targetstatus.h
```diff
@@ -46,13 +46,15 @@ class Window_TargetStatus : public Window_Base {
 	 * @param id ID of item/skill.
 	 * @param is_item true if ID for an item, otherwise for a skill.
```
Ghabry: needs @param for actor_index

src/game_actor.cpp
```diff
@@ -254,12 +260,27 @@ void Game_Actor::ChangeEquipment(int equip_type, int item_id) {
 	if (item_id != 0) {
 		Main_Data::game_party->RemoveItem(item_id, 1);
 	}
+
+	const RPG::Item* item = ReaderUtil::GetElement(Data::items, GetWeaponId());
+	const RPG::Item* item2 = ReaderUtil::GetElement(Data::items, GetShieldId());
```
Ghabry: Could you add a comment here that this unequips the other item when a two-handed item is equipped?

src/game_actor.h
```diff
@@ -423,6 +423,13 @@ class Game_Actor : public Game_Battler {
 	 */
 	const std::vector<int16_t>& GetWholeEquipment() const;
 
+	/**
+	 * Checks if the actor has a specific item equipped.
+	 *
```
carstene1ns: `@param equip_id` missing

src/game_player.cpp
```diff
@@ -716,14 +716,16 @@ void Game_Player::BeginMove() {
...
+			if (!terrain->on_damage_se || (terrain->on_damage_se && (terrain->damage > 0))) {
+				Game_System::SePlay(terrain->footstep);
+			}
+			if (terrain->damage > 0) {
+				for (auto hero : Main_Data::game_party->GetActors()) {
+					if (!hero->PreventsTerrainDamage()) {
+						red_flash = true;
+						hero->ChangeHp(-std::max<int>(0, std::min<int>(terrain->damage, hero->GetHp() - 1)));
+					}
 				}
```
Ghabry: This fix made me curious if the other vehicles play "footstep sounds": They do. Which means footstep sound is a bad naming in the editor \*g\*

src/sprite_timer.cpp
```diff
@@ -118,16 +128,6 @@ void Sprite_Timer::Update() {
...
-		SetY(SCREEN_TARGET_HEIGHT / 3 * 2 - 20);
-	}
-	else if (Game_Message::visible && Game_Message::GetRealPosition() == 0) {
-		SetY(SCREEN_TARGET_HEIGHT - 20);
-	}
-	else {
-		SetY(4);
-	}
-
 	SetZ(Priority_Timer);
```
Ghabry: I don't like the solution but great that you found this bug. This affects everything when returning from battle, e.g. "Change Vehicle Location" in a battle will show the vehicle at the wrong position during fade in.
The correct way to fix this is the same one that solved the same issue for Teleports:

Add ``spriteset->Update();`` to the end of Continue().

Ghabry:
(for everything else I give my approval)

PR #1401, opened 2018-07-14
Event Commands fixes

Albeleon:


src/game_character.cpp
```diff
@@ -766,7 +766,7 @@ void Game_Character::BeginJump(const RPG::MoveRoute* current_route, int* current
 	max_stop_count = (GetMoveFrequency() > 7) ? 0 : pow(2.0, 9 - GetMoveFrequency());
 	move_failed = false;
 
-	if (IsContinuous()) {
+	if (IsMiddleFrameAvailable()) {
```
Ghabry: The same bug is ``case RPG::MoveCommand::Code::stop_animation:``

src/game_actor.h
```diff
@@ -237,6 +237,13 @@ class Game_Actor : public Game_Battler {
 	 */
 	int GetSpriteIndex() const;
 
+	/**
+	 * Gets actor character sprite index.
+	 *
+	 * @return character sprite index.
+	 */
```
Ghabry: bogus doc comment

Ghabry:
Notes to self (not fully reviewed yet):

- BeginJump calculating the right direction at the end: Test
- When movement finishes, frequency wait is 0: Verify with event tracer
- Not stopping Loop+Skippable last movement if move_failed: Test
- Create Sprite Loop Duplicate graphics with the right Z position: ToDo
- Avoid a centered sprite's shake while slow scroll/movement: o.O
- Inmediately start call command when called by an EventCommand: ToDo
- When a battle, teleport or menu (by input) is called, turn MiddleFrame: ToDo

src/game_interpreter_map.cpp
```diff
@@ -258,6 +258,7 @@ bool Game_Interpreter_Map::CommandEnemyEncounter(RPG::EventCommand const& com) {
 
 	Game_Temp::battle_result = Game_Temp::BattleVictory;
 	Game_Temp::battle_calling = true;
+	Game_Temp::inmediate_call = true;
```
Ghabry: Typo: immediate

src/game_character.cpp
```diff
@@ -174,6 +174,36 @@ int Game_Character::GetScreenZ() const {
...
+	}
+
+	if (!ignore_loop && IsJumping()) {
+		int jump_height = (remaining_step > SCREEN_TILE_WIDTH / 2 ? SCREEN_TILE_WIDTH - remaining_step : remaining_step) / 8;
+		y -= (jump_height < 5 ? jump_height * 2 * TILE_SIZE : jump_height < 13 ? jump_height + 4 * TILE_SIZE : 16 * TILE_SIZE);
+	}
+
+	return y;
+}
+
```
Ghabry: There must be a way to do this with the normal GetScreenX/Y funcs without inventing a "Real" function. By inventing this we almost reach PHP function quality.

src/sprite_airshipshadow.cpp
```diff
@@ -52,6 +53,55 @@ void Sprite_AirshipShadow::RecreateShadow() {
...
+			airship->SetY(airship->GetY() - Game_Map::GetHeight() * TILE_SIZE * (y < src_rect.height ? 1 : -1));
+		}
+
+	}
+}
+
+void Sprite_AirshipShadow::DestroyClones() {
+	clone_x.reset();
+	clone_y.reset();
+	clone_xy.reset();
```
Ghabry: This SpriteClone approach is fine but is there a way to manage the clone creation in Sprite_AirshipShadow/Sprite_Character or in Spriteset Map? When you do it in Spriteset Map you could add them to the character_sprites array which means there Update function is called.
I mean you only need to create them once when the Sprite is created and then they basicly stay until the parent sprite is removed (on map change).
The 3 clones could manage there state automatically (e.g. when the sprite is not at a corner they are just Visible(false)).

src/sprite_clone.cpp
```diff
@@ -0,0 +1,62 @@
...
+	if (reference == nullptr)
+		return;
+	int old_x = reference->GetX();
+	reference->SetX(x);
+	int old_y = reference->GetY();
+	reference->SetY(y);;
+
+	reference->Draw();
+	reference->SetX(old_x);
+	reference->SetY(old_y);
```
Ghabry: Please move logic into Update(). When you have it in Draw() this will called less when it lags or not often enough when fast forward is on.

src/sprite_clone.h
```diff
@@ -0,0 +1,47 @@
...
+// Headers
+#include "sprite.h"
+
+/**
+ * DrawableClone class.
+ */
+class SpriteClone : public Sprite {
+public:
+	SpriteClone(Sprite* drawable, int nx, int ny, int nz);
+	~SpriteClone() override;
```
Ghabry: found a bug: In the sprite.h file change ``~Sprite() override;`` to ``virtual ~Sprite() override;`` otherwise the objects are not correctly deleted.

src/sprite_clone.h
```diff
@@ -0,0 +1,47 @@
...
+
+	void Draw() override;
+	int GetX() const;
+	int GetY() const;
+	int GetZ() const override;
+
+	DrawableType GetType() const override;
+
+private:
+	Sprite* reference;
```
Ghabry: SpriteRef

src/sprite_clone.cpp
```diff
@@ -0,0 +1,62 @@
...
+// Headers
+#include "graphics.h"
+#include "sprite_clone.h"
+#include "output.h"
+
+// Constructor
+SpriteClone::SpriteClone(Sprite* nreference, int nx, int ny, int nz) : reference(nreference), type(TypeClone), x(nx), y(ny), z(nz) {
+	if (reference == nullptr)
+		Output::Warning("Drawable reference null, it won't draw anything.");
+}
```
Ghabry: imo you can pass the shared ptr of the Sprite around (typedef'd as SpriteRef). Raw pointers always have to risk to be dangling and crash.
The reference == nullptr warning-check can be removed imo.

src/sprite_clone.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+// Destructor
+SpriteClone::~SpriteClone() {
+	Graphics::RemoveDrawable(this);
```
Ghabry: not needed, ~Sprite does it

src/game_character.cpp
Ghabry: Oh and could you provide a Test Project for this? Is not clear to me what this is fixing.

Ghabry:
...and that one.

Needs cross-check with #1439 (and maybe others) if some of the issues are also solved there.

Already four conflicts, I will hate myself in a week.

src/game_character.cpp
fmatthew5876: > Oh and could you provide a Test Project for this? Is not clear to me what this is fixing.

@Ghabry , see #1458

Ghabry:
only rebased for now

Ghabry:
Will also do the "battle strategy" here and split this in easier digestable PRs

fmatthew5876:
Looks like some of these are fixed or partially fixed in other PRs

carstene1ns:
yeah, this need a careful rebase :/

Ghabry:
I'm taking over the "CreateClones" change, I also found a much simpler way to do this.

src/game_actor.cpp
```diff
@@ -627,6 +627,10 @@ int Game_Actor::GetSpriteIndex() const {
 	return GetData().sprite_id;
 }
 
+int Game_Actor::GetSpriteFlags() const {
```
Ghabry: > Fix graphic transparency, refreshing and exact amounts. 

this is resolved already

carstene1ns:
Some things still need to be moved, e.g fix for #1113 (2afb65fca2b).

Ghabry:
Closing this in favor of 2 new PRs I create now.

@fmatthew5876 
Is https://github.com/EasyRPG/Player/pull/1401/commits/deeab4e017e5c37b6dfb5843ac6366f9e27dc34d already in?

Also I'm unsure about how much https://github.com/EasyRPG/Player/pull/1401/commits/d994f4c5d1f3dcaa79370b954a1a1c953ed390c0 is correct.

The rest is cherry-picked or already resolved.

fmatthew5876:
> @fmatthew5876
> Is [deeab4e](https://github.com/EasyRPG/Player/commit/deeab4e017e5c37b6dfb5843ac6366f9e27dc34d) already in?

I haven't done anything related to that. It also needs a deeper study. There are 4 actions that need to all be tested against each other.

1. Open menu from event
2. Battle from event
3. Open menu from `ESC` : `PartyLocation::menu_calling`
4. Battle from random encounter : `PartyLocation::encounter_calling`

I'm not sure if (1) and (2) can even happen same frame with multiple parallel events?
Does (2) cancel (3)?
Does (4) cancel (3)?

fmatthew5876:
Regarding the middle frame stuff, that also needs a deeper study to see whether its a feature in itself or merely a symptom of an interpreter timing issue.

If you're not going to address it in your new PRs right now please make a github issue so we can track it and fix later.

PR #1406, opened 2018-07-25
Window handling changes (zoom + fullscreen)

carstene1ns:


Ghabry:
Only tested SDL2 which works fine.
Did you make any SDL1 tests? Can this slowdown Wii build (by accidently enabling zoom for it again)

carstene1ns:
I have tested sdl1 under linux only. However, should not affect Wii at all.

Ghabry:
Under Windows there is now a minor artifact on startup:
For a second it shows a fullscreen black window with a white 640x480 window in the middle, after that second it renders the Logo Scene in fullscreen.
Will see if this can be improved.

Ghabry:
Untested but because this is likely an artifact due to a non redrawn Window it could be probably fixed via

if (fullscreen) {
  ToggleFullscreen();
  UpdateDisplay();
}

in the startup code

Ghabry:
Adding "RenderClear & RenderPresent" before SetLogicalRenderSize reduces a visual glitch on Windows. Otherwise is fine for me.

Fixes the MacOSX problem.

PR #1407, opened 2018-07-29
(hopefully) Support frameskip and framelimiting for emscripten

Ghabry:
Can't test the frame limiting code because I have no 120 Hz display, the artificial test (commenting out Sleep() and running locally in a busy loop) works.

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -248,12 +248,13 @@ void Player::Update(bool update_scene) {
 	static const double framerate_interval = 1000.0 / Graphics::GetDefaultFps();
 	next_frame = start_time + framerate_interval;
 
-	double cur_time = DisplayUi->GetTicks();
+	double cur_time = (double)DisplayUi->GetTicks();
 	if (cur_time < start_time) {
 		// Ensure this function is only called 60 times per second
 		// Main purpose is for emscripten where the calls per second equal the display refresh rate
 		return;
 	}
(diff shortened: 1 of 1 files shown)


fdelapena:
The [emscripten main loop test](https://github.com/kripken/emscripten/blob/950ebc6a19c66ef964f13deea1c59a2cf7955d76/tests/emscripten_main_loop.cpp) also mentions a 120Hz monitor may double FPS.

In that test there are interesting functions apparently not tried so far, like `emscripten_get_now`, `emscripten_get_main_loop_timing` and `emscripten_set_main_loop_timing`. The documentation says emscripten_set_main_loop_timing should be called after after `emscripten_set_main_loop` because emscripten_set_main_loop one overrides the value.

carstene1ns:
Need to remove the debug code here and watch out for possible conflicts with the libretro PR.

fdelapena:
Jenkins: test this please

Ghabry:
A promising solution for this is the new asynchify feature of the new LLVM emscripten backend.

https://emscripten.org/docs/porting/asyncify.html

By adding everything until Player::Update to ASYNCIFY_WHITELIST (is not that many, Player::Update is always close to the bottom of the stack) it should be possible to sleep normally.
(the old solution for this was the Emterpreter, but this was a custom bytecode interpreter that was very slow and threw errors when SDL events fired)

Ghabry:
closing this, the new Asynchifier (see #1917) works perfectly

PR #1415, opened 2018-09-10
MorphOS and AmigaOS4 port

BSzili:
I created a separate Makefile for simplicity, and also included an OS4-style icon, which is supported by all the Amiga-NG flavors.


src/utils.h
```diff
@@ -92,6 +92,7 @@ namespace Utils {
 	 */
 	std::string EncodeUTF(const std::u32string& str);
 
+#ifndef __amigaos4__
```
Ghabry: no wstring support?

src/sdl_ui.cpp
```diff
@@ -419,6 +419,90 @@ bool SdlUi::ShowCursor(bool flag) {
 void SdlUi::Blit2X(Bitmap const& src, SDL_Surface* dst_surf) {
 	if (SDL_MUSTLOCK(dst_surf)) SDL_LockSurface(dst_surf);
 
+#if defined(__MORPHOS__) || defined(__amigaos4__)
+	// Quick & dirty big endian 2x zoom blitter
```
Ghabry: May I ask what the problem with the normal Blit2x code is?

src/game_character.h
```diff
@@ -23,6 +23,10 @@
 #include "color.h"
 #include "rpg_moveroute.h"
 
+#ifdef __MORPHOS__
+#undef Wait
```
Ghabry: Where is all that namespace pollution coming from?
Is there a way to solve this in EasyRPG or is the system library the problem?

Ghabry:
Thanks for your PR.

We plan to get rid of all these Makefiles someday and replace everything with autotools or CMake.
Therefore this will probably need work again later in a few months.

src/game_character.h
BSzili: It's coming from a system header for exec.library. Shared library functions are defined as macros, and they often end up included by C library headers, causing namespace clashes like this.

src/sdl_ui.cpp
BSzili: It's slow, I barely got 20-25 fps on a Mac Mini G4 with it. With this one I'm getting a relatively stable 50 fps on the same machine. It's also fast enough to get 50 fps in WinUAE in the 2x zoom mode.

src/utils.h
BSzili: Not yet, and I didn't feel like hacking around, because these functions are only used on Windows.

BSzili:
Well, that's my luck. I guess I'll wait to see which one will be kept, and then I'll try to get it to build with that one.

src/sdl_ui.cpp
20kdc: https://github.com/EasyRPG/Player/blob/4bebc3d6dbbf0e297e04aa8188835a8fb49c66d5/src/bitmap.cpp#L1081-L1094
...could be why (arbitrary matrix transform? theoretically fastpathable but at some point the fastpath checks outweigh the benefits)
There doesn't seem to be anything in pixman to do such stretching - at best one could copy columns twice, then rows within the target buffer starting from the bottom (so it'll only overwrite already handled stuff)

Ghabry: Is the system running with 16 bit color scheme?
For some reason the blitting code of Pixman is extremely slow (at least twice) when operating on 16bit (also on little encian). Is probably our fault because this makes no real sense but nobody really understands the undocumented pixman API ^^'.
On the Wii we use a custom SDL build which forces a 32bit surface and this gives a gigantic speedup, maybe is the same problem here.

20kdc: Ghabry: For that particular case: Is it possible the source image surfaces are 32-bit but the destinations are 16 (necessitating conversion)?

Ghabry: In my tests even 16-16 was slow according to the profiler.
There are many non-obvious snail-paths in pixman, e.g. when you use ARGB instead of RGBA the code is suddenly 50% slower (could remember this wrong but at least some 32-bit pixel formats are much slower than others).

builds/amiga/Makefile
```diff
@@ -0,0 +1,72 @@
...
+SDL_CFLAGS = -I$(TOOLCHAIN_DIR)/include/SDL
+PIXMAN_LIBS = -lpixman-1
+PNG_LIBS = -lpng -lz
+endif
+
+ICU_LIBS = -licuuc -licui18n -licudata 
+
+LIBS = -llcf -lexpat \
+	$(PIXMAN_LIBS) $(PNG_LIBS) \
+	-lfreetype -lfreetype -lmpg123 \
```
carstene1ns: overlinking freetype here ;-)

src/main_data.cpp
```diff
@@ -83,7 +83,7 @@ void Main_Data::Init() {
 			// first set to current directory for all platforms
 			project_path = ".";
 
-#ifdef GEKKO
+#if defined(GEKKO) || defined(__MORPHOS__) || defined(__amigaos4__)
 			// Working directory not correctly handled under Wii
```
carstene1ns: This part should probably be rewritten, as is not Wii-specific anymore.

src/sdl_ui.cpp
carstene1ns: Not really happy with this either, however, is SDL1.2 only → does not really matter.
Maybe we could even use this under Wii in the meanwhile...

Ghabry:
> Not really happy with this either, however, is SDL1.2 only → does not really matter.
Maybe we could even use this under Wii in the meanwhile...

(don't have a reply button under your review comment o.O)

The Wii version does not use the 2x blitter because we set the screen resolution to 320x240 and let SDL do the magic.

carstene1ns:
I do not think there is a special case for Wii resolution.

src/sdl_ui.cpp
carstene1ns: Also, maybe we should hide this stuff deeply in some `src/platform/amiga/port.cpp`... :P

PR #1419, opened 2018-09-22
RM2k Battle fixes - turns and action switches

fmatthew5876:
This PR fixes some bugs in the rm2k battle system. It resolves some of the issues in #985 and makes hh3 playable past the intro (before the battle was impossible to lose).

I don't believe any of these code changes touch 2k3.


src/game_battlealgorithm.cpp
```diff
@@ -1640,5 +1647,6 @@ bool Game_BattleAlgorithm::NoMove::Execute() {
 
 void Game_BattleAlgorithm::NoMove::Apply() {
 	// no-op
```
carstene1ns: Not a no-op anymore ;-)

carstene1ns:
First commit does not really make sense to me, why does the turn counter change midway between player and enemy turn? Maybe is a general off-by-one instead?

src/scene_battle.cpp
```diff
@@ -427,7 +427,7 @@ void Scene_Battle::CreateEnemyAction(Game_Enemy* enemy, const RPG::EnemyAction*
 				enemy->GetBattleAlgorithm()->SetSwitchEnable(action->switch_on_id);
 			}
 			if (action->switch_off) {
-				enemy->GetBattleAlgorithm()->SetSwitchEnable(action->switch_off_id);
+				enemy->GetBattleAlgorithm()->SetSwitchDisable(action->switch_off_id);
```
carstene1ns: good find!

fmatthew5876:
> 
> 
> First commit does not really make sense to me, why does the turn counter change midway between player and enemy turn? Maybe is a general off-by-one instead?

I think the increment is in the wrong place. Before it changed after selecting all actions but before actually running them. The result is that enemy actions that are conditional on turns that are off by one but troop pages that are conditional on turns are correct.

I can look at moving it forward to before selecting any actions. Just need to be careful not to mess up the special "turn 0" which can be used to trigger troop pages at the beginning of the battle.

Ghabry:
Ah, now it also makes sense to me. Right, the EnemyAction that cares about the turn count was one turn behind. Good find!

Imo you can keep the way as it is now, I'm afraid any hugher change will mess up some interpreter handling :)

carstene1ns:
jenkins:ok to test!

PR #1420, opened 2018-09-22
Fix issue 1417

Ghabry:
also gitignore updates


Ghabry:
Uff, I wondered why .cia is already somehow on the ignore list but I couldn't find it in the .gitignore file xD

PR #1422, opened 2018-09-25
Add citro[23]d based 3dsui

### LINKED ISSUE #1398: 3DS: Replace sf2dlib
Because the devkitPro people love to break NIH stuff even if it makes your life easier we need a new graphic library for 3ds. When compiling sf2dlib currently you get tons of "API is deprecated" warnings in citro3d and some shader variables are not found anymore, which means it's not working anymore. 

Since sf2dlib was "ported" to citro3d it never really worked fine actually. The Player always had a strange flickering triangle visible during transitions e.g.

Guess we have to switch to the badly documented DKP invention "citro3d" because we decided in a previous evaluation that "citro2d" is not suitable for what we are doing because it focuses on sprite handling.

Ghabry:
Fix #1398 

Most of the code is by carstene1ns, I just did tons of performance tuning.


carstene1ns:
Open question is if we want to do the top screen stuff with pure c3d instead (the only thing c2d is doing for us here is centering the image and abstract away some texture stuff).

Also, maybe is not a good idea to ship the whole midi patchset with every 3dsx. Needs some hardware test, but i think shipping a minimal soundfont (GMGSx.sf2 - [info](https://lists.nongnu.org/archive/html/fluid-dev/2014-03/msg00006.html), [download](https://www.dropbox.com/s/qxdvoxxcexsvn43/GMGSx.sf2?dl=0)) and fluidlite might be better for 3DS.

Ghabry:
Good point. Do you have any working sample code for rendering the texture with citro3d?

I fully agree that this huge patch file is terrible and a smaller file would help almost every port. Also bonus points for "It sounded like standard Microsoft GM.dls from Roland". But we need upstream fluidlite support first ;).

src/3ds_ui.cpp
```diff
@@ -204,17 +232,19 @@ void CtrUi::ProcessEvents() {
 		Input::Keys::N0, Input::Keys::N0, Input::Keys::PERIOD, Input::Keys::ADD
 	};
 
+	// reset state
+	touch_state =
+		touch_state == 1 ? 2 :
+		touch_state == 2 ? 2 : 0;
```
carstene1ns: This does not really make sense to me.

src/3ds_ui.cpp
```diff
@@ -223,51 +253,106 @@ void CtrUi::ProcessEvents() {
...
+	// Using RGB8 as output format is faster and improves framerate ¯\_(ツ)_/¯
+	const u32 flags = (GX_TRANSFER_FLIP_VERT(0) |
+		GX_TRANSFER_OUT_TILED(1) |
+		GX_TRANSFER_RAW_COPY(0) |
+		GX_TRANSFER_IN_FORMAT(GX_TRANSFER_FMT_RGBA8) |
+		GX_TRANSFER_OUT_FORMAT(GX_TRANSFER_FMT_RGB8) |
+		GX_TRANSFER_SCALING(GX_TRANSFER_SCALE_NO));
+
+	// Doing this after FrameBegin corrupts the output, probably because this
+	// is asynchronous and FrameBegin will block until it finishes
```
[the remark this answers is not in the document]
Ghabry: I have no idea what C3D_FRAME_SYNC is supposed to do, I always got corrupted output when using it. It will wait for the GPU when you pass "0", But when you pass "0" it will wait until the GPU is ready, as the C3D_FRAME_NONBLOCK doc says: "Return false instead of waiting if the GPU is busy", therefore "0" should mean: Wait.

src/3ds_ui.cpp
Ghabry: This is for improving the frame rate by only redrawing the bottom screen when a touch happened (1) or one frame after a touch is finished (2). I can improve the naming.

carstene1ns: I understand what it should do, just the logic seems off here.
```bash
1 → 2
2 → 2 ?
0 → 0 ?
```

Ghabry: When 2 it stays in 2 because the reset to 0 is in "UpdateDisplay". I can't reset it to 0 here because when there is a frameskip "UpdateDisplay" will miss the chance for a redraw.

Ghabry: but maybe I can replace it with an if, because there is only one case where the value changes ;)

Ghabry: replaced that nonsense with an if ^^', don't be asleep and code I guess ;)

carstene1ns:
I am thinking of letting this state be 0.5.4. Further optimizations can come later with 0.6.0 (or when something breaks again)

Issue #886, opened 2016-05-05
Timer 1 in wrong position & the Event Touch is detected before


Br4ssman:
Hi there,
Look, **Timer 1** generally appears on the top-left screen corner. In this case, above the **Status Time** box.
About the **Event Touch**: it can be seen better when Alex touch the event from sides in EasyRPG Player.

RPG_RT
https://youtu.be/NqvziU82gSk

EasyRPG Player
https://youtu.be/56chY_mM-1A

By the way, in this mini-game test is more difficult to catch items in RPG_RT than EasyRPG

Testcase: [Collisions.zip](https://github.com/EasyRPG/Player/files/1220324/Collisions.zip)

---

Fixed:
- [x] timer1 position
- [ ] event touch

(Edited by @carstene1ns to make the status of this bug visible in the overview)

Zegeri:
The first issue is fixed here 3ffdd70b8ab. About the event touch issue, it might be hard to tell from the video. I've tried to isolate the issue in [this test project](https://dl.dropboxusercontent.com/u/30277047/TestTouch.7z) and I got the opposite impression. It looks like it would be more difficult to catch items in EasyRPG than in RPG_RT. Run that test project both with RPG_RT and Player: in RPG_RT, the hero collides with the moving event, while in Player doesn't.

carstene1ns:
Added the testcase from IRC link.

Ghabry:
The problem is that touch events only count when the move is finished:

![collision](https://user-images.githubusercontent.com/1331889/45061560-48aa1a80-b0a5-11e8-9bd1-0c9745dd4a67.gif)

Ghabry:
But one corner case: When the event is not moving the touch event triggers immediately (on the 1st frame of movement onto the tile)

Ghabry:
Moving the collision patch to 0.5.5/0.6 because this is a total mess:

New test case:

[Collisions+priorities.zip](https://github.com/EasyRPG/Player/files/2408629/Collisions%2Bpriorities.zip)

Going to prepare a table :/

Ghabry:
Because this is event collision related I will postbone this to 0.6.x and ping @fmatthew5876 as this is probably related to #1628

Br4ssman:
The events routes and collisions need to be fixed, but in SCH there is a new clue: the jewels, power-ups and power-down, are charsets with graphic from tilesets (16x16), Is it possible that EasyRPG use all event graphics like charsets graphics?

[...]

PR #1423, opened 2018-09-27
Fix Kanga Punch (1413), Shitigadianzu (1412) and minor stuff

Ghabry:
I discovered the bogus level sanity check when loading one of the savegames bundled with Shitigadianzu (they are > max_level).

The GetReal -> GetSprite change is the first step for my fix of the collision tests in #886 but the function renaming alone doesn't hurt imo.

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -1345,9 +1345,12 @@ void Game_Actor::RemoveInvalidData() {
 	// through database editing. Hopefully no game uses negative levels.
 	if (GetLevel() == 0) {
 		Output::Debug("Actor %d: Special handling for level 0", GetId());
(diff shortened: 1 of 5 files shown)


Br4ssman:
There's a new problem in this mini-game

In RPG_RT, the kangaroo can "hook" the enemy when he blocking hits for a limited time, in EasyRPG this hook is until the enemy is K.O.

By the way, this is not intentional, it is the normal behavior of events in RPG_RT.

**How to reproduce it**: especially with Happy Hippo (it's the more defensive enemy), launch punches when the enemy is moving until the enemy stay blocking hits during a time, then, the enemy will move in other direction and going forward the kangaroo again.

PR #1430, opened 2018-10-08
Support debug menu in battles

fmatthew5876:
Depends on: #1436 

This enables the F9 Debug menu during battles. Save is greyed out and disabled but Load is usable.

Testing this exposed a bug in `Scene_Load`, where the scene stack was not properly properly cleared. Any use of load from the debug menu in Player would keep pushing more stuff onto the stack.

In my case, because the `Scene_Battle` was never popped off the stack, the game loaded with `Game_Temp::battle_running == true` when on the map, which caused lots of errors to display on the screen.

The fix is to always unwind up to the title screen.

This was tested against a liblcf built from https://github.com/EasyRPG/liblcf/pull/242

### DIFF
--- a/src/scene_battle.cpp
+++ b/src/scene_battle.cpp
@@ -38,6 +38,7 @@
 #include "scene_battle_rpg2k.h"
 #include "scene_battle_rpg2k3.h"
 #include "scene_gameover.h"
+#include "scene_debug.h"
 
 Scene_Battle::Scene_Battle() :
 	actor_index(0),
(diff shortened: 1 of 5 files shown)


carstene1ns:
This looks really good! Currently in feature freeze for 0.5.4, but this looks like a good candidate for merging right after. Can the `Scene_Load` bug be exposed currently (e.g. from debug menu)?

fmatthew5876:
> Can the `Scene_Load` bug be exposed currently (e.g. from debug menu)?

The `Scene_Load` bug exists currently in master. If you load a game from the debug mode and then pause in a debugger you'll see the old map and menu from the previous game still on the Scene stack. If you keep loading games It'll just keep pushing more scenes on the stack and never removing the old ones.

Do you want the `Scene_Load` fix as a separate PR for 5.4?

carstene1ns:
jenkins: test this please. (service had a hiccup)

Ghabry:
merged as part of #1432

PR #1431, opened 2018-10-09
Additional features for DebugMenu

fmatthew5876:
This adds new features to the debug menu. This makes it less compatible with the original debug menu in rm2/3, but I don't think that's important compared to the benefit here.

* Uses a 2 level menu system instead of left - right scrolling through everything. Makes it faster to go through variables and switches.
* Add support for changing party gold
* Add support for changing party items
* Add support for starting battles
* Use instant transition, just like with call menu for faster response

debug
items

### DIFF
--- a/src/scene_debug.cpp
+++ b/src/scene_debug.cpp
@@ -26,9 +26,11 @@
 #include "game_switches.h"
 #include "game_map.h"
 #include "game_system.h"
+#include "game_battle.h"
 #include "scene_debug.h"
 #include "scene_load.h"
 #include "scene_save.h"
+#include "scene_map.h"
 #include "player.h"
 #include "window_command.h"
 #include "window_varlist.h"
@@ -36,6 +38,7 @@
 #include "bitmap.h"
 #include "game_temp.h"
 #include "game_party.h"
+#include "game_player.h"
 #include "data.h"
 
 Scene_Debug::Scene_Debug() {
@@ -82,6 +85,10 @@ void Scene_Debug::Update() {
 					prev_item_range_index = range_index;
 					prev_item_range_page = range_page;
 					range_index = 5;
+				} else if (mode == eBattle) {
+					prev_troop_range_index = range_index;
+					prev_troop_range_page = range_page;
+					range_index = 6;
 				} else {
 					range_index = 0;
 				}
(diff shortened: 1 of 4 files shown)


Ghabry:
I'm in general fine with any scene changes which improve the usability. As the game has no control over them and nothing can really break.
The new debug scene looks really cool, I'm going to play around with it :+1:

src/scene_debug.cpp
```diff
@@ -102,78 +205,105 @@ void Scene_Debug::Update() {
...
 		Game_Map::SetNeedRefresh(Game_Map::Refresh_All);
-	} else if (range_window->GetActive() &&  Input::IsTriggered(Input::RIGHT)) {
-		if (current_var_type != TypeGeneral) {
-			range_page++;
-		}
-		if (current_var_type == TypeSwitch && !Game_Switches.IsValid(range_page*100+1)) {
-			range_page = 0;
-			current_var_type++;
-		} else if (current_var_type == TypeInt && !Game_Variables.IsValid(range_page*100+1)) {
+	} else if (range_window->GetActive() && Input::IsTriggered(Input::RIGHT)) {
```
Ghabry: you could use IsRepeated here (Same for Input::LEFT)

Ghabry:
I don't know if this bug was there before but the game Unterwegs in Düsterburg has exactly 5000 switches but the debug scene renders one extra page from 5001 - 5100 which is completely empty.

In battle test the wrong sprites were displayed for me and after a GameOver I was returned to the map (but I guess this is intentional?)

Ghabry:
okay I return to the map with 0 HP, maybe the party could be fully recovered after failing a battle.

btw, a full recover option in the debug menu would be useful

Ghabry:
Forget the note about "Wrong sprites displayed", everything is fine here. What do you think about keeping the Debug scene open when starting a battle? This way you can directly do another battle.

Sorry, just dropping ideas here :)

fmatthew5876:
> Forget the note about "Wrong sprites displayed", everything is fine here. What do you think about keeping the Debug scene open when starting a battle? This way you can directly do another battle.

No only open, but with the same battle selected so you can go again. I think it's a great idea, but I don't know the implications of pushing a battle scene over a debug scene. There could be some hidden assumptions in battles that they always precede a map. Would need to test this carefully.

Also at least for HH3 (don't know other games), that would break as HH3 depends on an auto start common event to trigger after every battle to clean custom things up properly. We have to visit the map for that.

There's also a question of whether you want to configure first strike, fully heal the party, etc..

This isn't even close to being finished. These are just a few things I implemented to start and also to solicit feedback for the approach before going too deep into the rabbit hole.

Skills, add party members, etc.. etc..

I have a custom debug system in hh3 that just uses a number input. It has about 30 or so different actions but its clunky to use. I'll continue porting more of that into Player in the future.

If you're happy with whats there as is this can be merged. I'll make new PRs for continued improvements to the debug menu.

src/scene_debug.cpp
fmatthew5876: Thats in #1439

Ghabry:
Right, I didn't think about all the corner cases when not running the battle via the map.

Well this is the debug scene, therefore you get my approval for this, even if not everything works 100% (and there were already many bugs in the original debug scene).

fmatthew5876:
> Right, I didn't think about all the corner cases when not running the battle via the map.

It's definitely something that we should aim for. Later on when I'm actively using this debug menu for development I'll be able to test a lot more and add improvements.

carstene1ns:
rel. #948

PR #1435, opened 2018-10-11
A debug mode validation of scene stack correctness

fmatthew5876:


carstene1ns:
jenkins: test this please. (service had a hiccup)

src/scene.cpp
```diff
@@ -173,6 +177,7 @@ void Scene::Pop() {
 	for (size_t i = 0; i < instances.size(); ++i) {
```
Ghabry: you can remove this old /* */ block

src/scene.cpp
```diff
@@ -212,3 +219,27 @@ void Scene::DrawBackground() {
 Graphics::State &Scene::GetGraphicsState() {
 	return state;
 }
+
+
+inline void Scene::DebugValidate(const char* caller) {
+	if (instances.size() <= 1) {
+		//Scene of size 1 happens before graphics stack is up. Which can
+		//cause the following logs to crash.
```
Ghabry: minor style-nitpick: no space between // and text

src/scene.cpp
fmatthew5876: Done

src/scene.cpp
fmatthew5876: Fixed

src/scene.cpp
```diff
@@ -26,6 +26,13 @@
 #include "audio.h"
 #include "transition.h"
 
+#ifndef NDEBUG
```
carstene1ns: Not sure if this is always defined when using autotools/cmake. It should for release builds (disabling asserts), but would need a quick test.

Ghabry: for me it wouldnt ever matter if this is enabled in general, this check doesn't really hurt.

fmatthew5876: We use `assert()` in other places. So if `NDEBUG` is misconfigured in some cases, we already have a bug.

carstene1ns: `NDEBUG` is not enabled for autotools builds, but cmake enables it for release builds.
Our other Makefiles and MSVC solution also explicitly do this.
So is fine to have right now and maybe change/fix the autotools configuration later.

src/scene.cpp
```diff
@@ -26,6 +26,13 @@
...
+#define DEBUG_VALIDATE() Scene::DebugValidate(__PRETTY_FUNCTION__)
+#else
+#define DEBUG_VALIDATE() do {} while(0)
```
carstene1ns: This is actually not needed. Empty strings will get silently discarded.

PR #1436, opened 2018-10-11
Pop all scenes when loading game

fmatthew5876:
Fixes bugs when loading game debug menu. The stack would keep the old
maps and game state. This breaks when you try to load from battle.


carstene1ns:
Thank you!

Issue #1418, opened 2018-09-20
hh3 - Various sound and other issues.


fmatthew5876:
#### Name of the game:
Heaven and Hell Episode 3

#### Player platform:
Windows

All of these are observed in rm2k mode. I have no idea if rm2k3 is different. All of these were seen in the continuous binary.

- [x] - Calling "Open Menu Screen" command from an event plays the Select sound effect. This causes the sound to play twice from hh3's CMS.
- [x] - When you use a skill in the field that enables a switch, it's configured sound effect will not play.
- [x] - In battle when you select an action it plays the Select SE, but then when you select which enemy to target it does not play the Select SE as it does in RPG_RT.
- [x] - I have a spell which can only cure states that all end after battle. In RPG_RT, this spell is greyed out in the field menu. In EasyRPG, the spell is usable like any other healing spell.
- [x] - Characters with 2 handed weapons can also equip shields.
- [x] - When casting a healing spell from the menu, the MP amount doesn't render correctly. In this example, the cost should be 2. (see photo)
- [ ] - Mp3 music will stop playing when you take focus away from the game window. Midi music continues playing. The midi song will play to the end and then stop. It won't loop until you click back into the game window. This happens on Windows.

![bad_mp](https://user-images.githubusercontent.com/1198466/45789925-56889f80-bc50-11e8-984e-6573d3493292.png)

Ghabry:
Hello, thanks for the report.

We have multiple fixes in the queue that are not in master yet.
Could you please retest with this binary: https://ci.easyrpg.org/view/Pull%20Requests/job/player-win32-pr/1697/
At least equipment and the bogus number rendering should be fixed.

and anything battle related with this binary:
https://ci.easyrpg.org/view/Pull%20Requests/job/player-win32-pr/1698/

> Mp3 music will stop playing when you take focus away from the game window. Midi music continues playing. The midi song will play to the end and then stop. It won't loop until you click back into the game window.

The problem here is that we use the MIDI feature provided by Windows, therefore we have basicly zero control over it.
This is no issue on other platforms because we use our own midi player there but under Windows we would get complaints because "it sounds different" :/.
I'm not 100% sure about this but when you download this repository: https://github.com/ghabry/timidity_gus and put it in C:\timidity our internal Midi player will be used.

fmatthew5876:
> Could you please retest with this binary: https://ci.easyrpg.org/view/Pull%20Requests/job/player-win32-pr/1697/

[...]

PR #1439, opened 2018-10-12
Various engine logic fixes

fmatthew5876:
These are RPT_RT compatibility fixes.

Fixes some of the issues in #1418

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -210,17 +210,15 @@ std::string Game_BattleAlgorithm::AlgorithmBase::GetAttackFailureMessage(const s
 }
 
 std::string Game_BattleAlgorithm::AlgorithmBase::GetHpSpRecoveredMessage(int value, const std::string& points) const {
-	std::stringstream ss;
-
 	if (Player::IsRPG2kE()) {
-		ss << value;
 		return Utils::ReplacePlaceholders(
 			Data::terms.hp_recovery,
 			{'S', 'V', 'U'},
-			{GetTarget()->GetName(), ss.str(), points}
+			{GetTarget()->GetName(), std::to_string(value), points}
 		);
 	}
 	else {
+		std::stringstream ss;
 		std::string particle, particle2, space = "";
 
 		ss << GetTarget()->GetName();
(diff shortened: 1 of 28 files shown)


fmatthew5876:
I've tested with `RPG_RT.exe` and confirmed that battles count doesn't increment until the battle ends.

fdelapena:
Jenkins: test this please

src/game_party.h
```diff
@@ -311,9 +331,56 @@ class Game_Party : public Game_Party_Base {
...
+	++data().defeats;
+}
+
+inline int Game_Party::GetRunCount() const {
+	return data().escapes;
+}
+
+inline void Game_Party::IncRunCount() {
+	++data().escapes;
+}
```
Ghabry: Educational question:
I wonder about this inline void getter/setter. Are there any C++ references that suggest this? Is there an advantage in putting this in the header instead of cpp file?

fmatthew5876: Inlining is why C++ is so fast and yet enables high level abstractions. Functions which are very small should be inlined for performance. These `Game_Party` methods either just return or increment an integer. There is no benefit to paying the overhead of an entire function call for this.

The real power of this is when things are combined together because it enables the compiler optimizer to do smart things. For example, imagine you had a function `foo(int p)` and the first thing int does is check if `p == 0`. Imagine your calling code sets `int p = 3` and then calls `foo(p)`. If `foo()` is inlined, the compiler can see that `p` can never be zero, and therefore it could remove the entire `if (p == 0) { ... }` block from the code.

A good example is `unique_ptr`. All of `unique_ptr` is inlined, so in most cases if you write a piece of code using `unique_ptr` the compiler will cleanup the memory for you but when the optimizer hits you code the `unique_ptr` gets completely removed and it's as if you wrote `new` and `delete` by hand in all the right places.

src/game_battlealgorithm.cpp
```diff
@@ -348,17 +343,16 @@ std::string Game_BattleAlgorithm::AlgorithmBase::GetParameterChangeMessage(bool
...
 		Data::terms.parameter_decrease;
-	std::stringstream ss;
 
 	if (Player::IsRPG2kE()) {
-		ss << value;
 		return Utils::ReplacePlaceholders(
 			message,
 			{'S', 'V', 'U'},
-			{GetTarget()->GetName(), ss.str(), points}
+			{GetTarget()->GetName(), std::to_string(value), points}
```
Ghabry: We had problems with missing std::to_string on some platforms years ago (we have Utils::ToString as an alternative) but I will recheck if this is still problematic.

src/game_map.h
```diff
@@ -142,7 +142,7 @@ namespace Game_Map {
 	 * @param self Character to move.
```
Ghabry: force_through is missing in the doc comment

src/game_interpreter.cpp
```diff
@@ -176,6 +176,9 @@ void Game_Interpreter::Update() {
 			if (Main_Data::game_player->IsBoardingOrUnboarding())
 				break;
 
+			if (Main_Data::game_player->InVehicle() && Main_Data::game_player->GetVehicle()->IsAscendingOrDescending())
+				break;
```
Ghabry: Another timing issue with vehicles, we had an almost similiar one in Witch's Heart before: https://github.com/EasyRPG/Player/commit/9659baeac5e4c6be9046fac9597f4b0aa83fc0d2

src/game_map.h
fmatthew5876: Fixed.

fmatthew5876:
rebased to master

carstene1ns:
>The std::to_string issues we had before appear to be solved

Seems we should get rid of this in utils then.

fmatthew5876:
> Seems we should get rid of this in utils then.

Added a commit to remove `Utils::ToString`

Issue #1203, opened 2017-06-14
save_count 0x5B chunk not validated when a save is loaded


Ghabry:
Upon loading RPG_RT compares 0x5B of the current map with the save_count of the savegame and when it mismatches the events in the savegame are ignored and the map reloads instead.

In RPG Maker 2003 English Edition this was changed to 0x5A (or 0x5C, not 100% sure, needs a recheck).

CherryDT:
It's unrelated, it basically just aborts the foreground event (not parallel events)

That save count thing though basically resets the events completely, as if you had left and reentered the map.

CherryDT:
Actually it's two things:

The comparison of SavePartyLocation.map_save_count with Map.save_count resets the events on the map (their scripts and their map properties like their position, if I remember correctly)

The comparison of SavePartyLocation.database_save_count with System.save_count resets common events (not sure if it resets anything else)

(SaveSystem.save_count on the other hand is unrelated and just counts how often the user saved the game as far as I know)

PR #1441, opened 2018-10-13
Obey Map and Database save_count when loading games

fmatthew5876:
Resolves: #1203 

I tested this by creating a new project in Steam RM2k.

A map event with this code:
```
Text: before save
OpenSaveMenu
Text: After Save
```

Another map event which walks around randomly.

And a parallel common event with this code:
```
Text: Start
Loop
  Wait 4.0 seconds
EndLoop
```

This also correctly writes the database_save_count to the LSD file. Before it looks like it was never written to save games.

### DIFF
--- a/src/game_map.cpp
+++ b/src/game_map.cpp
@@ -132,7 +132,7 @@ void Game_Map::Quit() {
 }
 
 void Game_Map::Setup(int _id) {
-	SetupCommon(_id);
+	SetupCommon(_id, false);
 
 	Parallax::ClearChangedBG();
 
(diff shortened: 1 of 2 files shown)


fmatthew5876:
Added support for Map Chunk 0x5A.

I tested this with the following scenarios:
* rm2k steam game
* rm2k3 steam game converted from rm2k steam
* rm2k3 steam game converted from rm2k steam where the map was converted (saved) but the ldb was not. In this case rm2k3 doesn't reset the event states when it should and neither does player.

Now this depends on: https://github.com/EasyRPG/liblcf/pull/263

fdelapena:
Jenkins: test this please

src/game_map.cpp
```diff
@@ -258,9 +258,26 @@ void Game_Map::SetupCommon(int _id) {
 	pan_wait = false;
 	pan_speed = 0;
 
-	// Make RPG_RT happy
-	// Otherwise current event not resumed after loading
-	location.map_save_count = map->save_count;
+	auto map_save_count = map->save_count;
+	if (Data::system.ldb_id == 2003 && map->save_count_2k3e > 0) {
```
Ghabry: instead of ``Data::system.ldb_id == 2003`` use Player::IsRPG2k3

PR #1373, opened 2018-06-14
Fixes of Battle issues ( #1371 )

Albeleon:
I will be posting one fix for one issue per commit.

"1.2: RM2000: Animations executed in a battle event should not wait until the animation is over when the option "wait to end" is selected."

Solution: Eliminate in Game_Interpreter_Battle::ExecuteCommand and "if" that checks whether the animation is finished no matter if the option "wait to finish" is clicked or not. This difference is already treated and differentiated by itself, and this condition is just a left-over code from previous function structure that isn't needed anymore.

PR #1447, opened 2018-10-14
Albeleons's Battle fixes 1 of N

fmatthew5876:
Here is the first batch of rebased fixes from #1373

All of these are pretty simple changes so I'm making a first PR with these. They are not all in the same order as originally presented. I'm trying to pull out the ones that are simple and don't require a lot of merging so they can be quickly reviewed and merged into master.

Changes made:
 * 2.7 - Using `Player::escape_char` instead of `"\\"`
 * ~~4.1 - `Battle_BGM` was moved earlier so don't play it twice.~~
 * 2.15 - Added back `const`
 * 2.4 - Enhanced `Game_Actor::GetSkillName()` with skill selection logic.

These alone fix some issues from #1418 and #985

### DIFF
--- a/src/scene_battle_rpg2k.cpp
+++ b/src/scene_battle_rpg2k.cpp
@@ -593,12 +593,12 @@ void Scene_Battle_Rpg2k::ProcessInput() {
 			--actor_index;
 			SelectPreviousActor();
 			break;
-		case State_SelectEnemyTarget:
 		case State_SelectItem:
 		case State_SelectSkill:
 			Game_System::SePlay(Game_System::GetSystemSE(Game_System::SFX_Cancel));
 			SetState(State_SelectCommand);
 			break;
+		case State_SelectEnemyTarget:
 		case State_SelectAllyTarget:
 			Game_System::SePlay(Game_System::GetSystemSE(Game_System::SFX_Cancel));
 			SetState(previous_state);
--- a/src/scene_battle_rpg2k.cpp
+++ b/src/scene_battle_rpg2k.cpp
@@ -102,9 +102,7 @@ void Scene_Battle_Rpg2k::CreateBattleCommandWindow() {
 }
 
 void Scene_Battle_Rpg2k::RefreshCommandWindow() {
-	std::string skill_name = active_actor->GetSkillName();
-	command_window->SetItemText(1,
-		skill_name.empty() ? Data::terms.command_skill : skill_name);
+	command_window->SetItemText(1, active_actor->GetSkillName());
 }
 
 void Scene_Battle_Rpg2k::SetState(Scene_Battle::State new_state) {
--- a/src/scene_battle_rpg2k.cpp
+++ b/src/scene_battle_rpg2k.cpp
@@ -678,9 +678,6 @@ void Scene_Battle_Rpg2k::Escape() {
 		escape_success = escape_alg.Execute();
 		escape_alg.Apply();
 
-		if (escape_success)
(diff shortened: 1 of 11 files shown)


fmatthew5876:
I've reviewed these commits and done some basic testing. They all look good to me :+1:

Albeleon:
2.7: I remember there were some complications with the "\\" and "Player::escapeChar" after trying to change the codification from Occidental to Japanese that made me go back to "\\" again. I'll put a pin on this and recheck comparing both again.

4,1: The reason why PlayBGM was kept twice is because the battle music should sound also when you only do a battle test (without opening the map), and it doesn't reach Scene_Map::CallBattle in that case. I considered there was no problem with a double call because the function PlayBGM doesn't take much workaround if the song is the same with the same conditions. But this is up to discussion.

2.15: I don't remember why I took out "const", maybe it was during tests trying things to make it work and I just kept it, or there was a definition problem? I don't remember ^^u, I hope it was just a mistake for my part, but just in case later I'll recheck that commit to see whether there is a difference or not.

fmatthew5876:
@Albeleon why don't you take these over again? You have a much better understanding of your changes than me.

I can close out my PRs

Ghabry:
needs a rebase

fmatthew5876:
Rebased

src/scene_battle_rpg2k3.cpp
```diff
@@ -980,7 +980,7 @@ bool Scene_Battle_Rpg2k3::CheckWin() {
 		std::vector<int> drops;
 		Main_Data::game_enemyparty->GenerateDrops(drops);
 
-		Game_Message::texts.push_back(Data::terms.victory);
+		Game_Message::texts.push_back(Data::terms.victory + "\\|");
```
Ghabry: > 2.7: RM2000 + RM2003: The Victory message should have pauses after...

I'm almost sure that these "\\"-codes must be replaced with Player::escape_symbol, otherwise they won't be recognized in japanese or korean games.

src/scene_battle_rpg2k.cpp
```diff
@@ -102,9 +102,7 @@ void Scene_Battle_Rpg2k::CreateBattleCommandWindow() {
 }
 
 void Scene_Battle_Rpg2k::RefreshCommandWindow() {
-	std::string skill_name = active_actor->GetSkillName();
-	command_window->SetItemText(1,
```
Ghabry: > 2.4: RM2000 (not tested in RM2003): If you have the "special skill name...

Style question: As this is only used in combination with GetSkill what do you think about moving this "IsSkillRenamed" logic directly in "GetSkillName"?

fmatthew5876: I think that makes sense. From Player perspective there is only 1 skill name. The renaming stuff is a liblcf detail.

Ghabry:
> Cool had some problems with it. Also we talked about it in irc a while ago. I haven't tested it.

I have no idea what the original problem was but replacing "\\" with Player::escape\_symbol works for me.

You can use encoding "ibm-943_P130-1999" (Japanese with Yen as backslash) for testing.

fmatthew5876:
I've addressed the comments and updated the list of changes in the top comment of this PR.

Ghabry:
Jenkins: Test this please

fmatthew5876:
rebased onto current master

fmatthew5876:
Possible this and #1394 can merged this weekend? Of all the PRs out there Sormats are the most important in my opinion. 

They are large and likely will introduce regressions. We will need time after merging everything to test the new master.

PR #1448, opened 2018-10-14
Albeleon's Battle Fixes 2 of N

### LINKED ISSUE #1102: "monster died" message doesn't show in battle
[Video of the issue](https://www.youtube.com/watch?v=Sui9w49H_wM). This was made using the elf file of [build 120](https://ci.easyrpg.org/job/player-3ds/120/). Compare to [the original](http://i.imgur.com/KfbgX3D.png).

Seems that the message "\<monster name\> died" isn't being printed when a monster dies.

edit: whoops, forgot \< tags need escaping

fmatthew5876:
More of the smaller commits from #1373.

Depends on #1447

Changes:
* 2.26 Renamed `weapon` local variable to `item`.
* 3.14 Minor refactor for code clarity

Dropped:
* 1.4 - Since Ghabry wanted to have a discussion about it. Let's move it to a later PR and not hold up this one.
* 4.15 - Algorithm is not correct as per: #1454 - New fix for this in #1455

---

Fixes #1102.

### DIFF
--- a/src/game_battle.cpp
+++ b/src/game_battle.cpp
@@ -153,7 +153,26 @@ bool Game_Battle::CheckWin() {
 }
 
 bool Game_Battle::CheckLose() {
-	return !Main_Data::game_party->IsAnyActive();
+	if (!Main_Data::game_party->IsAnyActive())
+		return true;
+
+	// If there are active characters, but all of them are in a state with Restriction "Do Nothing" and 0% recovery probability, it's game over
+	// Physical recovery doesn't matter in this case
+	int character_number = 0;
+	std::vector<Game_Battler*> actors;
+
+	Main_Data::game_party->GetActiveBattlers(actors);
+	for (auto actor : actors) {
+		for (auto id_state : actor->GetInflictedStates()) {
+			RPG::State *state = ReaderUtil::GetElement(Data::states, id_state);
+			if (state->restriction == RPG::State::Restriction_do_nothing && state->auto_release_prob == 0) {
+				++character_number;
+				break;
+			}
+		}
+	}
+
+	return character_number == actors.size();
 }
 
 Spriteset_Battle& Game_Battle::GetSpriteset() {
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -1021,8 +1021,12 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 
 			this->success = (GetAffectedHp() != -1 && !IsAbsorb()) || (GetAffectedHp() > 0 && IsAbsorb()) || GetAffectedSp() > 0 || GetAffectedAttack() > 0
 				|| GetAffectedDefense() > 0 || GetAffectedSpirit() > 0 || GetAffectedAgility() > 0;
+
+			if (IsAbsorb() && !success)
+				return this->success;
 		}
 
+		// Conditions:
 		for (int i = 0; i < (int) skill.state_effects.size(); i++) {
 			if (!skill.state_effects[i])
 				continue;
@@ -1042,12 +1046,6 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 	}
 	else {
 		assert(false && "Unsupported skill type");
-	}
-
-	if (IsAbsorb() && sp != -1) {
-		if (GetTarget()->GetSp() == 0) {
-			this->success = false;
-		}
 	}
 
 	return this->success;
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -1127,6 +1127,10 @@ const RPG::Sound* Game_BattleAlgorithm::Skill::GetStartSe() const {
 	}
 }
 
+const RPG::Sound* Game_BattleAlgorithm::Skill::GetResultSe() const {
+	return !success && skill.failure_message != 3 ? NULL : AlgorithmBase::GetResultSe();
+}
+
 void Game_BattleAlgorithm::Skill::GetResultMessages(std::vector<std::string>& out) const {
 	if (!success) {
 		switch (skill.failure_message) {
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -1060,11 +1060,11 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 void Game_BattleAlgorithm::Skill::Apply() {
 	AlgorithmBase::Apply();
 
-	if (item) {
-		Main_Data::game_party->ConsumeItemUse(item->ID);
-	}
-	else {
-		if (first_attack) {
+	if (IsFirstAttack()) {
+		if (item) {
+			Main_Data::game_party->ConsumeItemUse(item->ID);
+		}
+		else {
 			source->ChangeSp(-source->CalculateSkillCost(skill.ID));
 		}
 	}
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -421,6 +421,12 @@ void Game_BattleAlgorithm::AlgorithmBase::GetResultMessages(std::vector<std::str
 				}
 			}
 		}
+
+		// If enemy is killed, it ends here
+		if (killed_by_attack_damage) {
+			out.push_back(GetDeathMessage());
+			return;
+		}
 	}
 
 	if (GetAffectedSp() != -1) {
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -641,7 +641,7 @@ bool Game_BattleAlgorithm::AlgorithmBase::IsTargetValid() const {
 		return false;
 	}
 
-	return (!GetTarget()->IsDead());
+	return GetTarget()->Exists();
 }
 
 int Game_BattleAlgorithm::AlgorithmBase::GetSourceAnimationState() const {
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -849,7 +849,7 @@ void Game_BattleAlgorithm::Normal::Apply() {
 		Game_Actor* src = static_cast<Game_Actor*>(source);
 		const RPG::Item* weapon = ReaderUtil::GetElement(Data::items, src->GetWeaponId());
 		if (weapon) {
-			source->ChangeSp(-weapon->sp_cost / src->GetSpCostModifier());
+			source->ChangeSp(std::ceil(-weapon->sp_cost / (float) src->GetSpCostModifier()));
 		}
 	}
 }
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -1312,7 +1312,8 @@ int Game_BattleAlgorithm::Item::GetSourceAnimationState() const {
 }
 
 void Game_BattleAlgorithm::Item::GetResultMessages(std::vector<std::string>& out) const {
-	AlgorithmBase::GetResultMessages(out);
+	if (success)
+		AlgorithmBase::GetResultMessages(out);
 }
 
 const RPG::Sound* Game_BattleAlgorithm::Item::GetStartSe() const {
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -458,11 +458,11 @@ void Game_Battler::RemoveBattleStates() {
 	// If death is non-permanent change HP to 1
 	if (IsDead() &&
 		non_permanent(1)) {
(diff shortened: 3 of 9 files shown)


src/game_battlealgorithm.cpp
```diff
@@ -428,6 +428,12 @@ void Game_BattleAlgorithm::AlgorithmBase::GetResultMessages(std::vector<std::str
 				}
 			}
 		}
+
+		// If enemy is killed, it ends here
+		if (killed_by_attack_damage) {
+			out.push_back(GetDeathMessage());
+			return;
+		}
```
fmatthew5876: This one causes a regression. when you kill the last enemy they don't fade out.

src/game_battle.cpp
```diff
@@ -81,6 +81,12 @@ void Game_Battle::Init() {
 		return false;
 	});
 
+	if (Game_Temp::battle_check_surprise_attack) {
+		Game_Temp::battle_first_strike = Utils::GetRandomNumber(1, 100) <=
+			(Main_Data::game_party->GetAverageAgility() > Main_Data::game_enemyparty->GetAverageAgility() ? 5 : 3);
+		//TODO: IMPLEMENT 2k3 DIFFERENT SURPRISE MODES
```
fmatthew5876: @CherryDT

Do you agree with 5% and 3% here?

fmatthew5876: Also maybe we should just move this logic to `Game_Map` when random encounters are triggered and set `battle_first_strike` there accordingly. This would avoid adding yet another `Game_Temp` global.

fmatthew5876:
I've reviewed and did some basic testing of these.

Other than the ones with comments on the commits, LGTM :+1:

src/scene_battle_rpg2k.cpp
```diff
@@ -56,6 +57,11 @@ void Scene_Battle_Rpg2k::Update() {
 	}
 
 	Scene_Battle::Update();
+
+	// If an event that changed status finishes without displaying a message window,
+	// we need this so it can update automatically the status_window
+	if (Game_Battle::GetInterpreter().finished_not_updated)
+		status_window->Refresh();
 }
```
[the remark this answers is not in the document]
fmatthew5876: I don't know the interpreter code well enough. Maybe there is a better way to achieve this.

src/game_battler.cpp
```diff
@@ -456,10 +456,11 @@ void Game_Battler::RemoveBattleStates() {
 	if (IsDead() &&
 		non_permanent(1)) {
 		ChangeHp(1);
+		states[1] = 0;
```
Albeleon: When using RemoveState or ReaderUtil::GetElement(), the index starts at 1. When using states[] directly in an array, the index starts at 0. This means in this case you are disabling the second state (that isn't Dead), not the first one. You can solve this by either:
states[0] = 0;
RemoveState(1);

src/game_battle.cpp
CherryDT: I spent a lot of time to try and figure out how all these battle types behave... this is the result: https://gist.github.com/CherryDT/8984e216dca3c0cd3c84f80b57006c16

Regarding your change though - I didn't find anything related to that, and I'm a bit unsure what it is doing...

src/game_battler.cpp
fmatthew5876: Fixed, thanks!

src/scene_battle_rpg2k.cpp
```diff
@@ -362,6 +362,10 @@ bool Scene_Battle_Rpg2k::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBase
 		}
 	}
 
+	if (Input::IsPressed(Input::CANCEL)) {
+		return false;
+	}
```
Ghabry: > "4.11: RM2000: If you keep CANCEL pressed while in State_Battle, the BattleActionState doesn't advance, it keeps paused."
Didn't even knew about this feature, I always smash enter to get it running faster \*g\*

src/game_battlealgorithm.cpp
```diff
@@ -645,7 +645,7 @@ bool Game_BattleAlgorithm::AlgorithmBase::IsTargetValid() const {
 		return false;
 	}
 
-	return (!GetTarget()->IsDead());
+	return (!GetTarget()->IsDead() && !GetTarget()->IsHidden());
```
Ghabry: > Make an "All enemy" attack not affect hidden enemies. 

statement can be simplified to a call to "Exists()"

src/scene_battle_rpg2k.cpp
fmatthew5876: Me too! But I learned from these PRs that you can just hold the key down instead of mashing it.

src/game_battler.cpp
```diff
@@ -459,10 +459,11 @@ void Game_Battler::RemoveBattleStates() {
 	if (IsDead() &&
 		non_permanent(1)) {
 		ChangeHp(1);
```
Ghabry: > 3.14: RM2000 + RM2003: After a battle is over, heal all the states that despite being "Persist after battle", still have more than 0% chance of recovery (turns and physical recovery don't matter)."

```cpp
ChangeHp(1)
states[0] = 0
```

With #1394 merged this will break (won't heal 1 HP anymore). Proposed solution: ``RemoveState(1)``

src/game_battler.cpp
```diff
@@ -696,12 +696,14 @@ std::vector<int16_t> Game_Battler::NextBattleTurn() {
...
 			states[i] += 1;
-
-			if (states[i] >= Data::states[i].hold_turn) {
+			if (states[i] > Data::states[i].hold_turn + 1) {
 				if (Utils::ChanceOf(Data::states[i].auto_release_prob, 100)) {
 					healed_states.push_back(i + 1);
 					RemoveState(i + 1);
 				}
+				else {
+					states[i] -= Data::states[i].hold_turn;
```
Ghabry: > 2.28: RM2000 + RM2003: If a battler is thrown a state again, the tur

Because the code before is
```cpp
states[i] += 1; 
if (states[i] > Data::states[i].hold_turn + 1) {
```

isn't this the same as ``states[i] = 1``?

src/game_battlealgorithm.cpp
```diff
@@ -504,21 +504,21 @@ Game_Battler* Game_BattleAlgorithm::AlgorithmBase::GetTarget() const {
 }
 
 float Game_BattleAlgorithm::AlgorithmBase::GetAttributeMultiplier(const std::vector<bool>& attributes_set) const {
-	float multiplier = 0;
-	int attributes_applied = 0;
+	float physical = -1001, magical = -1001;
```
Ghabry: > "3.16: RM2000: Attribute modifier is the maximum of the percentage da… 

The actual minimum in the editor (for E) is -9999, so this values should be at least -10000. (that you can set E to something negative is btw also totally non-obvious but documented in the help file)

src/game_battlealgorithm.cpp
```diff
@@ -839,7 +842,7 @@ void Game_BattleAlgorithm::Normal::Apply() {
 	AlgorithmBase::Apply();
 
 	source->SetCharged(false);
-	if (source->GetType() == Game_Battler::Type_Ally) {
+	if (source->GetType() == Game_Battler::Type_Ally && success) {
```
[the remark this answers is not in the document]
Ghabry: I tested this: I lose SP when the attack misses. @CherryDT did you fix this in RPG 2003 1.12?

src/game_battler.cpp
fmatthew5876: I don't think so.

`(s > ht + 1)` is equivalent to `(s - ht) > 1`. So `s >= 2` when that write happens.

src/game_battlealgorithm.cpp
Ghabry: Because #1394 copy-pastes this function into Game_Battler I suggest to remove it from here (or make it a thin wrapper)and just call the Game_Battler function. I will clean it up in the other PR afterwards.

fmatthew5876:
Changes:
* 3.14 - Used `RemoveState(1)`
* Make an All enemy... - Use `Exists()`


Dropped:
* 3.16 - Removed
* 2.30 - Moved to #1482
* 3.10 - Moved to #1482

Now depends on #1481

src/game_battler.cpp
Ghabry: I have no idea if RPG2000 does this different but for RPG2003 I question the whole code except ```states[state_id - 1] = std::max<int> (1, states[state_id - 1]); ``` (which prevents turn count rejection when the same state is applied again). I used a Poison with 10 turns & 20% setting and I got it healt after turn 16... (according to this commit it should be again at 20?)

fmatthew5876:
Moved 2.28 to #1482 as well.

Ghabry:
> 2.30 - Moved to #1482 

this commit is still in the list

Ghabry:
Because the questionable commits were wrong I will make a thorough test with some algorithm changes here, maybe can find more that don't match the real behaviour.

fmatthew5876:
> Because the questionable commits were wrong I will make a thorough test with some algorithm changes here, maybe can find more that don't match the real behaviour.

Ok that sounds prudent. Let me know if you don't finish them all. I can split out the ones we know are good into a separate PR so we can merge them and get the rebase mess down.

Ghabry:
Most of them were actually pretty easy to verify, but to be 100% sure I would like to delegate the question "How is the crit_chance of a normal battle attack calculated" to @CherryDT

fmatthew5876:
Extracted 3.17. Lets get these merged and then tonight I get 3, 4, and 5 cleaned up.

src/game_battler.cpp
```diff
@@ -458,11 +458,11 @@ void Game_Battler::RemoveBattleStates() {
...
 	if (IsDead() &&
 		non_permanent(1)) {
-		ChangeHp(1);
+		RemoveState(1);
 	}
 
-	for (size_t i = 0; i < states.size(); ++i) {
-		if (non_permanent(i + 1)) {
+	for (size_t i = 1; i < states.size(); ++i) {
+		if (non_permanent(i + 1) || ReaderUtil::GetElement(Data::states, i + 1)->auto_release_prob == 0) {
```
Ghabry: Oh I just remembered something!
The commit explanation is correct (> 0% chance of recovery will heal after battle) but the code is wrong, change == to > xD

CherryDT:
> Most of them were actually pretty easy to verify, but to be 100% sure I would like to delegate the question "How is the crit_chance of a normal battle attack calculated" to @CherryDT

X = (Actor|Enemy.critical_hit ? 100 / Actor|Enemy.critical_hit_chance : 0
X += max(weapon1's Item.critical_hit, weapon2's Item.critical_hit) [only for actors of course]
X /= 100

If the target is a friend and not a foe (can happen with confusion for example), the crit chance is always zero.

src/game_battler.cpp
fmatthew5876: Fixed

fmatthew5876:
Thanks Cherry, I'll review criticals in full in the other PR.

What is the base crit chance for enemies? 

For enemy, crit is only possible for normal attack and double attack yes?

CherryDT:
For enemies and actors it's the same thing, it's stored in critical_hit/critical_hit_chance in the Actor/Enemy structure in the LDB.

In general, crit is only possible for normal physical attacks (double attack is just two physical attacks one after another, which have their individual damage/crit/miss calculation)

src/game_battlealgorithm.cpp
```diff
@@ -641,7 +641,7 @@ bool Game_BattleAlgorithm::AlgorithmBase::IsTargetValid() const {
 		return false;
 	}
 
-	return (!GetTarget()->IsDead());
+	return GetTarget()->Exists();
```
carstene1ns: Commit message is bogus here.

7006d47b4bc

src/game_battlealgorithm.cpp
```diff
@@ -1127,6 +1127,10 @@ const RPG::Sound* Game_BattleAlgorithm::Skill::GetStartSe() const {
 	}
 }
 
+const RPG::Sound* Game_BattleAlgorithm::Skill::GetResultSe() const {
+	return !success && skill.failure_message != 3 ? NULL : AlgorithmBase::GetResultSe();
```
carstene1ns: This magic number should probably be resolved to some `SkillFailureMessage_Physical` in the future.

src/game_battlealgorithm.cpp
fmatthew5876: Fixed

carstene1ns:
alrighty

PR #1449, opened 2018-10-15
Albeleon's Battle Fixes 3+4 of N

fmatthew5876:
A few larger commits

~~Depends on #1448~~

### DIFF
--- a/src/battle_animation.cpp
+++ b/src/battle_animation.cpp
@@ -31,8 +31,8 @@
 #include "player.h"
 #include "game_temp.h"
 
-BattleAnimation::BattleAnimation(const RPG::Animation& anim) :
-	animation(anim), frame(0), frame_update(false)
+BattleAnimation::BattleAnimation(const RPG::Animation& anim, bool only_sound) :
+	animation(anim), frame(0), frame_update(false), should_only_sound(only_sound)
 {
 	SetZ(Priority_BattleAnimation);
 
@@ -139,6 +139,10 @@ void BattleAnimation::DrawAt(int x, int y) {
 	}
 }
 
+bool BattleAnimation::ShouldOnlySound() const {
+	return should_only_sound;
+}
+
 // FIXME: looks okay, but needs to be measured
 static int flash_length = 12;
 
@@ -155,6 +159,8 @@ void BattleAnimation::RunTimedSfx() {
 void BattleAnimation::ProcessAnimationTiming(const RPG::AnimationTiming& timing) {
 	// Play the SE.
 	Game_System::SePlay(timing.se);
+	if (ShouldOnlySound())
+		return;
 
 	// Flash.
 	if (timing.flash_scope == RPG::AnimationTiming::FlashScope_target) {
@@ -214,6 +220,8 @@ BattleAnimationChara::~BattleAnimationChara() {
 	Graphics::RemoveDrawable(this);
 }
 void BattleAnimationChara::Draw() {
+	if (ShouldOnlySound())
+		return;
 	//If animation is targeted on the screen
 	if (animation.scope == RPG::Animation::Scope_screen) {
 		DrawAt(SCREEN_TARGET_WIDTH / 2, SCREEN_TARGET_HEIGHT / 2);
@@ -231,13 +239,13 @@ bool BattleAnimationChara::ShouldScreenFlash() const { return true; }
 
 /////////
 
-BattleAnimationBattlers::BattleAnimationBattlers(const RPG::Animation& anim, Game_Battler& batt, bool flash) :
-	BattleAnimation(anim), battlers(std::vector<Game_Battler*>(1, &batt)), should_flash(flash)
+BattleAnimationBattlers::BattleAnimationBattlers(const RPG::Animation& anim, Game_Battler& batt, bool flash, bool only_sound) :
+	BattleAnimation(anim, only_sound), battlers(std::vector<Game_Battler*>(1, &batt)), should_flash(flash)
 {
 	Graphics::RegisterDrawable(this);
 }
-BattleAnimationBattlers::BattleAnimationBattlers(const RPG::Animation& anim, const std::vector<Game_Battler*>& batts, bool flash) :
-	BattleAnimation(anim), battlers(batts), should_flash(flash)
+BattleAnimationBattlers::BattleAnimationBattlers(const RPG::Animation& anim, const std::vector<Game_Battler*>& batts, bool flash, bool only_sound) :
+	BattleAnimation(anim, only_sound), battlers(batts), should_flash(flash)
 {
 	Graphics::RegisterDrawable(this);
 }
@@ -245,6 +253,8 @@ BattleAnimationBattlers::~BattleAnimationBattlers() {
 	Graphics::RemoveDrawable(this);
 }
 void BattleAnimationBattlers::Draw() {
+	if (ShouldOnlySound())
+		return;
 	if (animation.scope == RPG::Animation::Scope_screen) {
 		DrawAt(SCREEN_TARGET_WIDTH / 2, SCREEN_TARGET_HEIGHT / 3);
 		return;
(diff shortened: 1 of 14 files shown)


src/scene_battle_rpg2k.cpp
```diff
@@ -107,7 +107,8 @@ void Scene_Battle_Rpg2k::RefreshCommandWindow() {
 
 void Scene_Battle_Rpg2k::SetState(Scene_Battle::State new_state) {
 	previous_state = state;
-	state = new_state;
+	if (new_state != State_SelectEnemyTarget && new_state != State_SelectAllyTarget)
```
Ghabry: > 2.10: RM2000: When a skill or object is used, the skill/object windo… 

What I really dislike here is that this adds a filter to "SetState" which means SetState won't work as expected for all types of states. There must be a way to handle this in a better way.

src/scene_battle.cpp
```diff
@@ -163,17 +158,28 @@ void Scene_Battle::Update() {
...
+			command_window->Update();
+			help_window->Update();
+			item_window->Update();
+			skill_window->Update();
+			target_window->Update();
+
+			ProcessActions();
+			ProcessInput();
+		}
+	} while (changed_state != state && state == State_SelectOption);
```
Ghabry: > 2.1: RM2000: If no hero party is ready to attack and they can't choo

Uhm I need to test around to figure out what this is supposed to do

src/window_base.cpp
```diff
@@ -49,6 +65,26 @@ void Window_Base::Update() {
...
+		SetX(new_position[0]);
+		SetY(new_position[1]);
+	}
+}
+
+void Window_Base::Draw() {
+	if (IsMovementActive()) {
+		SetX(old_position[0] + (new_position[0] - old_position[0]) * current_frame / total_frames);
+		SetY(old_position[1] + (new_position[1] - old_position[1]) * current_frame / total_frames);
+	}
```
Ghabry: > 2.3: RM2000 + RM2003: The battle command window (both left [manual, 

Logic updates can't be done in Draw(), they will be skipped when frame skipping / lag happens or when Fast Forward is used

src/window_base.cpp
```diff
@@ -40,6 +40,22 @@ Window_Base::Window_Base(int x, int y, int width, int height) {
 	SetHeight(height);
 	SetStretch(Game_System::GetMessageStretch() == RPG::System::Stretch_stretch);
 	SetZ(Priority_Window);
+
+	current_frame = -1;
+	total_frames = -2;
```
Ghabry: > 2.3: RM2000 + RM2003: The battle command window (both left [manual, 

I assume the value is "-2" to ensure that ``IsMovementActive`` returns false. Imo they could be both set to "-1" and ``MovementActive`` gets an extra check for ">= 0"

fmatthew5876:
> (Also need to see the window animation in action to further comment on it)
> 
> Is it possible to detach these 4 commits conflict free from the others? (if not, keep it as is)

I tried and I start getting merge conflicts. I don't want to mess it up and get into rebase hell.

What we can do is move out anything questionable from #1448, merge those, and then rebase this onto master. It'll be much easier to work with then.

src/scene_battle_rpg2k.cpp
fmatthew5876: The one causes rebase hell too with the change of indentation,.

Ghabry:
The good news: 50% of the battle commits are in. The bad news: 50% of the battle commits are not in yet. 😅🙈

src/scene_battle.cpp
fmatthew5876: To reproduce the bug, create a troop page on turn 0 that sets a state like sleep to your entire party. When the battle starts, the battle menu will appear for a few frames and then disappear.

fmatthew5876: .. and testing it with this branch.. the bug appears to not be fixed. I think this one needs a refactor. Unless I misinterpreted the commit message?

fmatthew5876:
I've dropped 2.1 and 2.10 in favor of #1497

src/window_base.cpp
fmatthew5876: I moved this logic to `UpdateMovement()`. Seems to work fine so not sure why this was added to `Draw()` before.

src/window_base.cpp
fmatthew5876: Initialized both to 0 and added a check to `IsMovementActive()`

fmatthew5876:
I've addressed Ghabry's review comments. Hopefully this is now ready to be merged.

fmatthew5876:
To reduce rebase mess I've merged the old battle 4 back into this PR.

I've also addressed Ghabry's comment about `GetType()` in 2.11 by adding a virtual. I don't like this solution either, but it's good enough to get through these PR's until we can do a proper refactor.

src/scene_battle.cpp
```diff
@@ -123,7 +123,7 @@ void Scene_Battle::CreateUi() {
 	commands.push_back(Data::terms.battle_fight);
```
Ghabry: (Commit review TEST: commit 1)

src/game_battlealgorithm.cpp
```diff
@@ -174,6 +174,14 @@ bool Game_BattleAlgorithm::AlgorithmBase::IsFirstAttack() const {
 	return first_attack;
 }
```
Ghabry: (COMMIT review test: Comment 2)

src/game_battlealgorithm.cpp
```diff
@@ -1094,20 +1102,45 @@ std::string Game_BattleAlgorithm::Skill::GetStartMessage() const {
...
+std::string Game_BattleAlgorithm::Skill::GetSecondStartMessage() const {
+	if (Player::IsRPG2k()) {
+		if (item && item->using_message == 0) {
+			return "";
+		}
+		if (Player::IsRPG2kE()) {
+			return Utils::ReplacePlaceholders(
+				skill.using_message2,
+				{ 'S', 'O', 'U' },
+				{ GetSource()->GetName(), GetTarget()->GetName(), skill.name }
```
Ghabry: > 2.5: RM2000: If the second line of a Skill is not empty

``(target ? target->GetName() : "???")`` is missing

src/game_battlealgorithm.cpp
```diff
@@ -1094,20 +1102,45 @@ std::string Game_BattleAlgorithm::Skill::GetStartMessage() const {
...
+			return source->GetName() + skill.using_message1;
 		}
 	}
 	else {
 		return skill.name;
 	}
 }
 
+bool Game_BattleAlgorithm::Skill::IsSecondStartMessage() const {
+	return Player::IsRPG2k() && (!item || item->using_message != 0) && !skill.using_message2.empty();
```
Ghabry: > 2.5: RM2000: If the second line of a Skill is not empty

Grammar wise the name should be "HasSecondStartMessage".
This suggestion is maybe minimal more inefficient but the whole function could be removed and the code in battle_rpg2k.cpp replaced with an "empty()" check.

src/battle_animation.cpp
```diff
@@ -31,8 +31,8 @@
 #include "player.h"
```
Ghabry: > 2.11: RM2000: When an enemy uses a technique/ability, the animation,

Approved.
Only this commit message... 6. is e.g. misleading now

In general it is not helpful.
I would completely wipe it and simply mention that skills on allies play the sound of the animation instead of the attack sound.

Ghabry:
the last 2 commits look okay.

src/game_battlealgorithm.cpp
fmatthew5876: Great catch

fmatthew5876:
I've addressed all of the most recent set of review comments.

fmatthew5876:
The 1 frame flicker bug turned out to be simple. So I just pushed a new commit fixing it.

fmatthew5876:
Are there anymore comments? Rebasing 5, 6, and 7 are blocked until this one goes in.

Ghabry:
I would give @carstene1ns a short grace period for feedback as I havn't seen him around since a week.
If theres no response will merge it tomorrow.
I will start to understand Battle 5 in the meanwhile. this is a huge blocker :/

Ghabry:
merging this for now, in case of further feedback can be incorporated in the next PR, we have enough of them...

PR #1450, opened 2018-10-15
Albeleon's Battle Fixes 5 of 7

fmatthew5876:
Just one large commit

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -452,12 +452,13 @@ std::string Game_BattleAlgorithm::AlgorithmBase::GetStateMessage(const std::stri
 	}
 }
 
-void Game_BattleAlgorithm::AlgorithmBase::GetResultMessages(std::vector<std::string>& out) const {
+void Game_BattleAlgorithm::AlgorithmBase::GetResultMessages(std::vector<std::string>& out, std::vector<int>& out_replace) const {
 	if (current_target == targets.end()) {
 		return;
 	}
 
 	if (!success) {
+		out_replace.push_back(0);
 		out.push_back(GetAttackFailureMessage(Data::terms.dodge));
 		return;
 	}
@@ -466,14 +467,17 @@ void Game_BattleAlgorithm::AlgorithmBase::GetResultMessages(std::vector<std::str
 
 		if (IsPositive()) {
 			if (!GetTarget()->IsDead()) {
+				out_replace.push_back(0);
 				out.push_back(GetHpSpRecoveredMessage(GetAffectedHp(), Data::terms.health_points));
 			}
 		}
 		else {
 			if (critical_hit) {
+				out_replace.push_back(0);
 				out.push_back(GetCriticalHitMessage());
 			}
 
+			out_replace.push_back(0);
 			if (GetAffectedHp() == 0) {
 				out.push_back(GetUndamagedMessage());
 			}
@@ -489,6 +493,7 @@ void Game_BattleAlgorithm::AlgorithmBase::GetResultMessages(std::vector<std::str
 
 		// If enemy is killed, it ends here
 		if (killed_by_attack_damage) {
+			out_replace.push_back(1);
 			out.push_back(GetDeathMessage());
 			return;
 		}
@@ -501,6 +506,7 @@ void Game_BattleAlgorithm::AlgorithmBase::GetResultMessages(std::vector<std::str
 	}
 
 	if (GetAffectedSp() != -1) {
+		out_replace.push_back(0);
 		if (IsPositive()) {
 			out.push_back(GetHpSpRecoveredMessage(GetAffectedSp(), Data::terms.spirit_points));
 		}
@@ -515,18 +521,22 @@ void Game_BattleAlgorithm::AlgorithmBase::GetResultMessages(std::vector<std::str
 	}
 
 	if (GetAffectedAttack() > 0) {
+		out_replace.push_back(0);
 		out.push_back(GetParameterChangeMessage(IsPositive(), GetAffectedAttack(), Data::terms.attack));
 	}
 
 	if (GetAffectedDefense() > 0) {
+		out_replace.push_back(0);
 		out.push_back(GetParameterChangeMessage(IsPositive(), GetAffectedDefense(), Data::terms.defense));
 	}
 
 	if (GetAffectedSpirit() > 0) {
+		out_replace.push_back(0);
 		out.push_back(GetParameterChangeMessage(IsPositive(), GetAffectedSpirit(), Data::terms.spirit));
 	}
 
 	if (GetAffectedAgility() > 0) {
+		out_replace.push_back(0);
 		out.push_back(GetParameterChangeMessage(IsPositive(), GetAffectedAgility(), Data::terms.agility));
 	}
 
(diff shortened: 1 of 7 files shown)


Ghabry:
Which one of these commits is the large one?

fmatthew5876:
My rebase messed up. It's the one called big commit

Ghabry:
uff, this one looks painful to review. Doing this another day 😅

src/scene_battle_rpg2k.cpp
```diff
@@ -359,10 +359,16 @@ void Scene_Battle_Rpg2k::ProcessActions() {
 }
 
 bool Scene_Battle_Rpg2k::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBase* action) {
+	/**
+     * Order of execution of BattleActionState:
+	 * ConditionHeal > Execute > Apply > (ResultPop > ResultPush) > Death > Finished.
+	 **/
```
Ghabry: just doing a quick scroll ^^. This shouldn't be a doc-comment but two normal // comments.
But maybe before updating anything here I need to understand the new state stuff.

fmatthew5876:
I've reintroduced `AlgorithmBase::GetType()`. Its needed in big commit and other commits in 6 and 7.  Also we use this idiom in other places so it makes sense here.

This implementation of `GetType()` avoids virtual calls, so switching on type in some cases may faster than calling a virtual function call.

We ultimately probably want to use variant, but for now we have this hybrid approach.

fmatthew5876:
To help with reviewing, these are the bugs supposed to be fixed here:

> 3.1: RM2000: If an attack affects a group, the message display should be:
>· 1st line (+ 2nd line if a skill has one) always stay still.
>· 2nd/3rd line shows the damage results: this line is always replaced when it shows the next enemy damage.
· 3rd/4th line shows, if the enemy is not dead and the skill causes a condition (poison, etc.), it should show that. If it's various conditions, this line should be replaced by each new condition. This line disappears once it goes to the next enemy damage.
· 3rd/4th line shows, if the enemy dies, the message of the enemy dying at the same time the enemy vanishes. This line disappears once it goes to the next enemy damage.

> 2.2: RM2000: If a character recovers themself from a negative condition, the message should be printed and not empty.

>3.2: RM2000: When replacing a line (like in 3.1), that particular line should flash for 1 or 2 frames instead of changing inmediately. The skill line shouldn't flash.

From #1371

fmatthew5876:
I tested the 3.1 behavior by starting with a level 10 Alex and casting `Starlight 1` on 2 slimes

In master I see these messages:
> Alex cast Starlight 1
> Slime too X damage

> Alex cast Starlight 1
> Slime was defeated

> Alex cast Starlight 1
> Slime was defeated
> Same took Y damage

> Alex cast Starlight 1
> Slime was defeated

In this PR and RPG_RT I see

> Alex cast Starlight 1
> Same took X damage
> Slime was defeated

> Alex cast Starlight 1
> Same took Y damage
> Slime was defeated

fmatthew5876:
I tested the 2.2 behavior by setting the 2xSlime battle to give Alex "Shock" status on turn 0


In RPG_RT:
After the enemy names are displayed, I get the message:

> Alex recovered from the Shock!

In master and this PR:

After the enemy names are displayed, I get a screen flick (probably fixed by #1497, not tested) the message:

> Alex recovered from the Shock!

Not sure if this was already fixed or I misinterpreted the bug description?

fmatthew5876:
I tested 2.2 again. This time I changed Slime to always use an attack that inflicts shock 100%. I still see

>Alex recovered from the Shock!

There doesn't appear to be any different between this, master, and RPG_RT in displaying this message.

src/game_battlealgorithm.cpp
```diff
@@ -462,6 +466,7 @@ void Game_BattleAlgorithm::AlgorithmBase::GetResultMessages(std::vector<std::str
 
 		// If enemy is killed, it ends here
 		if (killed_by_attack_damage) {
+			out_replace.push_back(1);
```
fmatthew5876: This out_replace mechanism looks really brittle. Can't we just edit the vector directly?

fmatthew5876:
I've reviewed this. What I really don't like about it is how the battle state machine is driven by the state of the console instead of the other way around.

That being said, I don't have a plan to refactor it. I also wouldn't want to at this stage as that would make dependent 6+7 commits impossible to rebase and have to be re-written.

I guess I can live with it as it is now, assuming it doesn't introduce any regressions.

fmatthew5876:
I've reworked several commits ontop of this in #1451. This refactor seems fine to me for now.

fmatthew5876:
I fixed the doc comment Ghabry had mentioned.

fmatthew5876:
I know it's painful, but we have got to come to a resolution on this one soon. It's the next big blocker. Once this one is finished, the rest of the battle PRs will not take long. I'm already half way through with the remaining ones already.

fmatthew5876:
jenkins: test this please

src/scene_battle_rpg2k.cpp
```diff
@@ -467,52 +517,31 @@ bool Scene_Battle_Rpg2k::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBase
 				Game_System::SePlay(*action->GetStartSe());
 			}
 
-			battle_action_state = BattleActionState_Result;
+			action->Apply();
```
Ghabry: The moving down of the Apply() call should make no difference as the code before doesn't depend on the side-effects of Apply

src/scene_battle_rpg2k.cpp
```diff
@@ -467,52 +517,31 @@ bool Scene_Battle_Rpg2k::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBase
...
 			break;
-		case BattleActionState_Result:
-			battle_action_wait = GetDelayForWindow();
+		case BattleActionState_ResultPop:
+			battle_action_wait = std::min<int>(GetDelayForLine() / 2, battle_action_wait);
 
 			if (battle_result_messages_it != battle_result_messages.end()) {
+				default_result_lines = 1 + (action->HasSecondStartMessage() ? 1 : 0);
+				while (battle_message_window->GetLineCount() > (default_result_lines + *battle_result_order_it))
+					battle_message_window->Pop();
```
Ghabry: All this ugly out_vector stuff just for this while-loop thing which will always add 0 except for 1 case... oh come on Sormat :/

Ghabry:
Because the only place where this ugly replace vector thing even makes a difference is in the "killed_by_attack" case which can be queried via "IsKilledByAttack()" this whole stuff is unnecessary and complecated, you could just pop the messages based on this condition...

Good that the commit message is so verbose, otherwise I wouldn't have been able to follow the thought process of some code changes ^^'



semi-ot: Is there a smart diff-tool which can detect when code was just moved around? (like the ConditionHeal being moved up)

At least I worked in Bioinformatics with tools before which were able to detect and visualize Genome rearrangements which is more or less the same kind of problem, just instead of DNA letters you move sourcecode around. xD

fmatthew5876:
I think the `window_battlemessage` should just have a marker you increment, rollback to, and clear. Not sure we need all these complicated states about messages.

Ghabry:
Interestingly there is no merge conflict in the type commit even though this was just added to master as part of another PR.

fmatthew5876:
It's exactly the same commit, so I guess github is smart enough to figure it out.

fmatthew5876:
rebased to master

fmatthew5876:
jenkins: test this please

fmatthew5876:
@Ghabry This one is ready now

fmatthew5876:
jenkins: test this please

Ghabry:
cherry-picked the newlib commit, will merge it directly with 6 of 7 :+1:

PR #1451, opened 2018-10-15
Albeleon's Battle Fixes 6 of 7

fmatthew5876:


fmatthew5876:
This rebase is messed up. Please don't touch this until 1-5 are done.

fmatthew5876:
All commits that show up here have been reviewed, tested, and approved by me.

fmatthew5876:
The apply atk/spi commit fixes problems in #985 where the enemy's buffs boost the wrong values.

fmatthew5876:
jenkins: test this please

src/scene_battle.h
```diff
@@ -20,6 +20,7 @@
 
 // Headers
 #include <deque>
+#include <memory>
```
Ghabry: > 2.22: RM2000: Correct printing of state heal/affect messages 

Purpose of memory include?

src/scene_battle_rpg2k.cpp
```diff
@@ -15,6 +15,7 @@
  * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <iostream>
```
Ghabry: > 2.22: RM2000: Correct printing of state heal/affect messages 

and this one

fmatthew5876: Oops looks like I forgot to clean up my mess after debugging.

src/game_battlealgorithm.cpp
```diff
@@ -425,6 +426,39 @@ std::string Game_BattleAlgorithm::AlgorithmBase::GetStateMessage(const std::stri
...
+			{ GetTarget()->GetName(), ss.str() }
+		);
+	}
+	else {
+		std::string particle, space = "";
+		ss << GetTarget()->GetName();
+
+		if (Player::IsCP932()) {
+			// FIXME?? Verify if these particle are correct
+			particle = "の ";
```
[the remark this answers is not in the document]
Ghabry: Can't figure this out, will try to get the message directly from Japanese RPG_RT.exe.


No idea why there are sometimes whitespace after the particle btw, as Japanese has no spaces in sentences o.O.

src/game_battler.h
```diff
@@ -136,6 +136,15 @@ class Game_Battler {
...
 	void ShiftAttributeRate(int attribute_id, int shift);
 
+	/**
+	 * Gets the current modifier (buff/debuff) to an attribute rate.
+	 * The shift is cleared after the battle ended.
+	 *
+	 * @param attribute_id Attribute modified
+	 * @return shift Shift applied.
+	 */
+	int GetShiftAttributeRate(int attribute_id);
```
Ghabry: > 2.29: RM2000: Print "attribute resistance

Well the correct name would be "GetAttributeRateShift" because the original function is called "ShiftAttributeRate" because it shifts the attribute rate.

src/game_battlealgorithm.h
```diff
@@ -363,6 +363,7 @@ class AlgorithmBase {
 	std::string GetDamagedMessage() const;
 	std::string GetParameterChangeMessage(bool is_positive, int value, const std::string& points) const;
 	std::string GetStateMessage(const std::string& message) const;
+	std::string GetShiftAttributeMessage(bool is_positive, const std::string& attribute) const;
```
Ghabry: The is\_positive arg is useless, GetShiftAttribMessage itself could simply call IsPositive

src/game_battlealgorithm.cpp
```diff
@@ -1270,6 +1310,13 @@ void Game_BattleAlgorithm::Skill::GetResultMessages(std::vector<std::string>& ou
 	}
 
 	AlgorithmBase::GetResultMessages(out, out_replace);
+
+	// Attribute resistance / weakness + an attribute selected + can be modified
+	std::vector<int16_t>::const_iterator it_shift = shift_attributes.begin();
+	for (; it_shift != shift_attributes.end(); ++it_shift) {
```
Ghabry: > shift attrib

that new loop could be a foreach-loop

src/game_battlealgorithm.cpp
```diff
@@ -1158,16 +1205,9 @@ void Game_BattleAlgorithm::Skill::Apply() {
...
-		// failed (display a miss)
-
-		for (int i = 0; i < (int)skill.attribute_effects.size(); ++i) {
-			if (skill.attribute_effects[i]) {
-				GetTarget()->ShiftAttributeRate(i + 1, healing ? 1 : -1);
-			}
-		}
+	std::vector<int16_t>::const_iterator it_shift = shift_attributes.begin();
+	for (; it_shift != shift_attributes.end(); ++it_shift) {
+		GetTarget()->ShiftAttributeRate(*it_shift, healing ? 1 : -1);
```
Ghabry: > shift attrib

that new loop could be a foreach-loop

src/game_battlealgorithm.cpp
```diff
@@ -1134,6 +1168,19 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 				conditions.push_back(Data::states[i]);
 			}
 		}
+
+		// Attribute resistance / weakness + an attribute selected + can be modified
+		if (!success && skill.affect_attr_defence) {
+			for (int i = 0; i < skill.attribute_effects.size(); i++) {
+				if (skill.attribute_effects[i] && GetTarget()->GetShiftAttributeRate(i + 1) * (IsPositive() ? 1 : -1) != 1) {
```
Ghabry: > shift attrib

I find this ugly to read. Maybe there should be instead of GetShiftAttributeRate just a CanShiftAttributeRate which probes if a shift would be possible and returns a bool...

fmatthew5876:
I fixed 2.22 and did the requested refactors for 2.29.

I've retested 2.29. It will still need to be reviewed again to ensure no rebase mistakes.

I added the 'CanShiftAttribute()` method, but in general I don't like this approach of check if we can, set success state, and then apply the change. We basically have to duplicate our logic twice all the time.

I'd rather just the apply returned a bool and then we set success at the end if anything happened. I will look into such a refactor after battle 7 is done.

src/game_battlealgorithm.cpp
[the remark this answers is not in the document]
Ghabry: The particle is "ha" は without whitespace

src/game_battlealgorithm.cpp
```diff
@@ -1171,6 +1173,12 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 			}
 		}
 
+		//If resurrected and no HP selected, the effect value is a percentage:
```
Ghabry: That's is not visible in the diff but in the else-block is another "int effect" which shadows the effect declaration, I assume this is wrong.

missing whitespace after //

Ghabry: more "static analysis": The else-codepath is for "non-healing", so that if with "IsRevived" isn't taken then but why aren't these 4 new lines then not simply put into the healing if-block?

src/game_battlealgorithm.cpp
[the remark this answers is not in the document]
fmatthew5876: Fixed the particle

src/game_battlealgorithm.cpp
fmatthew5876: The effect code was confusing and ugly. I've cleaned it up.

I also agree with you that `IsRevived()` belongs in the upper healing block.

src/game_battlealgorithm.h
```diff
@@ -401,6 +406,7 @@ class AlgorithmBase {
 	bool killed_by_attack_damage;
 	bool critical_hit;
 	bool absorb;
+	bool revived = false;
```
Ghabry: Because an attack can have multiple targets all state variables must be reset in the Reset() function.

fmatthew5876: Great catch. This has been fixed.

fmatthew5876:
Rebased to master

PR #1452, opened 2018-10-15
Albeleon's Battle Fixes 7 of 7

fmatthew5876:
Changed:
* 3.13 - Remove printing states logic. This is already done in earlier PR.
* 2.21 - #1538 
* 2.41 - #1538

Dropped:
* 3.18 - already implemented in #1487

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -1781,7 +1781,7 @@ std::string Game_BattleAlgorithm::SelfDestruct::GetStartMessage() const {
 }
 
 int Game_BattleAlgorithm::SelfDestruct::GetSourceAnimationState() const {
-	return Sprite_Battler::AnimationState_Dead;
+	return Sprite_Battler::AnimationState_SelfDestruct;
 }
 
 const RPG::Sound* Game_BattleAlgorithm::SelfDestruct::GetStartSe() const {
--- a/src/game_enemy.cpp
+++ b/src/game_enemy.cpp
@@ -338,3 +338,7 @@ const RPG::EnemyAction* Game_Enemy::ChooseRandomAction() {
 
 	return nullptr;
 }
+
+bool Game_Enemy::IsTransparent() const {
+	return enemy->transparent;
+}
--- a/src/scene_battle_rpg2k.cpp
+++ b/src/scene_battle_rpg2k.cpp
@@ -435,9 +435,6 @@ bool Scene_Battle_Rpg2k::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBase
 			source_sprite = Game_Battle::GetSpriteset().FindBattler(action->GetSource());
 			if (source_sprite) {
 				source_sprite->Flash(Color(255, 255, 255, 100), 15);
-				source_sprite->SetAnimationState(
-					action->GetSourceAnimationState(),
-					Sprite_Battler::LoopState_DefaultAnimationAfterFinish);
 			}
 
 			auto* src = action->GetSource();
@@ -541,6 +538,22 @@ bool Scene_Battle_Rpg2k::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBase
 
 			action->Apply();
 			battle_action_state = BattleActionState_ResultPop;
+
+			if (action->GetSource()->GetType() == Game_Battler::Type_Enemy) {
+				if (action->GetType() == Game_BattleAlgorithm::Type::Escape) {
+					source_sprite = Game_Battle::GetSpriteset().FindBattler(action->GetSource());
+					source_sprite->SetAnimationState(
+							Sprite_Battler::AnimationState_Dead,
+							Sprite_Battler::LoopState_DefaultAnimationAfterFinish);
+				}
+
+				if (action->GetType() == Game_BattleAlgorithm::Type::SelfDestruct) {
+					source_sprite = Game_Battle::GetSpriteset().FindBattler(action->GetSource());
+					source_sprite->SetAnimationState(
+							Sprite_Battler::AnimationState_SelfDestruct,
+							Sprite_Battler::LoopState_DefaultAnimationAfterFinish);
+				}
+			}
 
 			if (!action->IsFirstAttack()) {
 				battle_action_wait = 0;
--- a/src/sprite_battler.h
+++ b/src/sprite_battler.h
@@ -101,6 +101,7 @@ protected:
 	void DoIdleAnimation();
 	void OnMonsterSpriteReady(FileRequestResult* result);
 	void OnBattlercharsetReady(FileRequestResult* result, int32_t battler_index);
+	int GetMaxOpacity() const;
 
 	std::string sprite_name;
 	int hue = 0;
@@ -111,6 +112,7 @@ protected:
 	std::string sprite_file;
 	int sprite_frame = -1;
 	int fade_out = 255;
+	int fade_out_incr = 15;
 	int flash_counter = 0;
 	LoopState loop_state = LoopState_DefaultAnimationAfterFinish;
 	bool old_hidden = false;
--- a/src/sprite_battler.h
+++ b/src/sprite_battler.h
@@ -31,7 +31,8 @@ class BattleAnimation;
 class Sprite_Battler : public Sprite {
 public:
 	enum AnimationState {
-		AnimationState_Idle = 1,
+		AnimationState_Null,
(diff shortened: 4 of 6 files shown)


fmatthew5876:
For 2.42 part of the commit message says:

> Solution: After doing several tests, aproximately the way it works is by varying the Agi for the  
> execution between (-10%, 10%), with 1 value more or less for each extreme (for example, if you 
> have two characters with Agility 10 and 8, the second one still has a few chances of going first).

@CherryDT 

For the random agility modifier to determine who goes first. Do you agree with +- 10%?

CherryDT:
@fmatthew5876 RPG_RT defines something like a "battle agility" for each turn which is AGI+X. X is a random number below (AGI/4 + 3). However, if the attack is with a weapon that has the "Start with Initiative" checkbox, this puts it into a class of its own which follows the same rules but can rival only other attacks with such weapons (I think this is done by simply adding a huge number like 100000 to the value in such a case). The turn order is then decided by this "battle agility", high to low.

src/game_battlealgorithm.cpp
```diff
@@ -945,11 +1004,31 @@ bool Game_BattleAlgorithm::Normal::Execute() {
...
+				Output::Warning("Algorithm Normal: Invalid weapon animation ID %d", weapon->animation_id);
+			}
+			if (!animation2) {
+				Output::Warning("Algorithm Normal: Invalid weapon animation ID %d", weapon2->animation_id);
+			}
 
-		if (weapon && !weapon->ignore_evasion) {
-			to_hit = (int)(100 - (100 - hit_chance) * (1 + (1.0 * GetTarget()->GetAgi() / ally->GetAgi() - 1) / 2));
-		} else {
+			hit_chance = (weapon->hit + weapon2->hit) / 2;
```
fmatthew5876: This should be `max()` not `mean()`

fmatthew5876:
I've rebased this onto #1450 and got it to compile.

src/game_battlealgorithm.cpp
```diff
@@ -1116,8 +1116,15 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 				// Determined via testing, the heal is always at least 50%
 				mul = 0.5f;
 			}
+			effect = skill.power +
+				source->GetAtk() * skill.physical_rate / 20 +
+				source->GetSpi() * skill.magical_rate / 40;
+			effect *= mul;
 
-			effect = (int)(skill.power * mul);
+			effect += (effect * Utils::GetRandomNumber(-skill.variance * 10, skill.variance * 10) / 100);
```
fmatthew5876: This one needs a review by @CherryDT

Ghabry:
Jenkins: test this please

fmatthew5876:
FYI I have only rebased and gotten this to compile. Have not reviewed the changes yet.

src/scene_battle_rpg2k.cpp
```diff
@@ -418,26 +418,28 @@ bool Scene_Battle_Rpg2k::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBase
...
+						message_to_print = state->message_recovery;
 						message_to_show = true;
 					}
 				}
 				for (auto state_id : states_remaining) {
 					// BattleAlgorithm verifies the states
 					const RPG::State* state = ReaderUtil::GetElement(Data::states, state_id);
 					if (!state->message_affected.empty()) {
-						battle_message_window->PushWithSubject(state->message_affected, action->GetSource()->GetName());
+						message_to_print = state->message_affected;
```
fmatthew5876: Need to test this. Might it be state priority instead of just the last one that dictates which message is shown?

fmatthew5876: This one is wrong. Fixed in #1451

fmatthew5876:
Commits Dropped:
* 3.18 - already implemented in #1487

src/game_battlealgorithm.cpp
```diff
@@ -1117,8 +1117,10 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 
 			if (skill.affect_hp)
 				this->hp = std::max<int>(0, std::min<int>(effect, GetTarget()->GetMaxHp() - GetTarget()->GetHp()));
-			if (skill.affect_sp)
-				this->sp = std::max<int>(0, std::min<int>(effect, GetTarget()->GetMaxSp() - GetTarget()->GetSp()));
+			if (skill.affect_sp) {
+				int prov_sp = GetSource() == GetTarget() ? source->CalculateSkillCost(skill.ID) : 0;
+				this->sp = std::max<int>(0, std::min<int>(effect, GetTarget()->GetMaxSp() - GetTarget()->GetSp() + prov_sp));
```
fmatthew5876: This `prov_sp` is super ugly but its required here with the way we clamp early. This whole structure needs to be refactored. This wart is a reminder.

Ghabry:
Just posting a compiler warning here, I think it is related to one of your recent PRs, could you check if this is a bug?

```
In file included from jni/src/../../../../../../..//src/window_paramstatus.cpp:22:
jni/src/../../../../../../..//src/window_paramstatus.h:50:6: warning: private field 'atk' is not used [-Wunused-private-field]
        int atk = 0;
            ^
jni/src/../../../../../../..//src/window_paramstatus.h:51:6: warning: private field 'def' is not used [-Wunused-private-field]
        int def = 0;
            ^
jni/src/../../../../../../..//src/window_paramstatus.h:52:6: warning: private field 'spi' is not used [-Wunused-private-field]
        int spi = 0;
            ^
jni/src/../../../../../../..//src/window_paramstatus.h:53:6: warning: private field 'agi' is not used [-Wunused-private-field]
        int agi = 0;
            ^
4 warnings generated.
```

src/game_battlealgorithm.cpp
Ghabry: what does "prov" even mean?

fmatthew5876: No idea :)

fmatthew5876:
I've hydra'd myself out again. These rest of these will need to wait for the other battle PRs before they can be reworked.

fmatthew5876:
rebased to master

fmatthew5876:
I've reviewed and tested everything here.

I'm not convinced this animation state stuff belongs in `Game_BattleAlgorithm`. Animation states are very specific to 2k (almost non-existant) vs 2k3 (used a lot).

I added a new commit to refactor the animation state stuff on the 2k side. Leaving 2k3 stuff alone.

I also tested and confirmed that when an enemy escapes or self destructs, you don't receive any EXP, GP, or items.

This is ready for review. With this and #1560 we're finally done with the Sormat battle PRs! :1st_place_medal:

src/sprite_battler.h
```diff
@@ -42,7 +43,8 @@ class Sprite_Battler : public Sprite {
 		AnimationState_WalkingLeft,
 		AnimationState_WalkingRight,
 		AnimationState_Victory,
-		AnimationState_Item
+		AnimationState_Item,
+		AnimationState_SelfDestruct
```
Ghabry: Adding anything after Item feels misleading btw because this enum maps exactly to "Animation 2 -> Animation Type" in the RPG Maker 2003 editor and it is possible to have up to 32 different poses (for Items and Skills).

The 2k3 battle system will never read this SelfDestruct enum for actors but this shows that Sormat didn't notice where the enum values come from.

And yes, I know that I heavily abused this state stuff to inject "animations" for enemies 😜 (but at least they mapped to existing 2k3 anim stuff)

Ghabry: > Shouldn't such an enum be in liblcf?

yes

That's not 2k business, this is monster business. For no logical reason they decided to keep the monsters static and only the party battlers can be animated. (there is "flying" but this only moves the static sprite up and down)

src/game_enemy.cpp
```diff
@@ -338,3 +338,7 @@ const RPG::EnemyAction* Game_Enemy::ChooseRandomAction() {
 
 	return nullptr;
 }
+
+bool Game_Enemy::GetTransparency() const {
+	return enemy->transparent;
```
Ghabry: would call this IsTransparent maybe

fmatthew5876: Done

src/sprite_battler.h
fmatthew5876: I moved SelfDestruct = 255, so it's out of the way of items.

fmatthew5876:
I've retested this after the rebase.

PR #1455, opened 2018-10-16
Random Encounters fixes

fmatthew5876:
Depends on:  #1441

Info from here: https://github.com/EasyRPG/liblcf/issues/261

* Correctly load encounter_rate from saved game
* Compute first strike chance on random encounter.

### DIFF
--- a/src/game_map.cpp
+++ b/src/game_map.cpp
@@ -20,6 +20,7 @@
 #include <iomanip>
 #include <sstream>
 #include <algorithm>
+#include <climits>
 
 #include "async_handler.h"
 #include "system.h"
@@ -77,6 +78,8 @@ namespace {
 	bool teleport_delay;
 
 	RPG::Chipset* chipset;
+
+	int last_encounter_idx = 0;
 }
 
 void Game_Map::Init() {
@@ -921,6 +924,11 @@ void Game_Map::UpdateEncounterSteps() {
 		return;
 	}
 
+	if (GetEncounterRate() <= 0) {
+		location.encounter_steps = 0;
+		return;
+	}
+
 	int x = Main_Data::game_player->GetX();
 	int y = Main_Data::game_player->GetY();
 
@@ -930,23 +938,59 @@ void Game_Map::UpdateEncounterSteps() {
 		return;
 	}
 
-	location.encounter_steps = location.encounter_steps - terrain->encounter_rate;
+	location.encounter_steps += terrain->encounter_rate;
 
-	if (location.encounter_steps <= 0) {
+	struct Row {
+		int ratio;
+		float pmod;
+	};
+
+#if 1
+	static constexpr Row enc_table[] = {
+		{ 0, 0.0625},
+		{ 20, 0.125 },
+		{ 40, 0.25 },
+		{ 60, 0.5 },
+		{ 100, 2.0 },
+		{ 140, 4.0 },
+		{ 160, 8.0 },
+		{ 180, 16.0 },
+		{ INT_MAX, 16.0 }
+	};
+#else
+	//Old versions of RM2k used this table.
+	//Left here for posterity.
+	static constexpr Row enc_table[] = {
+		{ 0, 0.5 },
+		{ 20, 2.0 / 3.0 },
+		{ 50, 5.0 / 6.0 },
+		{ 100, 6.0 / 5.0 },
+		{ 200, 3.0 / 2.0 },
+		{ INT_MAX, 3.0 / 2.0 }
+	};
+#endif
+	const auto encounter_rate = GetEncounterRate();
+	const auto ratio = location.encounter_steps / encounter_rate;
+
+	auto& idx = last_encounter_idx;
+	while (ratio > enc_table[idx+1].ratio) {
+		++idx;
+	}
+	const auto& row = enc_table[idx];
+
+	const auto pmod = row.pmod;
+	const auto p = (1.0f / float(encounter_rate)) * pmod * (float(terrain->encounter_rate) / 100.0f);
+	auto draw = std::uniform_real_distribution<float>()(Utils::GetRNG());
+
+	if (draw < p) {
 		ResetEncounterSteps();
 		PrepareEncounter();
 	}
 }
 
 void Game_Map::ResetEncounterSteps() {
-	int rate = GetEncounterRate();
-	if (rate > 0) {
-		int throw_one = Utils::GetRandomNumber(0, rate - 1);
-		int throw_two = Utils::GetRandomNumber(0, rate - 1);
-
-		// *100 to handle terrain rate better
-		location.encounter_steps = (throw_one + throw_two + 1) * 100;
-	}
+	last_encounter_idx = 0;
+	location.encounter_steps = 0;
 }
 
 std::vector<int> Game_Map::GetEncountersAt(int x, int y) {
--- a/src/game_map.cpp
+++ b/src/game_map.cpp
@@ -137,7 +137,7 @@ void Game_Map::Quit() {
 void Game_Map::Setup(int _id) {
 	SetupCommon(_id, false);
 	map_info.encounter_rate = GetMapInfo().encounter_steps;
-	ResetEncounterSteps();
+	SetEncounterSteps(0);
 
 	Parallax::ClearChangedBG();
 
@@ -206,7 +206,7 @@ void Game_Map::SetupFromSave() {
 	map_info.Fixup(GetMapInfo());
 	SetChipset(map_info.chipset_id);
 
-	ResetEncounterSteps();
+	SetEncounterSteps(location.encounter_steps);
 
 	// FIXME: Handle Pan correctly
 	location.pan_current_x = 0;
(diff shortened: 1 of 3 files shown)


fmatthew5876:
Hey @CherryDT

I wanted to ask you some questions to clarify the behavior with Random Encounters. Hopefully we can fix everything encounter related and get 100% compatibility in this PR.

1. SavePartyLocation 0x7C encounter steps

Is this value saved and loaded exactly? If I save my game with 5 steps left will I always hit a random encounter on exactly 5 steps when I load the game? Are there any special changes made to this value on loading the game?

2. What is the algorithm used for recomputing location encounter_steps in rm2k?

Player does this:

```
a = rand(0, rate -1)
b = rand(0, rate -1)
steps = (a + b + 1) * 100
```

I remember in old versions of rm2k you could get into an encounter in a single step if you were unlucky. I think they fixed that in newer versions. It looks like that same thing could happen here.

3. Is location steps always recomputed from scratch when you change maps?

4. What happens when the interpreter processes a `ChangeEncounterRate` command?

Does location steps get recomputed against the new rate or is it left alone until after the next battle? Is the custom encounter rate always reset when you change maps?

5. Any other wierd edge cases I should know about not covered here?

Thanks!

CherryDT:
Teleporting the party (also on the same map) resets the counter.

On every step, the SavePartyLocation.encounter_steps is increased by Terrain.encounter_rate.

There is then a probability calculation which is not linear, and I think it was different in the past - but now it is basically like this:

`SavePartyLocation.encounter_steps/Terrain.encounter_rate` ration threshold|Probability modifier
>0|0.0625
>20|0.125
>40|0.25
>60|0.5
>80|**0.5**
>100|**2**
>120|2
>140|4
>160|8
>180|16

(I marked the steps at which it behaves a bit strangely)

Then, the probability to trigger an encounter is calculated using:

Probability = (1 / SaveMapInfo.encounter_rate) * Probability modifier * (Terrain.encounter_rate / 100)

I'm not sure yet why the multiplication with the terrain encounter_rate is done at the end again though.

CherryDT:
Teleporting the party (also on the same map) resets the counter.

On every step, the SavePartyLocation.encounter_steps is increased by Terrain.encounter_rate.

There is then a probability calculation which is not linear:

**2k3 and modern 2k**

`SavePartyLocation.encounter_steps/Terrain.encounter_rate` ratio threshold|Probability modifier
-|-
0+|1/16
20+|1/8
40+|1/4
60+|1/2
100+|2
140+|4
160+|8
180+|16

**Old 2k** (I think before Value)

`SavePartyLocation.encounter_steps/Terrain.encounter_rate` ratio threshold|Probability modifier
-|-
0+|1/2
20+|1/1.5
50+|1/1.2
100+|1.2
200+|1.5

Then, the probability to trigger an encounter is calculated using:

Probability = (1 / SaveMapInfo.encounter_rate) * Probability modifier * (Terrain.encounter_rate / 100)

I'm not sure yet why the multiplication with the terrain encounter_rate is done at the end again though.

Also, RPG_RT does at the end `if(RandomInteger(0 to X-1) < Probability*X)`, and old 2k has less granularity here with an X of 1000, while modern 2k and 2k3 uses an X of 65536.

CherryDT:
To answer your questions:

* Ground vehicles don't affect the calculation.
* In the following cases, the calculation is not done ("freezing" the counter effectively): Ctrl pressed, map encounter rate is zero, inside airship
* An encounter resets the counter to zero.
* The algorithm is using floating point.

Also, a detail I realized: the encounter waiting flag is checked before this calculation. So if an encounter is scheduled, it would set this flag and then only next frame initiate the actual encounter

fmatthew5876:
Thanks for the clarifying info and the code review. I fixed the typos in the table.

> Also, a detail I realized: the encounter waiting flag is checked before this calculation. So if an encounter is scheduled, it would set this flag and then only next frame initiate the actual encounter

I'm guessing you refer to `SavePartyLocation::encounter_calling (0x7D)` flag here?

This one might be important to get right so encounters behave correctly w.r.t. to running events.

fmatthew5876:
Rebased to master

carstene1ns:
jenkins: test this please! :heart:

src/game_map.cpp
```diff
@@ -918,25 +935,61 @@ void Game_Map::UpdateEncounterSteps() {
 		return;
 	}
 
-	location.encounter_steps = location.encounter_steps - terrain->encounter_rate;
```
Ghabry: the reason this was written in such an ugly way (not using -=) was btw for support in my automagical refactor liblcf field-access to Get/Set function converter

src/game_map.cpp
```diff
@@ -918,25 +935,61 @@ void Game_Map::UpdateEncounterSteps() {
...
+	const auto& row = enc_table[idx];
 
-void Game_Map::ResetEncounterSteps() {
-	int rate = GetEncounterRate();
-	if (rate > 0) {
-		int throw_one = Utils::GetRandomNumber(0, rate - 1);
-		int throw_two = Utils::GetRandomNumber(0, rate - 1);
+	const auto pmod = row.pmod;
+	const auto p = (1.0f / float(encounter_rate)) * pmod * (float(terrain->encounter_rate) / 100.0f);
+	auto draw = std::uniform_real_distribution<float>()(Utils::GetRNG());
```
Ghabry: is std::uniform_real_distribution implementation standardized? I mean does it give the same values for the same seed independent of the C++ implementation (clang c++ lib, gcc c++ lib, MSVC)?

fmatthew5876: Good question. I can't find anything that guarantees this. 

`uniform_real_distribution` does not actually play a part in drawing random numbers. The rng stored in `Utils::GetRNG()` does that. The distribution is just a deterministic mathematical function which takes an input random number in some integer range and converts it to floating point. I guess standard library implementations could use different combinations of floating point expressions which could result is slightly different answers.

That being said, this is the right way to do this. Modulus and all the other tricks we used in the past with C `rand()` are bad. They don't actually give you good random numbers.

https://en.cppreference.com/w/cpp/numeric/random/uniform_real_distribution

fmatthew5876: Also, Here is a great cppcon talk on the C++11 random facilities. Worth the time to watch.

https://www.youtube.com/watch?v=6DPkyvkMkk8

fmatthew5876:
Rebased onto master.

The one commit that handled save data `encounter_steps` was dropped as this is now done by the lsd fixes in master.

src/game_map.cpp
fmatthew5876: I will change this to use `PercentChance()` when #1487 gets merged.

Ghabry:
Is this ratio-threshold table really correct? I have a hard time to understand this.

Let's say the map has a encounter rate of 25, then the Party will need 500 steps to reach a ratio of 20.

Which means the algorithm is always stuck at the lowest value for an extremely long time and has a constant encounter probability of 0.000025 (assuming terrain rate is 1). Is this really correct?

CherryDT:
It sounds for me like you are off by a factor of 100 - the terrain rate is in percent!

Ghabry:
Well, that's the code I use for the analysis:

```R
enc = matrix(c(
  0, 1/16,
  20, 0.125,
  40, 0.25,
  60, 0.5,
  100, 2.0,
  140, 4.0,
  160, 8.0,
  180, 16.0,
  1000000, 16.0
), ncol=2, byrow=TRUE)

player_steps = seq(from = 0, to = 10000, by = 10)
map_enc_steps = 25

ratio = player_steps / map_enc_steps

res = sapply(ratio, FUN = function(x) {
  idx = 1
  while (x > enc[idx+1,1]) {
    idx = idx + 1
  }
  return (idx)
})

pmod = enc[res, 2]
p = (1.0 / map_enc_steps) * pmod * (1.0 / 100.0)

plot(player_steps, p, lwd = 0.5)
```

rplot

Ghabry:
Oh, now I see in the editor what you mean... sec

CherryDT:
Which would be correct at first glance, except that the terrain rate of 1 (percent!) is a totally unrealistic value. 100 is the default.

Ghabry:
okay fixed. k, so the algorithm is just really bad, thanks for the confirmation xD.

CherryDT:
It's not that bad if you think about it. I think that stepping is weird but apart from that it does its job giving a perceptible rate that matches the one configured in the map while reducing the probability of very short and very long interval outliners, factoring in effects of the terrain in both accumulated and immediate effects.

fmatthew5876:
The way it steps up into infinity is pretty clever in my opinion. You can get randomness when you load a saved game without resetting the encounter steps.

CherryDT:
@Ghabry even the fixed graph is misleading because the player_steps isn't really steps, it's "accumulated terrain effects", with the effect value being in percent, so it's effectively still off by x100 on the x axis (either that or player_steps is a misnomer)

Ghabry:
At least I understand now why some would prefer to rename it from steps to rate... >.>

Ghabry:
Yeah, nevermind. After fixing this it makes more sense. THe stepping is still strange imo, but it gives a reasonable result now.

PR #1459, opened 2018-10-26
Fix a regression introduced in b7f307e which resulted in a glitchy airship sprite

Ghabry:
The observation by Sormat was correct that one of the two SyncWithPlayer calls is redundant.

This change keeps his bugfix (Ship "jumps" to the player while entering) and resolves the flickering.

@fmatthew5876 please test.


fmatthew5876:
I tested this and can confirm it fixes the bug for me.

Ghabry:
okay, thx for noticing that last-minute regression :)

PR #1462, opened 2018-10-27
Refresh status window when battle events affect actors - Previously Albeleon 1.4

fmatthew5876:
~~…sage display window is needed to display, the status window should be able to refresh inmediately without skipping a frame (for example, if "Event page: (Turn 0x + 0); DecreaseHP(Player1, 999, Dead)"."~~

~~Solution: Game_Battle only knows when the event ends when using UpdateEvents in the next frame, which mean we can't depend on that, if we refresh the status window then, for a frame we would see the previous state. That's why, to solve that frame of space while UpdateEvents is not updated:~~

~~1. We add to Game_Interpreter a "finished_not_updated" bool, initialized to false.
2. In CommandEnd, we set the variable to "true". In Game_Battle::UpdateEvents, we set it to "false". This means this variable will be active while the interpreted is finished but the events aren't active.
3. Right after the Scene_Battle::Update, we check if the variable is true. If that's the case, Refresh the status window. This way it's been updated before the status window is drawn when it has the previous data.~~

* When a battle event finishes it may have changes character HP/SP/State etc.. Therefore we need to refresh the status window `onCommandEnd()`.

### DIFF
--- a/src/scene_battle_rpg2k.cpp
+++ b/src/scene_battle_rpg2k.cpp
@@ -1051,3 +1051,4 @@ bool Scene_Battle_Rpg2k::CheckLose() {
 bool Scene_Battle_Rpg2k::CheckResultConditions() {
 	return CheckLose() || CheckWin();
 }
+
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -551,6 +551,8 @@ bool Game_Interpreter::CommandEnd() { // code 10
 			evnt->StopTalkToHero();
 	}
 
+	Scene::instance->onCommandEnd();
+
 	return true;
 }
 
--- a/src/scene.h
+++ b/src/scene.h
@@ -172,6 +172,9 @@ public:
 
 	Graphics::State& GetGraphicsState();
 
+	/** Called by the interpreter when an event finishes processing */
+	virtual void onCommandEnd() {}
+
 private:
 	/** Scene stack. */
 	static std::vector<std::shared_ptr<Scene> > instances;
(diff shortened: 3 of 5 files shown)


Ghabry:
iirc what I disliked here was that it introduces a new variable into Game_Interpreter just to handle this cosmetical corner case.

fmatthew5876:
I agree, more varibles make the code more complex. I'll take a look and see if it can be done differently.

fmatthew5876:
I am able to reproduce the bug this fixes. To do this create a troop page that activates on turn 0 and sets a status condition to the party. You won't see their status updated in the window until you do some action that overwrites it.

fmatthew5876:
I've added a new implementation for this. It adds `Scene::onCommandEnd()` hook and uses it to fix the bug.

I'm not sure if we need this hook in such a general context. Would it be useful for other things?

src/scene_battle_rpg2k.cpp
```diff
@@ -56,6 +57,11 @@ void Scene_Battle_Rpg2k::Update() {
 	}
 
 	Scene_Battle::Update();
+
+	// If an event that changed status finishes without displaying a message window,
+	// we need this so it can update automatically the status_window
+	if (Game_Battle::GetInterpreter().finished_not_updated)
+		status_window->Refresh();
 }
```
fmatthew5876: I don't understand why we can't just call `status_window->Refresh()` every time? @Ghabry do you know why we can't do that and have to invent some logic like this?

Ghabry: Because our text rendering is slow (you get noticable frame drops when redrawing each frame on slow systems).
Though considering all the performance improvements over the years maybe is not that bad anymore, should be profiled.

fmatthew5876:
Took another look at this.

Even if performance is better I agree its a bad idea to recompile all of that status window stuff every frame. For now this `onCommandEnd()` approach is the best I can come up with.

Ghabry:
For now this solution is fine but we should really come up with a more local solution.

PR #1463, opened 2018-10-28
Update splash logo image

### LINKED ISSUE #624: Use a better EasyRPG logo at startup

fdelapena:
Here's a logo refresh proposed for 0.6.0 and later with dark background and more discreet.

Not using bright colours on purpose to keep eyes happy and integrate better with dark games.

Maybe the text looks a bit blurry, it might be sharpened up if worth.

Test: https://easyrpg.org/play/pr1463/

Closes #624, where there's the original SVG and if we go high res later or to make tweaks.


Ghabry:
Imo the EasyRPG logo + text is too small, there is too much space wasted (huge borders left and right) and the font is hard to read, maybe make it a bit brighter?

fdelapena:
The large margin perception is due to non 4:3 screens, for white background the white shows clearly the delimitation, this is a side effect. The margin is mostly the same, but also thinking for CRT overscan "safe areas", but not sure if worth nowadays, also game messages are probably not aware of this.

About brighter text, it may be increased a bit, but there is a risk of getting more attention than the logo itself. I could try sharpening the antialiased text.

fdelapena:
I was using the original version, which is now the one for white backgrounds. Though the variant for dark backgrounds looks nice for the small web logo, the wave line looks low contrast in the "easy" part compared with the "RPG" part, also the "easy" part is maybe too bright compared with the last.

As commented on IRC, I'd vote for an unified logo tweak suitable for both light and dark backgrounds if possible, that would prevent possible misuse (like tentatively in this case).

Maybe thinking on a logo revamp (e.g. flat style for the nowadays hotness or some ideas from the Schoko variant) is worth for mid-future term.

fdelapena:
Updated logo header with brighter version and added the PNG in the resources directory as requested.

PR #1464, opened 2018-10-29
Libretro upstreaming

Ghabry:
I'm slowly annoyed by always rebasing this therefore I would like to finally upstream the libretro code.

The code is not very invasive, only adds required libretro_ui and _audio files and "USE_LIBRETRO" where needed.

CMake considerations:
The CMakeLists is sufficient to build a easyrpg_libretro.a/.so when you have all dependencies installed (I refuse to add dependency source code in the upstream repo), simply compile with ``-DPLAYER_TARGET_PLATFORM="libretro" -DBUILD_SHARED_LIBS=ON`` and you get a library that works with at least Linux RetroArch.

The exception is libretro-common which I added as a submodule with an appropriate CMakeLists.txt.

I'm a bit surprised how simple it was to add a non-SDL2 build with not too many changes (also a nice base now for other platforms) as I didn't intended that use-case when writing the file...

Biggest changes to the CMake file is the new "PLAYER_TARGET_PLATFORM"
Almost everything else is just adding tabulators because code was moved into if() but this renders badly in the diff viewer :/.

- [x] Apply useful commits from https://github.com/libretro/easyrpg-libretro/commits/master
- [x] Add Input handling code


Ghabry:
updated the readme with minimal instructions

Ghabry:
Jenkins: test this please

Ghabry:
should be ready now.

src/player.h
```diff
@@ -104,8 +104,11 @@ namespace Player {
 	/**
 	 * Resets all game objects. Faster then CreateGameObjects because
```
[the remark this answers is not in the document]
Ghabry: I havn't modified this line in this PR btw ;), but fixing it. sec

Ghabry:
the 3.7 friendly c++11 change doesn't work, will search a better way

Ghabry:
Jenkins: test this please

src/player.cpp
```diff
@@ -322,10 +316,31 @@ void Player::Update(bool update_scene) {
...
 	}
 
+#ifdef EMSCRIPTEN
+	Graphics::Draw();
+#else
+	cur_time = (double)DisplayUi->GetTicks();
+	if (cur_time < next_frame) {
+		Graphics::Draw();
+		cur_time = (double)DisplayUi->GetTicks();
+		// Don't use sleep when the port uses an external timing source
```
carstene1ns: Would the emscripten way not suffice for `libretro` too?

src/audio_libretro.h
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * EasyRPG Player is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef AUDIOLIBRETRO2_H
```
carstene1ns: `EP_AUDIO_LIBRETRO_H`

src/audio_libretro.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+retro_audio_sample_batch_t RenderAudioFrames=0;
+
+static const unsigned AUDIO_SAMPLERATE = 44100.0;
+
+namespace {
+   unsigned samples_per_frame = 0;
+   bool enable_audio = false;
+	LibretroAudio* instance = nullptr;
+  std::vector<uint8_t> buffer;
```
carstene1ns: whitespaces are totally messed up here.

src/filefinder.h
```diff
@@ -29,9 +29,9 @@
 
 #ifdef PSP2
 #  include <psp2/types.h>
-#  define Offset SceOff
+#  define EasyRPG_Offset SceOff
 #else
-#  define Offset off_t
+#  define EasyRPG_Offset off_t
```
carstene1ns: This is very ugly (also the other type bending in filefinder for PSVita). I hope we can change this to use the same types later and get rid of all these `Sce` stuff. The vitasdk should really abstract this away in their c library.

README.md
```diff
@@ -135,6 +135,23 @@ Read detailed instructions at:
 * https://wiki.easyrpg.org/development/compiling/player/windows/visual-studio
 
 
+### Building a libretro core:
+
+Building for libretro is based on the CMake method.
+
+Additional commands required before building:
```
carstene1ns: Generally, I would use a subrepo nowadays. https://github.com/ingydotnet/git-subrepo#benefits
(No additional maintenance burden, users can use the repository checkout as is.)

However, as libretro is not the normal build and the repository files would mostly be useless, a submodule is the better approach of course.

src/player.cpp
Ghabry: no, because this way you get frameskip. The emscripten way will slow down the engine when you are < 60 FPS.
I wanted to use the same technique for emscripten but GetTicks is a bit broken there and doesn't work correctly.

src/audio_libretro.cpp
Ghabry: Twinaphex, pls

src/filefinder.h
Ghabry: should be done in one go with the AMIGA symbol mess :/

Ghabry:
Havnt't worked on the reviews yet but I noticed that rthreads was an unused include, wiped the whole retro-common dependency... -_-

Ghabry:
Sorry for this sloppy PR, is really a mess... should've checked the code of the other devs.

Ghabry:
still cleaning up the chaos others left behind :-1: .

Remaining todo: Improve the use of the input API (button remapping) and remove unused API functions from libretro_ui.cpp.

Ghabry:
I don't want to maintain this optional GLES stuff, I'm not comfortable with OpenGL and there other "render to framebuffer" API seems to work properly.
And you have now lots of GLES code already in your Switch branch, therefore shouldn't be needed anymore here.

rogersachan:
Can someone rebase this?

Ghabry:
This is not ready yet. Use the master branch from https://github.com/libretro/easyrpg-libretro (or if you are a gamer: Get it from the core downloader in RetroArch)

Ghabry:
finally implemented the difficult part (input remapping and core options).
Needs some final polish, then this is finally ready.

Keyboard is passed through (I disabled stuff that is provided by libretro directly) and the Gamepad allows button remapping via the core options (this is not the intended way to do it but we have more possible buttons than gamepad buttons, so this hack works great).
Also the testplay mode can be toggled via the core options.

src/system.h
```diff
@@ -79,7 +79,13 @@
...
 #  define SUPPORT_JOYSTICK_HAT
 #  define SUPPORT_JOYSTICK_AXIS
-#else // Everything not catched above, e.g. Linux/*BSD/macOS
+#elif defined(USE_LIBRETRO)
+#  define SUPPORT_KEYBOARD
+#  define SUPPORT_JOYSTICK
+#  define SUPPORT_JOYSTICK_HAT
+#  define SUPPORT_JOYSTICK_AXIS
+#  define JOYSTICK_AXIS_SENSIBILITY 20000
+#else // Everything not catched above, e.g. Linux/*BSD/macOS/libretro
```
carstene1ns: libretro has an own section above.

Ghabry: Wups, guess you can imagine based on this comment how the code evolved over time :D

Ghabry:
This is now ready for review.
The only commit I will touch & test again is "libretro: Fix startup and shutdown crashes" because there were so many code changes in the meanwhile and maybe isn't needed anymore (ugly commit).

Ghabry:
The ugly startup/shutdown commit which added new conditions all around the Player update code is now gone. Also is much simpler this way.

Ready for review.

src/libretro_ui.cpp
```diff
@@ -0,0 +1,739 @@
...
+				button_remapper[i] = Input::Keys::JOY_10 + j;
+				break;
+			}
+		}
+		if (button_remapper[i] == Input::Keys::NONE) {
+			variables[i].value = none;
+		}
+	}
+
+	struct retro_input_descriptor desc[] = {
```
Ghabry: Guess you will miss this during the review, so I highlight this quirk here:
libretro requires that only the string-pointer (the last struct element) stays valid but as you can see this struct is on the stack and destroyed but as you know in C the struct doesn't contain the string but a pointer to the string and the strings are constant/known during compile time, so they are always pointing to the same valid storage in the data segment. :D

CMakeLists.txt
```diff
@@ -364,12 +360,43 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/builds/cmake/Modules"
 include(ConfigureWindows)
 include(PlayerFindPackage)
 
+# Platform setup
+set(PLAYER_TARGET_PLATFORM "SDL2" CACHE STRING "Platform to compile for. Options: SDL2 libretro")
+set_property(CACHE PLAYER_TARGET_PLATFORM PROPERTY STRINGS Default libretro)
```
carstene1ns: ↑ What is this supposed to do?

Ghabry: That's for the selection boxes in ccmake / cmake-gui but I got the names wrong, Default must be SDL2.

src/libretro_ui.cpp
```diff
@@ -0,0 +1,739 @@
...
+
+	if (!game)
+		return false;
+
+	if (!LibretroUi::environ_cb(RETRO_ENVIRONMENT_SET_PIXEL_FORMAT, &fmt)) {
+		log_cb(RETRO_LOG_INFO, "XRGB8888 is not supported.\n");
+		return false;
+	}
+
+	extract_directory(parent_dir, game->path, sizeof(parent_dir));
```
Ghabry: This crashes on Windows because path in game->path is not UTF-8. Don't tell me they use the local Codepage here m(

Ghabry: Nevermind, the path is UTF-8 encoded but liblcf can't handle such paths on Windows. In the GameBrowser we work around this with "SetCurrentDirectory" but I have a proper solution for this in my VFS branch, so I won't fix this. This is only a problem for the filename of the main game folder (where liblcf must have access to).

Ghabry:
This is also ready now. There were no new crashes or anything reported by libretro users.

The Player still crashes on Android when closing but I blame RetroArch for this, even my start/stop simplification didn't help here.

Ghabry:
I have an idea how to fix the audio thread crashes on core shutdown on Windows and Android (these are RetroArch bugs because they call code in the unloaded dll/so) but I need to wait for there awesome buildbot infrastructure first before I can test it ^^'

Ghabry:
okay the workaround works, will only update libretro-common to the latest submodule version, then we are good to go.

src/input_buttons_libretro.cpp
```diff
@@ -0,0 +1,129 @@
...
+#if defined(USE_LIBRETRO)
+
+// Headers
+#include "input_buttons.h"
+#include "keys.h"
+#include "libretro.h"
+
+void Input::InitButtons() {
+	// Basicly the desktop keyboard mapping but with some functionality
+	// disabled because it is provided by libretro or not usableon these
```
carstene1ns: missing space, typos.

src/libretro_ui.h
```diff
@@ -0,0 +1,91 @@
...
+ */
+class LibretroUi : public BaseUi {
+public:
+	/**
+	 * Constructor.
+	 *
+	 * @param width window client width.
+	 * @param height window client height.
+	 * @param title window title.
+	 * @param fullscreen start in fullscreen flag.
```
carstene1ns: dox for removed `param`s

src/player.cpp
```diff
@@ -405,6 +420,10 @@ void Player::ParseCommandLine(int argc, char *argv[]) {
 	touch_flag = false;
 	Game_Battle::battle_test.enabled = false;
 
+	if (argv == nullptr) {
```
carstene1ns: would like to have a comment here, this is not POSIX

src/system.h
```diff
@@ -100,14 +107,16 @@
...
-#    endif 
+#    endif
 #  endif
 
 #else
 
-#  ifdef WANT_FMMIDI
-#    if WANT_FMMIDI != 1
-#      error "WANT_FMMIDI must be set to 1 for non-SDL builds"
+#  ifndef HAVE_SDL_MIXER
```
carstene1ns: this makes no sense, because sdl-mixer is not used without SDL (see line 99)

PR #1467, opened 2018-11-01
GNU/Linux: Use default wildmidi config

carstene1ns:


src/decoder_wildmidi.cpp
```diff
@@ -175,13 +175,21 @@ WildMidiDecoder::WildMidiDecoder(const std::string file_name) {
 	config_file = "wildmidi.cfg";
 	found = FileFinder::Exists(config_file);
 
+	// wildmidi command line player default config
+	if(!found) {
```
fdelapena: nit: space between if and parentheses
```suggestion
	if (!found) {
```

carstene1ns: thanks, applied.

PR #1472, opened 2018-11-02
Increment SaveInventory::steps and fix state damage algo

fmatthew5876:
* Ticks when player moves with keyboard
* Ticks when plyaer moves with keyboard in a vehicle
* Does not tick for events forcing player movement
* Does not tick for moving once space to embark/disembark
* Tested and matches RPG_RT behavior

This counter ticks in RPG_RT and is stored in the save data, but outside of that it appears completely unused.

@CherryDT Is there anything we should be doing in the interpreter that depends on this?

### DIFF
--- a/src/game_party.cpp
+++ b/src/game_party.cpp
@@ -389,14 +389,6 @@ int Game_Party::GetActorPositionInParty(int actor_id) {
 	return it != data().party.end() ? std::distance(data().party.begin(), it) : -1;
 }
 
-int Game_Party::GetGold() {
-	return data().gold;
-}
-
-int Game_Party::GetSteps() {
-	return data().steps;
-}
-
 std::vector<Game_Actor*> Game_Party::GetActors() const {
 	std::vector<Game_Actor*> actors;
 	std::vector<int16_t>::const_iterator it;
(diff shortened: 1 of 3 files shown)


CherryDT:
No I think it's only for use in "change variable"

CherryDT:
You are right. That was my mistake. I confused it with RMXP which had this option.
RM2k(3) use this value for the terrain poison feature.

fmatthew5876:
Tested this against RPG_RT and behavior matches. Ready to be reviewed and merged.

Ghabry:
great, when I was discussing this with Sormat weeks ago we tried to figure out how RPG_RT saves this

src/game_party.cpp
```diff
@@ -596,18 +588,17 @@ bool Game_Party::ApplyStateDamage() {
...
-			state_steps_sp.resize(state_id);
-		}
+	const auto steps = GetSteps();
 
+	for (auto state_id : states) {
 		RPG::State *state = ReaderUtil::GetElement(Data::states, state_id);
 
-		if (state->hp_change_map_val > 0 && (++state_steps_hp[state_id - 1]) >= state->hp_change_map_steps) {
-			state_steps_hp[state_id - 1] = 0;
+		//NOTE: We do steps + 1 here because this gets called before steps are incremented.
```
carstene1ns:
```suggestion
		// NOTE: We do steps + 1 here because this gets called before steps are incremented.
```

PR #1475, opened 2018-11-03
Refactor Switches and Variables

fmatthew5876:
Depends: https://github.com/EasyRPG/liblcf/pull/278

* No more count fields
* Variables are signed
* Reads don't resize the array
* Writes to id <= 0 silently do nothing.
* Writes to any value N > 0, even > 5000 actually work. It writes a
  value which is saved to LSD and can be read back. This works in
  RPG_RT to enable any number of variables.
* Matches RPG_RT behavior and written to LSD in same way


fdelapena:
jenkins: test this please

Ghabry:
Thanks for the refactor, this also makes debugging read/writes of S/V easier as you don't need to use memory breakpoints anymore.

> Writes to any value N > 0, even > 5000 actually work. It writes a
value which is saved to LSD and can be read back. This works in
RPG_RT to enable any number of variables.

This already worked before btw.

I'm okay with deleting the "resized array to X elements" message as it gave no advantages but the "index is invalid" message should be kept.

You also removed this (totally arbitrary) limit of PLAYER_VAR_LIMIT/1000000. Does RPG_RT really support writes up to 2^31? I'm worried about memory grow and gigantic savegames.

fmatthew5876:
I added back the debug logging. I agree its helpful for finding bugs.

> You also removed this (totally arbitrary) limit of PLAYER_VAR_LIMIT/1000000. Does RPG_RT really support writes up to 2^31? I'm worried about memory grow and gigantic savegames.

Variables are already clamped to 999999(9) so game code won't be able to write past that anyway. I didn't test clamping on `\v[N`, but there we don't resize the array so it doesn't matter.

Ghabry:
This "report_limit" was btw added because Win Wind Windy used out-of-bounds variable writes for a replay function and this became so extremely spammy (up to 60 writes per second) that the framerate went to 0 on some platforms :/.


I made some testing with really high variables (via lcf2xml) and RPG_RT already gets problems around 100 million (Saving hangs without error?), with higher values it runs out of memory when writing the variable or saving.

Therefore there won't be any game using this but on the other hand this is also a really trivial way to crash our Player.

fmatthew5876:
4 * 9999999 is 40MB. With the clamping in `Set()` that's the worst damage a rogue rm2k3 `ChangeVariable` command could do. Not much on today's machines. Our current memory overhead on games like heroes realm is more than that.

The only way even more variables could get loaded into memory if someone hacked more into an LSD file. Crashing or performing badly on corrupted data seems a normal thing to do.

In the new Rpg makers I'm sure developers are free to allocate huge objects in RGSS and crash the system just the same.

I don't feel strongly either way. If you prefer to add additional checks I can do that. Otherwise I'm fine with it as is.

Ghabry:
you are right, I forgot that using a manipulated LSD would work, too. making this check useless.

Ghabry:
Jenkins: Test this please :\*

PR #1479, opened 2018-11-05
Albeleon's Battle Fixes -- Trashed

fmatthew5876:
Depends on #1449

Battle 3 commits are quite large, so I've split them in half.

### DIFF
--- a/src/scene_battle_rpg2k.cpp
+++ b/src/scene_battle_rpg2k.cpp
@@ -359,7 +359,7 @@ void Scene_Battle_Rpg2k::ProcessActions() {
 }
 
 bool Scene_Battle_Rpg2k::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBase* action) {
-	if (Game_Battle::IsBattleAnimationWaiting()) {
+	if (Game_Battle::IsBattleAnimationWaiting() && !Game_Battle::IsBattleAnimationOnlySound()) {
 		return false;
 	}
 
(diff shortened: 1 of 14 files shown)


src/scene_battle_rpg2k.cpp
```diff
@@ -441,10 +441,14 @@ bool Scene_Battle_Rpg2k::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBase
...
-					action->GetTarget()->GetType() == Game_Battler::Type_Enemy) {
-
-					action->PlayAnimation();
+				if (action->GetTarget()) {
+					if (action->GetTarget()->GetType() == Game_Battler::Type_Enemy) {
+						battle_action_wait = GetDelayForLine();
+						action->PlayAnimation();
+					} else if (action->GetTarget()->GetType() == Game_Battler::Type_Ally && action->GetType() == "Skill") {
+						battle_action_wait = GetDelayForLine();
+						action->PlaySoundAnimation();
```
Ghabry: afaik I already discussed these "sound" and "cutoff" thinks already with Sormat, so they are fine for me. I find this "GetType()" stuff really verbose, because this is really only used for Skill and only here, so how about adding a new virtual function "IsSkill" instead?

fmatthew5876: I don't like that it's a string compare also. Will look into it.

fmatthew5876:
Rebased onto master

src/scene_battle_rpg2k.cpp
fmatthew5876: Fixed in #1449

PR #1481, opened 2018-11-05
ChangeHP, AddState and RemoveState changes - single commit

fmatthew5876:
The way RPG_RT works:
 - You cannot change a battler's HP when they're dead.
 - If ChangeHP kills a character, it adds the Death state to that battler.
 - When Death state is cured, even if alone, it always sets the HP to 1, instead of other functions.
 - When a character is caused Death, all the other states are erased; the defense, charge, gauge and stat modifiers are reset, and the current attribute shifts erased.

We implement all of this, with these particular changes:
 - Now there is no difference between Game_Actor::ChangeHP and Game_Enemy::ChangeHP, so we delete this function from both and add the ChangeHP characteristics to the superclass Game_Battler.
 - In Game_Battler::UseItem, the function now doesn't have to set the HP. However, it needs to have an int "revived" that checks if the character was revived, so in case the function heals HP too after dead, it substracts 1 since that value is automatically put after the Death state is removed. We also move the state cure before the HP change, so the Death State is cured before ChangeHP is executed (if this function was executed before Death state was cured, due to our new conditions, it wouldn't do anything).
 - When you rest on an inn and when the command FullHeal, we move the RemoveAllStates function before the ChangeHP too.
 - We move in UseSkill the state before the ChangeHP too. We also add that, if the character is still dead, then the HP and SP changes can't happen (you can't throw a Cure spell to a dead character).

### DIFF
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -212,7 +212,7 @@ bool Game_Battler::UseItem(int item_id) {
 
 	if (item->type == RPG::Item::Type_medicine) {
 		bool was_used = false;
-
+		int revived = 0;
 		int hp_change = item->recover_hp_rate * GetMaxHp() / 100 + item->recover_hp;
 		int sp_change = item->recover_sp_rate * GetMaxSp() / 100 + item->recover_sp;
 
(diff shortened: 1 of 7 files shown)


fmatthew5876:
@Ghrabry I pulled this directly from an old copy of Albeleons original branch I had in my fork. You didn't make any changes to it in #1394 did you?

src/game_battler.cpp
```diff
@@ -221,19 +221,22 @@ bool Game_Battler::UseItem(int item_id) {
...
+				was_used |= HasState(Data::states[i].ID);
+				if (i == 0 && HasState(i + 1))
+					revived = 1;
+				RemoveState(Data::states[i].ID);
+			}
+		}
+
 		if (hp_change > 0 && !HasFullHp()) {
-			ChangeHp(hp_change);
+			ChangeHp(hp_change - revived);
```
Ghabry: You always find something by looking long enough at it...
Any idea what this "- revived" is supposed to do here? It substracts 0 or 1? o.O

Ghabry:
Okay this is also merged now but I would vote for adding a comment about this non obvious line of code in one of the other thousand battle PRs we have already open. xD

PR #1482, opened 2018-11-05
Fix Item usability in battle - Previously Albeleon 2.30

fmatthew5876:
>    2.30: RM2000 + RM2003: Mark as "unusable" the objects of type "speci
>
>If I understand this correctly the purpose of the new line in ÌsitemUsable is to ensure that the items are always selectable semi-ot: occasion_field1 is btw confusing, maybe we should think about a better name
>
>I see multiple flaws here:
>item.actor_set could be also item.class_set, therefore we should fine a way to get this handled by IsItemUsable imo. (maybe move the check from IsItemUsable to CheckEnable, then the codepath is free)
>
> Also actor_set could be smaller than the ID, resulting in an out-of-bounds read.

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -1243,6 +1243,12 @@ bool Game_BattleAlgorithm::Item::Execute() {
 
 	if (item.type == RPG::Item::Type_medicine) {
 		this->healing = true;
+		// RM2k3 BUG: In rm2k3 battle system, this IsItemUsable() check is only applied when equipment_setting == actor, not for class.
+		if (GetTarget()->GetType() == Game_Battler::Type_Ally && !static_cast<Game_Actor*>(GetTarget())->IsItemUsable(item.ID)) {
+			// No effect, but doesn't behave like a dodge or damage to set healing and success to true.
+			this->success = true;
+			return this->success;
+		}
 
 		// HP recovery
 		if (item.recover_hp != 0 || item.recover_hp_rate != 0) {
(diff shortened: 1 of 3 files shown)


fmatthew5876:
I've checked with rm2k and the following behavior exists in battles:
* Special: if the actor cannot use then they aren't selectable by that actor.
* Medicine: Anyone can use, but if used on actor who cannot use then there is no effect.

fmatthew5876:
I've reimplemented this from scratch. Tested the behavior against both rm2k and rm2k3. All matching.

I've opted to fix the rm2k3 class bug here.

src/game_battlealgorithm.cpp
```diff
@@ -1243,6 +1243,11 @@ bool Game_BattleAlgorithm::Item::Execute() {
 
 	if (item.type == RPG::Item::Type_medicine) {
 		this->healing = true;
+		if (GetTarget()->GetType() == Game_Battler::Type_Ally && !static_cast<Game_Actor*>(GetTarget())->IsItemUsable(item.ID)) {
+			//No effect, but doesn't behave like a dodge or damage to set healing and success to true.
```
Ghabry: whitespace after //

Ghabry:
Could you add a comment to Game_Battler:IsItemUsable which mentions this bug (that RPG2k3 only cares about actor_set for medicine in battle)? Better preserve this knowledge, if this is ever a problem later.

CherryDT:
Another thing which I'd put as "compat mode"...
There are so many bugs, especially in RM2k3, which are for me a conflict between "supporting existing RM games" and "building a game engine usable on its own which can be even better than RM"

Ghabry:
I find this bug especially silly because first they added this ridiculous global radio button and then they failed to implement it correctly. 😥

fmatthew5876:
Fixes comments. I put the bug comment in battlealgorithm, because `Game_Actor::IsItemUsable()` is also used by the menu in the field, and here it behaves correctly.

PR #1483, opened 2018-11-06
"3.10: RM2000: If an attack/skill misses, it shouldn't apply anything…

fmatthew5876:
From #1448 

I've changed this one to still reduce your SP if you attack with a weapon and miss. Confirmed thats the behavior on rm2k3 and rm2k steam.


fmatthew5876:
Tested this with Player, looks good to me.

Ghabry:
could you remove the wrong "4. In Normal::Apply, there shouldn't be side effects if the attack missed (in this case, decreasing SP for attack)." from the commit message?

fmatthew5876:
Fixed the commit message

PR #1484, opened 2018-11-06
"2.28: RM2000 + RM2003: If a battler is thrown a state again, the tur…

fmatthew5876:
…n count for recovery probability should be the same with no changes, it shouldn't "reset". Probability of state recovery probability should also be only each X turns, not each turn after it surpasses X turn. When the battler is first thrown into a state, if it has 1 turn recovery or more, there should be an extra turn where they do nothing (afterwards it's always a multiple. With 0 turn it happens even in the first turn)."

Solution:
1. To avoid "resetting" each time a state is readded, when the state is set, the turn count doesn't start on 1 but clamps it at 1 at minimum.
2. To make sure the turn where it tries to recover each probability is each X turns, multiples, we put that, if the recovery fails, the turn is decreased X, so in the next X turns it will try again.
3. Turn ">=" in ">" and adding +1 makes sure the turn appears two later. Before, a Turn 0, Turn 1 and Turn 2 were executed right after the battler had the state.

>  2.28: RM2000 + RM2003: If a battler is thrown a state again, the tur

>Because the code before is
>states[i] += 1; 
>if (states[i] > Data::states[i].hold_turn + 1) {
>
>isn't this the same as states[i] = 1?

>I don't think so.

>(s > ht + 1) is equivalent to (s - ht) > 1. So s >= 2 when that write happens.

>I have no idea if RPG2000 does this different but for RPG2003 I question the whole code except >states[state_id - 1] = std::max<int> (1, states[state_id - 1]); (which prevents turn count reduction when the >same state is applied again). I used a Poison with 10 turns & 20% setting and I got it healt after turn 16... >(according to this commit it should be again at 20?)

### DIFF
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -380,7 +380,7 @@ void Game_Battler::AddState(int state_id) {
 		states.resize(state_id);
 	}
 
-	states[state_id - 1] = 1;
+	states[state_id - 1] = std::max<int> (1, states[state_id - 1]);
 }
 
 void Game_Battler::RemoveState(int state_id) {
@@ -696,11 +696,13 @@ std::vector<int16_t> Game_Battler::NextBattleTurn() {
 	for (size_t i = 0; i < states.size(); ++i) {
 		if (HasState(i + 1)) {
 			states[i] += 1;
-
-			if (states[i] >= Data::states[i].hold_turn) {
+			if (states[i] > Data::states[i].hold_turn + 1) {
 				if (Utils::ChanceOf(Data::states[i].auto_release_prob, 100)) {
 					healed_states.push_back(i + 1);
 					RemoveState(i + 1);
+				}
+				else {
+					states[i] -= Data::states[i].hold_turn;
 				}
 			}
 		}


fmatthew5876:
@CherryDT I think we need your help with this.

What is the behavior with state recovery in battle? Does getting that state added again affect anything?

CherryDT:
> What is the algorithm with state recovery in battle? Does getting that state added again affect anything?

Inflicting a status always sets the value to 1 (regardless what it was before)

Also note that it checks which statuses need to be cleared (I think this is also little-known): Any status with a priority <= maxPriority-10 is cleared. (maxPriority would be the highest value of all currently active statuses' priorities.)

Ghabry:
Okay this basicly means the whole commit is wrong. Close it or fix as Cherry documented it, thx.

> Also note that it checks which statuses need to be cleared (I think this is also little-known): Any status with a priority <= maxPriority-10 is cleared.

Is this documented anywhere in the help file?
This means all states in the maxPriority area are permanent? woah.

CherryDT:
No maybe I wasn't clear here. They are still cleared based on their settings of course

But I when higher priority states are inflicted, lower priority states (10 apart ftpm highest) are cleared automatically

Ghabry:
Now I get it, thanks.

Which means this commit is completely unrelated, closing 👍

fmatthew5876:
I'll review our battle code tonight and make a PR if we don't obey Cherry's rules.

fmatthew5876:
@CherryDT does the priority -10 thing also happen when a state is set by an event?

CherryDT:
Yes, always.

> _if we don't obey Cherry's rules._
_That sounds wrong._

PR #1485, opened 2018-11-06
Support SavePanorama data

fmatthew5876:
* saves and loads scrolling panorama x and y position.
* Only supported by rm2k3e, but we enable for all modes.

I've tested and the numbers match rm2k3e. You can save and load a game between RPG_RT and Player and the panorama will have the same starting position.

One thing I don't understand is that the range of pan_x/pan_y are 2 times the size of the panorama picture. In order to make the numbers match rm2k3e I had to scale them by 2 in places.

This feature has a bug in rm2k3. If you edit the map, say to change the panorama or turn off scrolling and the load the game the offsets still get loaded. So you could have a non-scrolling panorama which is stuck at a weird offset. I reset the panorama data when `save_count` does not match.

### DIFF
--- a/src/game_map.cpp
+++ b/src/game_map.cpp
@@ -1582,6 +1582,7 @@ void Game_Map::Parallax::ChangeBG(const Params& params) {
 	map_info.parallax_vert = params.scroll_vert;
 	map_info.parallax_vert_auto = params.scroll_vert_auto;
 	map_info.parallax_vert_speed = params.scroll_vert_speed;
+	panorama = {};
 }
 
 void Game_Map::Parallax::ClearChangedBG() {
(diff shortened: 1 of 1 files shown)


CherryDT:
> So you could have a non-scrolling panorama which is stuck at a weird offset. I reset the panorama data when save_count does not match.
Good catch. The fact that it's in a different object than some other things makes it not reset on its own, and I didn't think of that when I implemented it.

Ghabry:
(needs rebase)

Guess the code is correct, I will just need a while to understand why the result of this code is the same as the old code 😅

fmatthew5876:
Rebased.

The only way I know it works was by testing.

Ghabry:
Fired some test data at it, when I substract the width and height from the original code I get the same result as in your new code. The X value is off by one pixel compared to the original but I can't say if this is correct or not.

carstene1ns:
maybe is a rounding difference.

PR #1487, opened 2018-11-06
Critical Hit fixes - Previously Albeleon 3.17

fmatthew5876:
Fixes critical hit behavior.

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -1288,7 +1288,21 @@ int Game_Actor::GetHitChance() const {
 }
 
 float Game_Actor::GetCriticalHitChance() const {
-	return GetActor().critical_hit ? (1.0f / GetActor().critical_hit_chance) : 0.0f;
+	auto& actor = GetActor();
+	float crit_chance = actor.critical_hit ? 1.0f / actor.critical_hit_chance : 0.0f;
+
+	float weapon_bonus = 0;
+
+	auto checkWeapon = [&](const RPG::Item* weapon) {
+		if (weapon) {
+			weapon_bonus = std::max(weapon_bonus, float(weapon->critical_hit));
+		}
+	};
+
+	checkWeapon(GetWeapon());
+	checkWeapon(Get2ndWeapon());
+
+	return crit_chance + (weapon_bonus / 100.0f);
 }
 
 Game_Battler::BattlerType Game_Actor::GetType() const {
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -1383,3 +1383,57 @@ bool Game_Actor::PreventsTerrainDamage() {
 
 	return false;
 }
+
+
+const RPG::Item* Game_Actor::GetWeapon() const {
+	auto* weapon = GetEquipment(RPG::Item::Type_weapon);
+	if (weapon && weapon->type == RPG::Item::Type_weapon) {
+		return weapon;
+	}
+	return nullptr;
+}
+
+const RPG::Item* Game_Actor::Get2ndWeapon() const {
+	// Checking of HasTwoWeapons() not neccessary. If true, the
+	// item equipped in this slot will never be a weapon from
+	// legitimate means.
+	auto* weapon = GetEquipment(RPG::Item::Type_shield);
+	if (weapon && weapon->type == RPG::Item::Type_weapon) {
+		return weapon;
+	}
+	return nullptr;
+}
+
+const RPG::Item* Game_Actor::GetShield() const {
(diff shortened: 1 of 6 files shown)


fmatthew5876:
From CherryDT

>X = Actor|Enemy.critical_hit ? 100 / Actor|Enemy.critical_hit_chance : 0
>X += max(weapon1's Item.critical_hit, weapon2's Item.critical_hit) [only for actors of course]
>
>Then, X is the chance in percent.
>
>If the target is a friend and not a foe (can happen with confusion for example), the crit chance is always zero.
>
> If the target has an equipped item with Item.prevent_critical, the chance is zero as well.
>
>Also, note that a critical hit does not stack with concentration. Concentration makes the attack x2, critical makes it x3, both together also make x3 and not x6.

> For enemies and actors it's the same thing, it's stored in critical_hit/critical_hit_chance in the Actor/Enemy structure in the LDB.
>
>In general, crit is only possible for normal physical attacks (double attack is just two physical attacks one after another, which have their individual damage/crit/miss calculation)

src/game_battlealgorithm.cpp
```diff
@@ -776,7 +781,7 @@ bool Game_BattleAlgorithm::Normal::Execute() {
 
 	// Damage calculation
 	if (Utils::GetRandomNumber(0, 99) < to_hit) {
-		if (!source->IsCharged() && Utils::GetRandomNumber(0, 99) < (int)ceil(crit_chance * 100)) {
+		if (Utils::GetRandomNumber(0, 99) < (int)ceil(crit_chance)) {
```
CherryDT: Should either use floats or a multiplier for the integer (RPG_RT uses 10000 instead of 100 as far as I know, since you may have crit chances of less than 1%)

fmatthew5876: I just changed that out before you commented. The float returned is now scaled 0-100.0f and the `* 100` here is gone.

CherryDT: Okay correction, less than 1% normally not because the box is capped to "1 in 100", but still there would otherwise be no difference between "1 in 100" and "1 in 90" for example.

fmatthew5876: Wait I think I see what you mean. This algo will always clamp to a minimum of 1%, which is not great.

fmatthew5876:
I added a `Utils::PercentChance(float)` function to be used generally for drawing random numbers against percentages in the range `[0.0, 1.0]`

src/game_actor.cpp
```diff
@@ -1383,3 +1383,57 @@ bool Game_Actor::PreventsTerrainDamage() {
...
+	}
+	return nullptr;
+}
+
+const RPG::Item* Game_Actor::GetAccessory() const {
+	auto* accessory = GetEquipment(RPG::Item::Type_accessory);
+	if (accessory && accessory->type == RPG::Item::Type_accessory) {
+		return accessory;
+	}
+	return nullptr;
```
Ghabry: (note: The type checks for 1st weapon, armor, helmet and accessory are not really necessary as we sanitize the garbage away but they don't hurt.)

fmatthew5876: This check doesn't cost much but does provide us some additional safety. When you call it you know the type is correct as a post condition and can rely on that.

CherryDT:
For the record, let's make bring this back to a readable state with line breaks:

```
Cherry10/23/2018
if I understand this right, then only under this conditions.....

if target is not able to move, success probability = 100, otherwise:
success probability = (skill success probability * (min(100, all the attacker's state accuracy rates)/100)
and then only if weapon doesn't have "ignore enemy dodge rate" option is enabled:
final success probability = 100 - (100 - success probability) * (1 + 0.5 * (target agility / attacker agility - 1))
and then only if target has an equipment with "increase dodge rate against physical attacks":
final success probability -= 25
"this condition": only if the failure_message is 3 and the target is an (or all) enemy
not sure why the enemy thing though. that means that if an enemy casts a spell, the agilities never matter...?
I think the failure_message 3 classifies it as a physical attack
in a way
so you can dodge it
actually it is illogical that this applies only to skills with enemies as target, yet it checks if the target has equipment with the "increase dodge rate" option
since enemies can't wear equipment
but I guess it's just copied from normal attack logic...
But this is for skills. Now, for physical attack, most of this is the same. But instead of the skill success probability, you have:

if attacker is actor: max(weapon1's Item.hit, weapon2's Item.hit) or 90 if no weapons equipped
else: if Enemy.miss then 70, otherwise 90
```

PR #1489, opened 2018-11-07
Fix for AddState()

fmatthew5876:
When a state is added, all states with priority <= max_priority - 10
should be cleared.

### DIFF
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -381,6 +381,20 @@ void Game_Battler::AddState(int state_id) {
 	}
 
 	states[state_id - 1] = 1;
+
+	int max_priority = 0;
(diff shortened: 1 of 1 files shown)


fmatthew5876:
@CherryDT

How about state recovery in battle? We just check `states[i] >= Data::states[i].hold_turn` and then roll the auto_release probability each turn until its healed.

Is that correct?

CherryDT:
State recovery: Yes, it's correct (assuming you do the per-turn increase of the counter before checking. otherwise you'd need `>`).

Immediately cleared: Yes.

Ghabry:
What should be also checked is if states that are induced in battle but autocleared because of this behaviour are still reported in the 2k battle msg.
But out of scope here.

PR #1497, opened 2018-11-14
Battle fixes - replaces some Albeleon battle 3 commits + new fixes

fmatthew5876:
Fixes all of the following scenarios from #1449. In master you see the battle menu flicker on the screen when these things happen.

1. Party starts the battle unable to move (recoverable) -> skip menu and start battle sequence
2. Party starts the battle unable to move (not recoverable) -> skip menu and start defeat sequence
3. Turn 0 event kills the party or immobilizes (not recoverable) the party -> skip menu and start defeat sequence
4. Turn 0 event immobilizes (recoverable) party -> skip menu and start battle sequence
5. Turn 0 event which creates a message window -> display message
6. Turn 0 event which waits -> battle window is empty while waiting

Also includes some minor cleanups

### DIFF
--- a/src/scene_battle_rpg2k.cpp
+++ b/src/scene_battle_rpg2k.cpp
@@ -238,8 +238,7 @@ void Scene_Battle_Rpg2k::ProcessActions() {
 	case State_Start:
 		if (DisplayMonstersInMessageWindow()) {
 			Game_Battle::RefreshEvents();
-			SetState(State_SelectOption);
-			CheckResultConditions();
+			SetState(State_Battle);
 		}
 		break;
 	case State_SelectOption:
@@ -290,7 +289,12 @@ void Scene_Battle_Rpg2k::ProcessActions() {
 			// Everybody acted
 			actor_index = 0;
 
-			SetState(State_SelectOption);
+			// Go right into next turn if no actors controllable.
+			if (!Main_Data::game_party->IsAnyControllable()) {
+				SelectNextActor();
+			} else {
+				SetState(State_SelectOption);
+			}
 		}
 		break;
 	case State_SelectEnemyTarget: {
--- a/src/scene_battle_rpg2k.cpp
+++ b/src/scene_battle_rpg2k.cpp
@@ -808,7 +808,7 @@ void Scene_Battle_Rpg2k::SelectPreviousActor() {
 	battle_actions.back()->SetBattleAlgorithm(std::shared_ptr<Game_BattleAlgorithm::AlgorithmBase>());
 	battle_actions.pop_back();
 
-	if (active_actor->GetAutoBattle()) {
+	if (active_actor->GetAutoBattle() || !active_actor->IsControllable()) {
 		SelectPreviousActor();
 		return;
 	}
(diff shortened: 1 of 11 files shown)


fmatthew5876:
The "Correctly handle win/loss" is also needed here. This fixes a more general bug.

In #985, its not mentioned in the issue but when testing this I found a bug with battle loss not being triggered correctly when the enemy kills the player. The enemy attacks are scripted battle events and sometimes you'll see the enemy kill you but then the battle continues for a few rounds. In fact if you use the skill "Drain Sword" which steals HP, you can see the enemy kill you, you use drain sword to recover HP back up to > 0, and then the battle continues!

fmatthew5876:
"Fixes for Select..." replaces Albeleon 2.10

src/scene_battle_rpg2k.cpp
```diff
@@ -808,7 +808,7 @@ void Scene_Battle_Rpg2k::SelectPreviousActor() {
 	battle_actions.back()->SetBattleAlgorithm(std::shared_ptr<Game_BattleAlgorithm::AlgorithmBase>());
 	battle_actions.pop_back();
 
-	if (active_actor->GetAutoBattle()) {
+	if (active_actor->GetAutoBattle() || !active_actor->IsControllable()) {
```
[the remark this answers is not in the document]
fmatthew5876: `IsControllable()` is different than can act. If you have a charm or berserk state you can act but you aren't controllable.

I'm not clear on the intention from the naming, but `GetSignificantRestriction()` appears to search all inflicted states and only returns normal if no other restriction is there.

Ghabry:
How is the prefered merge order for this? between #1449 and #1450?

fmatthew5876:
Before or after #1449 is fine.

Ghabry:
looks good to me, but I will check which of these fixes also apply to 2k3 battle.

fmatthew5876:
Anymore feedback for this one? Would be helpful for testing other battle PRs if this got in.

src/scene_battle_rpg2k.cpp
```diff
@@ -162,14 +162,16 @@ void Scene_Battle_Rpg2k::SetState(Scene_Battle::State new_state) {
...
+	if (state != State_SelectEnemyTarget && state != State_SelectAllyTarget) {
+		options_window->SetVisible(false);
+		status_window->SetVisible(false);
+		command_window->SetVisible(false);
+		target_window->SetVisible(false);
+		battle_message_window->SetVisible(false);
+		item_window->SetVisible(false);
+		skill_window->SetVisible(false);
+		help_window->SetVisible(false);
+	}
```
Ghabry: > commit "Fixes for select ally/target "

could you apply the whole commit to 2k3, code looks duplicated.

Ghabry:
All other commits don't apply to 2k3 or look hard to port.

src/scene_battle_rpg2k.cpp
fmatthew5876: I started looking at that and its not so simple. The behavior of battle windows for rm2k3 is different for the 3 battle modes. What is visible and what is not is different and the logic is tangled in many places outside of this function.

I'm going to push back on this. I don't want this PR getting bogged down on rm2k3 stuff. That should be a separate PR.

fmatthew5876:
Any other comments on this one? I prefer not to start down rm2k3 rabbit hole until later PRs.

PR #1498, opened 2018-11-18
Fix Saving and Loading of panning chunks

fmatthew5876:
**Don't merge until #1502 is merged**

Fixes saving and loading games in the following conditions. Saves were tested to be compatible between `RPG_RT` and `Player` for these behaviors.

* When the screen is panned.
* While the screen is panning.
   * Loading correctly continues the panning. This is broken in `RPG_RT`.
* While screen is locked

I also tested these scenarios
* Screen behavior when event unlocks when screen is in default centered position.
* Screen behavior when event unlocks when screen is panned.
* Teleporting to same map does not change any panning state
    * If pan lock is enabled, the screen will still change position when you teleport to the same map.
* Teleporting to different map resets pan position and speed, but not pan lock.
* Panning move screen event commands still work while screen is locked.
    * Including Loading while pan is occurring.
* Multiple consecutive no wait panning move command overwrite each other. All but the final one become no-ops.
* Verified return to origin panning still works.
    * Including all 4 directions of diagonal panning

This patch uses `RPG_RT` centered panning units inside `Game_Map` and converts to 0 based for use externally. This was done to minimize the amount of change on panning code.

### DIFF
--- a/src/game_map.cpp
+++ b/src/game_map.cpp
@@ -358,10 +358,7 @@ void Game_Map::ReserveInterpreterDeletion(std::shared_ptr<Game_Interpreter> inte
 	free_interpreters.push_back(interpreter);
 }
 
-void Game_Map::ScrollRight(int distance, bool ignore_pan_lock) {
-	if (!ignore_pan_lock && IsPanLocked()) {
-		return;
-	}
+void Game_Map::ScrollRight(int distance) {
 	int x = map_info.position_x;
 	AddScreenX(x, distance);
 	map_info.position_x = x;
@@ -372,10 +369,7 @@ void Game_Map::ScrollRight(int distance, bool ignore_pan_lock) {
 	pan_x = (pan_x - distance + pan_limit_x) % pan_limit_x;
 }
 
-void Game_Map::ScrollDown(int distance, bool ignore_pan_lock) {
-	if (!ignore_pan_lock && IsPanLocked()) {
-		return;
-	}
+void Game_Map::ScrollDown(int distance) {
 	int y = map_info.position_y;
 	AddScreenY(y, distance);
 	map_info.position_y = y;
@@ -850,7 +844,6 @@ int Game_Map::CheckEvent(int x, int y) {
 
 void Game_Map::Update(bool only_parallel) {
 	if (GetNeedRefresh() != Refresh_None) Refresh();
-	UpdatePan();
 	Parallax::Update();
 	if (animation) {
 		animation->Update();
@@ -875,6 +868,7 @@ void Game_Map::Update(bool only_parallel) {
 	}
 
 	Main_Data::game_player->Update();
+	UpdatePan();
 	GetInterpreter().Update();
 
 	for (Game_Event& ev : events) {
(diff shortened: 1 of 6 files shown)


fmatthew5876:
@CherryDT did I miss any edge case?

This stuff is tricky. Could use a second pair of hands to retest the cases mentioned any more you can think of that I didn't mention.

CherryDT:
Unfortunately I don't have time for this at the moment, maybe later

Ghabry:
one game to test is this one as it had multiple pan issues before: https://github.com/EasyRPG/Player/issues/1169#issue-229497518

fmatthew5876:
I tested some HH3 intro scenes and they work with this branch.

Another test case is you can produce diagonal panning using a no-wait scroll followed by a wait scroll. For example Up (no-wait), and Right (wait). I checked and this case works in Player.

src/game_map.cpp
```diff
@@ -151,6 +150,7 @@ void Game_Map::Setup(int _id) {
 		events.emplace_back(location.map_id, ev);
 	}
 
+	// pan_state does not reset when you change maps.
```
Ghabry: This could fix a bug in Ara Fell (have no savegame around for this) but the pan after loading a new map was completely off, which requires you to navigate blind to the next exit to fix it.

Ghabry:
also agree on the note of fdela to use 256 * X to make the calc more clear

fmatthew5876:
I spent some time debugging but still haven't figured it out. The last commit breaks the test case. I need to debug further to figure out a proper fix.

fmatthew5876:
I fixed the bug in that test case. You were right in that its related to the map position not being a multiple of 256.

fmatthew5876:
I did some testing of panning with looping maps. Everything appears to work.

Ghabry:
okay, will also do another quick test tomorrow from my side if I can still break it, havn't tested the Razas intro yet ^^

Ghabry:
I think my Razas testcase is broken, but I will recheck with RPG_RT because I'm not 100% sure anymore.

Ghabry:
This also reintroduces the "Pan too far" in Razas:
Unfortunately there are not any details available how the Panning is special, wasn't ever analyzed. Was fixed as part of https://github.com/EasyRPG/Player/pull/1197

Game download: https://razas.itch.io/razas

[Save13.zip](https://github.com/EasyRPG/Player/files/2651001/Save13.zip)

Load the save and on the next map in that cutscene there is pan downwards which goes too far until it reaches the map border.

fmatthew5876:
For a more concise test, just place your party starting position at map "69: Inexplored Lands 3" at (24,15) and start a new game.

fmatthew5876:
Rebased the last change. Now it fixes both Razas and Personal Project.

fmatthew5876:
Well you're right, it is still broken. Somehow I really messed up here.

Looking into it again..

fmatthew5876:
You guys are going to hate me for this.

Previously I was converting panning from (9,7) internally to (0,0) externally in order to minimize code changes. In order to fix these panning problems and get panning implemented the right way, I've had to completely refactor how player scrolling works from scratch.

Panning has a special behavior I didn't notice before. Lets say you are 2 tiles from the right edge of the map, and then you do a panning command to move to the right by 5. What actually happens is `RPG_RT` will only pan 2 tiles and then stop. Pan current will be +2 and pan finish will be +5. As soon as you move left enough, the pan will continue until it gets to +5.

Before we tracked how much we actually panned in a local variable (actual_pan_x/y) and kept incrementing pan_current each frame. This is wrong, pan_current is stops when it hits the edge of the map.

This behavior can be confirmed by saving your game and looking at the chunks. You'll see pan_current = +2 * 256 and pan_finish = +5 * 256.

Finally, we had 2 state variables `last_remaining_move` and `last_remaining_jump`. There are several more like this in the game character heriarchy and many are removed by #1502. These variables are not saved in LSD, and thus writing algorithms this way will produce inaccuracies when saving.

The panning refactor removes these 2 local variables. Once #1502 is in and `SaveMapEvent::remaining_steps` is used, we should have frame perfect position and panning in save games.

fmatthew5876:
I did some testing of panorama. It does not appear broken by this PR.

Ghabry:
This gets my approval now, works with that fancy test, Razas and my new map loop rendering code.
I just keep the X to prevent an accidental merge before 1502

fmatthew5876:
rebased to master

PR #1501, opened 2018-11-18
Screen flash fixes

fmatthew5876:


fmatthew5876:
Continuous flashing is broken in rm2k3 with save games in a unique way.

If you save your game, and then go to end game, and load without exiting `RPG_RT`. Then loaded game will load properly and keep the continuous flash going.

If you close `RPG_RT`, restart the program, and try to load your game, the current flash will finish but then it will not repeat.

It looks like some state is kept to keep the flash going which is left in memory when you go back to the title and reload.

I'm very confused how it even works. In order for you to be able to repeat the flash, you need to store the original time amount and saturation so that the program knows how to repeat the same flash. The `flash_current_level` and `flash_time_left` values get destroyed as the flash completes.

There appears no chunk in the save data to store these extra 2 values.

@CherryDT How are they doing it in rm2k3?

fmatthew5876:
Our algorithm for shake screen looks wrong. It doesn't work at all for some values of speed and power.

@CherryDT What is the algorithm used for computing `shake_position` each frame?

src/game_map.h
```diff
@@ -59,7 +59,7 @@ namespace Game_Map {
 	/**
 	 * Disposes Game_Map.
```
Ghabry: ``@param is_load_savegame``

fmatthew5876: Fixed and rebased to master.

Ghabry:
After rebasing the "force-pushed" diff is completely useless because it shows the whole diff to master in there which means you can still sneak in new code without anybody noticing ;)

PR #1502, opened 2018-11-18
Refactor Game_Character classes 1 of 2

### LINKED ISSUE #1499: Refactor and Fix saving and loading of SaveMapEvent chunks
Player does not correctly save and load `SaveMapEvent` data. This data needs to be loaded for the player, all 3 vehicles, and any map events.

On symptom can be observed by simply by saving and loading in Player before https://github.com/EasyRPG/liblcf/pull/279 is merged. Games saved by Player with this liblcf will incorrectly save `animation_type` to 1 for the player and all map events. If you load the save game in `RPG_RT`, you'll see all events including the player walking in place (correct load of bad data). If you load the save game in `Player`, the player and map events will be standing in place (incorrect load of bad data). This is because `animation_type` is never actually loaded from the `SaveMapEvent` data and applied to the `Game_` C++ objects.

In the previous paragraph we observed bugs with `animation_type`, but its likely that other `SaveMapEvent` fields also have bugs.

`SavePartyLocation` and `SaveVehicleData` conceptually inherit from `SaveMapEvent`. This relation needs to be respected and the common `SaveMapEvent` chunks need to be saved and loaded correctly for player, vehicles, and map events. Some of the logic to deal with `SaveMapEvent` is in `Game_Event` instead of the `Game_Character` base.

Fixing this properly might require some new inheritance construct in the liblcf schema for these 3 types.

fmatthew5876:
Depends on: https://github.com/EasyRPG/liblcf/pull/282

This PR refactors the `Game_Character` class heirarchy to use the `RPG::SaveMapEventBase` class heirarchy from liblcf.

The goals are:
- [x] Correct loading of fields from LSD
- [x] Correct saving of fields to LSD
- [x] Correct behavior when saving / loading between `RPG_RT` <-> `Player`.
- [x] Binary compatibility with `RPG_RT` even when data change appears not change behavior.
- [x] Devirtualize parts of `Game_Character` where not needed
- [x] Simplify and remove duplication of logic and state
- [x] Resolve #1499


src/game_character.cpp
```diff
@@ -851,12 +850,8 @@ bool Game_Character::IsInPosition(int x, int y) const {
 	return ((GetX() == x) && (GetY() == y));
 }
 
-void Game_Character::SetOpacity(int opacity) {
-	this->opacity = min(max(0, opacity), 255);
-}
-
 int Game_Character::GetOpacity() const {
-	return opacity;
+	return (8 - GetTransparency()) * 32 - 1;
```
fmatthew5876: @CherryDT Do you agree with this calculation?

Transparency is valued from 0 to 7, with 7 being almost completely transparent. This value can't go one more step to 8 to set fully transparent.

src/game_character.cpp
```diff
@@ -810,7 +809,7 @@ void Game_Character::CancelMoveRoute() {
 }
 
 int Game_Character::GetTileId() const {
-	return tile_id;
+	return GetSpriteName().empty() ? GetSpriteIndex() : 0;
 }
```
Ghabry: > Simplify GetTileId() 
should be tested against this test case: https://github.com/EasyRPG/Player/pull/1179

src/game_vehicle.cpp
```diff
@@ -296,7 +290,6 @@ void Game_Vehicle::Update() {
 			if (!IsDescending()) {
 				if (CanLand()) {
 					SetLayer(RPG::EventPage::Layers_same);
```
Ghabry: > Remove driving

Issue worth testing: https://github.com/EasyRPG/Player/issues/1349

fmatthew5876:
Unless I find more issues to fix, I think this PR is nearly done. It's a lot of commits, so once the liblcf dependencies are in it could be broken up if needed.

Here are some save chunks I'm still unsure about:

* `map_id` - For some reason `RPG_RT` always sets `map_id = 0` for all map events. The map_id for the player and vehicles is set correctly. I'm hesitant to do this as someone could call GetMapId() on a game_event in the code expecting the right Id and getting surprised.
* `processed` - This chunk is used somehow by `RPG_RT` to resolve movement, particularly when you have 2 events that want to move and event A wants to take event B's spot. It looks like some kind of way to not double compute event movement. How do we handle this situation in Player? At some point we should maybe also use this flag to ensure our saved games don't break in mysterious ways between RPG_RT and Player.
* `route_through` - Move commands that set/unset through also set this flag. I set it for binary compatibility on LSD but I don't use it. Why does this exist? What edge case do we need to differentiate between `through` and `route_through`? If there is such a case, fixing in it Player would likely fix esoteric event code bugs in some game.

fmatthew5876:
For `processed`, I decided to just write the flag so it looks like `RPG_RT`.

My observation is that `processed` will always be 1 in `RPG_RT` saves (except for vehicles you haven't seen yet). My assumption is that this is because you can never save at a time when events are resolving their moves and haven't finished processing.

Is that assumption correct?

src/game_vehicle.cpp
Ghabry: Doesn#t break the linked testcase and also not the Yume2kki music room.

carstene1ns:
jenkins: you know what to do, test this please :heart_eyes:

fmatthew5876:
@Ghabry and @carstene1ns 

Would you like me to break this one up into separate PRs?

carstene1ns:
> game_character.h(1100): error C2589: '(': illegal token on right side of '::'

The weird build error happens, because windows header has own macros for min() and max().
Needs [this](https://github.com/EasyRPG/Player/blob/master/src/sdl2_ui.cpp#L29-L31) stuff [here](https://github.com/EasyRPG/Player/blob/master/src/filefinder.cpp#L31).

Ghabry:
> Would you like me to break this one up into separate PRs?

@fmatthew5876 
No, I think it's fine. Because is mostly code simplification because of the new inheritance feature in liblcf I think we can handle this.
Will tell you when I find anything that could be extracted.

src/utils.h
```diff
@@ -234,6 +234,17 @@ namespace Utils {
 	 */
 	std::string ReplacePlaceholders(const std::string& text_template, std::vector<char> types, std::vector<std::string> values);
 
+	/**
+	 * @return value clamped between min and max
+	 */
+	template <typename T>
+		T Clamp(T value, const T& minv, const T& maxv);
```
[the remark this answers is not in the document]
fmatthew5876: Fixed.

fmatthew5876:
jenkins: test this please

Ghabry:
I'm requesting a split after the commit "Remove Game_Vehicle::driving" with the commit "Add compatibility hack for old Player save games" cherry-picked. (using commit msg instead of hashes here, less chaotic)
Afterwards the corner-case "this is how RPG_RT does it" stuff begins which needs more looking and testing.

Additional 3 commits also cherry-pickable imo from more down the commit log (if possible conflict free):
 - Remove unused Game_Character::animation_id 
 - Use SaveMapEventBase::anim_frame … 
 - Use SaveMapEventBase::anim_paused 

Only stuff I found: "Encapsulte" in one commit message.
And there are multiple doc comments missing the @param field.

Ghabry:
Well, maybe should also give some 👍 . Nice refactor, I found this huge code duplication with virtual inheritance ugly since I added it. xD

And great to see more LSD compat problems solved.

fmatthew5876:
I've rebased this and removed the commits after the requested break point. I've cherry-picked all the ones asked for, including `Utils::Clamp()` as that's needed to make this compile on windows.

Ghabry:
Jenkins: test this please  (vcpkg 😘 )

fmatthew5876:
This one starts to become a blocker now. For future LSD work.

src/player.cpp
```diff
@@ -772,6 +772,15 @@ static void OnMapSaveFileReady(FileRequestResult*) {
 		Main_Data::game_data.party_location.position_x,
 		Main_Data::game_data.party_location.position_y
 		);
+
+	// Compatibility hack for old EasyRPG Player saves.
```
carstene1ns: Maybe we should add some version string, e.g. `(pre 0.6)` here for reference.

fmatthew5876: Great idea. Fixed.

src/game_character.cpp
CherryDT: Almost. It is a bit weirder. It is `255 - GetTransparency() * 30`. Yes, **30** and not 32.

Issue #1444, opened 2018-10-14
Row flag seems to be in a weird state when loading from RPG_RT save


CherryDT:
Tested on Windows with continous builds.

Load a 2k3 game from an RPG_RT save. Try changing a battler's row in the main menu. Note that the first change does nothing, only the second one works, and from then on that hero behaves normally.

PR #1507, opened 2018-11-24
Fix SaveActor::row

fmatthew5876:
Depends on: https://github.com/EasyRPG/liblcf/pull/283
Resolves: #1444


carstene1ns:
jenkins: test this please.

carstene1ns:
I would like to have some code comment about why the count logic in "Row" command exists.
e.g. "Do not allow entire party in back row" or "Keep at least one member in front row"

PR #1508, opened 2018-11-24
Fixes for states

### LINKED ISSUE #1504: RM2000: Battle state bugs
- [x] States inflicted recovery turn count doesn't respect whether target has acted yet

Create an enemy that always inflicts a state with (1 turn, 100%) recovery. If the state hits before the actor's turn, the actor will recover the next turn (correct). If the state hits after the actor's turn, the actor will recover the next turn (wrong, should be turn after next)

fmatthew5876:
Resolves #1504

### DIFF
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -482,16 +482,9 @@ static bool non_permanent(int state_id) {
 
 void Game_Battler::RemoveBattleStates() {
 	std::vector<int16_t>& states = GetStates();
-
-	// If death is non-permanent change HP to 1
-	if (IsDead() &&
-		non_permanent(1)) {
-		RemoveState(1);
-	}
-
-	for (size_t i = 1; i < states.size(); ++i) {
+	for (size_t i = 0; i < states.size(); ++i) {
 		if (non_permanent(i + 1) || ReaderUtil::GetElement(Data::states, i + 1)->auto_release_prob > 0) {
-			states[i] = 0;
+			RemoveState(i + 1);
 		}
 	}
 }
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -38,9 +38,8 @@ Game_Battler::Game_Battler() {
 }
 
 bool Game_Battler::HasState(int state_id) const {
-	const std::vector<int16_t> states = GetInflictedStates();
-
-	return (std::find(states.begin(), states.end(), state_id) != states.end());
+	auto* state_turns = ReaderUtil::GetElement(GetStates(), state_id);
+	return state_turns && *state_turns > 0;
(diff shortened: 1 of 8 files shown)


src/game_battler.cpp
```diff
@@ -38,9 +38,8 @@ Game_Battler::Game_Battler() {
 }
 
 bool Game_Battler::HasState(int state_id) const {
-	const std::vector<int16_t> states = GetInflictedStates();
-
-	return (std::find(states.begin(), states.end(), state_id) != states.end());
+	auto& states = GetStates();
+	return state_id >= 1 && state_id <= states.size() && states[state_id-1] > 0;
```
Ghabry: We also have a sanity checking function array access function in ReaderUtil:
```
ReaderUtil::GetElement(states, state_id);
```
Is that suitable here?

Ghabry:
> Bugfix: Don't reset turns count when we already  have the state

Didn't Cherry mention the turn count is indeed reset?

https://github.com/EasyRPG/Player/pull/1484#issuecomment-436191120

fmatthew5876:
> > Bugfix: Don't reset turns count when we already  have the state
> 
> Didn't Cherry mention the turn count is indeed reset?
> 
> [#1484 (comment)](https://github.com/EasyRPG/Player/pull/1484#issuecomment-436191120)

Hmm. I tested this by making some enemy that only does a single attack skill which is 100% hit rate and inflicts a state. I observed that when the actor already has the state, the skill always fails.

Maybe that's some "already has state" logic which is specific to skill success/failure and forcing the state to be added, say by an event would cause the turn count to reset to 1?

I will do some further testing to verify the behavior.

Ghabry:
Maybe theres a difference between weapons and skills when inflicting states.

src/game_battler.cpp
fmatthew5876: If the `states` vector was not the same size as the `Data::states` then using `GetElement()` would be wrong. It could even result in a segfault.

My latest liblcf patch resizes `states` to always match `Data::states` so probably `GetElement()` is ok now, but maybe we will change the design later?

Doing it this way at least is bulletproof.

Ghabry: That's why I passed "states" (which is the local variable in yor code) to GetElement, not "Data::states".
Also you don't resize the states array of enemies, so an unchecked state access will still segfault (we had this before).

fmatthew5876: Oh I see now. I didn't realize just how general `GetElement()` is.

Yes I agree with you then, this is much better.

fmatthew5876:
Closing this for now. Will take another look once all Sormat battle PRs done.

PR #1511, opened 2018-11-30
Fix to_hit calculation - Previously Albeleon 2.34 + Weapon sp cost fix

fmatthew5876:
From Cherry:

>if target is not able to move, success probability = 100, otherwise: success probability = (skill success probability * (min(100, all the attacker's state accuracy rates)/100) and then only if weapon doesn't have "ignore enemy dodge rate" option is enabled: final success probability = 100 - (100 - success probability) * (1 + 0.5 * (target agility / attacker agility - 1)) and then only if target has an equipment with "increase dodge rate against physical attacks": final success probability -= 25 "this condition": only if the failure_message is 3 >and the target is an (or all) enemy not sure why the enemy thing though. that means that if an enemy casts >a spell, the agilities never matter...? I think the failure_message 3 classifies it as a physical attack in a way so you can dodge it actually it is illogical that this applies only to skills with enemies as target, yet it checks if the target has equipment with the "increase dodge rate" option since enemies can't wear equipment but I guess it's just copied from normal attack logic...
But this is for skills. Now, for physical attack, most of this is the same. But instead of the skill success probability, you have:
>
>if attacker is actor: max(weapon1's Item.hit, weapon2's Item.hit) or 90 if no weapons equipped else: if Enemy.miss then 70, otherwise 90

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -1284,6 +1284,15 @@ int Game_Actor::GetBattleAnimationId() const {
 }
 
 int Game_Actor::GetHitChance() const {
+	auto* weapon1 = GetWeapon();
+	auto* weapon2 = Get2ndWeapon();
+	if (weapon1 && weapon2) {
+		return std::max(weapon1->hit, weapon2->hit);
+	} else if(weapon1) {
+		return weapon1->hit;
+	} else if(weapon2) {
+		return weapon2->hit;
+	}
 	return 90;
 }
 
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -1451,6 +1451,13 @@ const RPG::Item* Game_Actor::GetAccessory() const {
 	return nullptr;
 }
 
+bool Game_Actor::AttackIgnoresEvasion() const {
+	auto checkEquip = [](const RPG::Item* item) {
+		return item && item->ignore_evasion;
+	};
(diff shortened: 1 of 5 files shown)


Ghabry:
Is ``HasHalfSpCost```used by anything?
We already have ``GetSpCostModifier``.

fmatthew5876:
> 
> 
> Is `HasHalfSpCost`used by anything?
> We already have `GetSpCostModifier`.

I missed that we have `GetSpCostModifier()`. I just added `HasHalfSpCost()` to match the style of the 3 other armor modifiers. I can change the sp cost cost to use the new function.

fmatthew5876:
I'm glad you said something, our weapon sp cost logic was actually broken. I've fixed it now.

fmatthew5876:
Added the other sp cost commit. Easy to test. Just attack with a weapon with sp cost and "attack all" attribute.

fmatthew5876:
I don't plan to add anymore commits to this so it's ready for review. Once this #1519, and #1450 are in I can rebase and finish refactoring battle 7. Then we will finally be done with these!

fmatthew5876:
Rebased to master

fmatthew5876:
jenkins: test this please

fmatthew5876:
jenkins: test this please

PR #1515, opened 2018-12-01
Fix issue 1514: Fixup wasn't called for last actor (off-by-1)

### LINKED ISSUE #1514: Makoto Mobius: player character is not visible after loading savegame
The game [Makoto Mobius](http://terriball-tl.tumblr.com/post/102600051034/makoto-mobius) has a reproducible issue while loading savegame.

To reproduce, skip intro with fast forward and accept all conversation, then enter the house and talk to the bunny to save. Once saved, load the saved game and the player character will become invisible.

Here's a saved game with the issue: [makoto-mobius-savegame.zip](https://github.com/EasyRPG/liblcf/files/2634978/makoto-mobius-savegame.zip)

Here's another saved game with Player from PR 1502: [makoto-mobius-savegame-pr1502.zip](https://github.com/EasyRPG/Player/files/2636063/makoto-mobius-savegame-pr1502.zip)

Currently, Player does not log when a savegame is saved or loaded, this would be nice to understand better what users are sending via bug reports.

Ghabry:
Fix #1514

### DIFF
--- a/src/game_actors.cpp
+++ b/src/game_actors.cpp
@@ -49,7 +49,7 @@ void Game_Actors::Fixup() {
 		}
 	}
 
-	for (size_t i = 1; i < data.size(); ++i) {
+	for (size_t i = 1; i <= data.size(); ++i) {
 		GetActor(i)->Fixup();
 	}
 }


Ghabry:
fdelapena raised the valid concern that there could be more such errors hidden and made a regexp search for that pattern.

https://xkcd.com/208/

I also did a search now:

```
ag "for.*1;.*<[^=].*"

scene_debug.cpp
381:                    for (int i = 1; i < 10; i++){

input.cpp
141:    for (unsigned i = 1; i < 10; i++) {

window_battlestatus.cpp
249:                    for (int i = 1; i < item_max; i++) {

game_interpreter.cpp
563:    for (unsigned index_temp = index + 1; index_temp < list.size(); ++index_temp) {

game_battler.cpp
492:    for (size_t i = 1; i < states.size(); ++i) {

image_bmp.cpp
83:     for (int i = 1; i < num_colors; i++) {

game_actor.cpp
528:    for (int i = 1; i < (int)exp_list.size(); ++i) {

player.cpp
404:    for (int i = 1; i < argc; ++i) {

bitmap.cpp
513:    for (int i = 1; i < PIXMAN_MAX_INDEXED; i++)
```

The only questionable is in game_battler.cpp but by looking at the context: https://github.com/EasyRPG/Player/blob/master/src/game_battler.cpp#L492 this is correct. Is just an optimisatition because the death-state was already handled a few lines above.

PR #1519, opened 2018-12-02
Correct battle turn order in rpg2k battle - previously Albeleon 2.42

fmatthew5876:
From Cherry:
> RPG_RT defines something like a "battle agility" for each turn which is AGI+X. X is a random number below (AGI/4 + 3). However, if the attack is a basic attack with a weapon that has the "Start with Initiative" checkbox, this puts it into a class of its own which follows the same rules but can rival only other attacks with such weapons (I think this is done by simply adding a huge number like 100000 to the value in such a case). The turn order is then decided by this "battle agility", high to low.

### DIFF
--- a/src/scene_battle_rpg2k.cpp
+++ b/src/scene_battle_rpg2k.cpp
@@ -449,7 +449,7 @@ bool Scene_Battle_Rpg2k::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBase
 				if (action->GetTarget()) {
 					if (action->GetTarget()->GetType() == Game_Battler::Type_Enemy) {
 						action->PlayAnimation();
-					} else if (action->GetTarget()->GetType() == Game_Battler::Type_Ally && action->IsSoundAnimationOnAlly()) {
+					} else if (action->GetTarget()->GetType() == Game_Battler::Type_Ally && action->GetType() == Game_BattleAlgorithm::Type::Skill) {
 						action->PlaySoundAnimation(false, 20);
 					}
 				}
--- a/src/scene_battle_rpg2k.cpp
+++ b/src/scene_battle_rpg2k.cpp
@@ -849,24 +849,20 @@ void Scene_Battle_Rpg2k::SelectPreviousActor() {
 	SetState(State_SelectActor);
 }
 
-static bool BattlerSort(Game_Battler* first, Game_Battler* second) {
-	if (first->HasPreemptiveAttack() && second->HasPreemptiveAttack()) {
-		return first->GetAgi() > second->GetAgi();
-	}
-
-	if (first->HasPreemptiveAttack()) {
-		return true;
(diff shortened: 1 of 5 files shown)


fmatthew5876:
I've tested this and confirmed the behavior matches RPG_RT

fmatthew5876:
jenkins: test this please

Ghabry:
This technique for "::Type" propagation through the base constructor is pretty neat btw.

src/scene_battle_rpg2k.cpp
```diff
@@ -849,24 +849,19 @@ void Scene_Battle_Rpg2k::SelectPreviousActor() {
...
-	return first->GetAgi() > second->GetAgi();
-}
-
 void Scene_Battle_Rpg2k::CreateExecutionOrder() {
-	std::sort(battle_actions.begin(), battle_actions.end(), BattlerSort);
+	// Define random Agility. Must be done outside of the sort function because of the "strict weak ordering" property, so the sort is consistent
+	for (auto battler : battle_actions) {
+		int battle_order = battler->GetAgi() + Utils::GetRandomNumber(0, battler->GetAgi() / 4 + 3);
+		if (battler->GetBattleAlgorithm()->GetType() == Game_BattleAlgorithm::Type::Normal && battler->HasPreemptiveAttack()) {
+			battle_order += 100000;
```
Ghabry: Could you add a comment that this is an "arbitrary huge number"?

fmatthew5876:
> 
> 
> This technique for "::Type" propagation through the base constructor is pretty neat btw.

This commit is from battle 5. I like this approach because you can make `GetType()` non-virtual and optimize a lot of usage.

I've added the requested comment.

src/scene_battle_rpg2k.cpp
```diff
@@ -849,24 +849,20 @@ void Scene_Battle_Rpg2k::SelectPreviousActor() {
...
-	return first->GetAgi() > second->GetAgi();
-}
-
 void Scene_Battle_Rpg2k::CreateExecutionOrder() {
-	std::sort(battle_actions.begin(), battle_actions.end(), BattlerSort);
+	// Define random Agility. Must be done outside of the sort function because of the "strict weak ordering" property, so the sort is consistent
+	for (auto battler : battle_actions) {
+		int battle_order = battler->GetAgi() + Utils::GetRandomNumber(0, battler->GetAgi() / 4 + 3);
+		if (battler->GetBattleAlgorithm()->GetType() == Game_BattleAlgorithm::Type::Normal && battler->HasPreemptiveAttack()) {
+			// This is an arbitrarily large number to separate preemptive vs non-premptive battlers
```
carstene1ns: typo. `non-preemptive`

fmatthew5876: Fixed

Issue #1443, opened 2018-10-14
Small UI discrepancies


CherryDT:
Tested on Windows with continous builds.

**Save screen**

- [X] For one frame, all saves seem to be selected at once before the latest one is selected
- [X] The menu jumps instead of scrolling
- [x] ~~The menu allows going from 1 to 15 by going back (and the other way round) while RPG_RT doesn't~~
- [X] The menu is missing the blinking scrolling arrows
- [X] The text of the hero name and stats is not aligned correctly with the save title

**Main menu**

- [ ] When the row is "front", the faces are centered as they were in 2k, although RPG_RT shows them shifted to the left (however, in my personal opinion, centered looks better there anyway)

**Skill screen**

- [X] In the window below the info window (first hero's status), the text is positioned a bit too high

**Equipment screen**

- [x] The arrows from old to new value are not displayed

**Status screen**

- [X] The stats are not aligned - the one's digit for the values for ATK/DEF/etc. should be aligned with the one's digit of HP/MP/EXP (left to the slash)

Ghabry:
just three quick comments, more in a week:

> For one frame, all saves seem to be selected at once before the latest one is selected

One of Sormats commits fixes this

> The menu allows going from 1 to 15 by going back (and the other way round) while RPG_RT doesn't

Imo features that can't break games are fine. :)

> The arrows from old to new value are not displayed

iirc I decided to completely omit them because what is shown depends on the translation of the engine ^^'.

fmatthew5876:
What needs to be done to get back the equipment menu arrows? It's pretty ugly and lacking without them.

carstene1ns:
original issue updated to reflect current development.

carstene1ns:
Can someone confirm, if this has been fixed:
> The arrows from old to new value are not displayed

fmatthew5876:
> 
> 
> Can someone confirm, if this has been fixed:
> 
> > The arrows from old to new value are not displayed

No it hasn't.

@Ghabry Can you enlighten us on whats the roadblock with this one?

It really seems the last piece missing from UI

Ghabry:
The "arrow" used differs based on the engine translation so we decided to simply omit it completely.

Will check the japanese version and report back

Ghabry:
Well English uses > and Japanese 2k uses ->, so doesn't really matter, no big difference.

[...]

PR #1520, opened 2018-12-02
More ui fixes

fmatthew5876:
Fixes more from #1443

### DIFF
--- a/src/scene_file.cpp
+++ b/src/scene_file.cpp
@@ -136,7 +136,6 @@ void Scene_File::Refresh() {
 	for (int i = 0; i < (int)file_windows.size(); i++) {
 		Window_SaveFile *w = file_windows[i].get();
 		w->SetActive(i == index);
-		w->SetVisible(i >= top_index -1 && i <= top_index + 3);
 		w->Refresh();
 	}
 }
--- a/src/scene_file.cpp
+++ b/src/scene_file.cpp
@@ -38,11 +38,24 @@ Scene_File::Scene_File(std::string message) :
 	index = 0;
 }
 
+static std::unique_ptr<Sprite> makeBorderSprite(int y) {
+	auto bitmap = Bitmap::Create(SCREEN_TARGET_WIDTH, 8, Cache::System()->GetBackgroundColor());
+	auto sprite = std::unique_ptr<Sprite>(new Sprite());
+	sprite->SetVisible(true);
+	sprite->SetZ(Priority_Window + 1);
+	sprite->SetBitmap(bitmap);
+	sprite->SetX(0);
+	sprite->SetY(y);
+	return sprite;
+}
+
 void Scene_File::Start() {
 	// Create the windows
 	help_window.reset(new Window_Help(0, 0, SCREEN_TARGET_WIDTH, 32));
 	help_window->SetText(message);
 	help_window->SetZ(Priority_Window + 1);
+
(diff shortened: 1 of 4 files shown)


Ghabry:
Quick Note: A misalignment of ~2 pixels on y could be also because of the built-in font used because our naive pixel font renderer doesn't use a baseline and not a problem with the scene.

There are two fonts. you can switch between them via the "system" graphic setting in the editor.

Best way to test is with a glyph which takes the whole 12 pixels of height.
Here is the pixel font: https://github.com/EasyRPG/Player/tree/master/resources/shinonome

fmatthew5876:
We seem to hard code a y offset of 2 in many window classes

Ghabry:
k, then I need to check this. Don't know which ones but for some windows I measured the font position via screenshots and paint ^^

fmatthew5876:
Take a look at Window_Help. It's almost exactly like this. 

Also my last PR for status had y = 2

fmatthew5876:
I added animation for the file menu slots but there is a problem. RPG_RT keeps a small blue rectangle above and below the 3 displayed file menu boxes.

Whats the best way to preserve this behavior in Player? Do we have a mechanism to define a clipping rectangle so that as the file box animates up/down it wont cover the border? 

If not, I guess the other way would be to actually render 2 blue rectangles ontop of the windows?

fmatthew5876:
How can I draw single color `Sprite()`?

I've tried this but I don't see anything. What am I missing?

```
auto* border = new Sprite();
border->SetSrcRect(Rect(0, 32, SCREEN_TARGET_WIDTH, 8));                      
border->SetVisible(true);
border->SetZ(Priority_Window + 10);
border->SetBitmap(Bitmap::Create(0, 32, SCREEN_TARGET_WIDTH, 8));
border->GetBitmap()->Fill(Color(255,255,255,255));
```

Ghabry:
SetSrcRect is not needed and GetBitmap is const, you must draw on the Bitmap before you give it to the sprite.

Ghabry:
You can use "SpriteRef". SetSrcRect is not needed, same for Update. You invoke the wrong Bitmap::Create constructor, guess this invokes width, height, transparent, bpp. Guess you want to invoke width, height, Color

Ghabry:
and the system background is Cache::System()->GetBackgroundColor()

fmatthew5876:
I added the border sprites and now the animation matches `RPG_RT`.

src/scene_file.cpp
```diff
@@ -109,23 +124,31 @@ void Scene_File::Start() {
...
+
 	index = latest_slot;
 
 	Refresh();
 	Update();
 }
 
 void Scene_File::Refresh() {
-	for (unsigned int i = 0; (size_t) i < file_windows.size(); i++) {
+	for (int i = 0; i < int(file_windows.size()); i++) {
```
carstene1ns: this looks really ugly.

fmatthew5876: Changed it to a cast. The loop variable should be signed as it is compared to signed values inside the loop.

fmatthew5876:
It's actually animating the windows super fast. The delta in space is x15 but the delta in time is the same.

Using PageUp / PageDown also scrolls 3 times faster

I like this better than an instant transition. All motions up, down, pageup, pagedown all have the same speed and delay.

PR #1523, opened 2018-12-04
Implement Get MIDI Ticks for non-MIDI formats by returning position in sec

### LINKED ISSUE #1517: tdf2018wildcompodemo: interpreter hangs or skips all map sequences
#### Name of the game:
It does not contain a game title, but the project name on GitHub and in the readme file is tdf2018wildcompodemo.

#### Player platform:
Any, probably...

Not needed

This amazing demo ([live video](https://twitter.com/njHvYnA6vIU9vkV/status/1068810920795635713)) has been created by @katai5plate and the scripted full source code (MIT licensed) is available at [katai5plate/tdf2018wildcompodemo](https://github.com/katai5plate/tdf2018wildcompodemo) on GitHub.

It hangs in the first scene after less than 40 seconds at normal speed.
An online version has been uploaded for convenience to the web player: [stable](https://easyrpg.org/play/?game=tdf2018wildcompodemo), [master](https://easyrpg.org/play/master/?game=tdf2018wildcompodemo).

Some few BMPs have been converted to PNG, these have been recompressed with zopfli and MP3 music converted to 24 kbps Opus for faster async download.

Additionally to this hang, player shows a warning instead of a debug message due to the empty title (it could be considered as a bug, as RPG_RT ignores this).

Ghabry:
Implemented for mpg123, libvorbisfile and libopusfile.
Works for EmptyAudio and SDLAudio by approximating it through the frame count.

When no BGM is playing the returned value is 0.

Fix #1517

### DIFF
--- a/src/audio.cpp
+++ b/src/audio.cpp
@@ -20,6 +20,7 @@
 #include "system.h"
 #include "baseui.h"
 #include "player.h"
+#include "graphics.h"
 
 AudioInterface& Audio() {
 	static EmptyAudio default_;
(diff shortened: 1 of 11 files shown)


Ghabry:
Couldn't figure out how to do this with libsndfile, don't see this exposed through the API :/

fdelapena:
> Couldn't figure out how to do this with libsndfile, don't see this exposed through the API :/

Perhaps the sf_vio_tell implementation which uses ftell may help somehow, at least knowing the WAV type for PCM and sample rate it could be guessed with acceptable precision. For MS ADPCM it is also possible because the bitrate is constant (somebody says around 352 kbps). However, MS ADPCM for music is not widely used because when MS ADPCM landed in RPG_RT (1.50/1.05) it also added support for MP3.

PR #1526, opened 2018-12-05
Fixes for skills and attributes

fmatthew5876:
Contains a commit from #1394. Also Sormat battle commit 3.16, which somehow got lost in the rebasing.

Sormat's observation that items that trigger skills always set the source battler as the highest level member of the party. If there are multiple members of the party with the highest level, the first is selected. From testing I found his observations to be correct, with the additional caveat that the source also has to be able to act (not dead or has a nomove state).

In `RPG_RT`, this logic is applied a bit too heavy handedly and has a bug. Skills with usage of "Self" also get their source set to the highest level member of the party. Since the target is self, that means the target also gets forced to the highest level member! When you use such an item the first actor is highlighted, you can't select another actor, and the effect is always on the highest leveled party member regardless.

I opted to break compatibility for self skills here. Instead I let you use them on anyone, with the source being that same actor instead of the highest level member.

I tested all of this by changing the levels of party members and hacking their SPI stats really high or low to determine which character was being used as the source.

The idea behind the highest level actor logic makes sense but its not very good. For example, if your highest level actor is a fighter type with low SPI and then you use an item which cases a healing spell that has SPI effect, it'll be wasted as your high level fighter's SPI gets used.

### DIFF
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -282,15 +282,25 @@ bool Game_Battler::UseItem(int item_id) {
 		return true;
 	}
 
-	switch (item->type) {
-		case RPG::Item::Type_weapon:
-		case RPG::Item::Type_shield:
-		case RPG::Item::Type_armor:
-		case RPG::Item::Type_helmet:
-		case RPG::Item::Type_accessory:
-			return item->use_skill && UseSkill(item->skill_id, Main_Data::game_party->GetHighestLeveledActor());
-		case RPG::Item::Type_special:
-			return UseSkill(item->skill_id, Main_Data::game_party->GetHighestLeveledActor());
+	bool do_skill = RPG::Item::Type_special
+		|| (item->use_skill && (
+				item->type == RPG::Item::Type_weapon
+				|| item->type == RPG::Item::Type_shield
+				|| item->type == RPG::Item::Type_armor
+				|| item->type == RPG::Item::Type_helmet
+				|| item->type == RPG::Item::Type_accessory
+				)
+				);
+
+	if (do_skill) {
+		auto* skill = ReaderUtil::GetElement(Data::skills, item->skill_id);
+		if (skill != nullptr) {
+			Game_Battler* source = this;
+			if (skill->scope != RPG::Skill::Scope_self) {
+				source = Main_Data::game_party->GetHighestLeveledActor();
+			}
+			UseSkill(item->skill_id, source);
+		}
 	}
 
 	return false;
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -326,10 +326,6 @@ bool Game_Battler::UseSkill(int skill_id, const Game_Battler* source) {
 
 		// Calculate effect:
 		float mul = GetAttributeMultiplier(skill->attribute_effects);
-		if (mul < 0.5f) {
-			// Determined via testing, the heal is always at least 50%
-			mul = 0.5f;
-		}
 
 		int effect = skill->power;
 		if (source != nullptr) {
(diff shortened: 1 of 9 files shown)


src/game_battler.cpp
```diff
@@ -315,6 +316,25 @@ bool Game_Battler::UseSkill(int skill_id) {
...
+		effect = source == nullptr ? skill->power : skill->power +
+			source->GetAtk() * skill->physical_rate / 20 +
+			source->GetSpi() * skill->magical_rate / 40;
+		effect *= mul;
+
+		effect += (effect * Utils::GetRandomNumber(-skill->variance * 10, skill->variance * 10) / 100);
+
+		if (effect < 0)
+			effect = 0;
+
```
fmatthew5876: I really don't like to see this logic duplicated here and in the battle code.

fmatthew5876: We should refactor this skill and item effect logic in some common place. I will do this only after battle PRs are merged.

src/game_battler.cpp
```diff
@@ -252,27 +286,38 @@ bool Game_Battler::UseItem(int item_id) {
...
+				|| item->type == RPG::Item::Type_shield
+				|| item->type == RPG::Item::Type_armor
+				|| item->type == RPG::Item::Type_helmet
+				|| item->type == RPG::Item::Type_accessory
+				)
+				);
+
+	if (do_skill) {
+		auto* skill = ReaderUtil::GetElement(Data::skills, item->skill_id);
+		if (skill != nullptr) {
```
Ghabry: This silent rejection of the skill can hide a game bug.

Needs a Output::Warning "IsSkillUsable: Can't use item %d with invalid skill %d"

src/scene_actortarget.cpp
```diff
@@ -52,18 +52,33 @@ void Scene_ActorTarget::Start() {
 		if (!item) {
 			Output::Warning("Scene ActorTarget: Invalid item ID %d", id);
 			Scene::Pop();
+			return;
```
Ghabry: good find!

fmatthew5876:
Leaving a note that I need to test what happens when highest level actors cannot use the item. What if Target cannot use?

src/game_battler.cpp
fmatthew5876: This has been added

fmatthew5876:
Turns out item usability plays into item skills.

If the highest level actor can't use the item, then the next highest level actor is chosen as the "source".

I've moved the clamping damage values commit to #1452 

This is ready now. I don't have anything more to add.

carstene1ns:
jenkins: test this please (newlib check).

Ghabry:
At least for "One Ally" and "Whole Party" it doesn't seem to be "Highest Leveled Actor".

When the Skill is of Magical Element and you set it to E (0%) for 3 of the 4 party members then only the one member who doesn't have E will be healt by the item.

Because that one member isn't the highest leveled actor the caster must be the actor himself.

fmatthew5876:
> When the Skill is of Magical Element and you set it to E (0%) for 3 of the 4 party members then only the one member who doesn't have E will be healt by the item.

But that's because elements are based on the target, not the caster. Try it with magical influence and super high/low SPI stats. That's how I tested it and it really is highest level actor.

Also we have this 50% thing in our healing skill effect logic. Maybe its wrong?

Ghabry:
ah that makes sense, but something must be wrong in our healing calculation because the healing should be 0. I will try with that 50% patched out.

Ghabry:
I determined via testing that this 50% is nonsense in both map and battle...

fmatthew5876:
Removed the the 50% thing.

I wouldn't worry too much about perfect accuracy at this stage. We have still more battle PRs which adjust the skill effect logic. After all are in I'd like to test for accuracy and merge the skill effect logic from battle and menu to one place.

PR #1527, opened 2018-12-05
Tweak manpage installation

AMDmi3:
- Manpages are needed not only on Linux but on all UNIX systems, including at least *BSDs
- Most non-linux systems install manpages under ${PREFIX}/man, not ${PREFIX}/share/man


fmatthew5876:
On my FreeBSD system, the base uses `/usr/share/man`. Ports and packages however appear to use `/usr/local/man`.

Is there no standard way to detect manpage path from cmake? Do all cross platform projects really have to add platform specific hacks like this?

Unless theres a better cmake way to do this manpath stuff, LGTM.

AMDmi3:
Just making MANDIR variable so it can be tweaked from outside would be nice. Or https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html

fmatthew5876:
> Just making MANDIR variable so it can be tweaked from outside would be nice. Or https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html

I think that would be a great idea. This PR logic can be the default, but it must be overridable. In particular that allows package maintainers to deal with these kinds of problems themselves.

Ghabry:
oh I didn't knew about the MANDIR variable, otherwise I would have used it.

fmatthew5876:
Does cmake use the same or similar convention as autotools for this? In a standard configure script you have `MANDIR`, `BINDIR`, `LOCALSTATEDIR` and many more for customizing installation on unix systems. Package maintainers just script their packages to use this interface and its easy and standard.

There must be a standard cmake convention for this??

AMDmi3:
Lemme come with simpler customizable variant

AMDmi3:
See https://github.com/EasyRPG/Player/pull/1533

Ghabry:
@fmatthew5876 

I think the "standard" is to load the script "GNUInstallDirs" and then use CMAKE_INSTALL_MANDIR which can be overwritten with a -D argument. Same applies to CMAKE_INSTALL_BINDIR which we already use (and CMAKE_INSTALL_LIBDIR in liblcf).

https://chromium.googlesource.com/external/github.com/g-truc/glm/+/refs/heads/master/cmake/GNUInstallDirs.cmake#135

carstene1ns:
Yep, this is what I used elsewhere too.

Ghabry:
I close this one in favor of #1533

PR #1529, opened 2018-12-05
Always clamp HP and SP on level change

fmatthew5876:
From #1394

* Ensures all calls to SetLevel() and ChangeLevel() will clamp
  hp/sp in all cases where max hp/sp goes down.
* Includes level down and level up but hp/sp is less
* Inspiried by Albeleon's commit, but we also fix it for SetLevel()


Ghabry:
Jenkins: Test this please  (for some reason the Wii PR builder wasn't triggered for the last builds, wtf)

Ghabry:
I'm not aware of any game with such a silly HP/SP curve, hopefully there isn't any.

Issue #1383, opened 2018-06-25
Missing or unloadable images in VH01 in EasyRPG for Windows


Yuuyatails:
#### Attach files (as a .zip archive or link them)
 [easyrpg_log.zip](https://github.com/EasyRPG/Player/files/2132072/easyrpg_log.zip)

#### Describe the issue in detail and how to reproduce it: 
The EasyRPG player will fail to load the correct image and shows the empty screen if the filename of said image files or the file paths contains non-English texts. It happens when the game tries to loads those files.

Note that this is Windows-specific issue, as I don't encounter such issue on the Android version of the EasyRPG player on my phone.

Ghabry:
Thanks for the report. This is probably a Unicode normalization problem again. Could be likely solved together with #839 ("Normalize").

Yuuyatails:
So it's a known issue, huh...

It might take a long time to fix it up, I guess.

Yuuyatails:
Unfortunately, the latest test build (811722bfd77793cc6c63dde757cfd39fcbc3df42) didn't solved the image not find issue. This was tested on PC running on Windows 10 Pro 1809 (64-bit)

Here's the log:
[easyrpg_log.zip](https://github.com/EasyRPG/Player/files/2726289/easyrpg_log.zip)

Ghabry:
based on the logfile I don't think this is normalization related but something with handling of subfolders (which only appears to affect Windows?)

Yuuyatails:
Tested with build 47447d33007c24e20d930023ab27695d4a96dca7, but still causes the same issue. Unlike the previous two logs, this log was created while running the build 180601 of the game.

[easyrpg_log.zip](https://github.com/EasyRPG/Player/files/2840533/easyrpg_log.zip)

Issue #839, opened 2016-03-20
Player can't find files with extended Latin characters?


dwmtz:
It seems like the Player has problems when loading some files with extended Latin characters.
This is what happens in [Blue Star Story](http://tsukuru.pl/index.php?link=gra&title=2k-bluestarstory):

> [2016-03-20 18:32:48] Debug: Loading Map Map0121.lmu
> [2016-03-20 18:32:48] Debug: Tree: PigVillage < Mapa Swiata
> [2016-03-20 18:32:48] Debug: Cannot find: ChipSet/WioskaŚwiniakówBezDrzew (!)
> [2016-03-20 18:32:48] Warning: Image not found: ChipSet/WioskaŚwiniakówBezDrzew

![screenshot_8](https://cloud.githubusercontent.com/assets/4052560/13907336/1d15826a-eeec-11e5-998d-e7a53c3d67fd.png)
RPG_RT can load it just fine.
I've been struggling to figure out whether it's something on my end, the Player, or the game's database.

Ghabry:
Which platform are you using? What does the "Detected Encoding" line say?
If the Encoding detection was wrong you can overwrite it via the RPG_RT.ini:

```
[EasyRPG]
Encoding=1252
```
Not sure what codepage extended latin is. 1252 or 1251

Ghabry:
Rename ```ChipSet/WioskaŚwiniakówBezDrzew``` to ```ChipSet/WioskanświniakówBezDrzew``` (small ś)

fdelapena:
This would require Unicode case-insensitive string comparison with `u_strcasecmp()` from ICU (`ustring.h`) and something equivalent for the iconv fallback.

Ghabry:
Or ```u_strToLower```

dwmtz:
Windows 7 64-bit, Player64.

> Debug: Detected encoding: ibm-5346_P100-1998

I had it set to 1250 in RPG_RT.ini before, changing to 1251 and 1252 didn't do anything. The file actually uses the lower case "ś" character by default. 
Does that mean the Player ignores files unless it gets a perfect match with what's in the database?

Sorry for the late reply.

Ghabry:
The Player does a case insensitive match but this currently only works for ASCII, not for anything extended.
This is for supporting platforms that are case sensitive... (all except Windows)

BlisterBoy:
I have a problem similar with some projects (not everyone) with the Android app.
[...]

PR #1532, opened 2018-12-06
Composite Characters Fix

### LINKED ISSUE #1530: Composite japanese characters cause files to be not found
Experiencing a bug related to encoding problems with japanese characters with filenames in Yume2kki:

***
#### Player platform:
macOS 10.13.6

#### Locale:
`ru_RU.UTF-8`

#### Save File and Log
[save-and-log.zip](https://github.com/EasyRPG/Player/files/2651405/save-and-log.zip)

#### Steps to reproduce:
Navigate to track No. 003 D in the PC music player.
***

In particular, the Player claims to not be able to find 夢オルゴールFC.

Interestingly enough, [grepping the music folder](https://pastebin.com/puBHLCUj) has shown that the track does not indeed exist under that name, but exists under the name 夢オルゴールFC instead.

Looks similar right?

Hence I have further copy/pasted the correct and the wrong name into text files and ran xxd against both of them producing:
```
$ xxd correct.txt
00000000: e5a4 a2e3 82aa e383 abe3 82b3 e382 99e3  ................
00000010: 83bc e383 ab46 430a                      .....FC.
```
```
$ xxd wrong.txt
00000000: e5a4 a2e3 82aa e383 abe3 82b4 e383 bce3  ................
00000010: 83ab 4643 0a                             ..FC.
```
which are clearly different.
```
wrong: e5a4a2e382aae383ab e382b4        e383bce383ab46430a
right: e5a4a2e382aae383ab e382b3 e38299 e383bce383ab46430a
```
In particular the difference is in the ゴ character as shown by [grepping the music folder](https://pastebin.com/puBHLCUj).

Potentially related issues are #1383 and #839

SmiVan:
Closes #1530 and probably #1383 and #839, by implementing normalization in place of lower-casing per @Ghabry's patch in #839.

Has to be accepted together with [the related liblcf pull request](https://github.com/EasyRPG/liblcf/pull/292) (EasyRPG/liblcf#292).

Requires more thorough testing, but is confirmed to work against the Yume2kki example in #1530.

***
Also needs to be tested that it doesn't break everything on [non-Mac platforms which use NFC as opposed to NFD](https://gist.github.com/JamesChevalier/8448512).

### DIFF
--- a/src/filefinder.cpp
+++ b/src/filefinder.cpp
@@ -80,6 +80,7 @@
 #include "registry.h"
 #include "rtp_table.h"
 #include "main_data.h"
+#include "reader_util.h"
 
 // MinGW shlobj.h does not define this
 #ifndef SHGFP_TYPE_CURRENT
@@ -111,11 +112,11 @@ namespace {
 			return em_file;
 #endif
 
-		std::string lower_dir = Utils::LowerCase(dir);
+		std::string corrected_dir = ReaderUtil::Normalize(dir);
 		std::string const escape_symbol = Player::escape_symbol;
-		std::string corrected_name = Utils::LowerCase(name);
+		std::string corrected_name = ReaderUtil::Normalize(name);
 
-		std::string combined_path = MakePath(lower_dir, corrected_name);
+		std::string combined_path = MakePath(corrected_dir, corrected_name);
 		std::string canon = MakeCanonical(combined_path, 1);
 		if (combined_path != canon) {
 			// Very few games (e.g. Yume2kki) use path traversal (..) in the filenames to point
@@ -147,10 +148,10 @@ namespace {
 		}
 #endif
 
-		string_map::const_iterator dir_it = tree.directories.find(lower_dir);
+		string_map::const_iterator dir_it = tree.directories.find(corrected_dir);
 		if(dir_it == tree.directories.end()) { return ""; }
 
-		string_map const& dir_map = tree.sub_members.find(lower_dir)->second;
+		string_map const& dir_map = tree.sub_members.find(corrected_dir)->second;
 
 		for(char const** c = exts; *c != NULL; ++c) {
 			string_map::const_iterator const name_it = dir_map.find(corrected_name + *c);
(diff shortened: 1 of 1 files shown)


carstene1ns:
Just a reminder for later: this might break some existing emscripten games (json cache).

carstene1ns:
ok to test.

Ghabry:
This will require another extremely annoying rebase in my VFS branch.

Keep Filefinder alone! 😭

fmatthew5876:
> 
> 
> This will require another extremely annoying rebase in my VFS branch.
> 
> Keep Filefinder alone! 😭

I understand your pain, but this is an important fix.

Ghabry:
Jenkins: test this please

fdelapena:
Jenkins: test this please

fdelapena:
I've tested the PR executable with #839, it does not fix that particular issue yet on Windows with the Blue Star Story game.

fmatthew5876:
Can you post the 2 filenames in hex? The one player tries to open, and the actual filename on disk.

carstene1ns:
Maybe we should use some lowercase function from ICU, I guess there is some dedicated for unistrings. Might be aware of such things.

SmiVan:
@carstene1ns [UnicodeString::toLower()](http://www.icu-project.org/apiref/icu4c/classicu_1_1UnicodeString.html#ac7fb1eeca89ad489988ef9847dc5cf09) or [UnicodeString::toLower(Locale)](http://www.icu-project.org/apiref/icu4c/classicu_1_1UnicodeString.html#a69ee6c3a17a91b52bacd592bb4975c13) might be what we need then.

Ghabry:
we already use icu::UnicodeString(str.c_str()).toLower() function in Normalize.
But you are Right, I forgot that lower and uppercasing depends on the language...

src/filefinder.cpp
```diff
@@ -627,16 +628,16 @@ bool FileFinder::IsValidProject(DirectoryTree const & dir) {
 
 bool FileFinder::IsRPG2kProject(DirectoryTree const& dir) {
 	string_map::const_iterator const
-		ldb_it = dir.files.find(Utils::LowerCase(DATABASE_NAME)),
-		lmt_it = dir.files.find(Utils::LowerCase(TREEMAP_NAME));
+		ldb_it = dir.files.find(ReaderUtil::Normalize(DATABASE_NAME)),
+		lmt_it = dir.files.find(ReaderUtil::Normalize(TREEMAP_NAME));
```
Ghabry: Here Normalize is unnecessary

src/filefinder.cpp
```diff
@@ -627,16 +628,16 @@ bool FileFinder::IsValidProject(DirectoryTree const & dir) {
...
 
 	return(ldb_it != dir.files.end() && lmt_it != dir.files.end());
 }
 
 bool FileFinder::IsEasyRpgProject(DirectoryTree const& dir){
 	string_map::const_iterator const
-		ldb_it = dir.files.find(Utils::LowerCase(DATABASE_NAME_EASYRPG)),
-		lmt_it = dir.files.find(Utils::LowerCase(TREEMAP_NAME_EASYRPG));
+		ldb_it = dir.files.find(ReaderUtil::Normalize(DATABASE_NAME_EASYRPG)),
+		lmt_it = dir.files.find(ReaderUtil::Normalize(TREEMAP_NAME_EASYRPG));
```
Ghabry: same here

src/filefinder.cpp
```diff
@@ -854,7 +855,7 @@ bool FileFinder::IsMajorUpdatedTree() {
 			string_map mem = tree->sub_members["music"];
 			for (auto& i : mem) {
 				std::string file = mem[i.first];
-				if (Utils::EndsWith(Utils::LowerCase(file), ".mp3")) {
+				if (Utils::EndsWith(ReaderUtil::Normalize(file), ".mp3")) {
```
Ghabry: and here

Ghabry:
To figure out what the filefinder has in the file list you will need to dump it, best is printf Debugging.

good place is here: https://github.com/EasyRPG/Player/blob/81249d3413c704e607dda5a7cb167f97ffdcbf69/src/filefinder.cpp#L811

carstene1ns:
@Ghabry: Yep, I just remembered reading what @SmiVan found about default locale and such.

fdelapena:
After taking a look, comparing with some arbitrary locale looks ugly, because we should map some locale with the detected encoding.

What about [caseCompare](https://stackoverflow.com/a/38469016) instead?

Ghabry:
This must be a Windows only problem, I just tested this PR against multiple stuff under Linux and it works as expected.

Ghabry:
rebased + applied my review comments

carstene1ns:
Let's break things.

PR #1533, opened 2018-12-06
Enable manpages on all UNIX and make man path customizable

AMDmi3:


AMDmi3:
I've just noticed that `GNUInstallDirs` is already included, so `CMAKE_INSTALL_MANDIR` is already available. So this should be a final variant.

Ghabry:
yepp we were discussing 10 min ago about the best way to solve this and your solution matches with ours. :+1:

Issue #1279, opened 2017-10-09
Windows Platform Cleanup


carstene1ns:
Just a small list of stuff that bugs me related to our Windows Port.

- [x] `NOMINMAX` and other preprocessor stuff (`*LEAN*`) should be moved to the MSVC solution
- [x] The minidump writer inside `player.cpp` should be moved into `utils_win.cpp`

Ghabry:
By now the NOMINMAX & friends could be moved in the CMakeLists because the solution is almost deprecated.

PR #1537, opened 2018-12-07
Remove MSVC project files

### LINKED ISSUE #1270: Replace Visual Studio Projects with CMake
When vcpkg lands the easiest way to build the Player will be through CMake instead of MSVC project files (how ironic).

### LINKED ISSUE #1536: Windows: Fix remaining CMake issues
Player:
Linker error for vcpkg SDL Mixer because Modplug and Flac are not linked which is strange, Find_SDL2Mixer.cmake handles them.

Install PDB files next to the binary.
According to stackoverflow this is possible via
``install(FILES $<TARGET_PDB_FILE:${PROJECT_NAME}> DESTINATION bin OPTIONAL)``

liblcf: /Ehsc (unwind semantics for exception handling) is not used and emits warnings but don't ask me why there are no warnings when compiling Player...

Ghabry:
Guess nobody will miss these XML files.  ✨🎈🎉🎆

All Jenkins jobs already use CMake, everything should work as before even without this merged.

The 2015_uwp was completely unused btw, was a failed attempt to get EasyRPG with UWP working, should be attempted again someday (when SDL adds XAML :/).

Fix #1536
Fix #1270


carstene1ns:
:+1:

Ghabry:
btw we should now really update the Wiki ;)

carstene1ns:
Readme is also wrong though. (liblcf)

Ghabry:
right, copy-paste error...

lib/README.md
```diff
@@ -1,10 +0,0 @@
-EasyRPG Player
-==============
-
-The following note only applies to these builds:
-
-- Visual Studio 2015 project (Windows)
-- OpenDingux Makefile (Dingoo, GCW Zero)
```
carstene1ns: This will break the current GCW0 build job. However, should be ok to put liblcf as shared library in the OPK or just link it in as static library.

CMakeLists.txt
```diff
@@ -384,6 +384,11 @@ if(${PLAYER_AUDIO_BACKEND} MATCHES "^SDL.*$")
 	endif()
 endif()
 
+# MSVC: Set UTF-8 encoding for all files (important for rtp table)
+if(MSVC)
+	target_compile_options(${PROJECT_NAME} PRIVATE "/utf-8")
```
carstene1ns: I would like to have some other stuff here too. (see #1279)

lib/README.md
Ghabry: oh, I assumed that this OpenDingux note is outdated...

PR #1538, opened 2018-12-08
Battle action and states fixes - Previous Albeleon 2.21 / 2.41 from battle 7

fmatthew5876:
A simpler reimplementation of Albeleon 2.21.

To test these, create a 1x 1+ turn battle event which does:

```
Cure provoke, berserk, death, shock
Show Choice:
  Death: Add dead
  Provoke: Add provoke
  Confuse: Add confuse
  Shock: Add shock
  Else:
```

To test the effects of multiple states, do the show choice twice.

To test state recovery, you can use the above event. You can also inflict states, and the have a very fast actor use a recovery item.

### DIFF
--- a/src/game_battlealgorithm.h
+++ b/src/game_battlealgorithm.h
@@ -20,6 +20,7 @@
 
 #include <string>
 #include <vector>
+#include "rpg_state.h"
 
 class Game_Battler;
 class Game_Party_Base;
@@ -355,6 +356,12 @@ public:
 	 */
 	Type GetType() const;
 
+	/**
+	 * @return The significant restriction of the source when this
+	 *      algorithm was created.
+	 */
+	RPG::State::Restriction GetSourceRestrictionWhenStarted() const;
+
 protected:
 	AlgorithmBase(Type t, Game_Battler* source);
 	AlgorithmBase(Type t, Game_Battler* source, Game_Battler* target);
@@ -408,6 +415,7 @@ protected:
 	bool absorb;
 	bool revived = false;
 	mutable int reflect;
+	RPG::State::Restriction source_restriction = RPG::State::Restriction_normal;
 
 	RPG::Animation* animation;
 
--- a/src/game_battlealgorithm.h
+++ b/src/game_battlealgorithm.h
@@ -293,6 +293,11 @@ public:
 	virtual int GetSourceAnimationState() const;
 
 	/**
+	* @return true if it is still possible to perform this action now.
+	*/
+	virtual bool ActionIsPossible() const;
+
+	/**
 	 * Gets the sound effect that is played when the action is starting.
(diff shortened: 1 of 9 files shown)


fmatthew5876:
This one is tested and ready. I have nothing more to add here.

carstene1ns:
jenkins: test this please (newlib check).

src/scene_battle_rpg2k.cpp
```diff
@@ -374,6 +373,9 @@ void Scene_Battle_Rpg2k::ProcessActions() {
 bool Scene_Battle_Rpg2k::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBase* action) {
 	// Order of execution of BattleActionState:
 	// ConditionHeal > Execute > Apply > (ResultPop > ResultPush) > Death > Finished.
+	if (action == nullptr) {
```
Ghabry: Is it possible that the action is nullptr?
If not add a noise Output::Warning here so we get bugreports.

src/scene_battle_rpg2k.cpp
```diff
@@ -371,6 +371,11 @@ void Scene_Battle_Rpg2k::ProcessActions() {
 }
 
 bool Scene_Battle_Rpg2k::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBase* action) {
+	// Immediately quit for dead actors no move. Prevents any animations or delays.
+	if (action->GetType() == Game_BattleAlgorithm::Type::NoMove && action->GetSource()->IsDead()) {
+		return true;
```
Ghabry: you access the action here before the nullptr check

src/game_battlealgorithm.h
```diff
@@ -355,6 +356,12 @@ class AlgorithmBase {
 	 */
 	Type GetType() const;
 
+	/**
+	 * @ret The significant restriction of the source when this
```
Ghabry: ```@return```

src/scene_battle_rpg2k.cpp
fmatthew5876: Thats an artifact I forgot to remove from an old implementation of this I tried.

Ghabry: okay then just remove it

src/scene_battle_rpg2k.cpp
fmatthew5876: nullptr check not needed, it's gone.

fmatthew5876:
Addressed review comments. I'm not in favor of adding extra nullptr checks. If you call `ProcessBattleActions()` with nullptr its a bug and a crash is the best way to get attention. Not logging (likely ignored), and then returning and continuing in some wierd broken state.

PR #1541, opened 2018-12-08
Support proper rendering of Sprites at map borders of looping maps

Ghabry:
Alternative to https://github.com/EasyRPG/Player/pull/1401/commits/d11a906a196ecda412a7adf8fc731eba9746fe56  as you can see my solution is a bit simpler ;).

This depends on #1498 because the Screen scroll changes/fixes made it much easier to implement this.

Test case: https://easyrpg.org/play/pr1541/?start-map-id=44

### DIFF
--- a/src/spriteset_map.cpp
+++ b/src/spriteset_map.cpp
@@ -39,14 +39,16 @@ Spriteset_Map::Spriteset_Map() {
 
 	ChipsetUpdated();
 
+	need_x_clone = Game_Map::LoopHorizontal();
+	need_y_clone = Game_Map::LoopVertical();
+
 	for (Game_Event& ev : Game_Map::GetEvents()) {
-		character_sprites.push_back(std::make_shared<Sprite_Character>(&ev));
+		CreateSprite(&ev, need_x_clone, need_y_clone);
 	}
 
 	airship_shadow.reset(new Sprite_AirshipShadow());
 
-	character_sprites.push_back
-		(std::make_shared<Sprite_Character>(Main_Data::game_player.get()));
+	CreateSprite(Main_Data::game_player.get(), need_x_clone, need_y_clone);
 
 	timer1.reset(new Sprite_Timer(0));
 	timer2.reset(new Sprite_Timer(1));
@@ -95,7 +97,7 @@ void Spriteset_Map::Update() {
 
 		if (!vehicle_loaded[i - 1] && vehicle->GetMapId() == map_id) {
 			vehicle_loaded[i - 1] = true;
-			character_sprites.push_back(std::make_shared<Sprite_Character>(vehicle));
+			CreateSprite(vehicle, need_x_clone, need_y_clone);
 		}
 	}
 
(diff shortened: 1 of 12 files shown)


Ghabry:
RPG_RT glitch exhibition (I altered the map slightly because I needed to test z-layering):

For reference: That's the map

1-fs8

When I scroll the map this way the z-layering between airship (the arrow) and the sprites around becomes messed up and *the staircase dissapears*

2-fs8

Same as previous, just with the map fixes, so that the staircase stays invisible (but still works)

3-fs8

The "nose" of the vehicle is not wrapped

5-fs8

When you move in the airship to the most-bottom row the *vehicle dissapears* and the shadow is not wrapped to the top.

failpan

The same applies at the right-most map border.

failpan2

fmatthew5876:
Nice work Ghabry :+1:

Ghabry:
Well most of the thanks goes to your new scrolling code because without it my code will result in strange scroll locks independent of the pan configuration.

fmatthew5876:
Thats good, I guess we could say this work constitutes a review of the new panning good.

fmatthew5876:
This will need a retest after #1554 is done.

Ghabry:
Jenkins: test this please

carstene1ns:
@fmatthew5876 Do you think this will be affected by panorama code? IMO this might have artifacts if scrolling is wrong, but generally should be fine, as only uses player/event/vehicle positions...

fmatthew5876:
The change that fixes the huge offset getting passed to `Game_Map::Scroll()` could affect this PR maybe? The map position still worked even with that bug present so it's maybe ok. I would retest this with the panorama branch to be absolutely sure.

Ghabry:
Only rebased, don't merge yet. Have to chekc if this still works and can be maybe simplified now with the new scroll code.

Ghabry:
Everything still works as is and don't think this can be reduced further.
So can be merged now.

src/spriteset_map.cpp
```diff
@@ -186,6 +192,38 @@ bool Spriteset_Map::RequireBackground(const Graphics::DrawableList& drawable_lis
...
+	if (create_x_clone) {
+		airship_shadows.push_back(std::make_shared<Sprite_AirshipShadow>(CloneType::XClone));
+	}
+	if (create_y_clone) {
+		airship_shadows.push_back(std::make_shared<Sprite_AirshipShadow>(CloneType::YClone));
+	}
+	if (create_x_clone && create_y_clone) {
+		airship_shadows.push_back(std::make_shared<Sprite_AirshipShadow>(
+			(CloneType)(CloneType::XClone | CloneType::YClone)));
+	}
```
carstene1ns: Correct me if I am wrong, but this seems to create useless clones when both x and y clones are enabled.

Ghabry: When both X and Y are enabled you need a 4th clone because when you are in the a map corner you can have up to for times the same sprite visible.

Issue #1446, opened 2018-10-14
Log the seed used for the random number generator


fdelapena:
Player should log into `easyrpg_log.txt` the RNG seed used for a particular run. This might help to reproduce in a closer way errors got by the reporter where random movements were involved when the user got an issue and it is not able to reproduce under later circumstances.

PR #1544, opened 2018-12-08
Log the RNG seed

fmatthew5876:
Fixes: #1446


Ghabry:
Because the seed is usually an integer (timestamp) I would suggest that the seed is printed as decimal.

The \n adds an extra linebreak and must be emoved.

fmatthew5876:
Fixed

PR #1545, opened 2018-12-08
Fixes for TintScreen and tint pictures

fmatthew5876:


fmatthew5876:
Fixed #1546 by re-ordering `Update()` calls.

src/tone.h
```diff
@@ -71,4 +61,23 @@ class Tone {
 	int gray;
 };
 
+inline Tone blend(const Tone& l, const Tone& r) {
```
Ghabry: Could I have this instead either

a) As a static/factory function in the Tone or
b) As a method on the Tone object?

fmatthew5876:
Renamed `blend()` to `Blend()`, to match our coding style.

src/tone.h
Ghabry: well everybody always tells you about encapsulation and namespacing, so a free function looks unfamiliar but when you see this as some kind of operator this argumentation makes sense.

fmatthew5876: The missing encapsulation is that we don't have a namespace around liblcf and Player.

Issue #1547, opened 2018-12-09
Determine correct screen_shake algorithm


fmatthew5876:
The screen shake algorithm is broken. It has 2 major problems:

* It stores an extra state variable 'shake_direction' in memory but this is not part of the save data. This makes correct loading not possible
* The algorithm computes a 0 shake for many common inputs.

I've opened this issue to share data analysis results to figure out what function to use.

I've been using this code in Rpg maker:

```
Shake Screen: 9, 5, @1.0s
Loop
  Open Save Menu
  Wait: 0.1 seconds
```

So far, I have this:

[RPG_RT+shake(1).pdf](https://github.com/EasyRPG/Player/files/2660329/RPG_RT%2Bshake.1.pdf)

fmatthew5876:
Here is a frame by frame breakdown of `Shake 1,1, @2.0s`

[RPG_RT+Shake+1%2C1+2.0s.pdf](https://github.com/EasyRPG/Player/files/2660311/RPG_RT%2BShake%2B1.2C1%2B2.0s.pdf)

PR #1548, opened 2018-12-09
Fix screen shake algo

fmatthew5876:
Fixes: #1547 

From Cherry:
```
and the Shake X comes from something like this:

ConvertedPower = 1 + 2*Power
Val = ConvertedPower * Sin((FrameCount * 4*(Speed + 2)) % 256) * pi / 128)
Cutoff = Speed * ConvertedPower/8

newX = min(prevX + Cutoff, max(prevX - cutoff, Val))

I'm not sure this is right because I don't understand why it's so complicated and includes the old value

and when frame count is zero (I guess it's also set to zero when shake is turned off?) it just resets shake X and Y to 0
```

I've tested the following cases so far and they are frame accurate with RPG_RT:

| Strength | Speed | Wait |
| --- | --- | --- |
| 1 | 1 | 2.0s |
| 1 | 9 | 2.0s |
| 9 | 1 | 2.0s |
| 9 | 9 | 2.0s |

Some data here:
[RPG_RT-shake_9_1_20.pdf](https://github.com/EasyRPG/Player/files/2660859/RPG_RT-shake_9_1_20.pdf)

### DIFF
--- a/src/game_picture.cpp
+++ b/src/game_picture.cpp
@@ -80,8 +80,15 @@ void Game_Picture::UpdateSprite() {
 		sprite->SetBitmap(sheet_bitmap);
 	}
 
-	sprite->SetX((int)data.current_x);
-	sprite->SetY((int)data.current_y);
+	int x = data.current_x;
+	int y = data.current_y;
+	if (data.flags.affected_by_shake) {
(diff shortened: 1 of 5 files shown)


fmatthew5876:
Fixed picture support too. With this I think we have everything we need related to shake.

PR #1552, opened 2018-12-10
Make PercentChance() newlib compatible

fmatthew5876:


Ghabry:
instead of long and int can't we use int32_t ?

fmatthew5876:
The problem is more general that calling `Utils::PercentChance()` with long is ambiguous. You can try adding a `Utils::PercentChance(1L)` without this patch and see it fail compilation. Newlib with its horrible configuration just exposes the issue.

If we ever did need to use `PercentChance()` with a variable of type long (maybe size_t, int64_t), it would break again.

PR #1553, opened 2018-12-10
Handle saving and loading of tile substitutions

### LINKED ISSUE #1266: Tile substitution is not restored from RPG_RT savegames and EasyRPG doesn't save the substitution at all
Tile substitution is the event command that replaces one tile with another.

fmatthew5876:
Tested:
* Loading tile substitutions from save game
* Change chipset event command after doing tile substitution
* Loading changed chipset

---

Fixes #1266.

### DIFF
--- a/src/game_map.cpp
+++ b/src/game_map.cpp
@@ -1330,20 +1330,23 @@ void Game_Map::RemoveAllPendingMoves() {
 	pending.clear();
 }
 
-void Game_Map::SubstituteDown(int old_id, int new_id) {
-	for (size_t i = 0; i < map_info.lower_tiles.size(); ++i) {
-		if (map_info.lower_tiles[i] == old_id) {
-			map_info.lower_tiles[i] = (uint8_t) new_id;
+static int DoSubstitute(std::vector<uint8_t>& tiles, int old_id, int new_id) {
+	int num_subst = 0;
+	for (size_t i = 0; i < tiles.size(); ++i) {
+		if (tiles[i] == old_id) {
+			tiles[i] = (uint8_t) new_id;
+			++num_subst;
 		}
 	}
+	return num_subst;
 }
 
-void Game_Map::SubstituteUp(int old_id, int new_id) {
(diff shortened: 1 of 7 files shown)


src/tilemap_layer.cpp
```diff
@@ -681,13 +685,6 @@ std::vector<unsigned char> TilemapLayer::GetPassable() const {
 void TilemapLayer::SetPassable(const std::vector<unsigned char>& npassable) {
 	passable = npassable;
 
-	if (substitutions.size() < passable.size())
-	{
-		substitutions.resize(passable.size());
-		for (uint8_t i = 0; i < substitutions.size(); i++)
-			substitutions[i] = i;
-	}
```
fmatthew5876: Was this just here for extra safety or is there some specific case its needed? Both these arrays should always be size 144.

carstene1ns: hmmm, fcdf1b1be548.

Ghabry: can't comment on this, the chatlogs of February 2012 aren't useful either

Ghabry: don't want to link our chatlog publicly but take a look at 2011/2011-01-16.
has some substitution discussion, though there everybody says it is 144 because the autotiles are ignored.

fmatthew5876: The LSD data from RPG_RT is always size 144.

src/game_map.cpp
```diff
@@ -1327,20 +1330,24 @@ void Game_Map::RemoveAllPendingMoves() {
 	pending.clear();
 }
 
-void Game_Map::SubstituteDown(int old_id, int new_id) {
+template <typename T>
+static int DoSubstitute(T& tiles, int old_id, int new_id) {
```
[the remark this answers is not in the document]
Ghabry: I know that this is you being lazy, but I talk about the variable tiles not being used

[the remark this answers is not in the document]
fmatthew5876: Oh yes, that is a bit more serious! Will fix. 

Great catch.

Issue #1534, opened 2018-12-07
Panoramas are not pixel perfect


fmatthew5876:
![parallax](https://user-images.githubusercontent.com/1198466/49627840-14e5d600-f9af-11e8-8661-10026ea9e166.png)

Left: RPG_RT, Right: Player

Look at the lower right corner. In RPG_RT you can see 2 stars, one of which is halfway obscured by the glass tile. On the right, you see only 1 star.

This issue  was produced by creating a 41x31 map and setting the panorama to Cosmos1 with no scrolling.

carstene1ns:
I remember the same thing years ago in wulfhammer. In the creator room there was also some moon background that behaved wrong. I think this got fixed along with yume nikki back then,there might be some closed issue about.

Ghabry:
Do you mean "jumping panorama" when crossing a map-loop border?
Can't remember the issue either tough.

fmatthew5876:
Nothing jumps, its just a big picture on a big map that is slightly off in how it scales when you walk around. If you weren't paying attention you wouldn't even notice it.

Probably a minor rounding or interpolation issue in our panorama code.

carstene1ns:
Is #961, unfortunately images and video offline, however when loading the game, you will know why I mentioned it ;)
https://easyrpg.org/play/master/?game=embric&start-map-id=89&start-party=1&start-position=40%2060

fmatthew5876:
Panoramas also have incorrect starting offset for looping maps.

![pano](https://user-images.githubusercontent.com/1198466/49704755-1993cf80-fbe5-11e8-8c4c-9d2e4dea3832.png)

This is the same 41x31 map, set to loop on both dimensions.

fmatthew5876:
Regarding the second issue:

When the map loops horizontal or vertical. 

When you enter the map the panorama gets fixed in the looping dimensions. The position of the panorama is different depending on which place you start on the map, but it doesn't change after that.

Walk far away, then save your game. If you load your game, the panorama stays in it's position. If you change the new game position to the same tile as the save game and start a new game, the panorama offset is different than your loaded game.

This means the save game data must be storing some way to get the panorama offsets for looping maps. Maybe `SaveScreen::pan_` ...?

fmatthew5876:
Here are what the starting positions are in RPG_RT

Map (41,31) Start at (30,23)

[...]

Issue #1550, opened 2018-12-09
Regression: Scrolling panoramas "jump" at the map boundaries when the map is looping


Ghabry:
Basicly a new version of #761 

Broken:
https://easyrpg.org/play/master?game=yume_nikki&start-map-id=18&start-party=1&new-game

Working:
https://easyrpg.org/play/?game=yume_nikki&start-map-id=18&start-party=1&new-game

#### Describe the issue in detail and how to reproduce it:

Walk to the left until you are ca. 5 tiles away from the candles, then walk to the right. you will see that the Panorama jumps.
Repeat this to get more jumps.

PR #1522, opened 2018-12-03
Refactor Game_Character classes 2 of 2

fmatthew5876:

PR #1554, opened 2018-12-11
More panorama fixes

fmatthew5876:
I found more issues with panorama.

- [x] Our panorama pan_x / pan_y were not compatible with rm2k3e. The means saving and loading games between them will be wrong. We were treating them as scrolling offsets, but they are actually the entire panorama image offset.
- [x] We weren't handling Looping and scrolling maps right
- [x] `SaveScreen::pan_y` was inverted (not used in panorama but similar code so added here)
- [x] Save data not loaded correctly in all cases
- [x] Scroll jumps when you move slowly and is sometimes 1 pixel off in frames.
- [x] Retest #1550 with this PR.
- [x] Fixes: #1534 

So I've had to refactor this yet again!

I've retested this by trying different combinations of maps, saving the game in various spots, and comparing the pan_x and pan_y of Player vs RPG_RT.

Testing performed:
* Map with no looping, no scroll - Panorama is 2x size and moves with relative position on map 
* Map with looping, no scroll - Panorama is screen size and doesn't move
* Map with looping, scroll - Panorama is screen size and moves with player
* Map with no loop, scroll - Panorama is screen size and moves with player
* Map with loop, scroll, auto scroll - moves with player and moves by itself

I'm going to need to ask you guys to help test this one. Try some different size maps and panorama pictures. I don't want to keep finding bugs and refactoring this over and over again.

### DIFF
--- a/src/game_map.cpp
+++ b/src/game_map.cpp
@@ -83,6 +83,9 @@ namespace {
 	RPG::Chipset* chipset;
 
 	int last_encounter_idx = 0;
+
+	// FIXME: Find a better way to do this
+	bool load_panorama_from_save = false;
 }
 
 void Game_Map::Init() {
@@ -142,6 +145,7 @@ void Game_Map::Setup(int _id) {
 	SetupCommon(_id, false);
 	map_info.encounter_rate = GetMapInfo().encounter_steps;
 	SetEncounterSteps(0);
+	load_panorama_from_save = false;
 	panorama = {};
 
 	Parallax::ClearChangedBG();
@@ -221,6 +225,12 @@ void Game_Map::SetupFromSave() {
 	SetChipset(map_info.chipset_id);
 
 	SetEncounterSteps(location.encounter_steps);
+
+	// If the later async code will load panorama, set the flag to not clear the offsets.
+	if (!map_info.parallax_name.empty()
+			|| (map->parallax_flag && !map->parallax_name.empty())) {
+		load_panorama_from_save = true;
+	}
 }
 
 
@@ -1533,8 +1543,23 @@ void Game_Map::Parallax::Initialize(int width, int height) {
 	parallax_width = width;
 	parallax_height = height;
 
-	ResetPositionX();
-	ResetPositionY();
+	Params params = GetParallaxParams();
+
+	// We want to support loading rm2k3e panning chunks
+	// but also not break other saves which don't have them.
+	// To solve this problem, we reuse the scrolling methods
+	// which always reset the position anyways when scroll_horz/vert
+	// is false.
+	// This produces compatible behavior for old RPG_RT saves, namely
+	// the pan_x/y is always forced to 0.
+	if (load_panorama_from_save) {
+		load_panorama_from_save = false;
+		ScrollRight(0);
+		ScrollDown(0);
+	} else {
+		ResetPositionX();
+		ResetPositionY();
+	}
 }
 
 void Game_Map::Parallax::ResetPositionX() {
--- a/src/game_map.cpp
+++ b/src/game_map.cpp
@@ -1192,6 +1192,7 @@ void Game_Map::SetPositionX(int x) {
 		x = std::max(0, std::min(map_width - SCREEN_WIDTH, x));
 	}
 	map_info.position_x = x;
+	Parallax::ResetPositionX();
 }
 
 int Game_Map::GetPositionY() {
@@ -1210,6 +1211,7 @@ void Game_Map::SetPositionY(int y) {
 		y = std::max(0, std::min(map_height - SCREEN_HEIGHT, y));
 	}
 	map_info.position_y = y;
+	Parallax::ResetPositionY();
 }
 
 Game_Map::RefreshMode Game_Map::GetNeedRefresh() {
--- a/src/game_map.cpp
+++ b/src/game_map.cpp
@@ -45,8 +45,8 @@
 #include "utils.h"
 
 namespace {
-	constexpr int default_pan_x = 9 * SCREEN_TILE_WIDTH;
-	constexpr int default_pan_y = 7 * SCREEN_TILE_WIDTH;
+	constexpr int default_pan_x = 9 * SCREEN_TILE_SIZE;
+	constexpr int default_pan_y = 7 * SCREEN_TILE_SIZE;
 	constexpr int default_pan_speed = 16;
 
 	RPG::SaveMapInfo& map_info = Main_Data::game_data.map_info;
@@ -387,7 +387,7 @@ void Game_Map::ScrollRight(int distance) {
 	}
 	// Unused, except compatibility with RPG_RT
 	auto& pan_x = Main_Data::game_data.screen.pan_x;
-	const auto pan_limit_x = 20 * SCREEN_TILE_WIDTH;
+	const auto pan_limit_x = 20 * SCREEN_TILE_SIZE;
 	pan_x = (pan_x - distance + pan_limit_x) % pan_limit_x;
 
 	Game_Map::Parallax::ScrollRight(distance);
@@ -402,7 +402,7 @@ void Game_Map::ScrollDown(int distance) {
 	}
 	// Unused, except compatibility with RPG_RT
 	auto& pan_y = Main_Data::game_data.screen.pan_y;
-	const auto pan_limit_y = 10 * SCREEN_TILE_WIDTH;
+	const auto pan_limit_y = 10 * SCREEN_TILE_SIZE;
 	pan_y = (pan_y - distance + pan_limit_y) % pan_limit_y;
 
 	Game_Map::Parallax::ScrollDown(distance);
@@ -418,7 +418,7 @@ static void ClampingAdd(int low, int high, int& acc, int& inc) {
 }
 
 void Game_Map::AddScreenX(int& screen_x, int& inc) {
-	int map_width = GetWidth() * SCREEN_TILE_WIDTH;
+	int map_width = GetWidth() * SCREEN_TILE_SIZE;
 	if (LoopHorizontal()) {
 		screen_x = Utils::PositiveModulo(screen_x + inc, map_width);
 	} else {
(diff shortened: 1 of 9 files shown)


fmatthew5876:
Because the `pan_x` and `pan_y` values are the total panorama offset and not just scrolling offset, loading save games becomes problematic.

We have to do the reset logic in `Game_Map::Parallax::Initialize()`. The problem is this function is called asynchronously. I added a global variable which is set in `SetupFromSave()` and then retrieved in `Initialize()` but this solution is disgusting.

Anyone have a better idea how to do this?

fmatthew5876:
The Yume Nikki bug is back. It's something more serious related to cross the edge of looping maps. When you cross the boundary, `Game_Map::ScrollRight(-28128)` is called.

The last panorama fix just hid the bug, it seems to affect other things too. 

I'm still looking into it.

fmatthew5876:
I've fixed the Yume Nikki bug for real this time

I don't like the solutions here. In particular the savegame global and modulus stuff added to `Player::UpdateScroll()`.

fmatthew5876:
I fixed the panorama algo. It had some frame by frame issue. I've verified it by using my frame by frame save generation script.

fmatthew5876:
I've tested the new algo with the following cases:

640x480 Image
* 41x31 Map - 1 tile more than pic
* 40x30 Map - Same size as pic
* 39x29 Map - 1 tile less than pic
* 30x15 Map - Smaller than pic
* 20x15 Map  - Default map size, pic doesn't move

Ghabry:
Jenkins: test this please

src/game_player.cpp
```diff
@@ -180,6 +180,30 @@ void Game_Player::UpdateScroll(int old_x, int old_y) {
 	int new_panx = new_x - screen_x;
 	int new_pany = new_y - screen_y;
 
+	// Detect whether we crossed map boundary.
+	// We need to scale down dx/dy to a single step
+	// to not message up further calculations.
+	// FIXME: This logic will break if something moves so fast
+	// as to cross half the map in 1 frame.
```
Ghabry: So this will break when you are in a Vehicle and "SetVehiclePosition" is used to jump across half the map?

fmatthew5876: Well the answer is.. Yes and No.

It doesn't break in the way you're thinking. SetVehiclePosition() uses MoveTo() which doesn't use this scroll logic. 

The problem is, `MoveTo()` needs to reset the panorama. This is what happens when you use the vehicle trick or transport yourself to the same map.

Unfortunately, we call `MoveTo()` when loading saved games, so I can't make that change now because it will break save games loading.

We need commit e468f3f8d1ab4a3d884d38cb85 from #1522. This commit removes the call to `MoveTo()` from the load game path. Once that's done I can augment `MoveTo()` to reset the panorama.

Not resetting the panorama when you move your vehicle is a minor glitch, so I think I can add that part after this and #1522 are merged.

fmatthew5876:
jenkins: test this please

src/game_map.cpp
```diff
@@ -1501,47 +1519,92 @@ std::string Game_Map::Parallax::GetName() {
...
+	Params params = GetParallaxParams();
+
+	panorama.pan_y = 0;
+	if (params.name.empty()) {
+		return;
 	}
+	if (params.scroll_vert || Game_Map::LoopVertical()) {
+		panorama.pan_y = map_info.position_y;
+	} else if (GetHeight() > 15 && parallax_height > SCREEN_TARGET_HEIGHT) {
+		panorama.pan_y = map_info.position_y * (parallax_height - SCREEN_TARGET_WIDTH) * 2 / ((GetHeight() - 15) * TILE_SIZE);
```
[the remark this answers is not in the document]
fmatthew5876: Great catch. Thanks!

carstene1ns:
Commit 5bdbeaa4e04e is already part of #1555.

fmatthew5876:
Now that #1522 is in, I've added support for resetting the panorama when `SetVehicleLocation()` or similar function which ends up calling `Game_Player::MoveTo()` is used.

With this, panorama should really be done now. I'd like to get this one out to the beta testers so we can catch any regressions before next release.

src/game_player.cpp
```diff
@@ -181,6 +177,30 @@ void Game_Player::UpdateScroll(int old_x, int old_y) {
...
+		auto w = Game_Map::GetWidth() * SCREEN_TILE_WIDTH;
+		if (std::abs(dx) > w / 2) {
+			dx = (w - std::abs(dx)) % w;
+			if (new_x > old_x) {
+				dx = -dx;
+			}
+		}
+	}
+	if (Game_Map::LoopVertical()) {
+		auto h = Game_Map::GetHeight() * SCREEN_TILE_WIDTH;
```
carstene1ns: `SCREEN_TILE_HEIGHT`?

fmatthew5876: It's misleading but there is no `SCREEN_TILE_HEIGHT`.

carstene1ns: Maybe we should generally rename it to `SCREEN_TILE_SIZE` then.

fmatthew5876: Done

fmatthew5876:
jenkins: test this please

Ghabry:
> FIXME: load_panorama_from_save

Can I have some further context? Which async-handler is invoked and causes issues? Note that on desktop the async handlers are synchronous (they always fire directly).

fmatthew5876:
> > FIXME: load_panorama_from_save
> 
> Can I have some further context? Which async-handler is invoked and causes issues? Note that on desktop the async handlers are synchronous (they always fire directly).

The core problem is that from when we need to communicate the fact that it's a save game from the invocation of the async handler to when this `Initialize()` gets called.

To make matters worse, it looks like the async handler is triggered within an update call. I hate this global, but I don't know a better way.

Ghabry:
Download:
[panoramatest.zip](https://github.com/EasyRPG/Player/files/2696786/panoramatest.zip)



For orientation: Top left: Just panorama, Top right: H Loop, Bottm Right, V loop, Bottom left; Both loop

screenshot_20181219_235741

Start the game on the non-looping 41x31 map.
Talk to the 5th tile on the top-left (700x600).
Then talk to the different tiles at bottom-left. You will see that the tiles are way off when you compare the colors.

screenshot_20181220_000021

Ghabry:
Also strange: When you talk twice to the same event the Panorama sometimes changes the position.

fmatthew5876:
RPG_RT has bugs in this. I'm really confused about this one:

pano

fmatthew5876:
I added a fix for the clamping behavior for large panoramas.

carstene1ns:
This looks more and more like the old code we had years ago... :P

Ghabry:
> that in some cases the ChangeBG() command doesn't reset the panorama offsets. This logic appears buggy

Considering that even RPG_RT doesn't fully agree on there Panorama stuff this is the last thing I would like to see fixed, as the Panorama "moving" when you use the same ChangeBG command (with loop) twice looks wrong.

fmatthew5876:
> as the Panorama "moving" when you use the same ChangeBG command (with loop) twice looks wrong.

I think that's why they did it this way, to stop the panorama from "popping" when you change it. 

Unfortunately once you start walking around it'll pop anyway. This behavior at least validates how we call `ResetPosition()` each frame in the non-looping case.

I will make a new test to confirm all the cases where panorama is reset or not reset and then add then I should be able to emulate RPG_RT in `ChangeBG()`.

Ghabry:
Hm? I only experienced this in our Player not in RPG_RT.

fmatthew5876:
> Hm? I only experienced this in our Player not in RPG_RT.

All those gifs I posted are RPG_RT. In particular I'm referring to popping behavior the first one when you move from Y=7 to 8.

The core problem is that ChangeBG doesn't reset the panorama position, but scrolling does. So in some cases when you change bg and then scroll it'll pop.

Ghabry:
here is a video: https://my.mixtape.moe/alrpex.webm

At 15s until the end is what I mean with "Jumping" when invoking the same command twice. in EasyRPG it takes 2 attempts to get it in the RPG_RT position.

fmatthew5876:
The double reset behavior you're seeing happens because of the funky way we reload panoramas.

The first time we call `ChangeBG()`, the panorama offsets go to zero. Then when `Initialize()` is called later, they get reset based on map position.

The second time we call `ChangeBG()`, the panorama offsets go to zero. But this time the panorama name didn't change, so the reload logic in `Update()` is never called.

fmatthew5876:
We should now be fully compatible with RPG_RT, bugs and all.

The panorama offsets will not reset when
* Change panorama on looping map
* Change to a scrolling panorama
* Change to same filename (RPG_RT compatible, can cause pops when
  scrolling)

We do reset when:
* Loading game with non-scrolling panorama
   * Even if panorama was saved with different offsets
* Changing panorama to non-scrolling

In particular we now emulate the first gif. The reason the panorama doesn't reset in that case is because the filename doesn't change. We happen to copy this bug because our logic in `SpriteSet_Map::Update()` is dependent on filename.

Furthermore, if you do the movement in the first gif, save your game after switching to the top panorama. Then reload, the panorama will be loaded at the correct (0,0) offset, not the one you see on the screen.

So the popping in the first gif looks like an RPG_RT bug, which for now we also emulate!

fmatthew5876:
Here is another test, which works:

1. Copy the scrolling 100x100 panorama event to (1, 23)
2. Trigger the 1000x1000 panorama at the top of the map
3. Walk down to (1,23)
4. pan_y is now 8192
5. Talk to the 100x100 event, the 100x100 panorama is displayed
4. pan_y is still 8192 (which is greater than the max panning offset allowed by this image)
5. Step Up 1 tile
6. pan_y snaps to 1536, scrolling is smooth and correct.

Ghabry:
Wow this is really impressive! Unfortunately the looping map case is behaving incorrectly :/.

New Panorama Test (basicly the old one but the pictures have now the size that is written in there filename...)
[panoramatest.zip](https://github.com/EasyRPG/Player/files/2705008/panoramatest.zip)

On Looping Map 41x32: Different scrolling
pana


When the other side of the looping map becomes visible the panorama jumps:
panc

The jumping doesn't happen when the panorama mode is "H/V" in that case only the scrolling is wrong.

fmatthew5876:
Pushed a fix for looping maps.

I didn't see it misbehaving on looping maps with H/V scroll set...?

fmatthew5876:
Do you have a specific example when it's off? Maybe a small numerical tweak is all that's needed

Ghabry:
just use the bottom right events on the 41x32 map. YOu will see that the Panorama doesn't match the RPG_RT position.

PR #1555, opened 2018-12-11
Crash Bugfix: Pictures not cleaned up

fmatthew5876:
If you load a save game which has fewer pictures than exist
currently, Game_Screen::Reset() will try to Erase() pictures
which don't exist and this causes a segfault.

No idea when this was introduced or how long it has existed.


Ghabry:
This problem sounds familiar to me

PR #1558, opened 2018-12-11
Fix bug in CommandChangeHp when whole party is damaged

Ghabry:
CommandChangeHp trashed the amount variable resulting in wrong HP changes when the whole party was affected


carstene1ns:
Well, I would do it, however not possible from here.

Issue #1565, opened 2018-12-14
Automatic event processing from past map continues if you load a new savefile from event command.


elsemieni:
When loading a savefile from command, running events from original map (before load) don't stop processing, cointinuing in the new map in the new "scene".

#### Player:
Tested in 0.5.4 and Nightly build (to Dec 13th).

#### Testcase:
An RPG Maker 2003 1.12a project. 
https://drive.google.com/file/d/1Uu-TJriFOgJ6wrHViKbze8re3wLqvKjP/view?usp=sharing

The testcase consist in 2 maps and one savefile. 
When you start the game, you will start in Map A with an automatic event that makes a beep sound each second until 10 and then it repeats, and a parallel event that wait 3 seconds before calling Load Savefile command.

When you load the savefile, you will end in Map B, with no events at all.

- In RPG_RT the automatic event will stop instantly.
- In Player, the automatic event will continue to process, and making beeps each second.

fmatthew5876:
Thank you for the bug report, this is great find.

This is another reason we need to move off of globals and use RAII. With automatic cleanup, bugs like this become impossible to write.

fmatthew5876:
This bug seems to only affect autostart events.

I tried adding this parallel event:
```
Tint Red 4.0s Wait
Tint Green 4.0s Wait
Tint Blue 4.0s Wait
```

After loading the game, the tint commands did not continue.

I changed the sound beep event from Auto start to OnAction and then talked to it with the player. The beeping did not continue.

Ghabry:
The ability of RPG Maker 2k3 to Load anywhere and skip the title scene really caused many problems

Issue #1559, opened 2018-12-12
OpenSaveMenu waits for movement


fmatthew5876:
Create the following autostart event:

```
Move Player: Right, Repeat
OpenSaveMenu
```

In `RPG_RT`, the save menu will popup every frame. In `Player`, the player will take a full step, and then the save menu will appear.

Ghabry:
https://github.com/EasyRPG/Player/pull/1401/commits/a9e47008466ff749574539974d32da50841e336e

fmatthew5876:
Looks like we have a good test case.

Ghabry:
though this patch is not really acceptable because it adds more to Game_Temp

PR #1566, opened 2018-12-14
Interpreter Fixes

fmatthew5876:
Fixes: #1565

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -91,6 +91,7 @@ void Game_Interpreter::Clear() {
 			child_interpreter.reset();
 	}
 	list.clear();
+	ResetEventCalling();
 }
 
 // Is interpreter running.
@@ -197,11 +198,11 @@ void Game_Interpreter::Update() {
 		}
 
 		if ((Game_Temp::battle_calling && !Game_Temp::battle_running) ||
-			Game_Temp::shop_calling ||
-			Game_Temp::name_calling ||
-			Game_Temp::menu_calling ||
-			Game_Temp::save_calling ||
-			Game_Temp::load_calling ||
+			IsShopCalling() ||
+			IsNameCalling() ||
+			IsMenuCalling() ||
+			IsSaveCalling() ||
+			IsLoadCalling() ||
 			Game_Temp::to_title ||
 			Game_Temp::gameover) {
 
@@ -3069,6 +3070,23 @@ bool Game_Interpreter::DefaultContinuation(RPG::EventCommand const& /* com */) {
 	index++;
 	return true;
 }
+
+void Game_Interpreter::ResetEventCalling() {
+	event_calling = {};
+	// FIXME: Need to separate immediate battle calls from events
+	// and SavePartyLocation::encounter_calling for random encounters.
+	Game_Temp::battle_calling = false;
+}
+
+
+bool Game_Interpreter::IsImmediateCall() const {
+	return event_calling.load
+		|| event_calling.save
+		|| event_calling.name
+		|| event_calling.shop
+		|| event_calling.menu
+		|| Game_Temp::battle_calling;
+};
 
 // Dummy Continuations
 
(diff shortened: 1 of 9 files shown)


fmatthew5876:
Fixes #1559 and implements 1 commit from #1401

fmatthew5876:
All of the calls are immediate except random encounter battles and menu triggered by `ESC`.

The non-immediate cases for menu and battle are not yet perfectly frame accurate but with this branch they are pretty close and exhibit the main features. The main one being esc menu waits for you to finish you step before triggering.

src/game_map.cpp
```diff
@@ -128,6 +128,7 @@ void Game_Map::Dispose(bool is_load_savegame) {
 
 	map.reset();
 	animation.reset();
+	interpreter.reset(new Game_Interpreter_Map(0, true));
```
Ghabry: Please edit the issue number this fixes in the commit

Ghabry:
Same aplies to other commits that has numbers.

Interpreter is very tricky, so we need good documentation here in case some lines cause regressions later.

src/game_map.cpp
Ghabry: also you can't simply reset the global Interpreter on map change, events from the previous map continue.

src/game_interpreter_map.h
```diff
@@ -38,6 +38,8 @@ class Game_Interpreter_Map : public Game_Interpreter
 public:
 	Game_Interpreter_Map(int _depth = 0, bool _main_flag = false);
 
+	~Game_Interpreter_Map() {}
```
Ghabry: any reason for this?
I bet some obscure corner case specified in the standard almost nobody knows for "C++ will leak this because it is inherited"?

src/scene_map.cpp
```diff
@@ -260,14 +266,12 @@ void Scene_Map::CallBattle() {
 }
 
 void Scene_Map::CallShop() {
-	Game_Temp::shop_calling = false;
```
Ghabry: Any reason the Game_Map::GetInterpreter().ResetImmediateCalling is in Game_Interpreter and not in the Call-functions as before?

src/scene_map.cpp
```diff
@@ -190,7 +190,7 @@ void Scene_Map::Update() {
 	}
 
 	if (!Main_Data::game_player->IsMoving() || Game_Map::GetInterpreter().IsImmediateCall()) {
-		if (Game_Temp::menu_calling || Game_Temp::menu_calling_immediate) {
+		if (Main_Data::game_data.party_location.menu_calling || Game_Temp::menu_calling_immediate) {
```
Ghabry: Maybe Main_Data::game_data.party_location could be exposed via functions in Game_Party?

src/game_map.cpp

fmatthew5876:
I've refactored this:

* The whole `Game_Map::Dispose(bool)` thing is ugly. I changed it to call `Clear()` inside `Game_Interpreter_Map::SetupFromSave()`.
* Moved the `Game_Temp` calling flags (except battle) into `Game_Interpreter`. Instead of trying to rebase them and messing up, I've smashed the sormat, menu_calling, and game_temp features into 1 commit. Apologies if its a bit large.

src/game_interpreter.cpp
```diff
@@ -197,11 +198,12 @@ void Game_Interpreter::Update() {
...
-			Game_Temp::name_calling ||
-			Game_Temp::menu_calling ||
-			Game_Temp::save_calling ||
-			Game_Temp::load_calling ||
+			IsShopCalling() ||
+			IsNameCalling() ||
+			IsMenuCalling() ||
+			IsSaveCalling() ||
+			IsLoadCalling() ||
+			Main_Data::game_data.party_location.menu_calling ||
```
fmatthew5876: Not sure if we need the ESC menu calling flag here.

@Ghrabry what do you think about this?

Ghabry:
Attack on Game_Temp has begun, 5 variables removed 👍

src/game_interpreter.cpp
fmatthew5876: Fixed

Ghabry:
Jenkins: test this please

PR #1572, opened 2018-12-16
Wasapi audio fixes

Ghabry:
Could be Hardware depended result for me:

When using SDL_Mixer everything worked, but with SDL it was broken.

Moved both backends to the new SDL_OpenAudio API.

### DIFF
--- a/src/scene.cpp
+++ b/src/scene.cpp
@@ -27,9 +27,9 @@
 #include "transition.h"
 
 #ifndef NDEBUG
-#define DEBUG_VALIDATE() Scene::DebugValidate(__PRETTY_FUNCTION__)
+#define DEBUG_VALIDATE(x) Scene::DebugValidate(x)
 #else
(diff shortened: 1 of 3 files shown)


src/scene.cpp
```diff
@@ -27,9 +27,9 @@
 #include "transition.h"
 
 #ifndef NDEBUG
-#define DEBUG_VALIDATE() Scene::DebugValidate(__PRETTY_FUNCTION__)
```
carstene1ns: Could also use [__FUNCDNAME__](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2017) or such for msvc

Ghabry: yeah, was not sure if I want to add another preprocessor if-block or simply hardcode the function names.

fmatthew5876: The scope of this is so limited it's probably fine to just hard code the names instead of doing nasty stuff with macros.

Issue #1570, opened 2018-12-15
Player stuck in infinite loop with long battle messages


fmatthew5876:
Create a skill with a start message of "%S%S%S%S%S%S%S%S%S%S%S%S%S%S%S"

When used by an enemy in `RPG_RT`, all of the %S will be replaced by the enemy name. The line will be truncated and not word wrapped.

In Player, the game will freeze in an infinite loop.

PR #1573, opened 2018-12-16
Refactor the battle console - WIP

fmatthew5876:
As promised after battle 7 was merged, this is a series of commits to refactor the battle console and cleanup some of the mess from battle 5.

Included is a refactor of the word wrapping algo to fix the hang in #1570.

Currently this uses gtest for unit testing the WordWrap() function. I will change it to doctest, but for now I've put this PR up for early feedback.


fmatthew5876:
So next task will be supporting word wrap properly. 
I've got a skill with usage message 1 as `1%S%S%S%S 2%S%S%S%S 3%S%S%S%S%S 4%S%S%S%S 5%S%S%S 6%S%S%S%S 7%S%S%S%S` and usage message 2 as `11111111112222222222333333333344444444445555555555`

Here is what `RPG_RT` does.

output

fmatthew5876:
Here is a good test case for the start of battle:

output-start

fmatthew5876:
Test cases for usage messages:

Skill Using Messages and Animation Timing:
usage1

fmatthew5876:
Test cases for weapons:

A weapon with attack all that crits and kills

attack-all1

fmatthew5876:
Escape and Self-Destruct:

escape
explode

fmatthew5876:
Now we need to figure out the testing framework question?

What is the best way to pull in doctest? I see we have a copy in liblcf. Are we going to base everything on the liblcf one? Or change everything to use upstream version?

fmatthew5876:
This is ready for review and testing. The timings of the console delays are not perfect. We need a frame accurate video capture to tweak it.

src/scene_battle_rpg2k.cpp
```diff
@@ -379,286 +376,354 @@ void Scene_Battle_Rpg2k::ProcessActions() {
...
+	if (action->TargetNext()) {
+		return ProcessActionExecute(action);
+	}
+
+	battle_message_window->Clear();
+	return true;
+}
+
+
+#if 0
```
Ghabry: if 0 - endif

Ghabry:
About the testing code: I'm fine with the current solution. This can be adopted to the liblcf way ("make check" and all available tests compiled in a single doctest executable) in a later PR as this also involves changes to the other tests we have.

src/scene_battle_rpg2k.cpp
fmatthew5876: Fixed

fmatthew5876:
@CherryDT

Do you have information on the frame timings for the various console messages in battle? We have some numbers in the code but I don't know if they are correct.

How about battle animations? I tried a weapon with an animation of 99 frames and in RPG_RT it appeared to run faster in RPG_RT than in Player.

There is the 20 frames cutoff for sound animations reported by Sormat, but maybe there are other things like this?

There are also some cases, like in between targets where it appears to clear the console for one frame before starting the messages for the next target.

Getting the timings down is the last bit of polish to having a feature complete 2k battle system.

fmatthew5876:
I have found more regressions here. This will need more testing.

fmatthew5876:
So after several refactors and re-orgs, I've tested this PR again. I re-checked all the test cases here and some more from the commit messages for Sormat's battle PRs.

I refactored the Execute stage code to be much more clear and obvious than the last iteration. I've also played through the intro of HH3 and everything looks perfect except some animation issue I will fix in another PR.

src/game_battlealgorithm.h
```diff
@@ -381,6 +379,11 @@ class AlgorithmBase {
 	 */
 	void SetRepeat(int repeat);
 
+	/*
```
fdelapena:
```suggestion
	/**
```

fmatthew5876:
I've addressed @carstene1ns review comments and rebased to master

Ghabry:
There is no way to [Approve] after a merge :-1:

PR #1578, opened 2018-12-17
Fix 1574, 1576 and improve title reload

Ghabry:
The title cache change removes one Game_Temp variable and should unbreak "Return To Title" event command.

### DIFF
--- a/src/scene_title.cpp
+++ b/src/scene_title.cpp
@@ -51,9 +51,9 @@ void Scene_Title::Start() {
 }
 
 void Scene_Title::Continue() {
-	if (Game_Temp::restart_title_cache) {
-		// Clear the cache when the game returns to the
-		// title screen e.g. by pressing F12, except the Title Load menu
+	if (restart_title_cache) {
+		// Clear the cache when the game returns to the title screen
+		// e.g. by pressing F12, except the Title Load menu
 		Cache::Clear();
 		AudioSeCache::Clear();
 
@@ -181,7 +181,6 @@ void Scene_Title::CommandNewGame() {
 	if (!CheckValidPlayerLocation()) {
 		Output::Warning("The game has no start location set.");
 	} else {
-		Game_Temp::restart_title_cache = true;
 		Output::Debug("Starting new game");
 		Game_System::SePlay(Game_System::GetSystemSE(Game_System::SFX_Decision));
 		Game_System::BgmStop();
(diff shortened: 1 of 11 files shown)


Ghabry:
jenkins: test this please

src/scene_title.cpp
```diff
@@ -51,9 +51,9 @@ void Scene_Title::Start() {
...
 
 void Scene_Title::Continue() {
-	if (Game_Temp::restart_title_cache) {
-		// Clear the cache when the game returns to the
-		// title screen e.g. by pressing F12, except the Title Load menu
+	if (restart_title_cache) {
+		// Clear the cache when the game returns to the title screen
+		// e.g. by pressing F12, except the Title Load menu
 		Cache::Clear();
 		AudioSeCache::Clear();
```
fmatthew5876: Does this logic have to be done in `Scene_Title::Continue()`? Possible to just do it directly in `OnGameLoad()` (rename it) and then call that directly on startup also?

We localized the restart flag which is good. Possible to eliminate it entirely?

fmatthew5876: Actually I see how this works now. Doing it this way probably makes sense to ensure we don't miss an edge case.

Ghabry: It should be possible to put some of this tear-down in the Scene_Map deconstructor, this way we don't depend on the Title scene for this but I don't see a simply way to get rid of this boolean for the "Don't restart BGM when returning from Load scene" edge case.

fmatthew5876: Better to add state to handle it than try to juggle the edge cases. I was originally mistaken. The way you did it here is probably best.

Ghabry:
Fixed the issue found by ENI, forgot to delete the ResetEventCalling added in Clear.

That this was the bug was obvious after I started looking at the diff: https://github.com/EasyRPG/Player/pull/1566/files

Ghabry:
Cherry-picked one of Sormats commits which is known to be correct. But it doesn't fix #1587 so this is something else :/

fmatthew5876:
Are we sure forcing the frame to middle is right? Maybe its a timing issue where the jump doesn't start until you're at the middle frame, not the other way around?

Do you have a test case on that one? I could verify with my inotify script.

Ghabry:
For a proper test you should also remove the old if(IsContinuous) section.

give your character a charset that is significantly different for "frame left/right" compared to "center", add a "On Touch" event which does "Start Jump -> Right -> Right -> End Jump"

Walk on the event multiple times, you will see it.

Ghabry:
Removed sormats commit as there were valid concerns

PR #1579, opened 2018-12-17
Regression Fix: RTP table mapping japanese <-> other

Ghabry:
Maybe you remember how I fixed the 3DS crash by replacing std::string with const char* in the std::map of the rtp-table. The result of this change was that the function size was significantly reduced (which was probably the crash reason, function section too large???).

Turns out std::map doesn't implement a proper operator < for char* which means our table didn't work anymore :-1: .

### DIFF
--- a/src/rtp_table.h
+++ b/src/rtp_table.h
@@ -20,9 +20,16 @@
 
 #include <map>
 #include <string>
+#include <cstring>
 
 namespace RTP {
-	using rtp_table_type = std::map<const char*, std::map<const char*, const char*>>;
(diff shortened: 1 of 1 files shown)


fmatthew5876:
map will use `operator<` for `const char*`, comparing the 2 pointers..

C++17 `string_view` is another option here if the strings don't need to be owned by the map.

Ghabry:
> map will use operator< for const char*, comparing the 2 pointers..

That's why C++ is a bad language ;)

Issue #1571, opened 2018-12-15
Pictures won't load when loading savefile in a project with title screen disabled. 


elsemieni:
When you have a RM2k3 >=1.10 project without Ttitle screen enabled in database, when you try to load a savefile with pictures in game, these won't are being loaded (and didn't shown in screen). However, if you enable Title Screen in Database editor, the images are shown without issue. 

#### Player platform:
Tested in 0.5.4 and Master (Dec 15th). 

#### Test case
Here's a test case: 
[imgBug.zip](https://github.com/EasyRPG/Player/files/2683010/imgBug.zip)

Web: https://easyrpg.org/play/master/?game=issue-1571&engine=rpg2k3e  (ignore the missing FaceSets)

 A RM2k3 1.12a project with title screen disabled. It goes directly to a almost empty map with a automatic event that shows a dummy image, then it deletes itself, and a helper event for save/load calling. 

- Save the current progress (with debug features or with the helper event).
- Then load the savefile. 

#### Expected behaviour (RPG_RT 1.12a):
The dummy image is being shown when the map loads.

#### Current behaviour (Player 0.5.4 , Master):
No image is shown.

Talking with @fdelapena , maybe this bug can be related with https://github.com/EasyRPG/Player/issues/1524

fmatthew5876:
This might also explain the missing background in the Razas demo savegame from #1498

Ghabry:
And another test case for the web, thx 👍

fmatthew5876:
The whole business about a missing title screen is a red herring. This is broken everywhere.

PR #1580, opened 2018-12-17
Save game and move route fixes

fmatthew5876:
Fixes: #1571

I tested and this also correctly loads the mountain background from Razas save game from #1498

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -3076,9 +3076,6 @@ bool Game_Interpreter::DefaultContinuation(RPG::EventCommand const& /* com */) {
 
 void Game_Interpreter::ResetEventCalling() {
 	event_calling = {};
-	// FIXME: Need to separate immediate battle calls from events
-	// and SavePartyLocation::encounter_calling for random encounters.
-	Game_Temp::battle_calling = false;
 }
 
 
--- a/src/game_screen.cpp
+++ b/src/game_screen.cpp
@@ -34,7 +34,7 @@ static constexpr int kShakeContinuousTimeStart = 65535;
 Game_Screen::Game_Screen() :
 	data(Main_Data::game_data.screen)
 {
-	Reset(false);
+	Reset();
 }
 
 void Game_Screen::CreatePicturesFromSave() {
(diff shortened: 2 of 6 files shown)


src/game_screen.cpp
```diff
@@ -49,23 +49,22 @@ void Game_Screen::CreatePicturesFromSave() {
...
-		data.flash_time_left = 0;
-		data.flash_current_level = 0;
-	}
+	data.flash_red = 0;
+	data.flash_green = 0;
+	data.flash_blue = 0;
+	data.flash_time_left = 0;
+	data.flash_current_level = 0;
 	flash_sat = 0;
 	flash_period = 0;
```
fmatthew5876: Below this line (you can't see it unless you expand), we reset the tint if its negative.

Is it actually the case a negative value ever gets written to the tint chunks? If so it would be better to do this in fixup in liblcf.

fmatthew5876:
Added a battle regression fix and a brand new vehicle bug fix

fmatthew5876:
There was another bug in HH3 where player movement followed by `ProceedWithMovement` would fail.

This last one has to do with `ProceedWithMovement` commands. Why do we only check for `IsMovementRepeated()` in `Game_Map::IsAnyMovePending()`?

This makes no sense, and I can't imagine how `ProceedWithMovement` ever worked. It looks like `Game_Map::IsAnyMovePending()` hasn't been updated in 4 years??

@Ghrabry Please help!! I'm very confused about this one. What are we supposed to be doing in `Game_Map::IsAnyMovePending()`???

 :confused:

Ghabry:
I don't get it. In the second to last commit you remove the ! from the if and in the commit after you conclude that this change is broken and delete it???

fmatthew5876:
Well thats what happens when you are writing code all night..

I fixed this. My test case works now, with the vehicle fix.

Ghabry:
I will make some tests concerning Moves that cross map boundaries

src/game_interpreter.cpp
```diff
@@ -3076,9 +3076,6 @@ bool Game_Interpreter::DefaultContinuation(RPG::EventCommand const& /* com */) {
 
 void Game_Interpreter::ResetEventCalling() {
 	event_calling = {};
-	// FIXME: Need to separate immediate battle calls from events
-	// and SavePartyLocation::encounter_calling for random encounters.
-	Game_Temp::battle_calling = false;
```
Ghabry: I don't see any other change related to battle calling. Is just deleting this really the solution?

Ghabry:
actually the load scene hang is gone in master, maybe got the wrong download from Jenkins. Also cherry-picking your stuff doesn't add the hang.

PR #1584, opened 2018-12-18
Fix uninitialized Transition::screen_erase

fmatthew5876:
This is used uininitialized on transition.cpp:74 the first
time it is called.


Ghabry:
ASAN did not find this :(

PR #1590, opened 2018-12-23
Fix shop equipment arrows

fmatthew5876:
* Algorithm was wrong
* Support 2 weapon characters


Ghabry:
When you are 2-handed and you have 2 one-handed weapons and want to buy a 2-handed weapon where the resulting attack is smaller than the sum of the 2 one-handed you will also get an arrow down.

Ghabry:
Last merge before Xmas 🎄🎅

PR #1591, opened 2018-12-23
Generate Skill sound for Escape & Teleport (+Transition). Albeleon 1/2

Ghabry:
From #1401 

In TransitionOut I added an else case, otherwise the new transition would be overwritten...

These two look legit.

### DIFF
--- a/src/scene_teleport.cpp
+++ b/src/scene_teleport.cpp
@@ -21,6 +21,7 @@
 #include "game_player.h"
 #include "game_system.h"
 #include "input.h"
+#include "transition.h"
 
 Scene_Teleport::Scene_Teleport(Game_Actor& actor, const RPG::Skill& skill)
 		: actor(&actor), skill(&skill) {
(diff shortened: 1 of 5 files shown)


src/scene_teleport.cpp
```diff
@@ -42,7 +43,7 @@ void Scene_Teleport::Update() {
 	teleport_window->Update();
 
 	if (Input::IsTriggered(Input::DECISION)) {
-		Game_System::SePlay(Game_System::GetSystemSE(Game_System::SFX_UseItem));
+		Game_System::SePlay(skill->sound_effect);
```
fmatthew5876: This causes a crash. It needs to be moved down into the `if (skill)` block.

fmatthew5876: And in the item block, we need to do `ReaderUtil::GetElement(Data::skills, item->skill_id)` and do the skill SFX there.

fmatthew5876:
Using an Escape skill item doesn't generate the sound effect here. Can you add that too?

src/scene_teleport.cpp
fmatthew5876: Somehow my comment disappeared...?

This causes a crash now. The `skill` pointer here is null when a teleport item is used. We need it in the else block and there we have to lookup the skill from the item. 

The code is confusing as is. I think it would make more sense if `skill` pointer was always valid and `item` was the optional piece.

PR #1592, opened 2018-12-23
Event fixes. Albeleon 2/2 (Questionable!)

Ghabry:
From #1401 

I'm unsure if anything of this makes sense.
Give feedback ^^


Ghabry:
Great. These were the last commits contributed by Albeleon, which means we are done.

Issue #937, opened 2016-07-16
Pom Gets Wi-Fi: Interpreter stops after a "Stop!" message


fdelapena:
The issue title is pun intended, but it really happens :laughing:.

[Game](http://rpgmaker.net/games/5441/). [Savegame](https://github.com/EasyRPG/Player/files/367408/Save01.lsd.zip).

To trigger the halt sequence, press the decision key with the dog or with the door. After some dog advice and a transition, the door will be opened and a dog will say a "Stop!" message and interpreter halts. Some dogs should appear and start a battle instead. Earlier master versions were working there, so this is a regression.

**Actual Bug**
When a event that is "Waiting" collides with another event the wait timer reduces faster (depending on how many are colliding), so when one collides the wait countdown is twice as fast.

scurest:
[Cause](https://github.com/EasyRPG/Player/pull/936/commits/94cbf21ea2fd768ba541a32fee8e82a0e1d5ffe5#diff-80d5e57884c2e60d4f4084c5ab5ed79fR479).

Zegeri:
Here's the explanation: In the map 21 (The Gate Open2), the event EV0004 tries each frame to walk where EV0005 is. The move route of EV0004 is skippable. EV0005 has a parallel event with a ProceedWithMovement.

In EasyRPG, EV0004 will try to walk where EV0005 is. Because of #884, EV0005 runs its interpreter early, executes ProceedWithMovement and, because EV0004 is still moving, doesn't continue. After that, EV0004's movement fail. Because it's a skippable move route, the move route ends.

In RPG_RT, things go the same way but after all that, **EV0005's interpreter gets updated a second time**. This time, EV0004 doesn't have an overwritten move route, so it passes the ProceedWithMovement command.

I'm not sure in what circumstances does an event get update twice.

Ghabry:
In the meanwhile we know that it's the collision which causes this behaviour.

As a notice for the laaaaaaaate future when the editor is ready: The Player should offer for EasyRPG games a "Engine: EasyRPG" mode which does not emulate such bugs. ;)

carstene1ns:
This one has been broken in #1601. Resetting milestone.

fmatthew5876:
> In RPG_RT, things go the same way but after all that, **EV0005's interpreter gets updated a second time**. This time, EV0004 doesn't have an overwritten move route, so it passes the ProceedWithMovement command.
> 
> I'm not sure in what circumstances does an event get update twice.

[...]

Issue #663, opened 2015-12-06
Player step in place RPG_RT bug doesn't reproduce in Player.exe


trotterdotp:
This is a funny bug in RPG_RT that makes the hero step in place.

Explanative picture:
![stepinplace](https://cloud.githubusercontent.com/assets/12444176/11612908/0f936dd2-9c0d-11e5-9ecd-7a6d39c8bc9f.gif)

You need 2 maps events.

In the first one, start automatically:
<>move hero: slow down movement
<>delete this event

In the second map event, parallel:
<>move hero: speed up movement, speed up movement
<>wait for all movements to end
<>move hero: speed down movement, speed down movement
<>wait for all movements to end

In RPG_RT: you now have the player character stepping in place.
In Player.exe: nothing special happens.

[easyRPGdebugHeroSteps.zip](https://github.com/EasyRPG/Player/files/53119/easyRPGdebugHeroSteps.zip)

 I think this is the only way to have the player step in place like events can do.

Thanks for your work!

CherryDT:
It is amazing how you guys find things in RM's behavior even I didn't know about (and in this particular case it doesn't even make sense to me technically why that would happen). Actually, just from how the code looks, shouldn't this make the hero go at max. speed all the time?

Ghabry:
The comment was incorrect. I just looked in the attached game and it is. 2x speedup, wait movement, 2x speed down, wait movement. I edited the issue message.

And it really works :o

fmatthew5876:
If you start a new game on a map with no events, `anim_count` ticks up to 5 and then stops.

Somehow these move routes cause `anim_count` to tick like you're walking which triggers `anim_frame` to change.

fmatthew5876:
This behavior is now enabled in #1601, however the player move speed is very slow.

That is likely due to interpreter timing issues.

fmatthew5876:
This is now properly "fixed" in #1628 

It behaves the same as RPG_RT

Issue #579, opened 2015-09-07
The Sacred Tears TRUE: Character doesn't move after (very long) intro


Ghabry:
The first time the character is playable after the intro (in the house) the movement keys don't work.
Only Escape and Shift do. The workaround is to hold Shift, then moving works.

KeyInputProc problem?

Can't provide a Savegame because the created savegame is corrupted (#556)

Demo: http://nyu-media.com/sacred-tears-true/

The LMT file is "encrypted", decryption code (Python):

``` Python
#!/usr/bin/env python
import sys
with open("RPG_RT.lmt", "rb") as f:
    with open("RPG_RT.lmt2", "wb") as g:
        b = f.read(1)
        while b != b"":
            args = [chr(ord(b)-1 if ord(b)>0 else 255)]
            if (sys.version_info > (3, 0)):
                args.append("latin-1")
            _ = g.write(bytes(*args))
            b = f.read(1)
```

carstene1ns:
Just a short addition, this ruby script can also be used to decrypt and crypt the file:hammer::

```ruby
#!/usr/bin/env ruby

tree = IO.read('RPG_RT.lmt')
chars = tree.unpack('c*')
dir = chars[0] == 11 ? -1 : 1
chars.map! {|c| c + dir}
tree = chars.pack('c*')
File.open('RPG_RT.lmt','w') {|f| f.write tree}
```

Zegeri:
I can't reproduce this. I can move without problems inside and outside the house.

Ghabry:
That's interesting. Maybe your parallel process fix fixed this, too.
Checking later

fdelapena:
I can reproduce it in upstream master. It does not happen always, however. I got blocked in some case after navigating a bit the menu and returning to the game.

Ghabry:
So the important part is to fix #556 to get a proper savegame for easier testing.

carstene1ns:
Here is a savegame saved in RPG_RT right after the opening: http://f4ke.de/dev/easyrpg/TSTT/Save03.lsd

carstene1ns:
@Zegeri: Is this also fixed by 0c3fb4cb2cdc08209e9fd54b7593e84df9f83558?

Zegeri:
I'm sure that commit can't fix hangs. But it seems that exiting the house after loading your save file would cause the characters to disappear. Some other commit in my branch must have fixed it.

Ghabry:
[...]

PR #1601, opened 2019-01-05
Move Route refactors

### LINKED ISSUE #1267: Enter Vehicle fails when the Vehicle is entered through EnterVehicle while a move route is active
Test case: https://easyrpg.org/play/?game=issue-1267
Press SHIFT, Alex should move into the sky.

The problem is that our "GetOnOffVehicle" checks for "IsMovable" and one case of is movable is: MoveRouteOverwritten. Which is the case in this game.

In RPG_RT the move route actually continues after entering the vehicle.

### LINKED ISSUE #1293: Ascendence: Visual bugs
#### Name of the game:
Ascendence RSD VI - https://rpgmaker.net/games/736/downloads/

#### Player platform:
Windows (32 bit) - Continuous build

#### Attach files
save

#### Issues:
- [X] Slight animation bugs in intro for characters Kell and Velto (pulling animation, walking headbob hair) (**fixed in #1601 for 0.6.1**)
- [X] City intro fade-in is too quick compared to original
- [x] Screen shake for first elevator is not emulated (**fixed in 0.6**)
- [X] Strange red box around player when entering/leaving save zone (see gif)

### LINKED ISSUE #1346: Script for 8-way movement not smooth in EasyRPG
#### Name of the game:
Test Project

#### Player platform:
Windows 32 & 64-Bit

#### Attached files:
https://www.dropbox.com/s/vcchdiyqiom6eow/Project1.zip?dl=1

Log :
```
[2018-03-27 08:08:22] Debug: EasyRPG Player 0.5.3 (git~73@0220488, 2018-03-09) started
[2018-03-27 08:08:22] Debug: =========================================================
[2018-03-27 08:08:22] Debug: CLI: 
[2018-03-27 08:08:22] Debug: Opened audio at 44100 Hz (stereo), format: S16LSB
[2018-03-27 08:08:22] Debug: Cannot find: Title/Title1 (!)
[2018-03-27 08:08:22] Debug: Detected encoding: ibm-5348_P100-1997. Files not found. Trying next.
[2018-03-27 08:08:22] Debug: Cannot find: Title/Title1 (!)
[2018-03-27 08:08:22] Debug: Detected encoding: UTF-8. Files not found. Trying next.
[2018-03-27 08:08:22] Debug: Cannot find: Title/Title1 (!)
[2018-03-27 08:08:22] Debug: Detected encoding: ibm-5346_P100-1998. Files not found. Trying next.
[2018-03-27 08:08:22] Debug: Cannot find: Title/呩瑬攱 (!)
[2018-03-27 08:08:22] Debug: Detected encoding: UTF-16BE. Files not found. Trying next.
[2018-03-27 08:08:22] Debug: Cannot find: Title/楔汴ㅥ (!)
[2018-03-27 08:08:22] Debug: Detected encoding: UTF-16LE. Files not found. Trying next.
[2018-03-27 08:08:22] Debug: Bad encoding: ISO-8859-9. Trying next.
[2018-03-27 08:08:22] Debug: Cannot find: Title/Title1 (!)
[2018-03-27 08:08:22] Debug: Detected encoding: ibm-943_P15A-2003. Files not found. Trying next.
[2018-03-27 08:08:22] Debug: Cannot find: Title/Title1 (!)
[2018-03-27 08:08:22] Debug: Detected encoding: windows-936-2000. Files not found. Trying next.
[2018-03-27 08:08:22] Debug: Bad encoding: EUC-JP. Trying next.
[2018-03-27 08:08:22] Debug: Cannot find: Title/Title1 (!)
[2018-03-27 08:08:22] Debug: Detected encoding: windows-949-2000. Files not found. Trying next.
[2018-03-27 08:08:22] Debug: Cannot find: Title/Title1 (!)
[2018-03-27 08:08:22] Debug: Detected encoding: Big5. Files not found. Trying next.
[2018-03-27 08:08:22] Debug: Bad encoding: IBM424_ltr. Trying next.
[2018-03-27 08:08:22] Debug: Bad encoding: IBM424_rtl. Trying next.
[2018-03-27 08:08:22] Debug: Detected encoding: ibm-5348_P100-1997
[2018-03-27 08:08:22] Debug: Using . as Game and Save directory
[2018-03-27 08:08:22] Debug: Loading game Untitled
[2018-03-27 08:08:22] Debug: Using RPG2k3 (English release, v1.11) Interpreter
[2018-03-27 08:08:22] Debug: RPG2k >= v1.50 / RPG2k3 >= v1.05 detected
[2018-03-27 08:08:22] Debug: Engine configured as: 2k=0 2k3=1 2k3Legacy=0 MajorUpdated=1 Eng=1
[2018-03-27 08:08:22] Debug: Adding C:\Program Files (x86)\Steam\steamapps\common\RPG Maker 2003\RTP to RTP path
[2018-03-27 08:08:22] Debug: Adding C:\Program Files (x86)\rpg2003\RTP to RTP path
[2018-03-27 08:08:22] Debug: Actor 1: Removing invalid item 44 (of type 2) from equipment slot 2 (needs type 1)
[2018-03-27 08:08:23] Debug: libpng: iCCP: known incorrect sRGB profile
[2018-03-27 08:08:23] Debug: libpng: cHRM: inconsistent chromaticities
[2018-03-27 08:08:25] Debug: Starting new game
[2018-03-27 08:08:25] Debug: Loading Map Map0001.lmu
[2018-03-27 08:08:25] Debug: Tree: MAP0001
[2018-03-27 08:08:36] Debug: Actor 1: Removing invalid item 44 (of type 2) from equipment slot 2 (needs type 1)
```

#### Video showcasing issue:
[](url)https://youtu.be/5m9MSr2wX10

#### Description:
This is an event allowing for 8-way movement. it works fine in the regular RM2k3, but as of march 26, 2018, the movement is not in the latest continuous version of EasyRPG. Maybe it has to do with the "Wait for All Movement" command.

### LINKED ISSUE #1429: EasyRPG's "disembark vehicle" cmd doesn't operate the same as RM2K's
Tested in the stable version and in a more recent dev build (2aa5292). Same version I used to test my last bug.

This affects a game I was playing but I've made a minimal test case that reproduces the problem.

---

#### Name of the game:
Vehicle test

#### Player platform:
OSX

 - [Vehicle.zip](https://github.com/EasyRPG/Player/files/2450919/Vehicle.zip)
 - [easyrpg_log](https://gist.github.com/msikma/cb695910be988891e95a0ffbc6436a4e)

The player is on a boat, sailing on the water. You hit an event and the following commands are executed:

- Move boat left, left, face up (the exact movement commands don't really matter, as long as you can walk a few tiles)
- Ride/Disembark vehicle

In RM2k, the "disembark vehicle" command is executed immediately, before the move command is over. After disembarking it's the hero character that performs the move command.

This appears to only occur when the event is placed by the shore so that disembarking is possible right at the start.

RM2k: disembark, then hero walks left two tiles and faces up.

screen shot 2018-10-05 at 17 13 08

EasyRPG: does not disembark, boat goes left two tiles and faces up.

screen shot 2018-10-05 at 17 21 38

In the game I was playing this led to me being teleported to another map while still being a boat. 😄

### LINKED ISSUE #1585: Events with Spin restart their animation when write any command and interact with it
**Tested in** : Player 0.5.4 (continuous).
**How to reproduce it**: Create a new event with animation type "Spin" and write any command in it.
**Platform**: Windows 7 (x64)

**Test case**:
[Restart spin](https://github.com/EasyRPG/Player/files/2820060/TestCase.-.Event-Spin.zip)

### LINKED ISSUE #1587: Error to changing an event graphic using event pages (missing and wrong graphics)
**Tested in**: Player 0.5.4 (continuous).
**How to reproduce it**: Create an event with several pages and graphics (fixed graphic). A parallel process event will change among pages using variables and will make "Set Move Route".
**Platform**: Windows 7 (x64)

**Test case**:
[Changing charset graphic via event pages](https://github.com/EasyRPG/Player/files/2820062/TestCase.-.Change.events.graphic.zip)

### LINKED ISSUE #1593: Lakria Legends: Movement hangs on the ramp
via Android bug reporting.

#### Name of the game: [Lakria Legends](https://rpgmaker.net/games/5562/downloads/9559/)  
Yet another REFMAP RPG

Start with ``--new-game --start-map-id 18 --start-position 17 5``

Walk on the ramp and the game will get stuck in an endless movement uphill/downhill loop.

Quick event analysis:

Start conditions are "event touch"

```
Top ramp event: (disabled when Going Up = ON)
@> Set Move Route: Player, Turn Right
@> Control Switches: [0032:Going Down] = ON
@> Set Move Route: Player, Move Bottom Right, Move Bottom Right
@> Control Switches: [0032:Going Down] = OFF

Bottom ramp event: (disabled when Going Down = ON)
@> Set Move Route: Player, Turn Left
@> Control Switches: [0031:Going Up] = ON
@> Set Move Route: Player, Move Upper Left, Move Left
@> Control Switches: [0031:Going Up] = OFF
```

The switch toggling is obviously a red hering, the switch is already OFF again when the movement starts. Can be deleted, result is the same.

Basicly the other ramp event is touched while moving and this triggers the downwards movement.

We already ignore "Player touch" events when a move route is active (matches RPG_RT), maybe the same must be done for "Event touch"? Needs testing. (Changing the events from Event to Player touch fixes it in Player).

Note: Event Touch is CheckCollisionEvent() in Game_Player

### LINKED ISSUE #1604: Move towards player move route is wrong
There are 2 kinds of "move to player" movements.

There is `move_type == toward`. This one will move to the player but also dodge obstacles

Then there is move commands to "Step Towards Player" commands. These are dumb, in that they will only try to step to the player and fail if anything is in the way.

In Player, both behave like the `move_type` case.

### LINKED ISSUE #1654: Manually spin event is not working
Create 2 events, one of them a charset with the graphic (the event to animate), the other one as parallel process with:

> @> Wait: 0.5 seconds
@> Set Move Route: [Event to animate], Turn Right
@> Wait: 0.3 seconds
@> Set Move Route: [Event to animate], Turn Down
@> Wait: 1.5 seconds
@> Set Move Route: [Event to animate], Turn Left
@> Wait: 0.3 seconds
@> Set Move Route: [Event to animate], Turn Up

In my case, the result is the charset graphic stay in the center position without animation.

### LINKED ISSUE #1671: Enemies stay "hooked" unlimitedly in KangaPunch mini-game
There's a new problem in this mini-game with Player 0.6

In RPG_RT, the kangaroo can "hook" the enemy when the enemy is blocking hits for a limited time, in EasyRPG this hook is until the enemy is K.O.

(This is not intentional, it's the normal behavior of events in RPG_RT)

How to reproduce it: especially with Happy Hippo (it's the more defensive enemy), launch punches when the enemy is moving until the enemy stay blocking hits during a time, then, the enemy will move in other direction and going forward the kangaroo again.

Here you can see the difference:

RPG_RT
https://youtu.be/rFYmSVbPHOc

Player
https://youtu.be/yWgGdY4miho

fmatthew5876:
This PR is refactor `SaveMapEventBase` move route related chunks to behave more like `RPG_RT` and improve support loading save games correctly on any frame of movement.

This still needs a lot more testing and probably more changes to be correct. 

Detailed test cases and research can be found in #1567

Fix #1267 
Fix #1293
Fix #1346
Fix #1429
Fix #1585
Fix #1587
Fix #1593
Fix #1604
Fix #1654
Fix #1671

Breaks #579
Breaks #937

Behaviors of `stop_count`
- [x] Continues ticking through the end of the last move route command into infinity, does not reset
- [x] Resets every time the player takes a step from the keyboard
- [x] Initialized to 0 on the first frame of a new game
- [x] Does not reset when we change maps
- [x] Move route advances *after* `stop_count == max_stop_count`
- [x] Initialized to `65535` when a `SetMoveRoute` command is processed.
- [x] Initialized to `0` when the next move command is processed, and then advances to `1` on the same frame.
- [x] Move routes are not processed on the first frame of a new map, and so `stop_count` will not tick for the first frame. This means it'll be `0` if on the first frame you do have a parallel event do `OpenSaveMenu` or `65535` if you do `SetMoveRoute, OpenSaveMenu`. The behavior affects the Player and Vehicles but not map events! :question: 
- [x] `stop_count` ticks while any (player, event, or vehicle) is idle and has no move route, unless an event is active in the foreground (autostart, talk).
- [x] Does not tick for vehicles not on the current map. Stays frozen to whatever the previous value was when you teleport away from a map that had the vehicle.
- [x] Does not tick for events which don't have any active page.
- [x] If the last move command is a step, we ignore `max_stop_count` and cancel the move route as soon as the move finishes, reseting back to idle state.
- [x] If the last move command is a jump, we ignore `max_stop_count` and cancel the move route as soon as the move finishes, reseting back to idle state.
- [x] When player `vehicle > 0` and `aboard = 1`, that vehicle `stop_count` is frozen.
- [x] Does not change when you talk to an event and that event changes direction (or doesn't change)
- [x] SetEventLocation and SetVehicleLocation do not reset `stop_count`
- [x] Teleport does not reset `stop_count`

Behaviors of `max_stop_count`
- [x] Player and vehicle initialized to 0 on new game.
- [x] Map event initialized to stepping value of event's `move_frequency`.
- [x] Step: `move_frequency == 8 ? 0 : (2 ^ (9 - move_frequency))`
- [x] Turn: `move_frequency == 8 ? 0 : (2 ^ (8 - move_frequency))`
- [x] Wait: `20 + (move_frequency == 8 ? 0 : (2 ^ (8 - move_frequency)))`
- [x] When a `SetMoveRoute` command is processed, initialized to stepping value for the `move_frequency` of the move route.
- [x] After move route finishes, resets to stepping value of the `original_move_frequency`
- [x] Does not change when the player steps from player controls (remains 0 when walking around after new game, or 128 when walking around after the Player has had 1 move route since the game started).

Behaviors of `move_frequency`
- [x] Player defaults to 2 (empty chunk) and resets to 2 after move route finishes
- [x] Vehicle defaults to 2 (empty chunk) and resets to 2 after move route finishes
- [x] Event defaults to active page move_frequency and resets to it after move route finishes 
- [x] Event defaults to 2 (empty chunk) if no active page
- [x] When loading a save during an active move route, Player `move_frequency` resets to 0 after the move route finishes. (RPG_RT bug)
    - [x] Fixed in Player
- [x] When loading a save during an active move route, Vehicle `move_frequency` resets to 0 after the move route finishes. (RPG_RT bug)
    - [x] Fixed in Player
- [x] Increase/decrease frequency commands don't last after move route finishes

Behaviors of `move_commands`
- [x] Populated with array of commands when move route starts
- [x] Not cleared after move route finishes. Remains forever until next move route.

Behaviors of `move_route_index`
- [x] Starts at 0
- [x] Resets to 0 when new move route starts
- [x] When move command is being processed, points to the next move command, or `move_commands.size()` if this is the last move command.
- [x] If the last move command is a step, resets to 0 when move route finishes
- [x] If the last move command is not a step, remains unchanged when move route finishes

Behaviors of `move_route_overwrite`
- [x] Set 1 when move route starts
- [x] Set to 0 on the frame after move route finishes

Behaviors of `jumping`
- [x] Set to 1 on the frame the jump starts
- [x] Set to 0 on the last frame of the jump (when remaining_step = 0). This is different than most things which trigger on the frame after the movement.
- [x] SetEventLocation and SetVehicleLocation do not reset `jumping` (RPG_RT bug :exclamation: )
    - [x] Fixed in Player
- [x] Teleport does not reset `jumping` (RPG_RT bug :exclamation: )
    - [x] Fixed in Player

Behaviors of `remaining_step`
- [x] SetEventLocation and SetVehicleLocation reset `remaining_step`
- [x] Teleport resets `remaining_step`

Behaviors of `anim_count`
- [x] When `animation_type` is `fixed_graphic` or `step_frame_fix`, `anim_count` is always 0.
- [x] limit non-continuous = [11, 9, 7, 5, 4, 3] (based on `move_speed`)
- [x] limit continuous = [15, 11, 9, 7, 6, 5] (based on `move_speed`)
- [x] limit spin = [23, 14, 11, 7, 5, 3] (based on `move_speed`)
- [x] When anim_count reaches appropriate limit, either freezes there or increments `anim_frame` and goes back to 0. (see cases below)
- [x] when event `animation_type` is non_continous, `remaining_step=0`, and `anim_frame=1 or 3`, `anim_count` counts up to the non_continuous limit and then freezes there.
- [x]  when event `animation_type` is non_continous, `remaining_step=0`, and `anim_frame=0 or 2`, `anim_count` counts up to the continuous limit and then `anim_frame` cycles to the next value (1 or 3)
- [x]  when event `animation_type` is continous, `remaining_step=0`, `anim_count` counts up to the continuous limit and then `anim_frame` cycles to the next value
- [x] When `remaining_step > 0`, `anim_count` counts up to the non_continuous limit and then `anim_frame` cycles to the next value. Note that `remaining_step` decrements first, so this rule relies on the `remaining_step` value at the beginning of this frame.
- [x] Decreasing `move_speed` will cause `anim_count` to increment further as the limit increases.
- [x] Increasing `move_speed` will reduce the limit but `anim_count` will not change.
- [x] If you increase the `move_speed` enough that your current `anim_count` is now above the continuous limit for the new speed, the `anim_frame` will increment. #663 
- [x] Setting a tile graphic doesn't change anything.
- [x] Change graphic move command has no effect on animations.
- [x] When `jumping=1`, then we force `anim_count=0` and `anim_frame=1`. This means when a jump is started, the animation resets.
- [x] On the last frame of the jump, `jumping=0` and `anim_count=1`. That is, the `anim_count` starts ticking on the last frame of the jump, instead of waiting for the next frame to start.
- [x] When no event page is active, `anim_count=0` and `anim_frame=1`
- [x] SetEventLocation and SetVehicleLocation do not reset `anim_count`
- [x] Teleport resets `anim_count`
- [x] Is reset when `anim_paused=1`

Behaviors of `anim_frame`:
- [x] 0 = Left
- [x] 1 = Middle
- [x] 2 = Right
- [x] 3 = Middle
- [x] Defaults to 1, and then cycles 1, 2, 3, 0, 1, 2, 3, 0, ... as mentioned above in `anim_count` cases.
- [x] SetEventLocation and SetVehicleLocation do not reset `anim_frame`
- [x] Teleport resets `anim_frame`
- [x] Is reset when `anim_paused=1`

Animation of Vehicles:
- [x] Boat and Ship limit is always 15 regardless of `move_speed` and always continuously animating `anim_frame` to this.
- [x] Airship `anim_count=0` and `anim_frame=1` while `aboard=0`. This includes getting reset when landing
- [x] Airship limit is 11 and continuously animates while boarded
- [x] No change when player boards a vehicle, except the move speed of the vehicle (airship) can change how the animation continues as described above.
- [ ] If you put any move route on the airship, it's `anim_count` limit changes to 15 instead of 11 and seems to stay there forever. If you quit and load and saved game it goes back to 11 until you make a new move route. RPG_RT bug?? :exclamation: 

Behaviors of spinning events:
- [x] Spinning animation only changes `sprite_direction`.
- [x] A Turn move command changes both `direction` and `sprite_direction`
- [x] If a Turn move command is started on the same frame as animation, the `direction` will be of the turn, and `sprite_direction = (direction + 1) % 4`
- [x] A Move move command changes only `direction`
- [x] Triggering a spinning event via Action, Player Touch, or Event Touch does not affect `direction` or `sprite_direction`
- [x] When a jump is started, both `direction` and `sprite_direction` are set in the direction of the jump
- [x] When a jump is in progress, the spin will still continue to animate.
- [x] If a jump is started on the same frame as animation, the `direction` will be of the jump, and `sprite_direction = (direction + 1) % 4`

Player Controls:
- [x] Player controls don't respond until movement and `stop_count` of last move command finishes.

Vehicle Move Route interactions:
- [x] Airship move route gets canceled 1 frame after player boards
- [x] An active Boat, Ship move route gets canceled when player finishes boarding and `aboard` is set to 1.
- [x] Boat, Ship position fixed to player position when boarding finishes. This can be seen by moving the ship with a move route while the player is boarding.
- [x] If `Player::vehicle > 0`, SetMoveRoute against that vehicle gets applied to the player instead
- [x] If a move route gets set on the player while boarding, it will wait until the player finishes boarding and then execute on the player in the vehicle.

TBD:
- [x] How overwritten move route interacts with Event original move_route
- [x] Behavior of counters when move fails
- [x] Devirtualize GetOriginalMoveRouteIndex()
- [x] Don't store copy of original_move_route
- [x] When active move route is interrupted by another move route
- [x] How move route interacts with teleport
- [x] How move route interacts with get on/off vehicle  #1429
- [x] Investigate `through` and `route_through` behavior when getting on/off vehicles
- [x] Effect of Erase Event and disabling all active pages. What happens when you do this mid move, `stop_count`, `anim_count`, etc...?
- [x] Effect of changing the active event page in the middle of a move or wait? Animations?
- [x] Effect of Animation Off move command during different animation frames.
- [x] Effect of failed jump on animation
- [x] Effect of wait for all movement
- [x] Effect of stop all movement
- [x] Effect of `move_type=random`

Please post more test cases if you can think of any not here.

### DIFF
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -86,6 +86,8 @@ bool Game_Character::IsLandable(int x, int y) const
 
 	if (GetThrough()) return true;
 
+	if (IsFlying()) return true;
+
 	if (!Game_Map::IsLandable(x, y, this))
 		return false;
 
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -707,27 +707,33 @@ void Game_Character::BeginJump(int32_t& current_index, const RPG::MoveRoute& cur
 		}
 	}
 
-	if (jump_plus_x != 0 || jump_plus_y != 0) {
-		if (std::abs(jump_plus_y) >= std::abs(jump_plus_x)) {
-			SetDirection(jump_plus_y > 0 ? Down : Up);
-			if (!IsDirectionFixed() && !IsFacingLocked()) {
-				SetSpriteDirection(GetDirection());
-			}
-		} else {
-			SetDirection(jump_plus_x > 0 ? Right : Left);
-			if (!IsDirectionFixed() && !IsFacingLocked()) {
-				SetSpriteDirection(GetDirection());
-			}
-		}
-	}
-
 	if (
 		// A character can always land on a tile they were already standing on
 		!(jump_plus_x == 0 && jump_plus_y == 0) &&
 		!IsLandable(new_x, new_y)
 	) {
-		// Reset to begin jump command and try again...
 		move_failed = true;
+	}
+
+
+	if (!move_failed || !current_route.skippable) {
+		if (jump_plus_x != 0 || jump_plus_y != 0) {
+			if (std::abs(jump_plus_y) >= std::abs(jump_plus_x)) {
+				SetDirection(jump_plus_y > 0 ? Down : Up);
+				if (!IsDirectionFixed() && !IsFacingLocked()) {
+					SetSpriteDirection(GetDirection());
+				}
+			} else {
+				SetDirection(jump_plus_x > 0 ? Right : Left);
+				if (!IsDirectionFixed() && !IsFacingLocked()) {
+					SetSpriteDirection(GetDirection());
+				}
+			}
+		}
+	}
+
+	if (move_failed) {
+		// Reset to begin jump command and try again...
 		SetJumping(false);
 
 		if (current_route.skippable) {
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -32,9 +32,7 @@
 #include <cassert>
 
 Game_Character::Game_Character(RPG::SaveMapEventBase* d) :
-	move_type(RPG::EventPage::MoveType_stationary),
 	move_failed(false),
-	move_count(0),
 	jump_plus_x(0),
 	jump_plus_y(0),
 	visible(true),
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -140,8 +140,6 @@ int Game_Character::GetScreenZ(bool apply_shift) const {
 }
 
 void Game_Character::UpdateMovement() {
-	bool was_move_route_overriden = IsMoveRouteOverwritten();
-
 	if (IsMoveRouteOverwritten()) {
 		UpdateMoveRoute(data()->move_route_index, data()->move_route);
 	}
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -36,7 +36,6 @@ Game_Character::Game_Character(RPG::SaveMapEventBase* d) :
 	original_move_frequency(-1),
 	move_type(RPG::EventPage::MoveType_stationary),
 	move_failed(false),
-	last_move_failed(false),
 	move_count(0),
 	jump_plus_x(0),
 	jump_plus_y(0),
@@ -149,31 +148,58 @@ int Game_Character::GetScreenZ(bool apply_shift) const {
 	return z;
 }
 
-void Game_Character::Update() {
-	if (GetStopCount() >= GetMaxStopCount()) {
-		if (IsMoveRouteOverwritten()) {
-			MoveTypeCustom();
-		} else {
-			// Only events
+void Game_Character::UpdateMovement() {
+	bool was_move_route_overriden = IsMoveRouteOverwritten();
+
+	if (IsMoveRouteOverwritten()) {
+		UpdateMoveRoute(data()->move_route_index, data()->move_route);
+	}
+
+	if (!IsMoveRouteOverwritten()) {
+		if (IsStopping() && !IsStopCountActive()) {
 			UpdateSelfMovement();
 		}
 	}
-}
 
-void Game_Character::UpdateSprite() {
+	bool moved = false;
+
 	if (IsJumping()) {
+		SetStopCount(0);
 		UpdateJump();
+		moved = true;
 		if (IsSpinning())
 			SetAnimCount(GetAnimCount() + 1);
 	} else if (IsMoving()) {
+		SetStopCount(0);
 		SetRemainingStep(GetRemainingStep() - min(1 << (1 + GetMoveSpeed()), GetRemainingStep()));
+		moved = true;
 		if (IsSpinning() || IsAnimated())
 			SetAnimCount(GetAnimCount() + 1);
 	} else {
-		data()->stop_count++;
+
+		if (IsMoveRouteOverwritten() || (!Game_Map::GetInterpreter().IsRunning() && !Game_Map::IsAnyEventStarting())) {
+			SetStopCount(GetStopCount() + 1);
+		}
 
 		if (IsAnimated() && (IsSpinning() || IsContinuous() || GetAnimFrame() != RPG::EventPage::Frame_middle))
 			SetAnimCount(GetAnimCount() + 1);
+	}
+
+	// These actions happen after movement has finished but before stop count timer.
+	if (IsStopping() && IsMoveRouteActive() && IsMoveRouteOverwritten()) {
+		const auto& move_route = GetMoveRoute();
+		if (!move_route.move_commands.empty() && GetMoveRouteIndex() >= move_route.move_commands.size()) {
+			if (move_route.repeat) {
+				SetMoveRouteRepeated(true);
+				SetMoveRouteIndex(0);
+			} else if(moved) {
+				// If the last command of a move route is a move, and
+				// the move just completed, cancel immediately and
+				// don't wait for stop_count.
+				CancelMoveRoute();
+				SetMoveRouteIndex(0);
+			}
+		}
 	}
 
 	if (GetAnimCount() >= GetSteppingSpeed()) {
@@ -202,6 +228,7 @@ void Game_Character::UpdateSprite() {
 
 		SetAnimCount(0);
 	}
+
 }
 
 void Game_Character::UpdateJump() {
@@ -212,56 +239,53 @@ void Game_Character::UpdateJump() {
 	}
 }
 
-void Game_Character::UpdateSelfMovement() {
-	// no-op: Only events can have custom move routes
-}
+void Game_Character::UpdateMoveRoute(int32_t& current_index, const RPG::MoveRoute& current_route) {
+	auto is_move_route_possible = [&]() {
+		return IsStopping() && !IsStopCountActive();
+	};
 
-void Game_Character::MoveTypeCustom() {
-	// Detect if custom movement or event overwrite
-	const RPG::MoveRoute* active_route;
-	int active_route_index;
-	bool overwrite_changed = IsMoveRouteOverwritten();
-	if (IsMoveRouteOverwritten()) {
-		active_route = &GetMoveRoute();
-		active_route_index = GetMoveRouteIndex();
+	const int last_move_route_index = current_route.move_commands.size();
+
+	if (current_route.move_commands.empty()) {
+		return;
 	}
-	else {
-		active_route = &original_move_route;
-		active_route_index = GetOriginalMoveRouteIndex();
+
+	if (!is_move_route_possible()) {
+		return;
 	}
+
 	MoveOption option =
-		active_route->skippable ? MoveOption::IgnoreIfCantMove : MoveOption::Normal;
+		current_route.skippable ? MoveOption::IgnoreIfCantMove : MoveOption::Normal;
 
-	if (IsStopping()) {
-		move_failed = false;
+	move_failed = false;
 
-		int original_index = active_route_index;
-		bool looped_around = false;
-		while (true) {
-			if (!IsStopping() || GetStopCount() < GetMaxStopCount())
-				break;
+	int num_cmds_processed = 0;
 
-			if (active_route_index == original_index && looped_around) {
-				// We've gone around a full loop; stop here for now
-				break;
-			}
-
-			if ((size_t)active_route_index >= active_route->move_commands.size()) {
-				if (active_route->repeat && !active_route->move_commands.empty()) {
-					looped_around = true;
-					active_route_index = 0;
-					SetMoveRouteRepeated(true);
-					if (original_index == 0) {
-						break;
-					}
-				} else {
-					break;
+	while (true) {
+		//Move route is finished
+		if (current_index >= last_move_route_index) {
+			if (current_route.repeat) {
+				current_index = 0;
+			} else {
+				if (IsMoveRouteOverwritten()) {
+					CancelMoveRoute();
 				}
+				break;
 			}
 
-			const RPG::MoveCommand& move_command = active_route->move_commands[active_route_index];
+		}
 
-			switch (move_command.command_id) {
+		// Repeating move route will end after doing one full cycle in a single frame.
+		if (num_cmds_processed >= current_route.move_commands.size()) {
+			if (IsMoveRouteOverwritten() && current_route.repeat) {
+				SetMoveRouteRepeated(true);
+			}
+			break;
+		}
+
+		const RPG::MoveCommand& move_command = current_route.move_commands[current_index];
+
+		switch (move_command.command_id) {
 			case RPG::MoveCommand::Code::move_up:
 			case RPG::MoveCommand::Code::move_right:
 			case RPG::MoveCommand::Code::move_down:
@@ -321,7 +345,7 @@ void Game_Character::MoveTypeCustom() {
 				Wait();
 				break;
 			case RPG::MoveCommand::Code::begin_jump:
-				BeginJump(active_route, &active_route_index);
+				BeginJump(current_index, current_route);
 				break;
 			case RPG::MoveCommand::Code::end_jump:
 				// EndJump();
@@ -391,52 +415,23 @@ void Game_Character::MoveTypeCustom() {
 			case RPG::MoveCommand::Code::decrease_transp:
 				SetTransparency(GetTransparency() - 1);
 				break;
-			}
-
-			if (move_command.command_id <= RPG::MoveCommand::Code::move_forward) {
-				any_move_successful |= !move_failed;
-			}
-
-			last_move_failed = move_failed;
-			if (move_failed) {
-				if (active_route->skippable) {
-					last_move_failed = false;
-				} else {
-					break;
-				}
-			}
-
-			++active_route_index;
 		}
 
-		if ((size_t)active_route_index >= active_route->move_commands.size() && IsStopping()) {
-			if (IsMoveRouteOverwritten()) {
-				CancelMoveRoute();
-				Game_Map::RemovePendingMove(this);
-				SetStopCount(0);
-				SetMaxStopCountForStep();
-			}
+		//FIXME: DO we need this?
+		if (move_command.command_id <= RPG::MoveCommand::Code::move_forward) {
+			any_move_successful |= !move_failed;
 		}
-	}
 
-	// When the overwrite status changed the active_index belongs to the
-	// current non-active move route
-	if (overwrite_changed != IsMoveRouteOverwritten()) {
-		if (IsMoveRouteOverwritten()) {
-			SetOriginalMoveRouteIndex(active_route_index);
+		if (move_failed && !current_route.skippable) {
+			break;
 		}
-		else {
-			SetMoveRouteIndex(active_route_index);
+
+		++current_index;
+		++num_cmds_processed;
+		if(!is_move_route_possible()) {
+			break;
 		}
-	}
-	else {
-		if (IsMoveRouteOverwritten()) {
-			SetMoveRouteIndex(active_route_index);
-		}
-		else {
-			SetOriginalMoveRouteIndex(active_route_index);
-		}
-	}
+	} // while (true)
 }
 
 void Game_Character::Move(int dir, MoveOption option) {
@@ -461,18 +456,24 @@ void Game_Character::Move(int dir, MoveOption option) {
 		}
 	}
 
-	if (move_failed) {
-		if (!CheckEventTriggerTouch(Game_Map::RoundX(GetX() + dx), Game_Map::RoundY(GetY() + dy)))
-			return;
-	} else {
-		SetX(Game_Map::RoundX(GetX() + dx));
-		SetY(Game_Map::RoundY(GetY() + dy));
-		SetRemainingStep(SCREEN_TILE_SIZE);
-		BeginMove();
+	const auto new_x = Game_Map::RoundX(GetX() + dx);
+	const auto new_y = Game_Map::RoundY(GetY() + dy);
+
+	if (IsMoveRouteActive()) {
+		SetMaxStopCountForStep();
 	}
 
+	if (move_failed) {
+		CheckEventTriggerTouch(Game_Map::RoundX(GetX() + dx), Game_Map::RoundY(GetY() + dy));
+		return;
+	}
+
+	SetX(new_x);
+	SetY(new_y);
+	SetRemainingStep(SCREEN_TILE_SIZE);
+	BeginMove();
+
 	SetStopCount(0);
-	SetMaxStopCountForStep();
 }
 
 void Game_Character::MoveForward(MoveOption option) {
@@ -486,13 +487,6 @@ void Game_Character::MoveRandom(MoveOption option) {
 void Game_Character::MoveTowardsPlayer() {
 	int sx = DistanceXfromPlayer();
 	int sy = DistanceYfromPlayer();
-
-	// Try in the same direction of the last failed move
-	if (last_move_failed) {
-		MoveForward();
-		if (!move_failed)
-			return;
-	}
 
 	if (sx != 0 || sy != 0) {
 		if ( std::abs(sx) > std::abs(sy) ) {
@@ -588,7 +582,7 @@ void Game_Character::Wait() {
 	SetMaxStopCountForWait();
 }
 
-void Game_Character::BeginJump(const RPG::MoveRoute* current_route, int* current_index) {
+void Game_Character::BeginJump(int32_t& current_index, const RPG::MoveRoute& current_route) {
 	SetBeginJumpX(GetX());
 	SetBeginJumpY(GetY());
 	jump_plus_x = 0;
@@ -597,8 +591,8 @@ void Game_Character::BeginJump(const RPG::MoveRoute* current_route, int* current
 
 	bool end_found = false;
 	unsigned int i;
-	for (i = *current_index; i < current_route->move_commands.size(); ++i) {
-		const RPG::MoveCommand& move_command = current_route->move_commands[i];
+	for (i = current_index; i < current_route.move_commands.size(); ++i) {
+		const RPG::MoveCommand& move_command = current_route.move_commands[i];
 		switch (move_command.command_id) {
 			case RPG::MoveCommand::Code::move_up:
 			case RPG::MoveCommand::Code::move_right:
@@ -668,7 +662,7 @@ void Game_Character::BeginJump(const RPG::MoveRoute* current_route, int* current
 
 	if (!end_found) {
 		// No EndJump found. Move route ends directly
-		*current_index = i;
+		current_index = i;
 		SetJumping(false);
 		return;
 	}
@@ -717,8 +711,8 @@ void Game_Character::BeginJump(const RPG::MoveRoute* current_route, int* current
 		move_failed = true;
 		SetJumping(false);
 
-		if (current_route->skippable) {
-			*current_index = i;
+		if (current_route.skippable) {
+			current_index = i;
 			return;
 		}
 
@@ -727,7 +721,7 @@ void Game_Character::BeginJump(const RPG::MoveRoute* current_route, int* current
 
 	SetX(new_x);
 	SetY(new_y);
-	*current_index = i;
+	current_index = i;
 
 	SetRemainingStep(SCREEN_TILE_SIZE);
 	SetStopCount(0);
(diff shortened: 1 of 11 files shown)


Ghabry:
can't tell you the reason for "last_move_failed" yet but about "any_move_successful": Based on the comment in Game_player this was for fixing broken Ice physics in the game "Glimpse". Maybe you can try removing that handling and playing the game again maybe you fixed it in a better way by now.

And the other corner case is standing on a "on touch" handler which executes a skippable move command but that command fails, then the "on touch" handler doesn't fire again.

fmatthew5876:
This `last_move_failed` stuff seems not neccessary. I setup an event with move towards player with an obstacle in the way. As soon as I move the player around the event immediately changes direction to step towards the player.

fmatthew5876:
I've tested all of the cases from 1 to 14 in #1567. This PR matches RPG_RT with some exceptions

* `move_type == random` has some wierd behavior that needs to be investigated further
* There are some timing issues with the interpreter, such as player move routes being skipped for the very first frame, resulting in some off by 1 frame mismatches from the test cases. These are interpreter timing issues and not move route logic issues, so I will consider them separately.

fmatthew5876:
New changes, and now this PR passes test cases 15 and 16 from #1567.

Only caveat is again more instances of movement not being processed in the first frame.

fmatthew5876:
Fixed #1429

fmatthew5876:
Now passes Player #1567 Test cases 17 and 19

fmatthew5876:
This now passes #1567 Test cases 18

fmatthew5876:
This now passes test cases 22 and 23 (SetEventLocation and SetVehicleLocation). The jumping bug in RPG_RT is fixed in Player.

fmatthew5876:
I've redone test cases 1 - 8. The remaining issues are:

- [ ] 8.1 - `stop_count` is 0 in Player. (autostart should be delayed by 1 frame)
- [ ] 8.2 - `stop_count` stops incrementing at 2 instead of 3 (same issue as above)
- [ ] 8.4 - `stop_count` freezes at 2 (autostart event should frame delays between each invocation)
- [ ] 8.5 - `stop_count` freezes at 0 (same issue as above)

fmatthew5876:
I've gone through all of the test cases of #1567 again and marked the ones with differences in the previous comment.

All of the remaining issues are not really move route bugs but timing issues with the interpreter. They will be addressed in later PRs.

So this is ready for review and merge after 0.6.

fmatthew5876:
Rebased to master

Ghabry:
Can you sum up what you changed in bea23502701d0b019ff2b18b851c5675dcd89ee0 
That's really complex too review.

fmatthew5876:
> Can you sum up what you changed in [bea2350](https://github.com/EasyRPG/Player/commit/bea23502701d0b019ff2b18b851c5675dcd89ee0)
> That's really complex too review.

Can I say "Make the test cases pass"? Really there was no way to break it down in to nice digestable chunks. I just kept fixing logic until it was matching RPG_RT. I apologize for that nasty commit.

fmatthew5876:
For this one, instead of reviewing the diffs, it may be better to view the end results of `game_character`, `game_player`, `game_vehicle`, and `game_event` along with the test cases. There are so many changes here so that I don't think its possible to review them on a diff basis.

Ghabry:
Removing ``IsBlockedByMoveRoute`` breaks moving even when you have a move route under certain circumstances.

Example, have a parallel event:
``
SetMoveRoute Player "Turn Right"
``

Result: Player should be able to walk normally but doesn't trigger Touch or Collision events anymore.

Ghabry:
also breaks #937

fmatthew5876:
> 
> 
> Removing `IsBlockedByMoveRoute` breaks moving even when you have a move route under certain circumstances.
> 
> Example, have a parallel event:
> `SetMoveRoute Player "Turn Right"`
> 
> Result: Player should be able to walk normally but doesn't trigger Touch or Collision events anymore.

I plan to fix those things more generally in #1628 later. But for now I can remove the `IsBlockedByMoveRoute` commit from here if thats blocking this one from merge,

fmatthew5876:
I took out that commit and rebased to master

Ghabry:
hmm, removing that commit didn't help, is still broken.

You also break #937 because you fixed a RPG_RT bug ;)

fmatthew5876:
So there are 2 ways we can approach this:

1. Wait until I have 80+ commits from this, #1628, and others. Ensure absolutely no regressions and then review all of it and merge it all in one shot.
2. Reopen #937 (and any other regressions) to be fixed again properly in #1628. Then merge this PR soon.

I'm not in favor of (1) because it'll be a beast to review. Also getting some changes in piecemeal gives us time for further testing while the later interpreter work is pending,.

src/game_player.cpp
```diff
@@ -640,6 +640,7 @@ void Game_Player::UnboardingFinished() {
 	Unboard();
 	if (InAirship()) {
 		SetDirection(RPG::EventPage::Direction_down);
+		SetSpriteDirection(RPG::EventPage::Direction_down);
```
Ghabry: > Player sprite should face down after exiting airship 

This could cause issues when the Player has a the facing locked. Did you test this?

src/game_character.cpp
```diff
@@ -86,6 +86,8 @@ bool Game_Character::IsLandable(int x, int y) const
 
 	if (GetThrough()) return true;
 
+	if (IsFlying()) return true;
```
Ghabry: > Airship can jump and land anywhere 

Just a comment: It is confusing that "IsLandable" is not about Airship but Jump landing :D

src/game_player.cpp
```diff
@@ -464,12 +464,14 @@ bool Game_Player::GetOnVehicle() {
 	data()->preboard_move_speed = GetMoveSpeed();
 	if (type != Game_Vehicle::Airship) {
 		data()->boarding = true;
-		if (!GetThrough()) {
-			SetThrough(true);
-			MoveForward();
-			SetThrough(false);
-		} else {
-			MoveForward();
+		if (!IsMoving() && !IsJumping()) {
```
Ghabry: > Fix ship disembark behavior 

Disembark onto water? How is this possible? Is that a bug, if yes could you add a comment somewhere.

src/game_character.cpp
```diff
@@ -749,10 +745,6 @@ void Game_Character::BeginJump(int32_t& current_index, const RPG::MoveRoute& cur
 	SetStopCount(0);
 	SetMaxStopCountForStep();
 	move_failed = false;
-
-	if (IsContinuous()) {
-		SetAnimFrame(RPG::EventPage::Frame_middle);
-	}
```
Ghabry: Before a jump starts the animation is reset to the Middle frame.
Did you move this fix somewhere else?

src/game_character.cpp
```diff
@@ -415,9 +410,6 @@ void Game_Character::UpdateMoveRoute(int32_t& current_index, const RPG::MoveRout
 				data()->route_through = false;
 				break;
 			case RPG::MoveCommand::Code::stop_animation:
-				if (IsContinuous()) {
-					SetAnimFrame(RPG::EventPage::Frame_middle);
-				}
```
Ghabry: > Animtype fixes 

Are you sure with this deletion? It was added because when "Stop Animation" is called the movement frame is  supposed to reset to middle.
I suppose the new fix is the commit after this one? (currently 09eff58 until a rebase)

src/game_player.cpp
```diff
@@ -269,7 +269,8 @@ void Game_Player::Update(bool process_movement) {
 			}
 			SetMoveSpeed(vehicle->GetMoveSpeed());
 			vehicle->SetDirection(GetDirection());
-			vehicle->SetSpriteDirection(GetSpriteDirection());
+			vehicle->SetSpriteDirection(Left);
+			SetSpriteDirection(Left);
```
Ghabry: > Fix directions when finish boarding boat ship 

Same problem here with fixed graphics I suppose?

src/game_character.cpp
```diff
@@ -224,21 +224,18 @@ void Game_Character::UpdateAnimation(bool was_moving) {
 	const auto stationary_limit = stationary_limits[step_idx];
 	const auto continuous_limit = continuous_limits[step_idx];
 
-	if (GetAnimCount() >= continuous_limit) {
-		IncAnimFrame();
-		return;
+	if (IsContinuous()
+			|| was_moving
+			|| data()->anim_frame == 0 || data()->anim_frame == 2
```
Ghabry: > Refactor animation count 

Where 0 and 2 left and right? --> enum

src/game_character.h
```diff
@@ -1083,6 +1086,14 @@ inline void Game_Character::IncAnimCount() {
 
 inline void Game_Character::IncAnimFrame() {
 	data()->anim_frame = (data()->anim_frame + 1) % 4;
+	SetAnimCount(0);
+}
+
+inline void Game_Character::ResetAnimation() {
+	SetAnimCount(0);
+	if (GetAnimationType() != RPG::EventPage::AnimType_fixed_graphic) {
+		SetAnimFrame(1);
```
Ghabry: use the enum here RPG::EventPage::Frame_middle

src/game_event.cpp
```diff
@@ -625,3 +626,7 @@ const RPG::SaveMapEvent& Game_Event::GetSaveData() {
 
 	return *data();
 }
+
+bool Game_Event::IsMoveRouteActive() const {
+	return true;
```
Ghabry: > Move route refactors

Why is this always true for events?

src/game_character.cpp
```diff
@@ -149,33 +148,60 @@ int Game_Character::GetScreenZ(bool apply_shift) const {
...
 }
 
-void Game_Character::Update() {
-	if (GetStopCount() >= GetMaxStopCount()) {
-		if (IsMoveRouteOverwritten()) {
-			MoveTypeCustom();
-		} else {
-			// Only events
+void Game_Character::UpdateMovement() {
+	bool was_move_route_overriden = IsMoveRouteOverwritten();
```
Ghabry: > Move Route refactors
Is this unused? Can't find it in the diff

Ghabry:
> So there are 2 ways we can approach this

I accept that the two issues can be fixed in a later PR if this becomes to chaotic one if them is a special interaction with the interpreter.
I added all fixed and broken issues to the first comment (10 fixes, 2 breakages, don't think we ever had so many issues fixed at once)

Also marked everything that I would like to have discussed, when this is handled I accept this PR. thx.

src/game_player.cpp
fmatthew5876: It changes player direction even when facing locked flag is set. Player will be locked to down direction after landing airship.

src/game_character.cpp
fmatthew5876: This mess is getting cleaned up in #1628

src/game_player.cpp
fmatthew5876: Its in the test cases in #1429 

The code doesn't have anything specific to the "disembark into the water" bug.  It's just that it doesn't move again if you're already moving.

Do you still want some bug comment about that?

src/game_character.cpp
fmatthew5876: This was moved to `UpdateAnimation()` under `if (IsJumping())`.

The logic here is wrong. It always resets the `anim_frame` (and `anim_count` too), regardless of your movement type.

src/game_character.cpp
fmatthew5876: This was moved to `UpdateAnimation()`. Same comments as the jumping case.

src/game_player.cpp
fmatthew5876: Flag is ignored, just like airship disembark.

src/game_character.h
fmatthew5876: I can change these but we have 1 complication in that there are 2 frame middle values.

0 - left
1 - middle
2 - right
3 - middle

src/game_character.cpp
fmatthew5876: Its unused, will remove

src/game_event.cpp
fmatthew5876: It may be badly named, but I believe all the cases where its checked are always true for events but only true for player/vehicles with an active move route.

* Move route cancel - repeat logic - events can have custom move type so we need to check this
* `SetMaxStopCountForStep()` in `Game_Character::Move()` - This value is not set when the player moves via direction key. But is set for any event movement.
* `ForceMoveRoute()` - Need to clean up for events with custom move type.

This could use a refactor.

src/game_character.cpp
fmatthew5876: Also I verified if you're movement type is `fixed_graphic`, jump doesn't reset the `anim_frame` or `anim_count`. That's handled `UpdateAnimation()`

src/game_player.cpp
fmatthew5876: I added comments for this

fmatthew5876:
I've addressed the review comments.

Very thorough review. I appreciate it!

src/game_character.h
fmatthew5876: I added `Frame_middle2` in https://github.com/EasyRPG/liblcf/pull/316

The logic in `sprite_character.cpp:62` can be changed to use that enum value when its merged.

src/game_player.cpp
Ghabry: no is fine, thanks for the explanation

src/game_character.h
Ghabry: is in now :)

fmatthew5876:
Added middle2 enum

src/game_character.cpp
```diff
@@ -940,3 +940,19 @@ bool Game_Character::MakeWayDiagonal(int x, int y, int d) const {
 	return ((MakeWay(x, y, dy + 1) && MakeWay(x, y + dy, -dx + 2)) ||
 			(MakeWay(x, y, -dx + 2) && MakeWay(x + dx, y, dy + 1)));
 }
+
+void Game_Character::SetMaxStopCountForStep() {
+	const auto freq = GetMoveFrequency();
+	SetMaxStopCount(freq >= 8 ? 0 : 1 << (9 - GetMoveFrequency()));
```
carstene1ns: `GetMoveFrequency()` → `freq`?

fmatthew5876: Fixed and rebased to master. Thanks!

PR #1605, opened 2019-01-26
Battle 2k3: Visual fixes for CogDis

Ghabry:
This needs more testing to determine the correct logic but it seems that the condition for choosing between background and terrain is the terrain ID, not a non-empty background.

CogDis animates the background in a fancy way: the front image is only black + transparent and the background is a moving (downwards) mask image which defines the color. And via that trick they get 3 frames of animated stars on the background.

the lack of ``background->Update()`` completely broke any background scrolling in battles (upps) and the scrolling was too slow.

The rest fixes issues between Traditional vs Alternative style (still not fully matching but gets closer).

### DIFF
--- a/src/spriteset_battle.cpp
+++ b/src/spriteset_battle.cpp
@@ -77,6 +77,7 @@ void Spriteset_Battle::Update() {
 		}
 	}
 	background->SetTone(new_tone);
+	background->Update();
 
 	for (auto sprite : sprites) {
 		Game_Battler* battler = sprite->GetBattler();
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,6 +127,7 @@ add_library(${PROJECT_NAME}
 	src/weather.cpp
 	src/window_about.cpp
 	src/window_actorinfo.cpp
+	src/window_actorsp.cpp
 	src/window_actorstatus.cpp
 	src/window_actortarget.cpp
 	src/window_base.cpp
--- a/Makefile.am
+++ b/Makefile.am
@@ -262,6 +262,8 @@ libeasyrpg_player_la_SOURCES = \
 	src/window_about.h \
 	src/window_actorinfo.cpp \
 	src/window_actorinfo.h \
+	src/window_actorsp.cpp \
+	src/window_actorsp.h \
 	src/window_actorstatus.cpp \
 	src/window_actorstatus.h \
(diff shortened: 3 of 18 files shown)


Ghabry:
The first commit is probably wrong, currently doing some more testing.

Ghabry:
How the animation works, nice trick:

cogdis

Overlayed and moving (ingame the background moves, not the foreground, but result is the same)

cogdis

src/spriteset_battle.cpp
```diff
@@ -31,11 +31,11 @@
...
 Spriteset_Battle::Spriteset_Battle() {
 	// Create background
-	if (!Game_Temp::battle_background.empty()) {
-		background.reset(new Background(Game_Temp::battle_background));
-	} else {
+	if (Game_Battle::GetTerrainId() > 0) {
 		// Background verifies that the Terrain ID is valid
 		background.reset(new Background(Game_Battle::GetTerrainId()));
+	} else {
+		background.reset(new Background(Game_Temp::battle_background));
```
fmatthew5876: Did you verify this still works for event commands which start a battle with a fixed battle background?

fmatthew5876:
We need to verify the first commit doesn't break battle background for:
* Random encounters with terrain set
* Random encounters on map with fix background
* Events which trigger battles with fixed background
* Whatever the hell the rpg maker editor does for battle test??

Other than that, LGTM :+1:

src/spriteset_battle.cpp
Ghabry: havn't verified anything yet as this is the not-fun part

Ghabry:
To sum up the expected behaviours:

Random encounters:
- [ ] Terrain Based (uses Terrain setting of the tile)
- [ ] Specify: Uses a fixed background
- [ ] Same as parent map

Event command:
- [ ] Defined Map setting
- [ ] Specific background
- [ ] Terrain

Ghabry:
BattleTest will be inconvenient to fix as the arguments come via command line as args 5, 6 and 7...

Arg 5: 0 Terrain, 1 Shallow, 2 Deep
Arg 6: 0 Normal, 1 Preempt, 2 Surprise, 3 Sourround, 4 Pincer
Arg 7: When Arg 5 = 0 this is the Terrain ID

Ghabry:
this is now ready for review

PR #1615, opened 2019-02-02
Shop + Other UI Fixes

### LINKED ISSUE #1611: Use half-width spaces in description window
Some scenes use a single-line to display item descriptions and maybe other top messages in battle. This line is usually using half width spaces (similar to the `\_` command) to display text.

In Player, some texts don't fit because it uses normal space width. Maybe the simplest solution is to replace spaces with `\_` when using this specific window type.

menu

Reported via https://community.easyrpg.org/t/monigote-fantasy-several-issues/638

### LINKED ISSUE #1614: Shop scene: missing face in message window
Screenshot comparison between Player and RPG_RT:

menu

Reported via https://community.easyrpg.org/t/monigote-fantasy-several-issues/638

fmatthew5876:
Fixes #1614 
Fixes #1611

### DIFF
--- a/src/window_shopparty.cpp
+++ b/src/window_shopparty.cpp
@@ -27,7 +27,8 @@
 Window_ShopParty::Window_ShopParty(int ix, int iy, int iwidth, int iheight) :
 	Window_Base(ix, iy, iwidth, iheight) {
 
-	SetContents(Bitmap::Create(width - 16, height - 16));
+	SetBorderX(4);
+	SetContents(Bitmap::Create(width - GetBorderX() * 2, height - 16));
 
 	cycle = 0;
 	item_id = 0;
--- a/src/window_shopparty.cpp
+++ b/src/window_shopparty.cpp
@@ -77,6 +77,14 @@ static int CmpEquip(const Game_Actor* actor, const RPG::Item* new_item) {
 	return 0;
 }
 
+static bool IsEquipment(const RPG::Item* item) {
+	return item->type == RPG::Item::Type_weapon
+		|| item->type == RPG::Item::Type_shield
+		|| item->type == RPG::Item::Type_helmet
+		|| item->type == RPG::Item::Type_armor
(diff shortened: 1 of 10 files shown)


src/window_shopparty.cpp
```diff
@@ -92,13 +100,18 @@ void Window_ShopParty::Refresh() {
 		if (phase == 3) {
 			phase = 1;
 		}
-		bool equippable = actor->IsEquippable(item_id);
-		BitmapRef bm = bitmaps[i][equippable ? phase : 1][equippable ? 1 : 0];
+		bool usable = actor->IsEquippable(item_id);
```
fmatthew5876: This `IsEquippable()` method is suspect.

src/window_shopparty.cpp
```diff
@@ -27,7 +27,8 @@
 Window_ShopParty::Window_ShopParty(int ix, int iy, int iwidth, int iheight) :
 	Window_Base(ix, iy, iwidth, iheight) {
 
-	SetContents(Bitmap::Create(width - 16, height - 16));
+	SetBorderX(4);
+	SetContents(Bitmap::Create(width - GetBorderX() * 2, height - 16));
```
fmatthew5876: Maybe we need a `SetContentsDefault()` method which does `SetContents(Bitmap::Create(width - border_x*2, height - border_y*2))`

fmatthew5876:
This is ready now. I've gone through each menu using the below system graphic. (Flip the lines 90 degrees to test vertical vs horizontal alignment)

Please while reviewing this check the alignment of the text in the all the menus.


system

fmatthew5876:
Rebased to master

Ghabry:
the way you implemented the half-spacing will also work later with freetype/harfbuzz. thx

Ghabry:
Based on the stacktrace I can also find "Crash when selling empty item" in the Android crash list.

But not that common, only 12 users affected in 60 days. 👍

PR #1620, opened 2019-02-04
Vehicle bgm fix

### LINKED ISSUE #1619: Vehicle BGM should continue after teleport with vehicle and play new map BGM on disembark
> This one has to do with music. Here’s an [example](https://github.com/EasyRPG/Player/files/2825890/Example2.zip): 
Once you get on a vehicle, the reference to the music that was playing is stored, and if you move to another map for whatever the reason while still on the vehicle (for example, going to the deck of your ship), the music changes to the new map’s specific BGM (while RPG Maker player keeps playing the “vehicle music” despite the map change). 
If you get off the vehicle in the new map, then EasyRPG will start playing the stored one (music #1), while RPG Maker changes from the vehicle music to the new map’s music. 
(Sorry if it sounds repetitive and redundant :stuck_out_tongue:).

[Source](https://community.easyrpg.org/t/monigote-fantasy-several-issues/638/5).

fmatthew5876:
Fixes #1619 

Note: I use the aboard flag here because that's what tells us the player is actually inside the vehicle. You could have an event that teleports you while boarding and then the bgm effect would not happen.

I tested this parallel event:
```
GetOnOffVehicle
Wait 0.0s (x3)
Teleport other map
```

The player teleports to the other map, the other map music plays, and then the player boards the ship on that map and the vehicle bgm never plays.

I'll fix that behavior in #1601

### DIFF
--- a/src/game_map.cpp
+++ b/src/game_map.cpp
@@ -371,7 +371,7 @@ void Game_Map::PlayBgm() {
 		current_index = GetMapIndex(Data::treemap.maps[current_index].parent_map);
 	}
 
-	if ((current_index > -1) && !Data::treemap.maps[current_index].music.name.empty()) {
+	if ((current_index > 0) && !Data::treemap.maps[current_index].music.name.empty()) {
 		if (Data::treemap.maps[current_index].music_type == 1) {
 			return;
 		}
--- a/src/game_map.cpp
+++ b/src/game_map.cpp
@@ -375,7 +375,12 @@ void Game_Map::PlayBgm() {
 		if (Data::treemap.maps[current_index].music_type == 1) {
 			return;
 		}
-		Game_System::BgmPlay(Data::treemap.maps[current_index].music);
+		auto& music = Data::treemap.maps[current_index].music;
+		if (!Main_Data::game_player->IsAboard()) {
+			Game_System::BgmPlay(music);
+		} else {
+			Main_Data::game_data.system.before_vehicle_music = music;
+		}
 	}
 }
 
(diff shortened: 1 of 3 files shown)


fmatthew5876:
Jenkins test this please

carstene1ns:
What about the second part of #1619? Seems memorized bgm should be cleared on map change.

fmatthew5876:
> What about the second part of #1619? Seems memorized bgm should be cleared on map change.

You're right, I overlooked that. We actually have `SaveSystem::before_vehicle_bgm` chunk designed exactly for this purpose, which Player doesn't use currently.

I tested these cases:

* Get on, Get off -> plays map music
* Change map music, Get on, Get off -> plays changed music
* Get on, change map, Get off -> plays other map music
* Change map music, Get on, change map (no change music), Get off -> plays other map music
* Get on, change music, get off -> plays map music
* Get on, change music, change map (no change music), get off, plays other map music

fmatthew5876:
And another one, similar to the escape/teleport/save edge case fixed previous.

fmatthew5876:
> 
> 
> Looks good to me but I don't fully understand the difference between the aboard flag and "InVehicle()".

The difference matters when boarding the ship. As soon as you start boarding, the player `vehicle` chunk is set to the ship or boat. At this time, the player is still visible and the music behavior is the same if you teleport. Only once the boarding actually finishes does the `aboard` flag get set and then the vehicle music behavior takes effect.

carstene1ns:
I guess the other occurrences of `InVehicle()` might not care about this corner case. Maybe can be refactored/simplified.

fmatthew5876:
> 
> 
> I guess the other occurrences of `InVehicle()` might not care about this corner case. Maybe can be refactored/simplified.

There are times when you only need to know if the vehicle is "in use" and other times when it is "boarded". After #1601 we should do a review of this and simplify the API to make this distinction obvious. I expect the majority of cases required boarded.

PR #1622, opened 2019-02-04
Fix emscripten & OSX build

Ghabry:
Have no idea if the Metal.framework dep breaks running on OSX < 10.11 but this is a SDL problem then and I have no way to test this. (ObjC does lazy binding... should work in theory)

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -159,7 +159,7 @@ void Player::Init(int argc, char *argv[]) {
 
 	// Create initial directory structure in our private area
 	// Retrieve save directory from persistent storage
-	EM_ASM(
+	EM_ASM(({
 
 		FS.mkdir("easyrpg");
 		FS.chdir("easyrpg");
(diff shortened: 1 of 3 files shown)


fdelapena:
These macOS versions are unsupported, I guess we may go with this until somebody complains :+1:.

Ghabry:
Jenkins: test this please
Works now, also needed some jenkins fixes.
I forced emscripten to asm.js by now as WASM is the default now.

src/player.cpp
```diff
@@ -159,7 +159,7 @@ void Player::Init(int argc, char *argv[]) {
 
 	// Create initial directory structure in our private area
 	// Retrieve save directory from persistent storage
-	EM_ASM(
+	EM_ASM(({
```
carstene1ns: The added braces (`(`) seem superfluous here, the curly braces should be enough.

Ghabry: Unfortunately they are needed here because the Javascript code contains a "," :/

>  The C preprocessor does not have an understanding of JavaScript tokens, and as a result, if the code block contains a comma character ,, it may be necessary to wrap the code block inside parentheses. For example, code EM_ASM(return [1,2,3].length); will not compile, but EM_ASM((return [1,2,3].length)); does.

Issue #1623, opened 2019-02-07
Timing Bug: Wait commands and event movement off by 1 frame


fmatthew5876:
Create a map event with trigger = parallel and `anim_type = continuous`

Set this code
```
WAIT
OpenSaveMenu
```

Chunks of the event with different WAIT's at the time you save

| WAIT | RPG_RT stop count | Player stop count | RPG_RT anim_count | Player anim_count | RPG_RT anim frame | Player anim frame |
| -- | -- | -- | -- | -- | -- | -- |
| | 1 | 1 | 1 | 1 | 1 | 1 |
| 0.0s | 2 | 1 | 2 | 1 | 1 | 1 |
| 0.0s, 0.0s | 3 | 2 | 3 | 1 | 2 | 1 |
| 0.1s | 7 | 6 | 1 | 2 | 0 | 2 |
| 1.0s | 61 | 60 | 1 | 3 | 0 | 3

So there is an off by 1 error somewhere as shown in `stop_count` and `anim_count`. Either we're really waiting -1 frames or the order of updates with when the wait starts/ends is wrong somehow.

The `anim_count` results were taken with #1601 as master has these completely wrong to begin with.

fmatthew5876:
This bug looks like it only happens when the first thing you do in the event code is wait.

For example, the below code works correctly
```
OpenSaveMenu
WAIT
OpenSaveMenu
```

fmatthew5876:
For further investigation, I made a test with the code

```
Show Picture
Loop: 
  Wait
```

| Event Type | Result |
| -- | -- |
| Parallel Map Event | Picture appears before transition in |
| Autostart Map Event | Picture appears after transition in |
| Common parallel | Picture appears before transition in |
| Autostart common | Picture appears after transition in |

Now I change the code to this:

```
OpenSaveMenu
```


| Event Type | Player stop count | Event stop count |
| -- | -- | -- |
| Parallel Map Event | 0 | 1 |
| Autostart Map Event | 1 | 1 |
| Common parallel | 0 | 0 |
| Autostart common | 0 | 1 |

fmatthew5876:
Multiple events:

Here is a more revealing test:

1. Create 2 events with code

```
OpenSaveMenu
Loop:
  wait
```

| EV1 Trigger | EV2 Trigger | EV1 stop count | EV2 stop count | EV1 anim count | EV2 anim count |
| -- | -- | -- | -- | -- | -- |
[...]

Issue #1595, opened 2018-12-30
Grimps: Squeaking of the Dead: Bombs don't explode


Ghabry:
#### Name of the game:

Grimps: Squeaking of the Dead: Bombs don't explode

https://rpgmaker.net/games/10802/

#### Describe the issue in detail and how to reproduce it:

Load map ID 4 and push the bomb at the bottom right upwards. The bomb will move but not explode.

Seems to be some timing problem with move routes resulting in a wrong screen coordinate.
The check on Page 3 of Event 9 (the bomb) fails because Var 24 contains the wrong value.

Ghabry:
tested against 1628 with 28 commits: not fixed (Yet)

fmatthew5876:
Is it a touch or collision event? I haven't attempted collision yet

Ghabry:
it starts with a hero touch which works, the problem is as stated "Seems to be some timing problem with move routes resulting in a wrong screen coordinate."

fmatthew5876:
This is fixed by #1628

Issue #1247, opened 2017-08-13
COLORS: Lost Memories: loop when displaying 2nd page in custom textbox


carstene1ns:
#### Name of the game:
COLORS: Lost Memories ([download](https://rpgmaker.net/games/9973/))

#### Attach files (as a .zip archive or link them)
Savegame: [Save01.zip](https://github.com/EasyRPG/Player/files/1220620/Save01.zip)

#### Describe the issue in detail and how to reproduce it:
Go right, read the sign (or go right, up the tower, collect the red stone and go back). In both cases a Picture containing a message is shown, press the Action key to advance to second page, then get stuck.
There is running an endless loop, you can hear the message box sound repeating.

![screenshot_0](https://user-images.githubusercontent.com/4691314/29251900-f6483ca4-805d-11e7-8bec-cee2f76e8241.png) → ![screenshot_1](https://user-images.githubusercontent.com/4691314/29251899-f6458202-805d-11e7-9bb3-06db4e1c6efe.png)

This is made of the sign event with 2 pages (39) and 2 common events (8 and 9). So could be a timing problem. Common Event 8 waits for key input and sets some Variables and Switches, Common Event 9 could be ignored, as it only displays and hides the indicator arrow. The Map Event 39 is triggered on Action button, enables the Common Event 8 and on page 2 is set to Autorun on Switch (set by Common Event 8).

Ghabry:
The code after the Key Input "Check, ENTER" in CE 8 is strange because it waits for ENTER but afterwards the code checks if enter was pressed (thats the only possible case) and afterwards if enter was not pressed (impossible)

Ghabry:
Probably the relevant event is Auto-CE 40 which is activated through Switch 20 by Auto-Event 39 Page 2.

CE 40 is responsible for disabling autorun event 39P2 after ENTER is pressed. So my guess is that the common event never gets a chance to run due to a timing problem and E39P2 endlessly repeats instead (as indicated by the endless SE playback)

Ghabry:
The RPG_RT behaviour makes no sense (o'rly). Theres no obvious reason why one Autorun CE can run when an Event CE finishes.
Observed a 2nd behaviour: When the 1st Event CE doesn't yield the interpreter a 2nd Auto-Event CE starts (otherwise not >.<)

Ghabry:
According to my tests the behaviour is (all (common) events are autostart). Adding Parallel processes is left as an exercise ;):
```
E1: Var X+1
[...]

PR #1628, opened 2019-02-09
Refactor Event Update Ordering

fmatthew5876:
The PR refactors a number of interpreter and event related items

* Adds support for `triggered_by_decision_key` chunk
* Adds support for `pause` chunk
* Adds support for `waiting_execution` chunk
* Adds support for `processed` chunk
* Fixes the execution order of event update routines to match RPG_RT.
* Correctly handles "PreUpdate" which is the partial update routine done at startup which only updates events but not player or vehicles. This also happens when you teleport or use escape/teleport from the menu. *I haven't addressed the teleport PreUpdate() cases yet and may not do so in this PR*.

Fix: #1623 
Fix: #1247
Fix: #1595 
Fix: #663
Fix: #1664

Breaks: #937

### DIFF
--- a/src/game_commonevent.cpp
+++ b/src/game_commonevent.cpp
@@ -38,19 +38,14 @@ void Game_CommonEvent::SetSaveData(const RPG::SaveEventData& data) {
 
 void Game_CommonEvent::Refresh() {
 	if (GetTrigger() == RPG::EventPage::Trigger_parallel) {
-		if (GetSwitchFlag() ? Game_Switches.Get(GetSwitchId()) : true) {
-			if (!interpreter) {
-				interpreter.reset(new Game_Interpreter_Map());
-			}
-			parallel_running = true;
-		} else {
-			parallel_running = false;
+		if (!interpreter) {
+			interpreter.reset(new Game_Interpreter_Map());
 		}
 	}
 }
 
-void Game_CommonEvent::UpdateParallel() {
-	if (interpreter && parallel_running) {
+void Game_CommonEvent::Update() {
+	if (interpreter && IsWaitingBackgroundExecution()) {
 		if (!interpreter->IsRunning()) {
 			interpreter->Setup(this, 0);
 		}
@@ -97,6 +92,13 @@ RPG::SaveEventData Game_CommonEvent::GetSaveData() {
 bool Game_CommonEvent::IsWaitingForegroundExecution() const {
 	auto* ce = ReaderUtil::GetElement(Data::commonevents, common_event_id);
 	return ce->trigger == RPG::EventPage::Trigger_auto_start &&
+		(!ce->switch_flag || Game_Switches.Get(ce->switch_id))
+		&& !ce->event_commands.empty();
+}
+
+bool Game_CommonEvent::IsWaitingBackgroundExecution() const {
+	auto* ce = ReaderUtil::GetElement(Data::commonevents, common_event_id);
+	return ce->trigger == RPG::EventPage::Trigger_parallel &&
 		(!ce->switch_flag || Game_Switches.Get(ce->switch_id))
 		&& !ce->event_commands.empty();
 }
--- a/src/game_commonevent.cpp
+++ b/src/game_commonevent.cpp
@@ -49,22 +49,6 @@ void Game_CommonEvent::Refresh() {
 	}
 }
 
-void Game_CommonEvent::Update() {
-	if (GetTrigger() != RPG::EventPage::Trigger_auto_start)
-		return;
-
-	for (int i = 0; i < 500; ++i) {
-		if (GetSwitchFlag() ? Game_Switches.Get(GetSwitchId()) : true) {
-			if (!Game_Map::GetInterpreter().IsRunning()) {
-				Game_Map::GetInterpreter().Setup(this, 0);
-				Game_Map::GetInterpreter().Update();
-				continue;
-			}
-		}
-		return;
-	}
-}
-
 void Game_CommonEvent::UpdateParallel() {
 	if (interpreter && parallel_running) {
 		if (!interpreter->IsRunning()) {
@@ -108,4 +92,11 @@ RPG::SaveEventData Game_CommonEvent::GetSaveData() {
 	}
 
 	return event_data;
+}
+
+bool Game_CommonEvent::IsWaitingForegroundExecution() const {
+	auto* ce = ReaderUtil::GetElement(Data::commonevents, common_event_id);
+	return ce->trigger == RPG::EventPage::Trigger_auto_start &&
+		(!ce->switch_flag || Game_Switches.Get(ce->switch_id))
+		&& !ce->event_commands.empty();
 }
--- a/src/game_event.cpp
+++ b/src/game_event.cpp
@@ -79,8 +79,8 @@ void Game_Event::SetThrough(bool through) {
 	data()->through = through;
 }
 
-void Game_Event::ClearStarting() {
-	starting = false;
+void Game_Event::ClearWaitingForegroundExecution() {
+	data()->waiting_execution = false;
 }
 
 void Game_Event::Setup(const RPG::EventPage* new_page) {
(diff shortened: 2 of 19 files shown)


Ghabry:
Jenkins: Test this please

fmatthew5876:
This PR is now becoming very experimental as I refactor how and when the interpreter starts. Please don't recommend people to run it in continuous builds until we have a good suite of passing test cases.

fmatthew5876:
This PR is currently being redone. The problem was I had many commits attempting to to figure out `trigger=touch` and `trigger=collision` logic. Trying one thing, undoing it for another, etc.. Many useless commits.

This is reorganizing the other stuff (pause, processed, set waiting flag and run interpreter later, make way, etc..) first. And then I will redo the triggering logic last.

Ghabry:
I would prefer to have everything starting from "Refactor MakeWay" moved to another PR because one extremely complex interpreter commit ("Refactor how foreground interpreter is triggered") per PR is more than I feel comfortable to handle.

fmatthew5876:
Sure, but in that incomplete state I'm not sure if we cause more regressions.

I'll reorganize the PRs soon.

fmatthew5876:
This is ready for review.

All test cases pass in #1664 except:
- [ ] 1.3 - The pause and waiting flags disapear on the 4th save instead of the 3rd. This is likely a timing issue for when exactly we change event pages. If I get to this soon maybe I'll put it in this PR, or save it for later.

fmatthew5876:
Now that we have #1689 I've reintroduced the "Remove IsBlockedByMoveRoute()" commit from #1601 as that's really not needed anymore.

Ghabry:
This PR breaks faster decrement of the Wait counter when an interpreter is preempted when walking against events. I also checked against #1687 with same result.

Have two parallel events on same-as-hero-layer:

``
@> Wait: 5.0 seconds
@> Text: A

@> Wait: 5.0 seconds
@> Text: B
``

Walk against one of the events the whole time, this event will show the message after 2.5 seconds.
(This can be fixed in a later PR, as this is a quite obscure "feature")

Ghabry:
I can confirm that this really breaks Pom Gets Wifi, so the Pom-Bug is actually two problems in one: Movability while a move route is active + this wait bug.

So to track this will simply reopen #937 when this gets merged.

Ghabry:
"Check for map refresh before exiting **intpreter** Update()"

src/game_player.cpp
```diff
@@ -202,6 +200,16 @@ void Game_Player::UpdateSelfMovement() {
 		if (IsMovable()) {
 			const auto old_x = GetX();
 			const auto old_y = GetY();
+			const bool force_through = (Player::debug_flag
+					&& Input::IsPressed(Input::DEBUG_THROUGH)
+					&& !GetThrough());
+			if (force_through) {
+				SetThrough(true);
+			}
+			auto sg = makeScopeGuard([&](){
```
Ghabry: is the usage of a scope guard really needed here?
There is only one exit path and we don't throw/catch exceptions.
Though what actually bugs me is that this comes from liblcf. We already have other borderline-stuff in liblcf (INI) we call from Player but imo here the header should be copied into Player and made a private/not-installed header in liblcf.

fmatthew5876: I always get paranoid whenever there is a thing like this where you temporarily set some state, call a function, and then restore it. This gets optimized away and we ever change the code later to throw it's still safe.

Overly paranoid for sure but that's usually how I do these things as a habit. Can take it away if its really objectionable.

src/game_event.cpp
```diff
@@ -553,58 +549,24 @@ void Game_Event::Update() {
...
 		if (!interpreter->IsRunning()) {
 			interpreter->Setup(this);
 		}
 		interpreter->Update();
-	}
 
-	// Placed after the interpreter update because multiple updates per frame are allowed.
-	// This results in waits to finish quicker when an event collides with this event and
-	// emulates a RPG Maker bug)
-	// Only update the event once per frame
```
Ghabry: just marking this here because this comment must be readded when this is fixed again

```
// Placed after the interpreter update because multiple updates per frame are allowed.
// This results in waits to finish quicker when an event collides with this event and 
// emulates a RPG Maker bug) 
```

src/player.cpp
```diff
@@ -306,7 +306,11 @@ void Player::Update(bool update_scene) {
 	int speed_modifier = GetSpeedModifier();
 
 	for (int i = 0; i < speed_modifier; ++i) {
+		bool was_transition_pending = Graphics::IsTransitionPending();
 		Graphics::Update();
+		if (was_transition_pending && !Graphics::IsTransitionPending() && Scene::instance->IsInitialized()) {
```
Ghabry: I don't find this intuitive. Could you add a comment to explain that this is used by Scene Map to handle scene calling before the first update?

src/game_player.cpp
Ghabry: I understand the intention but I want to discourage that liblcf becomes a general collection of convenience headers that are used by Player because we don't use C++17.
So I would be already happy if you put a copy of scopeguard into the Player src folder, too.

Ghabry:
Well this PR was much easier to review than the STop Count one but MakeWay will be hell again

src/player.cpp
fmatthew5876: I don't find it intuitive at all. In fact I hate it, but I don't know how else to solve this.

I will add the comment.

fmatthew5876:
I've added #1724 to discuss `ScopeGuard` and other utilities. We can follow up with another PR whatever we decide there.

For the wait bug, we can reopen #937 and put details in there until its fixed again after #1687.

I've addressed the other comments.

src/game_commonevent.cpp
```diff
@@ -109,3 +88,17 @@ RPG::SaveEventData Game_CommonEvent::GetSaveData() {
...
+		(!ce->switch_flag || Game_Switches.Get(ce->switch_id))
+		&& !ce->event_commands.empty();
+}
+
+bool Game_CommonEvent::IsWaitingBackgroundExecution() const {
+	auto* ce = ReaderUtil::GetElement(Data::commonevents, common_event_id);
+	return ce->trigger == RPG::EventPage::Trigger_parallel &&
+		(!ce->switch_flag || Game_Switches.Get(ce->switch_id))
+		&& !ce->event_commands.empty();
+}
```
carstene1ns: Maybe these could be combined?

src/game_interpreter.cpp
```diff
@@ -142,11 +142,18 @@ void Game_Interpreter::SetContinuation(Game_Interpreter::ContinuationFunction fu
 	continuation = func;
 }
 
+bool Game_Interpreter::ReachedLoopLimit() const {
+	return loop_count >= 10000;
```
carstene1ns: This should be a constant here already. I have seen this in another commit.

src/game_event.cpp
```diff
@@ -315,45 +315,41 @@ bool Game_Event::GetActive() const {
...
-			return;
-		}
-		Start();
+		SetAsWaitingForegroundExecution(false, false);
+		return;
 	} else if (trigger == RPG::EventPage::Trigger_collision) {
 		CheckEventTriggerTouch(GetX(),GetY());
+		return;
 	}
 }
```
carstene1ns: These returns seem unnecessary now.

src/game_commonevent.cpp
fmatthew5876: Sure, I'll drop a commit for that into the next one #1687

src/game_interpreter.cpp
fmatthew5876: Thats fixed in #1713

src/game_event.cpp
fmatthew5876: Will fix in #1687 along with the previous comment

fmatthew5876: This call gets blown away in #1697 so I'm leaving it alone for now.

PR #1630, opened 2019-02-10
Make RTP warnings more noticeable

### LINKED ISSUE #1274: Verify RTP paths
RTP paths are currently not verified, only existence is tested.
This means when a path not containing an RTP is added,
no warning is shown about the missing RTP on game startup
leading to false positive bug reports (like #1264).

Relevant code:
https://github.com/EasyRPG/Player/blob/923f7f9da1eda092769929925172e790320024fc/src/filefinder.cpp#L533

Ghabry:
By telling the user everytime a RTP asset is missing. Sound and Music is also a visual warning, because we got already many reports that our Player fails to play sound...
Additionally when the RTP is not installed and the game never uses the RTP (but has no fullpackageflag=1) the user won't receive this annoying warning on startup.

Is kinda-RFC, maybe you have a better idea.

Also pretty sure I fixed a bug because search_paths.clear(); was not called when the game loaded has FullPackageFlag=1

a-fs8

Fixes #1274.

### DIFF
--- a/resources/easyrpg-player.6.adoc
+++ b/resources/easyrpg-player.6.adoc
@@ -28,7 +28,8 @@ for some additional binary patches.
   Disable audio (in case you prefer your own music).
 
 *--disable-rtp*::
-  Disable support for the Runtime Package (RTP).
+  Disable support for the Runtime Package (RTP). Will lead to checkerboard
+  graphics and silent music/sound effects in games depending on the RTP.
 
 *--encoding* 'ENCODING'::
   Instead of auto detecting the encoding or using the one in RPG_RT.ini, the
--- a/src/filefinder.cpp
+++ b/src/filefinder.cpp
@@ -225,12 +225,8 @@ namespace {
 			std::string msg = "Cannot find: %s/%s. " + std::string(search_paths.empty() ?
 				"Install RTP %d to resolve this warning." : "RTP %d was probably not installed correctly.");
 
-			Output::Warning(msg.c_str(),
-				dir.c_str(), name.c_str(), Player::IsRPG2k() ? 2000 : 2003);
+			Output::Warning(msg.c_str(), dir.c_str(), rtp_name.c_str(), Player::EngineVersion());
 		}
-
-		Output::Debug("Cannot find: %s/%s (%s)", dir.c_str(), name.c_str(),
-						name == rtp_name ? "!" : rtp_name.c_str());
 
 		return std::string();
 	}
@@ -476,11 +472,7 @@ void FileFinder::InitRtpPaths(bool no_rtp) {
 		return;
 	}
 
-	std::string const version_str =
-		Player::IsRPG2k() ? "2000" :
-		Player::IsRPG2k3() ? "2003" :
-		"";
-
+	std::string const version_str =	Player::GetEngineVersion();
 	assert(!version_str.empty());
 
 #ifdef GEKKO
--- a/src/filefinder.cpp
+++ b/src/filefinder.cpp
@@ -97,6 +97,8 @@ namespace {
 	search_path_list search_paths;
 	std::string fonts_path;
 	bool disable_rtp = true;
+	bool disable_rtp_warnings = false;
+	bool warning_broken_rtp_game_shown = false;
 
 	std::string FindFile(FileFinder::DirectoryTree const& tree,
 										  const std::string& dir,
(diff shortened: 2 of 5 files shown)


carstene1ns:
Btw. without #1274, this will not display anything when I have added an empty RTP dir. :grin:

Ghabry:
hmm, good point. I will issue a different warning when a RTP is found but incomplete.

src/filefinder.cpp
```diff
@@ -212,6 +220,14 @@ namespace {
...
 
+		if (!disable_rtp) {
+			std::string msg = "Cannot find: %s/%s. " + std::string(search_paths.empty() ?
+				"Install RTP %d to resolve this warning." : "RTP %d was probably not installed correctly.");
+
+			Output::Warning(msg.c_str(),
+				dir.c_str(), name.c_str(), Player::IsRPG2k() ? 2000 : 2003);
+		}
+
 		Output::Debug("Cannot find: %s/%s (%s)", dir.c_str(), name.c_str(),
```
carstene1ns: This one is redundant, as we have dedicated warning vs. debug messages for both now.

src/filefinder.cpp
```diff
@@ -212,6 +220,14 @@ namespace {
 			if (!ret_rtp.empty()) { return ret_rtp; }
 		}
 
+		if (!disable_rtp) {
+			std::string msg = "Cannot find: %s/%s. " + std::string(search_paths.empty() ?
+				"Install RTP %d to resolve this warning." : "RTP %d was probably not installed correctly.");
+
+			Output::Warning(msg.c_str(),
+				dir.c_str(), name.c_str(), Player::IsRPG2k() ? 2000 : 2003);
```
carstene1ns: Since I have seen this logic more than once by now, maybe we should add some `Player::EngineVer()` that returns either 2000 or 2003.

carstene1ns:
There is btw. still a bug when you use the gamebrowser and load a game with `FullPackageFlag=1` and then one without.

carstene1ns:
The mentioned bug has been fixed. Also added additional logic to make Player still use RTP like RPG_RT when `FullPackageFlag=1` has been set (as [mentioned by Cherry](https://github.com/EasyRPG/Player/issues/1274#issuecomment-430418125)). Please review.

resources/easyrpg-player.6.adoc
```diff
@@ -28,7 +28,8 @@ for some additional binary patches.
   Disable audio (in case you prefer your own music).
 
 *--disable-rtp*::
-  Disable support for the Runtime Package (RTP).
+  Disable support for the Runtime Package (RTP). Will lead to checkerboard
+  graphics and silent music in games depending on the RTP.
```
fdelapena: A bit nitpicky, but also silent sound effects. LGTM as is, too,

src/filefinder.cpp
```diff
@@ -191,30 +195,49 @@ namespace {
...
 		if (!ret.empty()) { return ret; }
 
-		const std::string& rtp_name = translate_rtp(dir, name);
+		// Try RTP if enabled and available
+		if (!disable_rtp)
+			rtp_name = translate_rtp(dir, name);
+
+		if (!rtp_name.empty()) {
+			// RPG_RT will even load RTP files when the game disables it
+			if(disable_rtp_warnings && !warning_broken_rtp_game_shown) {
```
fdelapena: space after if

src/filefinder.cpp
```diff
@@ -191,30 +195,49 @@ namespace {
...
+				warning_broken_rtp_game_shown = true;
+				Output::Warning("This game claims it does not need the RTP, but actually uses files from it!");
+			}
 
-		for(search_path_list::const_iterator i = search_paths.begin(); i != search_paths.end(); ++i) {
-			if (! *i) { continue; }
+			for(search_path_list::const_iterator i = search_paths.begin(); i != search_paths.end(); ++i) {
```
fdelapena: nit: space after for (yes, it was in the original, but it's a good chance)

src/player.cpp
```diff
@@ -644,6 +645,10 @@ void Player::CreateGameObjects() {
 	title << GAME_TITLE;
 	DisplayUi->SetTitle(title.str());
 
+	if(no_rtp_warning_flag) {
```
fdelapena: space after if

src/filefinder.cpp
```diff
@@ -191,30 +195,49 @@ namespace {
...
+				std::string const ret_rtp = FindFile(*(*i), dir, rtp_name, exts);
+				if (!ret_rtp.empty()) { return ret_rtp; }
+			}
 		}
 
-		Output::Debug("Cannot find: %s/%s (%s)", dir.c_str(), name.c_str(),
-						name == rtp_name ? "!" : rtp_name.c_str());
+		if (!disable_rtp_warnings && !rtp_name.empty()) {
+			std::string msg = "Cannot find: %s/%s. " + std::string(search_paths.empty() ?
+				"Install RTP %d to resolve this warning." : "RTP %d was probably not installed correctly.");
```
fdelapena: Debatable: `RTP %d` or `%d RTP`?

src/filefinder.cpp
```diff
@@ -191,30 +195,49 @@ namespace {
...
-		const std::string& rtp_name = translate_rtp(dir, name);
+		// Try RTP if enabled and available
+		if (!disable_rtp)
+			rtp_name = translate_rtp(dir, name);
+
+		if (!rtp_name.empty()) {
+			// RPG_RT will even load RTP files when the game disables it
+			if(disable_rtp_warnings && !warning_broken_rtp_game_shown) {
+				warning_broken_rtp_game_shown = true;
+				Output::Warning("This game claims it does not need the RTP, but actually uses files from it!");
```
fdelapena: Though this is nice to be shown and helps a lot users, there are a significant amount of games with missing audio files and this could annoy users as we had before. I think this should be fine with non-audio related assets, but output them as debug for audio.

resources/easyrpg-player.6.adoc
carstene1ns: Sure, can add this.

src/filefinder.cpp
carstene1ns: Yeah, I was unsure and thought about making this `Debug` only as well.

src/filefinder.cpp
Ghabry: please don't fix whitespacing in existing code.
This makes rebasing even worse

Ghabry:
The check for Sound and Music was incorrect in the "rtp is disabled" path.

LGTM :+1:

PR #1639, opened 2019-02-16
Refactor move type random

### LINKED ISSUE #1638: MoveType random is not frame accurate
Some details in test case 10.3 in #1567 

move type random has some unique behaviors. Needs more investigation and fix to replicate RPG_RT

fmatthew5876:
Fix #1638 

~~This is derived from what I've seen empirically, but the random algo may not be right.~~

The MV Algo of [0:1] Random, [2:4] Forward, [5] Reset is definately wrong.

~~I know for a fact we never reset stop count, but we do sometimes wait an extra frame.~~ The MV algo moves forward too much. This can be observed simply by watching an event walk around.

### DIFF
--- a/src/game_character.h
+++ b/src/game_character.h
@@ -374,6 +374,24 @@ public:
 	int GetMaxStopCount() const;
 
 	/**
+	 * @return stepping max_stop_count for the given frequency
+	 * @param freq input movement frequency
+	 */
+	static constexpr int GetMaxStopCountForStep(int freq);
+
+	/**
+	 * @return turning max_stop_count for the given frequency
+	 * @param freq input movement frequency
+	 */
+	constexpr int GetMaxStopCountForTurn(int freq);
+
+	/**
+	 * @return waiting max_stop_count for the given frequency
+	 * @param freq input movement frequency
+	 */
+	constexpr int GetMaxStopCountForWait(int freq);
+
+	/**
 	 * Sets the max_stop_count
 	 *
 	 * @param sc the new max stop count
@@ -584,6 +602,26 @@ public:
 	/** @return the direction we would need to face away from hero. */
 	int GetDirectionAwayHero();
 
+	/**
+	 * @param dir input direction
+	 *
+	 * @return the direction 90 degrees to the left of dir
+	 */
+	static constexpr int GetDirection90DegreeLeft(int dir);
+
+	/**
+	 * @param dir input direction
+	 *
+	 * @return the direction 90 degrees to the right of dir
+	 */
+	static constexpr int GetDirection90DegreeRight(int dir);
+
+	/**
+	 * @param dir input direction
+	 *
+	 * @return the direction 180 degrees to the of dir
+	 */
+	static constexpr int GetDirection180Degree(int dir);
 
 	/**
 	 * Character looks in a random direction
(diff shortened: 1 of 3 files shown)


fmatthew5876:
@CherryDT

Can you help with this one? Random stuff is very hard to reverse engineer and I'm sure my algo is not correct.

CherryDT:
https://github.com/EasyRPG/Player/issues/1638#issuecomment-466666082

fmatthew5876:
I've updated this with Cherry's comments and my empirical observations.

I'd like to know more about what happens in the "wait" case for lower move frequencies, but my save game method doesn't work there as it can be 100's of frames in between anything happening.

Ghabry:
Jenkins: Test this please

Ghabry:
Jenkins: Test this please

src/game_character.h
```diff
@@ -581,6 +581,27 @@ class Game_Character {
...
+	 * @return the direction 90 degrees to the right of dir
+	 *
+	 * @param dir input direction
+	 * */
+	static int GetDirection90DegreeRight(int dir);
+
+	/**
+	 * @return the direction 180 degrees to the of dir
+	 *
+	 * @param dir input direction
```
Ghabry: no idea if doxygen will mess this up because the normal order is ``@param @return``

Also ``* */`` instead of ``*/``

src/game_event.cpp
```diff
@@ -412,26 +412,30 @@ void Game_Event::UpdateSelfMovement() {
...
-	}
-	if (move_failed && !starting) {
-		SetDirection(last_direction);
-		if (!(IsDirectionFixed() || IsFacingLocked()))
-			SetSpriteDirection(last_direction);
 	} else {
-		SetMaxStopCount(GetMaxStopCount() / 5 * Utils::GetRandomNumber(3, 6));
+		// As per Cherry, RPG_RT supposedly waits for rand(0.. max_stop_count) inclusive in this case.
+		// This has not been observed empirically, as save data shows it only waiting 1 frame.
+		return;
```
Ghabry: guess this is a good use case for dynrpg, will see if I can get this somehow dumped.

CherryDT: *I believe* that `max_stop_count` is the incorrectly named `RPG::Character::movementFramesLeft` and `stop_count` is (also not quite correctly named) `RPG::Character::stepFrameCounter`.

fmatthew5876:
rebased and addressed comments. This still needs more research and verification.

src/game_character.h
fmatthew5876: Doxygen doesn't care about the order of tags. I fixed the `* */`

src/game_event.cpp
fmatthew5876: Looking at dumps I believe `RPG::Character::movementFramesLeft` is our `remaining_step`

fmatthew5876:
I setup a dump with dynrpg and started investigating this. I made a map with 8 events, each has frequency 1 to 8 and walks randomly.

At first I kept seeing many cases where `stop_count > max_stop_count`. But then I realized the cause.. The events were attempting to walk into one another and collision checks were causing them to wait extra frames.

After setting them all to `Through On`, now the only case where I see `stop_count >max_stop_count` is when `frequency == 8`, and every time it's `stop_count == 1` and `max_stop_count == 0`.

This can be explained very simply. When the wait case occurs, even though it calculates a `max_stop_count` of 0, there is still no movement until the next frame.

Cherry's algorithm is correct. I verified the values I get for `max_stop_count` match. I also saw the code for it in the dissassembly.

fmatthew5876:
This is finally done. I verified Cherry's algo with csv data in pandas and looked at the assembly in x64dbg. I also did some tests.

This is also a low risk change. I say we put it in 6.2!

PR #1646, opened 2019-02-24
Misc (cleanup, build system)

carstene1ns:


configure.ac
```diff
@@ -61,21 +63,12 @@ PKG_CHECK_MODULES([ZLIB],[zlib])
...
-		AC_CHECK_LIB([WildMidi],[WildMidi_Init],[AC_DEFINE(HAVE_WILDMIDI,[1],[Enable Midi support provided by libwildmidi]) WILDMIDI_LIBS=[-lWildMidi]],[auto_wildmidi=0])
-	],[
-		AC_MSG_CHECKING([for WILDMIDI])
-		AC_MSG_RESULT([assuming yes])
-		AC_DEFINE(HAVE_WILDMIDI,[1],[Enable Midi support provided by libwildmidi])
-	])
+AC_ARG_WITH([libwildmidi],[AS_HELP_STRING([--without-libwildmidi],
+	[Disable Midi support provided by libwildmidi. @<:@default=auto@:>@])])
+AS_IF([test "x$with_libwildmidi" != "xno"],[
+	PKG_CHECK_MODULES([WILDMIDI],[wildmidi],[AC_DEFINE(HAVE_WILDMIDI,[1],[Enable Midi support provided by libwildmidi])],[auto_wildmidi=0])
```
fdelapena: Setting minimum version might be useful for completeness (and maybe for other modules, if apply):

```suggestion
	PKG_CHECK_MODULES([WILDMIDI],[wildmidi >= 0.4.1],[AC_DEFINE(HAVE_WILDMIDI,[1],[Enable Midi support provided by libwildmidi])],[auto_wildmidi=0])
```

For testing this, it should generate a proper `Requires:` in the .pc file.

carstene1ns: I have added pkg-support in WildMidi for 0.4.1, so this check is useless.

carstene1ns:
(rebased, added exe_reader.cpp to cmakelists)

CMakeLists.txt
```diff
@@ -711,11 +693,11 @@ if(${PLAYER_AUDIO_BACKEND} MATCHES "^SDL2.*$")
 		else()
 			message(STATUS "MIDI playback: WildMidi")
 		endif()
-	elseif(${PLAYER_ENABLE_FMMIDI} MATCHES "ON")
+	elseif(PLAYER_ENABLE_FMMIDI)
```
Ghabry: this will not work because the next check is for "Fallback" and "Fallback" is also true.
Maybe swap the elseifs

CMakeLists.txt
```diff
@@ -711,11 +693,11 @@ if(${PLAYER_AUDIO_BACKEND} MATCHES "^SDL2.*$")
...
 		message(STATUS "MIDI playback: built-in (FmMidi)")
 	elseif(${PLAYER_ENABLE_FMMIDI} MATCHES "Fallback")
 		message(STATUS "MIDI playback: SDL2_mixer. Fallback to built-in (FmMidi)")
-	elseif(${PLAYER_ENABLE_FMMIDI} MATCHES "OFF")
+	elseif(NOT PLAYER_ENABLE_FMMIDI)
```
Ghabry: because that's the last elseif this could be changed to else(), no idea why I wrote it that way.

src/system.h
```diff
@@ -38,85 +38,73 @@
  */
 #include "options.h"
 
-/*
- * Smart pointer header.
- */
+// Smart pointer header.
```
Ghabry: thx for then cleanup here 👍

PR #1647, opened 2019-02-26
Battle 2k3: Bugfixes

### LINKED ISSUE #1490: Abnormal state

---
#### Player platform:Android
Windows, Linux, Android, ...

www.mediafire.com/file/k0969lkmqqax6hu/

Life recovery and mana recovery turns to decrease instead

Ghabry:
~~Don't merge yet, the heal/inflict logic still has some bug. I know where it is but touching battle at this time of the day is a bad idea~~ Decided to not fix this for skills, battlealgortihm needs a refactor first

and in #821 it ticks one more option:
- Items (Other): The 2nd option (heal instead of inflict) for states

This isn't really a battle thing, this just means that the item causes a permanent state that can't be removed except by unequipping.

Also improved the equip sanity check, handles now more corner cases.

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -1359,12 +1359,23 @@ void Game_Actor::RemoveInvalidData() {
 		RPG::Item::Type_accessory
 	};
 
-	for (int i = 1; i <= 5; ++i) {
-		const RPG::Item* item = GetEquipment(i);
-		if (item && item->type != eq_types[i - 1]) {
-			Output::Debug("Actor %d: Removing invalid item %d (of type %d) from equipment slot %d (needs type %d)",
-			GetId(), item->ID, item->type, i, eq_types[i - 1]);
-			SetEquipment(i, 0);
+	auto& equipment = GetWholeEquipment();
+	for (size_t i = 0; i < equipment.size(); ++i) {
+		int eq_id = equipment[i];
+		RPG::Item* item = ReaderUtil::GetElement(Data::items, eq_id);
+
+		if (!item && eq_id != 0) {
+			Output::Debug("Actor %d: Removing invalid item %d from equipment slot %d",
+			GetId(), eq_id, eq_types[i]);
+			SetEquipment(i + 1, 0);
+		} else if (item && item->type != eq_types[i]) {
+			Output::Debug("Actor %d: Removing item %d (of type %d) from equipment slot %d (needs type %d)",
+			GetId(), item->ID, item->type, i + 1, eq_types[i]);
+			SetEquipment(i + 1, 0);
+		} else if (item && !IsItemUsable(item->ID)) {
+			Output::Debug("Actor %d: Removing item %d from equipment slot %d (Not equippable by this actor)",
+			GetId(), item->ID, i + 1);
+			SetEquipment(i + 1, 0);
 		}
 	}
 
--- a/src/game_battle.cpp
+++ b/src/game_battle.cpp
@@ -247,8 +247,8 @@ void Game_Battle::NextTurn(Game_Battler* battler) {
 
 void Game_Battle::UpdateGauges() {
 	std::vector<Game_Battler*> battlers;
-	Main_Data::game_enemyparty->GetBattlers(battlers);
-	Main_Data::game_party->GetBattlers(battlers);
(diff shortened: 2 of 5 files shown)


src/game_actor.cpp
```diff
@@ -325,16 +334,55 @@ void Game_Actor::AddState(int state_id) {
 
 void Game_Actor::RemoveState(int state_id) {
 	Game_Battler::RemoveState(state_id);
+	AddEquipmentStates();
 }
 
 void Game_Actor::RemoveBattleStates() {
 	Game_Battler::RemoveBattleStates();
+	AddEquipmentStates();
```
Ghabry: Todo: AddEquipmentStates only needed in RemoveState, all code paths go through it (except RemoveAllStates, refactor?)

src/game_battlealgorithm.cpp
```diff
@@ -1215,7 +1216,16 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 		skill.scope == RPG::Skill::Scope_party ||
 		skill.scope == RPG::Skill::Scope_self;
 
-	this->revived = this->healing
+	bool heals_states = this->healing;
```
fmatthew5876: At some point I'd like to get rid of the `healing` and `positive` concepts here. Battle actions should be more general and just have effects (+-stats, inflict/heal states, etc..)

fmatthew5876:
Just a general comment.

It sounds like this could be exploited. Say you had an armor that inflicted poison. It sounds like whenever someone gets poisoned in battle, you could just equip and then unequip that armor to remove the poison.

src/game_battlealgorithm.cpp
```diff
@@ -1190,7 +1190,8 @@ bool Game_BattleAlgorithm::Skill::IsTargetValid() const {
 		skill.scope == RPG::Skill::Scope_party) {
 		if (GetTarget()->IsDead()) {
 			// Cures death
-			return !skill.state_effects.empty() && skill.state_effects[0];
+			return !skill.state_effects.empty() && skill.state_effects[0] &&
+				!(Player::IsRPG2k3() && skill.state_effect);
 		}
```
fmatthew5876: I'd actually like to see this feature as an EasyRPG extension to 2k battle system. But that can be added later.

src/game_actor.cpp
fmatthew5876: Instead of removing and adding the states back, would it make more sense to just filter in the remove? Like `if (!isEquipState(state)) Game_Battler::RemoveState(state);`

Ghabry:
> It sounds like this could be exploited

Yepp that's exactly how my test cases looked. Another 2k3 feature with bad execution.
When you have a "dead bringing shield" you can even revive actors by equipping and unequipping.
Imo this is only usable in combination with curse.

src/game_battlealgorithm.cpp
Ghabry: Healing and positive really don't work. Won't touch this here, too easy to break, but this needs a refactor

Ghabry:
The "Equipment that gives states" had many interesting corner cases but I think I got all.
E.g. a state with "persists after battle" set to false is not inflicted on the map but on battle start.

I also fixed a 2k3 bug: The Equipment scene ignores half/double stats when showing the preview.

fmatthew5876:
> I also fixed a 2k3 bug: The Equipment scene ignores half/double stats when showing the preview.

I guess this one is sort of debatable.

If I equip some armor with double stats, it looks very good but now I can't be buffed with double stat skills as they don't stack. Similar for half stat effects.

I'm not sure whats the best option here. To display or not.

Ghabry:
Well the equipment scene in RPG 2k3 will update as soon as you equip it. So imo this is a visual glitch because the buff is shown, just not in the preview :/

fmatthew5876:
I see what you mean, in that case I agree with you.

fmatthew5876:
Do you feel confident about no regressions? Since its a new feature and not a bug fix, maybe its safer to postpone to after the release?

carstene1ns:
I agree with @fmatthew5876 here, this looks like it could cause problems.
Maybe keep c73f9f67b6c9 until after release and only apply the rest?

Ghabry:
I feel condident with this one but I agree with your observation that this commit is far too complex for one day before release.

PR #1649, opened 2019-02-28
Fix #1631

### LINKED ISSUE #1631: Panorama pictures with color issues
Not tested the real reason, but it could be related with transparency issues or applying SDL color key when it shouldn't or even some pixman related glitch (like #357):

scarr57

The white color (255,255,255) used to fill the text uses the latest palette index (255), however for some reason it shows this value black:

sin titulo

Ghabry:
diff not readable:

I just removed the else-block and moved the transparent handling into the loop:

```
uint8_t alpha = (idx == 0 && transparent) ? 0 : 255;
```


carstene1ns:
https://github.com/EasyRPG/Player/pull/1649/files?w=1 ;)

PR #1655, opened 2019-03-02
Battle 2k3: Skills + Equip State infliction

### LINKED ISSUE #1490: Abnormal state

Ghabry:
the two WIP commits.

Not mergable yet.

Fix #1490

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -23,6 +23,7 @@
 #include "game_battle.h"
 #include "game_message.h"
 #include "game_party.h"
+#include "game_temp.h"
 #include "main_data.h"
 #include "output.h"
 #include "player.h"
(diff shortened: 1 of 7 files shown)


fmatthew5876:
We should verify what values are in the state chunks when this feature is active. Do they get locked at 1? Some really high number? 0 but still treated as active anyway?

Otherwise it could break save games.

Ghabry:
I already tested the +10 stuff, works like all other AddState cases

fmatthew5876:
If #1726 is adopted as is, this would change. The new paradigm is to call `AddState()` and `RemoveState()` and check the return value if it was successful. In your case it would check for equipment states and return false.

Ghabry:
see matthews battle prs which include this

PR #1656, opened 2019-03-02
Support EasyRPG Savegame version chunk

Ghabry:
Filled with 600 to reflect 0.6.0 savegame type.

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -812,10 +812,12 @@ static void OnMapSaveFileReady(FileRequestResult*) {
 	// Compatibility hack for pre 0.6 EasyRPG Player saves.
 	// Old savegames accidentally wrote animation_type as
 	// continuous for all events.
-	Main_Data::game_player->SetAnimationType(Game_Character::AnimType::AnimType_non_continuous);
-	Game_Map::GetVehicle(Game_Vehicle::Boat)->SetAnimationType(Game_Character::AnimType::AnimType_non_continuous);
-	Game_Map::GetVehicle(Game_Vehicle::Ship)->SetAnimationType(Game_Character::AnimType::AnimType_non_continuous);
-	Game_Map::GetVehicle(Game_Vehicle::Airship)->SetAnimationType(Game_Character::AnimType::AnimType_non_continuous);
+	if (Main_Data::game_data.easyrpg_data.version == 0) {
+		Main_Data::game_player->SetAnimationType(Game_Character::AnimType::AnimType_non_continuous);
+		Game_Map::GetVehicle(Game_Vehicle::Boat)->SetAnimationType(Game_Character::AnimType::AnimType_non_continuous);
+		Game_Map::GetVehicle(Game_Vehicle::Ship)->SetAnimationType(Game_Character::AnimType::AnimType_non_continuous);
+		Game_Map::GetVehicle(Game_Vehicle::Airship)->SetAnimationType(Game_Character::AnimType::AnimType_non_continuous);
+	}
 
 	Main_Data::game_player->Refresh();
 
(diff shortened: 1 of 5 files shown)


src/version.h
```diff
@@ -32,4 +32,16 @@
...
+ * Version written to the easyrpg_data.version field in savegames.
+ * Update this number when savegame changes cause incompatibility with old
+ * savegames.
+ *
+ * 1st digit (if not 0): Major
+ * 2nd digit: Minor
+ * 3rd digit: Patch
+ * 4rd digit: Usually 0. Increment only when having a good reason.
+ */
+#define PLAYER_SAVEGAME_VERSION 600
```
fmatthew5876: This should just be something like `PLAYER_MAJOR * 1000 + PLAYER_MINOR * 100 + PLAYER_PATCH * 10`

Then we don't need to update it separately every time and your comments become unnecessary.

Ghabry: but we will not change the savegame format with every release.

fmatthew5876: We think we won't change it. That doesn't mean a save game bug can't show up in one particular release inadvertently. If we don mark every release, we wouldn't be able to tag this after the fact.

Ghabry: hm actually this could be also interesting for statistics, because then we can see in the logs which player version was used before for creating the save.
I will think about a solution.

fmatthew5876: Yeah, that would be a great boon to debugging save game bug reports.

Ghabry:
added the discussed change, improved debug logging and added a warning when a save is loaded that is newer than the used EasyRPG Player version. (up to 9.9.9.9 supported, future proof ;))

This will emit a few saves with "540" until the 0.6 release PR is merged but shouldn't matter because that's only in continious builds.

src/version.h
```diff
@@ -32,4 +32,10 @@
 #define PLAYER_ADDTL ""
 #define PLAYER_VERSION TO_STRING(PLAYER_MAJOR) "." TO_STRING(PLAYER_MINOR) "." TO_STRING(PLAYER_PATCH)
 
+/**
+ * Version written to the easyrpg_data.version field in savegames.
+ * Increment the last digit (0) only when having a good reason.
+ */
+#define PLAYER_SAVEGAME_VERSION PLAYER_MAJOR * 1000 + PLAYER_MINOR * 100 + PLAYER_PATCH * 10 + 0
```
fmatthew5876: Put `()` around the expression. Otherwise someone could get a surprise when doing something like `PLAYER_SAVEGAME_VERSION * 10`

src/player.cpp
```diff
@@ -825,12 +827,37 @@ static void OnMapSaveFileReady(FileRequestResult*) {
...
+	std::stringstream verstr;
+	int32_t ver = save->easyrpg_data.version;
+	if (ver == 0) {
+		verstr << "RPG_RT or EasyRPG Player Pre-0.6.0";
+	} else if (ver >= 10000) {
+		verstr << "Unsupported";
+	} else {
+		verstr << "EasyRPG Player ";
+		char verbuf[6];
+		sprintf(verbuf, "%d.%d.%d", ver / 1000 % 10, ver / 100 % 10, ver / 10 % 10);
```
fmatthew5876: Use `snprintf()` for extra safety

Ghabry: I wanted to but I somehow expect that at least one platform port has a broken stdlib, well I will change it an dwe will see what happens.

fmatthew5876: Also I would just make the buffer large like 64 bytes. What if we have `10.0.0` or something bigger one day? Are you going to remember to resize the buffer? I know I wouldn't.

fmatthew5876: I see, if you make it bigger (like 32 or more), `sprintf()` is probably fine.

Ghabry:
Jenkins: test this please

fmatthew5876:
looks good. Great feature! Thanks for getting this in before release.

PR #1658, opened 2019-03-02
Refactor scene calling

### LINKED ISSUE #1642: Calling multiple scenes on the same frame has wrong behavior
Create 2 parallel map events with this code:
```
OpenSaveMenu
EraseEvent
```

Start a new game on the map.

In RPG_RT, the save menu will be called only once ❗️  and then both events dissapear.
In Player, the save menu is called, then EV1 dissapears, save menu, EV2 dissapears.

This also happens if you teleport into the map.

fmatthew5876:
Attempt to fix #1642

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -57,7 +57,7 @@ namespace {
 	static Game_Interpreter* transition_owner = nullptr;
 }
 
-Game_Interpreter::EventCalling Game_Interpreter::event_calling = {};
+Scene::SceneType Game_Interpreter::scene_call = Scene::Null;
 
 Game_Interpreter::Game_Interpreter(int _depth, bool _main_flag) {
 	depth = _depth;
@@ -3074,16 +3074,12 @@ bool Game_Interpreter::DefaultContinuation(RPG::EventCommand const& /* com */) {
 	return true;
 }
 
-void Game_Interpreter::ResetEventCalling() {
-	event_calling = {};
+void Game_Interpreter::ResetSceneCalling() {
+	scene_call = Scene::Null;
 }
 
 bool Game_Interpreter::IsImmediateCall() {
-	return event_calling.load
-		|| event_calling.save
-		|| event_calling.name
-		|| event_calling.shop
-		|| event_calling.menu
+	return scene_call != Scene::Null
 		|| Game_Temp::battle_calling;
 };
 
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -198,15 +198,7 @@ void Game_Interpreter::Update() {
 			break;
 		}
 
-		if ((Game_Temp::battle_calling && !Game_Temp::battle_running) ||
-			IsShopCalling() ||
-			IsNameCalling() ||
-			IsMenuCalling() ||
-			IsSaveCalling() ||
-			IsLoadCalling() ||
-			Game_Temp::to_title ||
-			Game_Temp::gameover) {
-
+		if (Game_Temp::to_title || Game_Temp::gameover) {
 			break;
 		}
 
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -3079,8 +3079,7 @@ void Game_Interpreter::ResetSceneCalling() {
 }
 
 bool Game_Interpreter::IsImmediateCall() {
-	return scene_call != Scene::Null
-		|| Game_Temp::battle_calling;
+	return scene_call != Scene::Null;
 };
 
 // Dummy Continuations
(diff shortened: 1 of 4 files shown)


fmatthew5876:
So this fixes #1642 with a caveat..

In RPG_RT, the save menu appears once, and then both events dissapear.
In this PR, both events dissapear and then the save menu appears once.

This is because of #1623 causing parallel events to run twice in the first frame of Player.

fmatthew5876:
@CherryDT

If you're around. I've noticed some very strange behavior with event code delaying by frames in ways that don't make sense..

When an event calls a menu scene, does this have any affect on how events execute that frame? Or does it just let them all execute like normal and then calls the menu at the end?

CherryDT:
> When an event calls a menu scene, does this have any affect on how events execute that frame? Or does it just let them all execute like normal and then calls the menu at the end?

It does have an effect. A quite complex effect actually, I believe.

It sets the scene value to menu immediately. This then has a couple of effects:

* The drawing of the scene is skipped (I'm not sure if the drawing of the map scene has any side-effects)*
* The map scene's update method will do a fade-out at the end and set a flag to remember to reinitialize the next time it is called
* The scripting engine will immediately exit the main loop [described here](https://github.com/EasyRPG/Player/issues/1247#issuecomment-322495397), the same way as if 10000 iterations had passed or there was no more code/events to process (**this is probably the behavior you see**)

Now to that last point: It's not as simple as just stopping all event execution. Remember that in general, there are two kinds of workers, the "parallel process workers" (where one exists for each such event) and the "global autostart event worker" (where only one exists) which handles autostart events and events otherwise triggered (e.g. by action button) and has more complexity in its main loop than the parallel process workers.

Also remember that the overall order is this (from my other post):

> - Reset the "already acted" flag for all events on the map (which is used to prevent events to trigger twice on the same frame in edge cases)
> - Ask all common events' parallel process workers to run
> - Ask all map events' parallel process workers to run
> - Ask all map events (including vehicles and party) to act (unless its "already acted" flag is set): Acting is taking an action like moving, but it will also check if the event is autostart and the start conditions are met, in this case its "waiting" flag is set. The "already acted" flag is also set. (If an event is triggered in another way, for example by action button, it would also get its "waiting" flag set.)
> - Ask the global autostart event worker to run. The global autostart event worker will do a loop of up to 10000 times trying the following: [...]

Now, both kinds of workers will check _at the end of their loop_ whether the scene is no longer the map and would exit the loop in that case. However, since this happens _after_ command execution, that means that even if the scene is changed, each worker may execute _one command_ before reaching that check and exiting.

Therefore, if a parallel process changes the scene, you will see other events still executing one single command - those events who are parallel process events coming after the scene-changing one (and all map events come after all common events) and the first autostart/otherwise-triggered event, if any. (You should not see anything weird if an autostart/triggered event calls the menu, since the global autostart worker is the last one to be run.)

Note: I did not empirically test this, this is just the result of logical deduction based on what the code in the binary appears to be doing.

-----

_*: To help understanding what this means: The main loop normally does things in this order: Increase frame counter -> Update key input -> Update current scene -> Draw current scene -> Wait for next frame. The drawing can be skipped in case the updating used too much CPU time recently, or when the scene was changed during the updating of the old scene. (And there are occasions, like transitions and some things in battle, where a "mini-version" of this loop is run inside of the scene updating itself, without updating the frame counter!)_

CherryDT:
Note that this also means that **you need to be very careful with your save-file-based testing method** because opening the save screen has the same effect!

fmatthew5876:
>  each worker may **execute one command** before reaching that check and exiting.

Thanks for confirming this Cherry, this matches exactly what I've seen empirically. And checking at the end of the loop makes more sense as to why this would be the case.

fmatthew5876:
Given the amount of testing I've done against this + Cherry's confirmations, this is ready for review and merge.

There is one more small piece of this where we don't load new events in the interpreter after the 1 command executes, but that's implemented in #1628. It can't be done here without that PR's refactors first.

Even without that, this is an improvement and good to merge.

fmatthew5876:
Note: This doesn't treat the special game over case mentioned in #1642.

I'll address that later as I don't want to cause a game over regression when this is more for fixing the interpreter.

fmatthew5876:
Rebased to master

Ghabry:
commit msg typo: "Fixes bug there " -> "Fixes bug where "

fmatthew5876:
Addressed review comments

carstene1ns:
Commit message of c463dad is wrong now :see_no_evil:
>* Add eBattle enum values

fmatthew5876:
> Commit message of [c463dad](https://github.com/EasyRPG/Player/commit/c463dad7ef74397c13b4c5d5d0f3cfb9bfd09c80) is wrong now see_no_evil
> 
> > * Add eBattle enum values

Fixed, thanks!

Issue #1659, opened 2019-03-02
Horizontal / Vertical movement with "Collision Trigger" not right


fmatthew5876:
Create an event with `trigger=collision`, no event event code, and `move_type=horizontal`.

Move the player in the way of the event.

In RPG_RT, the event will stay there next to the player until the player moves. In Player, the event will turn around.

This behavior only happens if `trigger=collision`, otherwise the event will always turn around.

Finally, the stop count stuff near https://github.com/EasyRPG/Player/blob/master/src/game_event.cpp#L460 looks really suspect.

I've opened this issue to get some frame dumps of this movement type to fix it.

fmatthew5876:
Test cases
----------

1. Create EV01 with `layer=same`, `move_type=horizontal`, `trigger=action`, `move_frequency=7`, `move_speed=6`, and no code

Create EV02 parallel with this code
```
Loop:
  OpenSaveMenu
```

Setup the map like this horizontally

| Tile | Event |
| -- | -- |
| Grass | Player |
| Grass | EV01 |
| Grass | |
| Water | |

Now start a new game

| Save | EV01 stop count | EV01 max stop count | EV01 remaining step | EV01 direction |
| -- | -- | -- | -- |
| 1 | 1 | 4 | 0 | 2 |
| 4 | 4 | 4 | 0 | 2 |
| 5 | 0 | 4 | 128 | 1 |
| 6 | 0 | 4 | 0 | 1 |
| 30 | 24 | 4 | 0 | 1 |
| 31 | 0 | 4 | 128 | 3 |
| 32 | 0 | 4 | 0 | 3 |
| 56 | 24 | 4 | 0 | 3 |
| 57 | 0 | 4 | 128 | 1 |

Result: When event hits a barrier, waits until `stop_count == 24` and then turns around.

2. Change EV01 `trigger=collision`


| Save | EV01 stop count | EV01 max stop count | EV01 remaining step | EV01 direction |
| -- | -- | -- | -- |
| 1 | 1 | 4 | 0 | 2 |
| 4 | 4 | 4 | 0 | 2 |
| 5 | 0 | 4 | 128 | 1 |
| 6 | 0 | 4 | 0 | 1 |
| 30 | 24 | 4 | 0 | 1 |
| 31 | 0 | 4 | 128 | 3 |
| 32 | 0 | 4 | 0 | 3 |
| 36 | 4 | 4 | 0 | 3 |
| 37 | 1 | 4 | 0 | 3 |
[...]

PR #1661, opened 2019-03-02
Fixes for horizontal and vertical movement.

fmatthew5876:
Fixes behavior for #1659, but not frame perfect as there are interpreter timing issues to be worked out.

This however is correct for horizontal / vertical move type

### DIFF
--- a/src/game_event.cpp
+++ b/src/game_event.cpp
@@ -449,20 +449,32 @@ void Game_Event::MoveTypeCycle(int default_dir) {
 	SetMaxStopCount((GetMoveFrequency() > 7) ? 0 : (1 << (9 - GetMoveFrequency())));
 	if (GetStopCount() < GetMaxStopCount()) return;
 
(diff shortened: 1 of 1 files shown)


fmatthew5876:
This is also ready for review and merge. I have nothing more to add here.

Ghabry:
can you link me the issue where algorithm was discussed? (want to know where +20 and +60 come from)

fmatthew5876:
They weren't discussed, the +20 and +60 were there from before. However, my tests confirm they are correct.

https://github.com/EasyRPG/Player/issues/1659

Ghabry:
oh, you are right. I think I will really do a 1-day break.

carstene1ns:
https://github.com/EasyRPG/Player/issues/1115#issuecomment-468884117 ?

fmatthew5876:
This fixes whats described in that comment

PR #1670, opened 2019-03-05
Android: Set black background

Ghabry:
Now I just need a billing address to send Google the invoice


carstene1ns:
Both issues fixed. There are still slight glitches when sliding the controls or menu bar in, but much better than before.

PR #1673, opened 2019-03-06
Fix: Search Windows Registry for installed RTP

carstene1ns:
Broken here: #1646


Ghabry:
yeah the first serious regression

PR #1682, opened 2019-03-10
MorphOS and AmigaOS4 build fix

BSzili:
I adjusted the Makefile for the latest OS4 compiler that needs some extra flags for the C++11 threads. I had to add a hack to implement to_string, I'm not sure if system.h is the right place for it, but it's used in quite a few places.

### DIFF
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -32,13 +32,7 @@
 #include <cassert>
 
 Game_Character::Game_Character(RPG::SaveMapEventBase* d) :
-	last_pattern(0),
-	original_move_frequency(-1),
-	move_type(RPG::EventPage::MoveType_stationary),
 	move_failed(false),
-	last_move_failed(false),
-	move_count(0),
-	wait_count(0),
 	jump_plus_x(0),
 	jump_plus_y(0),
 	visible(true),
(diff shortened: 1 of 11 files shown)


Ghabry:
So there are still platforms that dont implement to_string, how unfortunate.
Not tested but another way which allows keeping the definition out of system.h is:

1. Create a string.h in the amiga folder
2. Add the amiga folder as an include dir
3. Put the stuff you added to system.h into string.h:

```cpp
INCLUDE_GUARD

#include_next <string>

#include <sstream> 
namespace std { 
	template <typename T> string to_string(const T &n) { 
		ostringstream strm; 
		strm << n; 
		return strm.str(); 
	} 
}

END INCLUDE_GUARD
```

Include_next is a GNU extension that includes the next header with this name, which is the system header and all Player code will include this file first and get std::to_string this way.

fmatthew5876:
You can just call it `string` instead of `string.h`. Then a normal `#include <string>` in the code just works transparently. Once compilers upgrade, our fake `string` header can be removed and the code doesn't need to be changed,

I've done this `#include_next` trick before like that. Hacky but it works.

Ghabry:
yes that was a typo, I meant "string", not "string.h". thx

BSzili:
Mea culpa, I should have checked the C++ headers before copying the workaround I found on the forums. _GLIBCXX_USE_C99_STDIO and _GLIBCXX_USE_C99 did the trick.

Ghabry:
I approve but can't squash merge on mobile GitHub ^^'

carstene1ns:
Sorry, git took my master branch instead of yours when force-pushing.
(would have needed to add a different refspec)

Issue #1675, opened 2019-03-07
Make Way test cases


fmatthew5876:
Test Cases 1: Diagonal movement
-------------

1. Setup a Tile passable right and down only.

| Start from | Step | Result |
| -- | -- | --|
| On the tile | Upper left | Blocked |
| On the tile | Upper right | ok |
| On the tile | Lower left | ok |
| On the tile | Lower Right | ok |
| Left of the tile | Upper right | ok |
| Left of the tile | Lower right | ok |
| Above of the tile | Lower left | ok |
| Above of the tile | Lower right | ok |
| Up left from the tile | Lower right | Blocked |

Result: Both directions must be blocked to stop movement

2. Setup blocking tiles as described

| Right Of | Below Of | Step | Result |
| -- | -- | -- | -- |
| | | Lower Right | Ok |
| Blocked | | Lower Right | ok |
| | Blocked | Lower Right | ok |
| Blocked | Blocked | Lower Right | Blocked |

Result: At least 1 of the adjacent tiles must not be blocked to allow movement.

3. Setup 5 events in a star formation as shown below:

All of these events should be `layer=same`, `trigger=parallel`.

| Tile | Tile | Tile |
| -- | -- | -- |
| | EV05 | |
| EV04 | EV01 | EV02 |
| | EV03 | |

Set the code for EV02 to EV05 to Message "N" where N is the event number.

Now set the code of EV01 to a move route with the following movements:

| MoveRoute | Result |
| -- | -- |
| Lower Right | 3, 2, 3, 2, 3, 2, 3, 2, 3, 2,  |
| Lower Left | 3, 4, 3, 2, 3, 4, 3, 3, 4, 3, 2, 3, 4, 3 |
| Upper Right | 5, 2, 5, 2, 5, 2, 5, 2, 5, 2,  |
| Upper Left | 5, 4, 5, 2, 5, 4, 5, 3, 5, 4, 5, 2, 5, 4, 5, 4, 5, 2, 5, 4 |
| Right | 2, 2, 2, 2, 2, 2, 2 |
| Down | 3, 2, 3, 3, 2, 3, 3 |
| Left | 4, 2, 4, 3, 4, 2, 4, 4, 2, 4, 3, 4, 2, 4, 4, |
| Up | 5, 2, 5, 3, 5, 2, 5, 4, 5, 2, 5, 3, 5, 2, 5, 5 |

Result: Vertical direction is checked first!

4. Lets verify message timing didn't skew our results

[...]

Issue #1612, opened 2019-01-31
Event jump misbehaves with other events using the same priority


fdelapena:
Reported by BitBrosGames in a [forum post](https://community.easyrpg.org/t/monigote-fantasy-several-issues/638):

> In RPG Maker, three events can freely move over the same tile as long as their priorities are different. I mean, a character can walk over some event whose priority is “under the character”, and, let’s say, a bird on the “above character” layer can fly over said character. However, two characters can’t naturally occupy the same tile, in the same way as two “birds” couldn’t do so. This is implemented correctly in EasyRPG when it comes to walking, but it’s not when JUMPING.

[Example.zip](https://github.com/EasyRPG/Player/files/2819236/Example.zip)

fmatthew5876:
This is fixed in #1628

Issue #1606, opened 2019-01-27
Interpreter timing issues


fmatthew5876:
Here are a list of issues discovered that are related to the event interpreter timing. These have to do with different events running at the wrong time or incorrect frame delays in execution timing.

- [x] Parallel events triggering off of airship aboard during ascent are off by 1 frame

Create this parallel map event
```
If boarded airship:
  OpenSaveMenu
```

Start a new game, and make the player board the airship.

In the save game, `remaining_ascent` will be 248 in RPG_RT, but 240 in Player. That is because 1 frame has passed between when the player boarded (ascent started) and when the conditional triggers. Once the conditional branch is entered, ascending will complete in 32 frames in RPG_RT and 31 frames in Player. This can cause bugs in event code which wants to do things immediately after ascent finishes.

This also occurs if you wrap the above event in a `Loop:`

- [x] Movement is not processed on the first frame of a new game for Player and Vehicles

Many examples in #1567 show strange behavior on very the first frame of a new game. Move routes and GetOnOffVehicle on the first frame setup the action, but don't animate it for that frame like they would when executed any other time.

* Move Route: cases where `stop_count=65535`, `move_route_overwritten=1`, and `move_route_index=0` on first frame.
* GetOnOffVehicle: cases where `remaining_step=256` on first frame

Adding any kind of delay such as `Wait 0.0s` to the beginning of the event avoids these behaviors.

This is also the likely cause of #1588

- [x] Timing error with touch events and Vehicles

Setup these vertical tiles:

| Tile | Event |
| -- | -- |
| Grass | |
| Water | |
| Water | Event (see below) |
| Water | Player aboard ship |

Create this touch event

```
Move Ship, Freq=8, Up
Wait 0.0s
GetOnOffVehicle
```

In RPG_RT, the player will disembark into the top water tile and the ship will also be in the top water tile. In Player (with #1601), the player will disembark into the top water tile and the ship will remain on the event's water tile.

Removing the wait or waiting for more than 1 frame, the behaviors are the same.

fmatthew5876:
The third test case is fixed by #1628 and the first test case is fixed by #1687 

So after #1687 this issue is fixed

PR #1687, opened 2019-03-11
Refactor MakeWay

fmatthew5876:
~~Depends on #1628~~

This currently passes all test cases in #1675 ~~Except:~~

- [x] 1.2 - Collision parts pass, but the order of messages goes wrong later on. This is a deeper non-makeway related issue

Fix: #1675
Fix: #1612
Fix: #1606
Fixes a game hang reported in: #985

Retested: #884 
Retested: #1082


fmatthew5876:
This is now ready for review. 

I believe I have covered every possible collision case for player and vehicles here.

carstene1ns:
Better diff: https://github.com/fmatthew5876/Player/compare/event_collide...fmatthew5876:event_makeway

Ghabry:
Even the better diff is very hard to follow, too much code moved around everywhere.

fmatthew5876:
For this one I basically ripped out every collision related function in `game_map.cpp`.

Therefore for reviewing, I would suggest maybe just reading the final version of `game_map.cpp` and only using diffs for the other files touched.

fmatthew5876:
I added a commit that not only fixed #937, but also test case 1.2 from #1675 mentioned in the original post!

Ghabry:
When you ignore the diff and open old and new file instead and do full text search you notice that it's actually not that bad, mostly is really just code moved around (Sometimes with minor additions because of new behaviour matthew figured out), e.g.:

- Map::AirshipLandOk & Vehicle::CanLand & Map::IsLandable -> Map::CanLandAirship
- Vehicle::Update (partly moved to) -> Game_Vehicle::AnimateAscentDescent
- Map::IsPassableVehicle is now integrated in Map::IsPassableTile
- New func Map::IsPassableLowerTile uses code ripped out of Map::IsPassableTile

The Jump code was mostly rewritten, but this is a local thing so no big deal to review this.

So the ones with real changes are WouldCollide, MakeWayCollideEvent and MakeWay, looking deeper into them.

src/game_map.cpp
```diff
@@ -572,202 +572,226 @@ static CollisionResult TestCollisionDuringMove(
...
-
-	// A character can move to a position with an impassable tile by
-	// standing on top of an event below it. These flags track whether
-	// we stepped off an event and therefore don't need to check the
-	// passability of the tile layer below.
-	bool stepped_off_event = false;
-	bool stepped_onto_event = false;
-
-	for (Game_Event& other : GetEvents()) {
-		CollisionResult result = TestCollisionDuringMove(x, y, new_x, new_y, d, self, other);
```
Ghabry: Seems that TestCollisionDuringMove and the CollisionResult enum are now dead code and can be removed?

src/game_event.cpp
```diff
@@ -521,11 +521,9 @@ void Game_Event::Update() {
...
 	}
 
-	if (IsProcessed()) {
-		return;
-	}
-	SetProcessed(true);
-
+	// RPG_RT runs the parallel interpreter everytime Update is called.
+	// That means if the event updates multiple times due to makeway,
+	// the interpreter will run multiple times per frame.
```
Ghabry: > Run parallel events everytime update is called 

Could you also add a comment that this also emulates a bug because the Wait count also decrements

src/game_map.cpp
```diff
@@ -572,202 +572,226 @@ static CollisionResult TestCollisionDuringMove(
...
-		Output::Warning("IsPassableVehicle: Invalid terrain at (%d, %d)", x, y);
-	} else if (vehicle_type == Game_Vehicle::Boat) {
-		if (!terrain->boat_pass)
-			return false;
-	} else if (vehicle_type == Game_Vehicle::Ship) {
-		if (!terrain->ship_pass)
-			return false;
-	} else if (vehicle_type == Game_Vehicle::Airship) {
-		return terrain->airship_pass;
+		Output::Warning("MakeWay: Invalid terrain at (%d, %d)", x, y);
```
Ghabry: Warning msg: MakeWay -> CanLandAirship

src/game_map.cpp
```diff
@@ -794,6 +818,80 @@ bool Game_Map::IsPassableTile(int bit, int tile_index) {
...
+bool Game_Map::IsPassableTile(const Game_Character* self, int bit, int x, int y) {
+	if (!IsValid(x, y)) return false;
+
+	auto vehicle_type = (self != nullptr)
+		? self->GetVehicleType() : Game_Vehicle::None;
+
+	if (vehicle_type != Game_Vehicle::None) {
+		const auto* terrain = ReaderUtil::GetElement(Data::terrains, GetTerrainTag(x, y));
+		if (!terrain) {
+			Output::Warning("MakeWay: Invalid terrain at (%d, %d)", x, y);
```
Ghabry: Warning msg: MakeWay -> IsPassableTile

src/game_event.cpp
fmatthew5876: The wait count decrementing is a result of running the interpreter again. Do you still want a special comment for that?

Ghabry: because the wait is in frames and not in interpreter invocations I don't find this intuitive so yes, explicitly mentioning this is better.

Note, old comment was:
> This results in waits to finish quicker when an event collides with this event and
> emulates a RPG Maker bug)

src/game_map.cpp
fmatthew5876: Removed

src/game_map.cpp

src/game_map.cpp

src/game_event.cpp
fmatthew5876: Added

Ghabry:
Maybe three things that could be retested are:

- [ ] #1002
- [ ] #1079
- [ ] #1349

fmatthew5876:
I retested all those games. All passed.

PR #1689, opened 2019-03-12
Allow player input on same frame move route ends

fmatthew5876:
Create a parallel event: `SetMoveRoute: Player, Turn Left`

RPG_RT, you can still walk around. Before this fix the player
was stuck.

Fix: #579
Fix: #937


carstene1ns:
nice!

Ghabry:
I'm a bit surprised how this fixed Pom Gets Wifi as zegeri stated the problem is with Make Way and interpreter preemption but maybe is a combination of multiple issues: Confirm it works :)

Issue #1684, opened 2019-03-10
DBZ Legend of Z: Charge UP percentage skill cost


OmegaMagnusDA:
Name of the game: **Dragon Ball Z: Legend of Z RPG**

Player platform: **Android** and possibly others

-Download **DBZ LOZ (No music version)** for quickest download (third option) https://gamejolt.com/games/dragon-ball-z-legend-of-z-rpg/145494

-**Save(s)**: https://drive.google.com/open?id=1zb7NObjqhmKu_hjwl-TZ45SeExRvO2la (see skill in menu and/or roam around current map to test in battle)
https://drive.google.com/open?id=1nnymRrmyPsXMfICy5NARPjVZQsACRH-o (save with higher party levels)

-**easyrpg_log.txt log file**: https://drive.google.com/open?id=1O7Hr7P-puoby0-YWZQHLB1pKCR2AKyHJ

**Issue in detail** and how to reproduce it:
Issue from #1374 is half fixed. Charge UP now gives players the right amount of KI (MP), in recent update 0.6, but the skill cost is still stuck at 1 for all characters at all levels (percentage skill costs). 

![image](https://user-images.githubusercontent.com/48193819/54091221-bb3cd100-433a-11e9-8859-f9bb552f4694.png)

fmatthew5876:
This is caused by this bad line of code:
https://github.com/EasyRPG/Player/blob/master/src/game_battler.cpp#L424

It should be `Player::IsRPG2k3()` and not checking a specific version.

I have a different idea for a fix however..

carstene1ns:
Hmm, this has been in there since 2011, way before we had different engine modes (2k3e did not exist).

PR #1690, opened 2019-03-12
Allow percent_sp feature for all 2k3 versions

fmatthew5876:
* Fix for all versions of 2k3
* Allow in 2k (if the chunk is present) as an EasyRPG extension

Fix: #1684 

Vote: Do we feel unsafe enabling this chunk for 2k games in EasyRPG?

### DIFF
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -421,8 +421,7 @@ int Game_Battler::CalculateSkillCost(int skill_id) const {
 		return 0;
 	}
 
-	return (Player::engine == Player::EngineRpg2k3 &&
-			skill->sp_type == RPG::Skill::SpType_percent)
+	return (Player::IsRPG2k3() && skill->sp_type == RPG::Skill::SpType_percent)
 		? GetMaxSp() * skill->sp_percent / 100
 		: skill->sp_cost;
 }


carstene1ns:
I guess if there is a game converted to 2k3, but shipped with RPG_RT from 2k, this breaks, since is ignored by RPG_RT. We have a workaround that prevents such things elsewhere, but not applicable here.
Chances are too minimal to care... (and likely other things are broken in battles too in this case)

fmatthew5876:
In general, I'd like to be able to use these chunks in 2k games in the future.. But yes the situation you describe is potentially a break.

Still they'd have to convert the game, change sp cost to percentage style, and still keep the old RPG_RT to make this break.

Thats a pretty determined series of steps!

carstene1ns:
> I'd like to be able to use these chunks in 2k games in the future.

This means either using a different editor (e.g. setting easyrpg mode anyway) or hacking stuff in. So not really detectable.

I feel more safe with this guarded, but as we both said, chances are low.

fmatthew5876:
So I want these custom features for HH3, but I'm very far away from that. I've fixed this to use `Player::IsRPG2k3()` so we can just fix the bug for now.

Will revisit the customizations question at a later date with a more comprehensive plan. Maybe an ini setting for example?

Issue #1693, opened 2019-03-12
Move to variable location same map has wrong behavior


fmatthew5876:
Use a `MoveToVariableLocation` command to teleport somewhere on the same map.

In `RPG_RT` this is like the normal teleport command. It transitions out and in, resets player sprite etc.. In player it just immediately moves the player and doesn't do the teleport checks.

fmatthew5876:
Fixed in #1694

Issue #1624, opened 2019-02-08
Teleport Test Cases


fmatthew5876:
Parallel events are processed in the middle of teleporting

Setup this parallel common event

```
OpenSaveMenu
Teleport other map
OpenSaveMenu
Loop:
  Wait 0.0s
```

In RPG_RT, it will save, transition out, teleport, transition in, and save again

In Player, it will save, transition out, save :exclamation: , teleport, transition in

This bug occurs with parallel common events and parallel map events.

It does not occur with autostart common, autostart map, or triggered map events.

fmatthew5876:
Here is something even more strange

```
OpenSaveMenu
Teleport other map
OpenSaveMenu
OpenSaveMenu
```

In RPG_RT, it will save, transition in, teleport, transition out, save

In other words, despite 3 open save menu commands, the save menu is only summoned twice!

Hypothesis
----------

Perhaps the events are supposed to run in the middle of the teleport, and so the second OpenSaveMenu triggers, it sets a flag to summon the save menu, but since we're teleporting it doesn't come up. Then we finish the teleport, process events again, and the 3rd opensavemenu command triggers, setting the flag again, and then the save menu finally appears.

fmatthew5876:
Here are more test cases which expose information about how the interpreter works

```
Wait 0.0s
Teleport
OpenSaveMenu
CallMenu
```

The save menu is called

```
Wait 0.0s
Teleport
CallMenu
OpenSaveMenu
```

The menu is called

```
Wait 0.0s
Teleport
CallShop
ChangeName
```

Shop is called

```
Wait 0.0s
Teleport
ChangeName
CallShop
```

Name change menu is called

```
Wait 0.0s
Teleport
CallShop
Battle
```

Shop is called

```
Wait 0.0s
Teleport
CallShop
Battle
```

Shop is called

```
Wait 0.0s
Teleport
Battle
CallShop
```

Battle is started

Result
------

In all cases, the first action takes place and the second action is ignore.

Conclusion
------

RPG_RT probably uses one variable to store which menu/scene to summon. If its not set to `(None)` any subsequent request to call a menu or battle in the same frame is ignored.

fmatthew5876:
Here is something more telling

```
Wait 0.0s
Teleport
OpenSaveMenu
SetVariable
OpenSaveMenu
Loop:
  Wait 0.0s
```

In this case we call save menu twice!



Hypothesis
------

[...]

Issue #1316, opened 2017-12-07
Final Tear 3: Endless intro loop


Ghabry:
#### Name of the game: Final Tear 3
https://rpgmaker.net/games/3983/

Start a new game and Choose "Skip intro". This will still show the (far too long) credits which shows the authors. Simply fast forward through it until you reach a yellow epilepsy-like (at least at 10x :D) thunderstorm scene. This one will loop infinitely after the next name is displayed.

Quote from 20kdc in the forum

> Map 0719, “You will never disrupt < Intro pounder” is a night scene, just after the first bit of voice acting (which appears to be involuntary for the intro???). It’s driven by E2/Event talk, a parallel event.
> After it’s done, it goes to Map0954, “MAP0954”, driven by E1/EV0001 (also a parallel event), which shows the “STEVIE V.D LAAR / PRESENTS” scene.
Map 0954 then goes back to Map 0719 for a sec, shows map, and then event execution stops there, because Map 0719’s event starts up again, and on EasyRPG Player it interrupts map 0954’s event.
> I assume that on RPG_RT, it does not, and the map transition goes quickly enough that the intro lines don’t play, or alternatively, Map 0719 is the event that gets interrupted.

Ghabry:
FYI: The fix for log.[in] does not fix this.

fdelapena:
This game uses DynRPG patches and also uses some modified headers while trying to test with plain RPG_RT (it returns invalid signature), so these headers should be restored and play with unpatched RPG_RT.exe to make sure DynRPG patches are not the problem.

fmatthew5876:
This is fixed in #1694

Issue #1416, opened 2018-09-11
Log.[in]: Wrong event execution order after Erase Screen event


Ghabry:
Please fill in the following fields before submitting an issue:
#### Describe the issue in detail and how to reproduce it:
Start the game, go to the PC, leave the PC. Start the PC again, screen stays black.

Caused by Event 2 which does on Page 2 (Autorun) a Hide Screen, this results in a yield because we can't have inner loops. After the transition parallel events run and one enables Switch 5 again, which is wrong.

Possible solution: Prevent parallel event execution after returning from a hide/show screen effect, the main event continues without running anything else.

20kdc:
...It seems like only the interpreter-thread that's responsible for the transition should get priority, no matter which interpreter-thread that is... Maybe allow transitions to have a "responsible interpreter" attached?

Ghabry:
Valid point, just traced parallel events and they behave exactly the same, therefore your suggestion is better.

PR #1694, opened 2019-03-12
Refactor Teleport

fmatthew5876:
~~Depends on #1687~~

Fix: #1624
Fix: #1693 
Fix: #1316
Fix: #1567

### DIFF
--- a/src/teleport_target.h
+++ b/src/teleport_target.h
@@ -52,40 +52,40 @@ class TeleportTarget {
 		/** @return whether this is active */
 		bool IsActive() const;
 	private:
-		int _map_id = 0;
-		int _x = 0;
-		int _y = 0;
-		int16_t _d = -1;
-		bool _active = false;
+		int map_id = 0;
+		int x = 0;
+		int y = 0;
+		int16_t d = -1;
+		bool active = false;
 };
 
 
 inline TeleportTarget::TeleportTarget(int map_id, int x, int y, int d)
-	: _map_id(map_id)
-	  , _x(x)
-	  , _y(y)
-	  , _d(d)
-	  , _active(true)
+	: map_id(map_id)
+	  , x(x)
+	  , y(y)
+	  , d(d)
+	  , active(true)
 { }
 
 inline int TeleportTarget::GetMapId() const {
-	return _map_id;
+	return map_id;
 }
 
 inline int TeleportTarget::GetX() const {
-	return _x;
+	return x;
 }
 
 inline int TeleportTarget::GetY() const {
-	return _y;
+	return y;
 }
 
 inline int TeleportTarget::GetDirection() const {
-	return _d;
+	return d;
 }
 
 inline bool TeleportTarget::IsActive() const {
-	return _active;
+	return active;
 }
 
 #endif
--- a/src/teleport_target.h
+++ b/src/teleport_target.h
@@ -1,0 +1,91 @@
+/*
+ * This file is part of EasyRPG Player.
+ *
+ * EasyRPG Player is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * EasyRPG Player is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef EP_TELEPORT_TARGET_H
+#define EP_TELEPORT_TARGET_H
+
+#include <cstdint>
+
+/**
+ * Defines a target for teleportation
+ */
+class TeleportTarget {
+	public:
+		/** Construct an inactive target */
+		TeleportTarget() = default;
+
+		/** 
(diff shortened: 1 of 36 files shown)


fmatthew5876:
This passes all the test cases and is ready for review.

Now some trivia. If you `MoveToVariableLocation` to (-1,8) in RPG_RT on horizontal looping map, your position will look correct (as if you're at x=10) but the `position_x` chunk has the value of -1. :exclamation: Step left or right and x will fix itself to the correct looped value.

Not sure what kind of bugs that might cause with collisions in RPG_Rt.

This RPG_RT bug is fixed in Player.

carstene1ns:
Better diff: https://github.com/fmatthew5876/Player/compare/event_makeway...fmatthew5876:event_teleport

Ghabry:
This causes multiple issues:

https://easyrpg.org/play/pr1694/?game=yume_nikki
After the instructions the screen stays black.

https://easyrpg.org/play/pr1694/
Talk to the dude next to the bridge, the message box closes after the teleport.

fmatthew5876:
I fixed the test game issue, but the Yume Nikki thing is still outstanding.

It only happens on web, and not on my local build in linux. The first "Refactor Teleport" commit moved things around and touched some code with comments about emscripten, so this is probably a web only regression.

@Ghabry What exactly is the issue here with async loading and transitions that breaks emscripten?

fmatthew5876:
I've fixed the hang in Yume Niiki. But now there is another issue. 

The intro is supposed to fade in smoothly with "Working... Please Wait..."

What happens instead is a fade in screen transition, and then the black picture pops immediately after.

The intro has this autostart code:
```
Transfer Player Map3
ShowPicture 4 (all black)
ShowPicture 5 (Working... Please Wait...)
Show Screen
MovePicture4 Transparent 0% -> 100% 1.0s
MovePicture5 Transparent 0% -> 100% 1.0s Wait
```

In emscripten, the pictures don't finish async loading until after the transition for show screen starts. When the transition starts it calls `Transition::Init()`, and this is what snaps the transition end bitmap without the pictures since they haven't loaded yet.

I'm not sure this is actually a "bug" but maybe a side effect of calling the update routines and interpreters at the right time after teleport.

To fix this, we should somehow need to block on any graphical assets aysnc loading before starting a transition.

My log shows this:
```
Debug: PRE UPDATE pr1694:52:11
Debug: MAP UPD 3 1 pr1694:52:11
Debug: FORCE ERASE SKIP pr1694:52:11
Debug: MAP UPD 3 0 pr1694:52:11
Debug: COMMAND SHOW PIC pr1694:52:11
Debug: BIND FUNC OBJ pr1694:52:11
Debug: COMMAND SHOW PIC pr1694:52:11
Debug: BIND FUNC OBJ pr1694:52:11
Debug: SHOW SCREEN pr1694:52:11
Debug: CALL LISTENERS pr1694:52:11
Debug: USER TRANSITION START pr1694:52:11
Debug: MAP UPD 3 0 pr1694:52:11
Debug: CALL LISTENERS pr1694:52:11
Debug: PIC SPRITE READY pr1694:52:11
Debug: CALL LISTENERS pr1694:52:11
Debug: PIC SPRITE READY
```

Regardless, the fix for async was complex and nasty. I now need to retest all the teleport cases and pre-update scene calling again to make sure I didn't break it.

Ghabry:
Yeah you need one delay-frame, otherwise emscripten will load black because it has no time to do "blocking asset loading". THis was all discussed in the comments you deleted:

```
// Fade Out not handled here but in StartTeleportPlayer because otherwise 
// emscripten hangs before fading out when doing async loading... 
 
// Async loading note: 
// Fade In must be done before finish teleport, otherwise chipset is not 
// loaded and renders black while fading -> ugly 
 
// The delay is only needed for one frame to execute pending transitions, 
// the interpreters continue on the old map afterwards 
```

You can fake async loading on PC by changing to line in asynchandler.cpp:

Remove the comment after ``// remove comment for fake download testing`` and comment the line after ``// add comment for fake download testing``. Then you get a fake delay which simulates async on a bad connection

fmatthew5876:
> Yeah you need one delay-frame,

Only one delay frame is enough? What if the asset takes more than 1 frame time to download? Or will emscripten block calling the update loop when there are pending loads?

Ghabry:
That's what the "IsImportantFilePending" check does.
When any Request has the "SetImportant(true)" set the Updates won't be called.
(that's how Database and Map downloads work, they set this flag).

You could e.g. set it on any asset that is requested while the screen is faded out, then the fade-in won't happen until the Transitions are done.

fmatthew5876:
Async issues with Yume Nikki are now fixed.

I have to redo the teleport test cases now. But this can now be properly reviewed at the same time.

I think we have a path forward with transitions from this, which is great news. That'll be done proper in a future PR.

fmatthew5876:
I've redone all the test cases and fixed some issues.

This is ready now.

Ghabry:
woah the yume nikki fadein is for the first time looking correct. great.

src/teleport_target.h
```diff
@@ -0,0 +1,91 @@
...
+		int GetDirection() const;
+
+		/** @return whether this is active */
+		bool IsActive() const;
+	private:
+		int _map_id = 0;
+		int _x = 0;
+		int _y = 0;
+		int16_t _d = -1;
+		bool _active = false;
```
Ghabry: minor nitpick: We don't use vars prefixed with _

Ghabry:
Tested #1416: Works.

Log.[in] seems to be broken in the web player in general (in all branches), not tested there.

fmatthew5876:
I fixed the style comments

carstene1ns:
ok

PR #1695, opened 2019-03-13
Correctly render events at negative coordinates in non-looping maps

### LINKED ISSUE #1676: Events with negative coordinates are rendered
EasyRPG renders events with negative coordinates at map width + x and map height + y.

RPG_RT seems to only do this if the event's position is in the repeat direction of the map.
In those cases, the event is always passable and can't be interacted with in RPG_RT while it behaves normal as if it was positioned there in EasyRPG.

fmatthew5876:
Fix #1676


carstene1ns:
jenkins: test this please.

PR #1698, opened 2019-03-15
Refactor move towards / away from player

### LINKED ISSUE #1644: MoveType toward/away logic is not entirely accurate
RM logic for towards/away move types is:

* If event is not in sight, move randomly
* If event is in sight, move towards/away from hero with 80% chance, randomly with 10% chance, forwards without direction change with 10% chance.

Definition for "in sight": Screen coordinates (the same that are retrieved in the "Change Variable" event) of event are `x >= -32 && x <= 352 && y >= -32 && y <= 272`.

fmatthew5876:
Fix #1644

### DIFF
--- a/src/game_event.cpp
+++ b/src/game_event.cpp
@@ -470,66 +470,50 @@ void Game_Event::MoveTypeCycleUpDown() {
 	MoveTypeCycle(Down);
 }
 
-void Game_Event::MoveTypeTowardsPlayer() {
-	int sx = DistanceXfromPlayer();
-	int sy = DistanceYfromPlayer();
-	int last_direction = GetDirection();
+void Game_Event::MoveTypeTowardsOrAwayPlayer(bool towards) {
+	int sx = GetScreenX();
+	int sy = GetScreenY();
 
-	if ( std::abs(sx) + std::abs(sy) >= 20 ) {
-		MoveRandom();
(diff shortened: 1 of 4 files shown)


src/game_event.cpp
```diff
@@ -480,68 +480,52 @@ void Game_Event::MoveTypeCycleUpDown() {
...
+		SetStopCount(0);
 	}
 }
 
+void Game_Event::MoveTypeTowardsPlayer() {
+	MoveTypeRelatePlayer(true);
+}
+
+void Game_Event::MoveTypeAwayFromPlayer() {
+	MoveTypeRelatePlayer(false);
```
Ghabry: imo "Relate" sounds strange here because it is not a preposition.

src/game_event.h
```diff
@@ -178,37 +178,14 @@ class Game_Event : public Game_Character {
...
-	/**
-	 * Walks to the player.
-	 */
+	void MoveTypeRelatePlayer(bool towards);
 	void MoveTypeTowardsPlayer();
-
-	/**
-	 * Walks to the player.
-	 */
 	void MoveTypeAwayFromPlayer();
```
Ghabry: why where the comments deleted?

fmatthew5876:
Rebased and addressed the review comments.

Issue #1064, opened 2016-10-19
Celestial Silfade Story: Freeze after attack command in custom battle system


kleusbalut:
thank you for greatest app development.
i've got the bug. report them.
#### Name of the game:

Celestial Silfade Story ("シルフェイド幻想譚") by SmokingWOLF
#### Player platform:

Android
#### Attach files (as a .zip archive or link them)
- fleeze(no action) on after attack commands. (see attachments)

[CvGVGC9UIAETBDc.zip](https://github.com/EasyRPG/Player/files/537875/CvGVGC9UIAETBDc.zip)

Download: http://www.silversecond.net/contents/game/silfade_gensou/

#### Describe the issue in detail and how to reproduce it:

after character making, encount the enemy, input attack command , to freeze.

note: SmokingWOLF apps is very unique battle logic. not on RPG Maker default battle.

Ghabry:
The bug here is again a misalignment when scrolling the screen (this broke smash bros e.g.)

Relevant Code in Common Event 109:
```
@> Control Variables: [0441:中心座標X] = Player's X Coordinate
@> Control Variables: [0442:中心座標Y] = Player's Y Coordinate
@> Control Variables: [0443:アニメ最終表示X] = Player's Screen X
@> Loop
  @> Conditional Branch: Variable [0443:アニメ最終表示X] > 152
    @> Conditional Branch: Variable [0443:アニメ最終表示X] == 152
      @> Break Loop
      @>
     : Branch End
    @> Control Variables: [0441:中心座標X] -= 1 
    @> Control Variables: [0443:アニメ最終表示X] -= 16 
    @>
   : Else
    @> Conditional Branch: Variable [0443:アニメ最終表示X] == 152
      @> Break Loop
      @>
     : Branch End
    @> Control Variables: [0441:中心座標X] += 1 
    @> Control Variables: [0443:アニメ最終表示X] += 16 
    @>
   : Branch End
  @>
 : Repeat Above
```

The code basicly takes the X Screen position of the Player and adds/substracts 16 until the variable is 152 (which is the break condition).

For this code to terminate the X position must be on a 16-pixel grid (and 8 added because 152 has is not a dividing 16), as a formula the X position must be a solution of this equation.
```A * 16 + 8```

In Player the X position is 151, which is obviously wrong. :/

Ghabry:
But doing arbitrary tests on a TestGame map the result is actually 152. So this is probably a timing issue (battle event executes before the map finished scrolling).

PR #1700, opened 2019-03-16
Fix event screen x and y to match RPG_RT

### LINKED ISSUE #1699: Game_Character::GetScreenX() and Y are wrong on looping maps
On a looping map, the event commands that retrieve an events x and y screen coordinates into a variable return the wrong values in player.

Here are what they should be in the edge

| Local (screen) Tile position | Screen pos |
| -- | -- |
| x=0 | x=8 |
| x=-1 | x=-8 |
| x=-2 | x = (map_width - 1) * 16 + 8 |
| y = 0 | y = 16 |
| y = -1 | y = 0 |
| y = -2 | y = (map_height - 1) * 16 |

This also affects move towards / away from player in #1698

fmatthew5876:
Fix #1699

Retested #1676

### DIFF
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -99,7 +99,12 @@ int Game_Character::GetJumpHeight() const {
 }
 
 int Game_Character::GetScreenX(bool apply_shift) const {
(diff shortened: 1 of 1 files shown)


Ghabry:
Other screen test candidates:
#1064
#1169

PR #1166, opened 2017-05-16
Fix #1165

Ghabry:
See the commit message for an explanation.

PR #1701, opened 2019-03-16
Fix Screen Effects + Timers

### LINKED ISSUE #1662: もしもスマブラだったら: Actor (Picture) position and zoom incorrect
This is a new issue caused by cd4fff9 but based on tests the behaviour in this commit is correct.

The real problem must be somewhere else.
Simply start a battle, you will see the problem - the actors are completely off and not zoomed.

#### Name of the game: もしもスマブラだったら (https://rmarchiv.tk/games/1412)

Web player: https://easyrpg.org/play/master/?game=smashbros

### LINKED ISSUE #1702: Timers are not frame accurate
We don't run timer updates at the right time. As per Cherry:

```
Main update methods for common events run (they just check if they are parallel process and the condition is met and then invoke the parallel process worker for them)
Main update methods for map events run (they do a bunch of things, described in my previous post - but actually even more, because after the check for whether there is any active event page, they will run their parallel process worker and then check once more for active)
Message window updates
Timer updates
Screen effects update
Pictures update
Foreground worker runs
Panorama updates
```

Also I did some quick test and saw that our lsd chunks for timers are not frame accurate.

In HH3, I noticed some case where I would load an RPG_RT save game and the timer would appear with "00:01" when it shouldn't.

Leaving this issue as a placeholder to investigate timer behavior and setup test cases.

fmatthew5876:
~~Depends on #1628~~

Fix #1662
Fix #1702

### DIFF
--- a/src/game_event.cpp
+++ b/src/game_event.cpp
@@ -271,14 +271,14 @@ bool Game_Event::AreConditionsMet(const RPG::EventPage& page) {
 
 	// Timer
 	if (page.condition.flags.timer) {
-		int secs = Main_Data::game_party->GetTimer(Main_Data::game_party->Timer1);
+		int secs = Main_Data::game_party->GetTimerSeconds(Main_Data::game_party->Timer1);
 		if (secs > page.condition.timer_sec)
 			return false;
 	}
 
 	// Timer2
 	if (page.condition.flags.timer2) {
-		int secs = Main_Data::game_party->GetTimer(Main_Data::game_party->Timer2);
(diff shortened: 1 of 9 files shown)


Ghabry:
Adds two new commit "Fix game screen updating time " and "Refactor timers"

carstene1ns:
→ https://github.com/fmatthew5876/Player/compare/event_collide...event_screen

Ghabry:
another test case: Intro of "The Sacred Tears: TRUE", this was completely synchronized via timers.
The fix was in #1166

Going to review this tonight. thx.

Issue #1613, opened 2019-01-31
 Scene with automatic event skippable by keypress should also end after some seconds


fdelapena:
Reported by BitBrosGames in a [forum post](https://community.easyrpg.org/t/monigote-fantasy-several-issues/638):

> This is a bit tricky to explain, so maybe it’s better if you [open the project yourselves and take a look](https://github.com/EasyRPG/Player/files/2819236/Example.zip). It’s about the autorun event that runs at the beginning of the example. 
With RPG Maker you can skip the scene by pressing <kbd>Enter</kbd> or just wait 3 seconds for the picture to disappear, while with EasyRPG you always need to push <kbd>Enter</kbd> to carry on. I think it must be some kind of problem with the key input processing method. 
But this one is particularly weird, because if you simply add a message instruction into the event, it’ll stop working that way in RPG Maker. (??) 
(The message in the log is “`Event exceeded execution limit`”)

fmatthew5876:
Fixed by #1718

fmatthew5876:
This is actually fixed by #1713 !

It appears that the 2 events are so perfectly timed that the 2nd event changes the variable right after KeyInputProc but before the conditional branch.

So this is one example where `loop_count` is used precisely.

PR #1713, opened 2019-03-17
Branch loop limit workaround

### LINKED ISSUE #1707: Interpreter execution test cases
~~Events run for up to 10k iterations and then force exit. I found one test case where we don't match, so I'm writing test cases here.~~

General set of set cases for how the interpreter actually executes event code.

fmatthew5876:
Depends on #1628 

Fixes: #1613

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -57,6 +57,9 @@ namespace {
 	static Game_Interpreter* transition_owner = nullptr;
 }
 
+// 10000 based on: https://gist.github.com/4406621
+constexpr int loop_limit = 10000;
+
 Scene::SceneType Game_Interpreter::scene_call = Scene::Null;
 
 Game_Interpreter::Game_Interpreter(int _depth, bool _main_flag) {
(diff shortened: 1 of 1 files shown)


Ghabry:
You fixed loop limit stuff but you ignored the comment in game_interpreter mentioning this:
Guess this commit also means that the skipping is in general a problem and the real fix is skipping to the "End" command instead of the command after.
```
// FIXME?
// After calling SkipTo this index++ will skip execution of e.g. END.
// This causes a different timing because loop_count reaches 10000
// faster then Player does.
// No idea if any game depends on this special case.
```

fmatthew5876:
I missed that one. I just did quick tests of some commands with conditionals. Will take a deeper look at this. Maybe better to fix skip than add hacks like this.

Ghabry:
I can already tell you that I event traced this and it will indeed execute the "End" event.
Not sure about "Else" though, need to recheck this but wouldn't make sense when this executes.

fmatthew5876:
So we can test this SkipTo behavior with the way opensavemenu allows other events to run only a single command. Essentially it would let us single step through event code and see where the index goes each time.

I will setup tests in #1707

fmatthew5876:
So the real reason we aren't matching the loop limit here is because we don't correctly emulate branching commands and the `subcommand_path` chunks.

I will fix that in a later PR. For now I am leaving this one open as the current solution is a workaround which fixes #1613.

Issue #1691, opened 2019-03-12
Correctly handle menu_calling (ESC Menu) and encounter_calling flags (Random Encounter)


fmatthew5876:
We are not frame accurate with our handling of `menu_calling` and `encounter_calling`

`menu_calling`

- [x] Trigger ESC key to `menu_calling` at the right time within frame and under the right conditions
- [x] What gets blocked when `menu_calling` is active
- [x] What clears `menu_calling`
- [x] Frame accurate summoning of the menu

`encounter_calling`

- [x] Trigger `encounter_calling` on the right frame of movement with random encounters
- [x] Trigger `encounter_calling` at the right time within the frame and under the right conditions
- [x] What gets blocked when `encounter_calling` is active
- [x] What clears `encounter_calling`
- [x] Frame accurate summoning of the battle.

From Cherry:

Conditions that prevent party movement on map

    Airship is currently lifting off or landing
    Save.events.commands.length > 0 (Note that this condition will, if fulfilled, clear menu_calling before aborting)
    Message window is visible
    Party has move_route_overwrite
    Party has pause (irrelevant)
    encounter_calling flag was set and was processed now
    menu_calling flag was set and was processed now
    Any common or map event has waiting_execution (so you are right about that!)

Conditions that prevent decision key on map

    All of the above
    Party's remaining_step > 0 (checked after potentially initiating movement from directional keys)

Conditions that prevent escape key on map

(This happens after all of the above, without any of the above conditions though)

    Menu is not allowed (of course)
    Message window is visible
    Save.events.commands.length > 0

(Note that this won't immediately invoke the menu, but instead set the menu_calling flag.)

>    Why were these flags invented instead of just changing the scene directly like the event commands do?
>
> I guess in order to give those things a lower priority than others, especially event-controlled ones, and maybe to allow events to react on ESC press too, before the menu is actually called...? I'm not sure.
>
>    When does menu_calling get checked and the menu scene called? At the end of the same frame we detected ESC key?
>
>At the next frame, at the point where my post said "menu_calling flag was set and was processed now" above.
>
>    Are there any special cases with menu_calling outside of that? Anything which will later clear the flag? How does it interact if a scene call from an event happened on the same frame?
>
>Yes, only for random encounters. It is cleared when the player position changes by something other than movement (usually teleport), when a foreground event is active, when the encount_calling was set and is being processed.
>
[...]

Issue #1457, opened 2018-10-18
Teleports using input key as Cancel open the menu staying blocked


gadesx:
#### Player platform:
Windows 7 64bits

#### Attach files (as a .zip archive or link them)
[Sample_problem_menuprohibition.zip](https://github.com/EasyRPG/Player/files/2493396/Sample_problem_menuprohibition.zip)

#### Describe the issue in detail and how to reproduce it:
Trying my old project I've seen that the Go to memorized map create a situation that open the menu first, staying as blocked. There's must be something that explain that different behavior that usually seems not happen with 2k3 and break custom menus.

Just open the menu with any key, referenced by 6 internally.

fmatthew5876:
I did some testing of this in rm2k and this is a very subtle timing issue.

You can simplify the test case down to this code in a parallel event:

```
Key Input Processing [0001], Wait
Conditional Branch, Variable [0001] == 6
    Transfer Player Somewhere 
Branch End
```

In Player it will open the Menu first and then teleport you.

In rm2k3, it will just teleport you and never open the menu.
If you insert a Wait 0.0, rm2k3 will teleport and then open the menu immediately after.
If you insert a Wait 0.1, rm2k3 will open the menu and then teleport you.


In general depending on interpreter behavior like this in rm2k/3 games is bad practice. As you can see by making a minor change to the code completely changes the behavior in non-obvious ways and breaks your CMS. If you want to override the cancel key, then disable the menu before running your loop. 

Also doing too many things in Parallel events is asking for trouble. In general I always try to do the minimum required and get out of parallel context and into autostart context (via a switch + autostart event) ASAP. Autostart is synchronized with other foreground events and avoids subtle bugs. Parallel is really tricky and dangerous to play with.

That being said, we can't go and fix all the of the games so this still should be compatible in player.

CherryDT:
If I remembe correctly, it works along the lines of this:

each frame:
- event commands are executed until one requires waiting (such as the teleport)
- menu_calling is executed when it's set (but only if no blocking action is active)
- key trigger state is checked (setting menu_calling if necessary, but only if no blocking action is active)
- teleport is executed when it's pending

so, without wait, menu_calling is not set because the pending teleport is a blocking action (just like an open message box), then the teleport is executed
with wait 0.0, menu_calling is set, then next frame the teleport executes, and the following frame the menu is opened. (if there were an autostart event on the new map, the menu would not open.)
[...]

PR #1697, opened 2019-03-13
Refactor event triggering

fmatthew5876:
~~Depends on #1694~~

Finally we bring back the `touch`, `collision`, and `autostart` trigger fix changes.

Tests #1626 
- [x] 1-29 tested
- [ ] 2.20 - EV01 `stop_count` is 2 instead of 3. - Due to unsolved issue in #1624 Test 9.2

Fix: #886
Fix: #1626
Fix: #1696 
Fix: #1732

PR #1714, opened 2019-03-17
Refactor player update handling + scene calling

fmatthew5876:
~~Depends on #1697~~ 

Fix: #1691 
Fix: #1457 

Fixes the gameover part of #1642. The message vs gameover stuff is not fixed yet, but that looks like a message bug to be fixed in another PR.

I also haven't studied too deeply the immediate "to title" behavior.

2 more game temp variables killed!

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -60,8 +60,6 @@ namespace {
 // 10000 based on: https://gist.github.com/4406621
 constexpr int loop_limit = 10000;
 
-Scene::SceneType Game_Interpreter::scene_call = Scene::Null;
-
 Game_Interpreter::Game_Interpreter(int _depth, bool _main_flag) {
 	depth = _depth;
 	main_flag = _main_flag;
@@ -180,7 +178,7 @@ void Game_Interpreter::Update(bool reset_loop_count) {
 
 		// If something is calling a menu, we're allowed to execute only 1 command per interpreter. So we pass through if loop_count == 0, and stop at 1 or greater.
 		// RPG_RT compatible behavior.
-		if (loop_count > 0 && IsImmediateCall()) {
+		if (loop_count > 0 && Scene::instance->HasRequestedScene()) {
 			break;
 		}
 
@@ -3079,14 +3077,6 @@ bool Game_Interpreter::DefaultContinuation(RPG::EventCommand const& /* com */) {
 	index++;
 	return true;
 }
-
-void Game_Interpreter::ResetSceneCalling() {
-	scene_call = Scene::Null;
-}
-
-bool Game_Interpreter::IsImmediateCall() {
-	return scene_call != Scene::Null;
-};
 
 // Dummy Continuations
 
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -211,7 +211,7 @@ void Game_Interpreter::Update(bool reset_loop_count) {
 			break;
 		}
 
-		if (Game_Temp::to_title || Game_Temp::gameover) {
+		if (Game_Temp::to_title) {
 			break;
 		}
 
@@ -282,7 +282,9 @@ void Game_Interpreter::Setup(Game_CommonEvent* ev, int caller_id) {
 void Game_Interpreter::CheckGameOver() {
 	if (!Game_Temp::battle_running && !Main_Data::game_party->IsAnyActive()) {
 		// Empty party is allowed
-		Game_Temp::gameover = Main_Data::game_party->GetBattlerCount() > 0;
+		if (Main_Data::game_party->GetBattlerCount() > 0) {
+			Scene::instance->SetRequestedScene(Scene::Gameover);
+		}
 	}
 }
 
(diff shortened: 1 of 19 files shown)


src/game_interpreter.cpp
```diff
@@ -1540,8 +1542,9 @@ bool Game_Interpreter::CommandGameOver(RPG::EventCommand const& /* com */) { //
 	if (Game_Message::visible) {
 		return false;
 	}
-	Game_Temp::gameover = true;
-	SetContinuation(&Game_Interpreter::DefaultContinuation);
+
+	Scene::instance->SetRequestedScene(Scene::Gameover);
+	++index;
```
Ghabry: > Refactor Gameover handling 

What's this new ++index doing here?
Is that for aborting Game Over calls when another scene call overwrites in the same frame?

src/game_player.cpp
```diff
@@ -218,8 +258,21 @@ void Game_Player::UpdateSelfMovement() {
...
+				}
+			}
+		}
+	}
+
+	// ESC-Menu calling
+	if (Game_System::GetAllowMenu()
+			&& !Game_Message::visible
+			&& !Game_Map::GetInterpreter().IsRunning())
+	{
```
Ghabry: > Refactor player update routine order 

I'm not sure about the change for ESC-Menu calling.

I thought the menu can't be accessed when the route is overwritten or a message is pending to be shown but both these conditions were removed.

src/game_interpreter.cpp
fmatthew5876: That's right, otherwise we'd be stuck calling it over and over.

src/game_player.cpp
fmatthew5876: `UpdateSelfMovement()` is not called by `Game_Character` if `IsMoveRouteOverwritten() == true`. 

For the message pending part, I'll need to check how that overlaps with visible.

fmatthew5876: It's confusing as we don't have a standard way to ask "Is there a message active?"

`message_waiting` goes from when the `ShowMessage` command was invoked up until the message box starts to close. `visible` goes from the first frame of the actual message until the message close animation is done. So in most cases you probably want both.

I changed this to check for both. Also I added back a check for `!IsMoveRouteOverwritten()` I guess in theory if you press a direction and ESC at the same time, the player directional move could cause an event to run, which then sets a move route, which then prevents the player from opening the menu in the same frame...

Ghabry:
Has deps so not mergeable but except for the menu calling and message waiting which I find suspicious and will do more testing with I give my :+1: here.

> Refactor player update routine order based on Cherry

Can you find the post where Cherry explained this?

fmatthew5876:
More info here https://github.com/EasyRPG/Player/issues/1691

My original conversation with Cherry is here:
https://github.com/EasyRPG/Player/issues/1626#issuecomment-469976087

fdelapena:
Diff between PR1697 and PR1714: https://github.com/fmatthew5876/Player/compare/event_trigger...fmatthew5876:player_scene

fmatthew5876:
jenkins: test this please

Ghabry:
hm the moving is really blocked until the box is closed, not until it starts closing. Nobody noticed this for 10 years.

src/game_player.cpp
```diff
@@ -185,9 +185,50 @@ void Game_Player::UpdateScroll(int old_x, int old_y) {
...
 void Game_Player::UpdateSelfMovement() {
-	if (!Game_Map::GetInterpreter().IsRunning() && !Game_Map::IsAnyEventStarting()) {
-		if (IsMovable()) {
+	if (!IsBoardingOrUnboarding()
+			&& !Game_Map::GetInterpreter().IsRunning()
+			&& !Game_Message::message_waiting
+			&& !Game_Message::visible
+			&& !IsMoveRouteOverwritten()
+			&& !IsPaused() // RPG_RT compatible logic, but impossible to set pause on player
+			&& !Game_Map::IsAnyEventStarting())
```
Ghabry: imo this could be in "IsMovable". When you look into IsMovable almost all of this is already in there (Graphics::TransitionPending is nonsense and can be left out).
Also IsMovable is currently dead code, nothing calls it.
IsMovable even contains a case you forgot:

bla

(fancy APNG, should be animated)

Ghabry:
also when the airship is asc/desc'ing the menu is delayed until the animation finishes, in Player is immediate, not a regression, just for the notes.

src/game_player.cpp
fmatthew5876: Nice catch! Thanks

fmatthew5876:
I fixed the vehicle issues, including the menu thing.

Those are good catches, thanks for that.

I opted to just remove `IsMovable()` since it isn't really applicable anywhere else. Keeps the logic localized in one place and easier to follow imo.

fmatthew5876:
Test 4 in #1691 proves we need to check `message_waiting`

If I remove the `message_waiting` check, player will print "hello" and then call the menu after the message box dissapears.

fmatthew5876:
It looks like Ghabry's original concerns about `IsMoveRouteOvewritten()` were valid. You can press ESC during move route and menu will be summoned when the move route finishes.

Also more edges cases related to this are fixed, and described in test cases.

fmatthew5876:
jenkins: test this please

Ghabry:
Oh, so the vehicle menu calling was just an edge case, good finds :+1:

PR #1716, opened 2019-03-21
RFC: Synchronous Transitions

fmatthew5876:
Depends on #1714 

This PR changes screen transitions to act immediately by calling a function and doing an internal main loop while the transition runs.

This matches RPG_RT and allows for correct interpreter behavior when events call screen transitions. It also simplifies the scene stack logic as we don't need to handle asynchronous transitions anymore.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -253,6 +253,7 @@ add_library(${PROJECT_NAME}
 	src/switch_ui.cpp
 	src/switch_ui.h
 	src/system.h
+	src/teleport_target.h
 	src/text.cpp
 	src/text.h
 	src/tilemap.cpp
--- a/Makefile.am
+++ b/Makefile.am
@@ -241,6 +241,7 @@ libeasyrpg_player_a_SOURCES = \
 	src/spriteset_map.cpp \
 	src/spriteset_map.h \
 	src/system.h \
+	src/teleport_target.h \
 	src/text.cpp \
 	src/text.h \
 	src/tilemap.cpp \
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -395,11 +395,6 @@ void Game_Character::UpdateMoveRoute(int32_t& current_index, const RPG::MoveRout
 				break;
 		}
 
-		//FIXME: DO we need this?
-		if (move_command.command_id <= RPG::MoveCommand::Code::move_forward) {
-			any_move_successful |= !move_failed;
-		}
-
 		if (move_failed && !current_route.skippable) {
 			break;
 		}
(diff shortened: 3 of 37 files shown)


fmatthew5876:
There is a lot of history with this and some conversations on IRC and Discord. Essentially the problem is that some frameworks like emscripten and libretro want to take ownership of the main loop, where we provide them a callback function. They do the waiting and call us each frame. Supporting this means we cannot have synchronous transitions.

Synchronous transitions are desirable because they provide guaranteed accuracy with emulating RPG_RT. However they prevent integration with platforms that want to own the main loop.

I can think of 3 options, all of them suck.

1. Continue using async transitions.

In order to get RPG_RT accuracy here we're going to need an extremely delicate "suspend" and "resume" mechanism for the `Scene_Map` `Update()` and `PreUpdate()` loop. Basically everywhere we'll need to check if we're resuming a transition and exit if so. A transition can be called from any background or foreground event. Just off the top of my head I can think of many things we have to protect:
* Call Update() again after transition but don't increment the frame count.
* If PreUpdate() does a transition, make sure we go back to PreUpdate() and not Update()
* Don't process event/player/vehicle update routines twice
* Don't process screen, timer, pictures, etc.. update routines twice
* Make sure the event that caused the transiton is properly resumed, and that events that would normally go before it in the update loop don't run.
* Probably more things..

2. Synchronous transitions with using threads to save the stack

This would involve creating 2 threads that never actually run in parallel. Their only purpose is to maintain 2 stacks. It would be something like: 

Thread A runs the main loop, when it enters a transition, it goes to sleep on a mutex/condition variable, wakes up the main thread which returns. Then the next iteration, the main thread runs the transition loop. Once the transition is done, main thread goes to sleep and thread A resumes the main loop.

This is probably not feasable in emscripten:
https://emscripten.org/docs/porting/pthreads.html

It would also cost a mutex, condition variable, and thread switch on every non-transition frame which sucks :(

3. Something custom using setjmp / longjmp and saving the stack

This would be some really low level hacky solution using setjmp, longjmp, and some custom code to save the stack.

Basically when a transition is called, we need to save the stack when from EasyRPG got invoked until the current stack pointer to the heap. Then we longjmp back to the entry point and return. We get invoked again, do the transition frames. Finally, when the transition is done, we take our old saved stack and copy it back into stack memory ontop of our current stack pointer, and then longjmp back into the transition routine. This second longjmp would need very special care as the stack pointer would need a relative adjustment instead of an absolute one.

This seems terrifying and not debuggable, but maybe it could work.

As I said, all of these suck. Probably we'll have to settle for doing the best we can with (1).

fmatthew5876:
Like it or not I think we're stuck with async transitions. Closing this.

PR #1717, opened 2019-03-21
Scene + Transition fixes

fmatthew5876:
~~Depends on #1714~~

Fixes some transition issues and provides a refactor that eliminates more `Game_Temp` variables.

### DIFF
--- a/src/scene.h
+++ b/src/scene.h
@@ -85,8 +85,10 @@ public:
 	 * Continue processing.
 	 * This function is executed when returning from a
 	 * nested scene (instead of Start).
+	 *
+	 * @param prev_scene The previous scene
 	 */
-	virtual void Continue();
+	virtual void Continue(SceneType prev_scene);
 
 	/**
 	 * Resume processing.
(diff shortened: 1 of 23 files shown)


Ghabry:
I'm not a big fan of all these
```
if
elseif
else
```
to
```
if
  return
elseif
  return
[else-code]
```
changes.

Makes it really easy to add bugs by accident when a new if-condition is added and a return is forgotten.

fdelapena:
Diff between PR1714 and PR1717: https://github.com/fmatthew5876/Player/compare/player_scene...fmatthew5876:scene_fixes

src/scene.h
```diff
@@ -107,14 +111,18 @@ class Scene {
 	/**
 	 * Does the transition upon starting or resuming
 	 * the scene
+	 *
+	 * @param prev_scene - the scene we transitioned from
```
Ghabry: remove these dashes

fmatthew5876: Fixed

PR #1718, opened 2019-03-22
Fix KeyInputProc bug

fmatthew5876:
Don't set the variable if not waiting.

Fix: #1613

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -2457,7 +2457,7 @@ bool Game_Interpreter::CommandKeyInputProc(RPG::EventCommand const& com) { // co
 		}
 	}
 
-	if (var_id > 0) {
+	if (result > 0) {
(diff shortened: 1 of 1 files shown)


Ghabry:
Do you have a minimal test that shows this behaviour?

I did

```
@> Wait: 0.1 seconds
@> Control Variables: [0001] = 1
@> Key Input Processing: [0001] (Don't wait)
```

and after the Key Input Variable 1 is 0.

The Wait is just to ensure that no button is buffered

fmatthew5876:
I used #1613

I can check again a minimal test tonight

fmatthew5876:
You're right
```
Wait 1.0s
Var1 = 20
Key Input (No Wait) -> Var1
Text \v[1]
```

Prints 0. So something more subtle is happening in #1613. I will investigate further and change this.

fmatthew5876:
This one is totally wrong, so trashed.

Issue #1666, opened 2019-03-04
Android: Collection of (impossible) stacktraces


Ghabry:
Before patching them please find a way first to generate them, simply adding failure handling to silence the stacktrace is not helpful.

(with the correct debug symbols you can get the lines, until 0.6 there are only debug symbols for armeabi-v7a and x86. Newer releases also provide aarch64): https://easyrpg.org/downloads/player/dbgsym/easyrpg-player-0.6.0-android-dbgsym.tar.gz

Starting from Android 8 the C++ symbols are demangled btw.
Sorry if the stacktrace is formated incorrect, you can't copy as text from the dev console... thx google

Solved:

- [ ] 1 Input::ResetKeys()+128
- [ ] 2 SDL_MinimizeWindow_REAL+73
- [ ] 3 Scene_Battle_Rpg2k::ProcessActions()+411
- [ ] 4 _ZNK12Game_Battler8HasStateEi+31
- [ ] 5 Graphics::SortDrawableList(Drawable const*, Drawable const*)+20
- [ ] 6 ZN6BitmapC2ERKNSt6__ndk112basic_string....
- [ ] 7 _ZN13Spriteset_Map6UpdateEv+68
- [ ] 8 Graphics::GlobalDraw(int)+193
- [ ] 9 Player::Update(bool)+196
- [x] 10 _ZN6Bitmap4BlitEiiRKS_RK4RectRK7Opacity+152
- [x] 11 _ZN14MessageOverlay10AddMessageERKNSt6__ndk112basic_stringIcNS0_11char_traitsIcEENS0_9allocatorIcEEEE5Color+239
- [x] 12 _ZNK10Game_Actor11GetFaceNameEv+16
- [ ] 13 _ZN4Text4DrawER6BitmapiiiNSt6__ndk110shared_ptrI4FontEERKNS2_12basic_stringIcNS2_11char_traitsIcEENS2_9allocatorIcEEEENS_9AlignmentE+31
- [ ] 14 GameScanner.isRpg2kGame (GameScanner.java:119)
- [ ] 15 arena_run_reg_alloc+109
- [ ] 16 _WM_load_gus_pat+592

Ghabry:
Affected 3 users:

```
*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
pid: 0, tid: 0 >>> org.easyrpg.player <<<
backtrace:
[...]

Issue #858, opened 2016-04-07
Reduce Output::Error calls and invent new error "game error"


Ghabry:
We still have Output::Error calls that are recoverable, like invalid BMP/XYZ/PNG.

And because we have a game browser now maybe we should differentiate between Output::Critical for stuff like "SDL init failed" and Output::Error should just clean up nicely and return to the GameBrowser (or shutdown if not on the stack).

The only unrecoverable game caused error I'm seeing is "Invalid Map", the rest could be handle gracefully.

fdelapena:
Related: #432.

PR #1719, opened 2019-03-22
EnterHeroName on bad actor? Report error instead of crash

fmatthew5876:
Fixes Crash 12 of #1666


Ghabry:
Well I accept this but I will say again that I dislike the inconsistent use of Output::Error vs. Output::Warning.
See also #858

fmatthew5876:
The inconsistent use will be tackled when I do #1720. I want to use Error or some equivalent special error hanndler which logs event code and terminates the game on bad event code.

Issue #1725, opened 2019-03-24
Battle2k: Death state infliction animation and message issues


fmatthew5876:
Create a skill which reduces HP, MP, ATK, inflicts Death, and shifts down attributes Fire and Ice.

Use the skill on an enemy where the HP damage is not fatal.

RPG_RT:
1. HP Damage message
2. Enemy damage animation
3. MP Reduce Message
4. Atk Reduce Message
5. Death Message
6. Death animation

Player:
1. HP Damage message
2. Enemy damage animation
3. Enemy fade out death animation with no sound :exclamation:
3. Fire down message :exclamation:
4. Ice down message :exclamation:
5. Death Message
6. Death sound effect

Issue #1677, opened 2019-03-09
Fix Battle Console timings in 2k battles


fmatthew5876:
Our timings for the rm2k battle console messages are not frame accurate.

Cherry was nice enough to outline a method for extracting these timings from RPG_RT in #1568 

I've done this, and my findings are here:

Startup:
Slime appeared
Wait(4, 4) - Wait before text
Wait(30, 70) - Wait after first line

Slime appeared - first strike
Wait(4, 4) - Wait before text
Wait(30, 70) - Wait after first line
Wait(30, 70) - first strike msg

2 slimes appeared
Wait(4, 4) - Wait before text
Wait(8, 8) - Wait after first line
Wait(30, 70) - Wait after last line

Very long word wrapped lines
Wait(4, 4) - Wait before text
Wait(8, 8) - Wait after first line
Wait(8, 8) - 2nd line
Wait(8, 8) - 3rd line
Wait(30, 70) - Full page
Wait(4, 4) - Empty page
Wait(8, 8) - 1st line
Wait(30, 70) - 2nd line - done

Escape:
Multilines printed immediately
Wait(10,60)

Hero Attack - enemy none
Wait(4,4) - empty
Wait(20,40) - Attacks (text not on screen when wait called)
Wait(4,4) - next action?
Wait(4,4) - next action? (after critical maybe)
Wait(20,40) - Damage
Wait(0, 10) - post dmg
Wait(4, 4) - Enemy none action

Attack critical - enemy none
Wait(4,4) - empty
Wait(20,40) - Attacks (text not on screen when wait called)
Wait(4,4) - next action?
Wait(10,30) - critical
Wait(4,4) - next action?
Wait(20,40) - Damage
Wait(0, 10) - post dmg
Wait(4, 4) - enemy none

Enemy none - attack critical miss
Wait(4, 4) - Enemy nothing action
Wait(4, 4) - pre
Wait(20,40) - Attacks (text not on screen when wait called)
Wait(4,4) - delay
Wait(4,4) - 2nd delay likely intended for critical but still happens with no crit
Wait(20,60) - Dodge

Attack - Double crit
Wait(4, 4)
Wait(20, 40) Attakcs
Wait(4, 4)
Wait(10, 30) crtic
Wait(4, 4)
Wait(20, 40) damage
Wait(0, 10)
Wait(4, 4) Pre - attack msg shown
Wait(10, 30) crtic 2
Wait(4, 4)
Wait(20, 40) damage
[...]

PR #1726, opened 2019-03-24
Refactor Battle2k message timings + state fixes 1 /  2

fmatthew5876:
This is yet another 2k battle system refactor.

In order to emulate all of the waits RPG_RT does in #1677 I've had to add more states and substates.

I've also taken this opportunity to do some cleanup. In particular I've removed `GetResultMessages()` from `Game_BattleAlgorithm` and nearly all of the 2k message processing logic into `SceneBattleRpg2k`.

I also fixed a self-destruct bug and #1725 using a hack that hopefully can be dismantled later.

Related: #1725 
Related: #1677
Related: #1748
Related: #1770
Related: #821
Related: #1490
(they are all closed with #1785)

Tested:
- [x] Test all message timing cases in #1677 
- [x] Test all weapon state cases in #1677
- [x] Test all skill state cases in #1677
- [x] Test all item state cases in #1677
- [x] Test all self-destruct state cases in #1677
- [x] Test the last state for off by 1 errors.

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -116,6 +116,7 @@ void Game_BattleAlgorithm::AlgorithmBase::Reset() {
 	healing = false;
 	success = false;
 	lethal = false;
+	killed_by_dmg = false;
 	critical_hit = false;
 	absorb = false;
 	revived = false;
@@ -288,6 +289,10 @@ bool Game_BattleAlgorithm::AlgorithmBase::IsSuccess() const {
 
 bool Game_BattleAlgorithm::AlgorithmBase::IsLethal() const {
 	return lethal;
+}
+
+bool Game_BattleAlgorithm::AlgorithmBase::IsKilledByDamage() const {
+	return killed_by_dmg;
 }
 
 bool Game_BattleAlgorithm::AlgorithmBase::IsCriticalHit() const {
@@ -1016,6 +1021,7 @@ bool Game_BattleAlgorithm::Normal::Execute() {
 		if (GetTarget()->GetHp() - this->hp <= 0) {
 			// Death state
 			lethal = true;
+			killed_by_dmg = true;
 		}
 		else {
 			// Conditions healed by physical attack:
@@ -1293,6 +1299,7 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 				if (GetTarget()->GetHp() - this->hp <= 0) {
 					// Death state
 					lethal = true;
+					killed_by_dmg = true;
 				}
 			}
 
@@ -1831,6 +1838,7 @@ bool Game_BattleAlgorithm::SelfDestruct::Execute() {
 	if (GetTarget()->GetHp() - this->hp <= 0) {
 		// Death state
 		lethal = true;
+		killed_by_dmg = true;
 	}
 
 	healed_conditions = GetTarget()->BattlePhysicalStateHeal(GetPhysicalDamageRate());
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -557,13 +557,16 @@ std::string Game_BattleAlgorithm::AlgorithmBase::GetAttributeShiftMessage( const
 	}
 }
 
+std::string Game_BattleAlgorithm::AlgorithmBase::GetFailureMessage() const {
+	return GetAttackFailureMessage(Data::terms.dodge);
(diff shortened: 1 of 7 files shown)


fmatthew5876:
This is ready now, I've run through all the test cases and they all pass. 

- [ ] There is a bug where enemies with no actions will not print messages for their states. I believe they won't heal states either. If this takes a while to merge I might fix it here, or punt it until a later PR.

Ghabry:
Can you explain why ``GetResultMessages`` and ``GetResultSe`` were removed and ``GetFailureMessage`` and ``GetFailureSe`` were added.

How is success handled now?

fmatthew5876:
The GetResult methods had too much 2k specific logic in them so I moved them out.

In the damage and death commits I already had to start ripping stuff out of GetResultMessages(). So it already became partially implemented.

Success is computed in Execute(), so it shouldn't change.

Ghabry:
> Fix broken states healed by SelfDestruct logic 
What was broken there?


I don't really understand the cloning.
After Cloning the Actor it will still modify the same liblcf data as the original actor.

fmatthew5876:
> > Fix broken states healed by SelfDestruct logic
> 
> What was broken there?

Self Destruct is supposed to heal states like confuse, just like a normal attack.

> 
> I don't really understand the cloning.
> After Cloning the Actor it will still modify the same liblcf data as the original actor.

Ouch, that's a pretty bad oversight on my part. I will need to re-evaluate this one.

Ghabry:
the solution in #1655 by me was copying hp and states:

https://github.com/EasyRPG/Player/pull/1655/files#diff-4d15087b272c578c97d02363c9ea80d0R91

Ghabry:
Today I learned: enums can be declared in functions, cool.

Ghabry:
Actually this is not as complicated as I thought on first glance.

> Battle2k: Don't animate death until we display the message 

That should be fixed later as part of 2k3 battle refactor

> AddState() and RemoveState() return bool  AND Refactor states handling 

Has open issue with Clone

> Battle2k: Refactor results messages processing 

_Not reviewed yet as this looks complicated_

I see a view random doc comments again which contain a "-" after ``@param arg``

fmatthew5876:
Rebased and fixed the comments. 

Clone fix TBD

fmatthew5876:
The clone issue is now fixed. I need to retest state logic now.

fmatthew5876:
This now implements equipment that inflicts states.

Still testing, but I don't expect anymore major refactors/features. So this can be reviewed.

Apologies for such a huge PR.

fmatthew5876:
This now passes all test cases of #1770

src/scene_equip.cpp
```diff
@@ -89,24 +89,45 @@ void Scene_Equip::UpdateStatusWindow() {
...
+				(other_old_item->two_handed || current_item->two_handed)) {
+			add_item(other_old_item, -1);
+		}
+		add_item(current_item, 1);
+
+		atk = Utils::Clamp(atk, 1, 999);
+		def = Utils::Clamp(def, 1, 999);
+		spi = Utils::Clamp(spi, 1, 999);
+		agi = Utils::Clamp(agi, 1, 999);
+
```
Ghabry: >  Refactor scene_equip 

But this change doesn't handle the case where a state is inflicted which halfes the SP, or does it?

Ghabry:
I sak for a split after "Fixes for physically healed conditions" because the state code exploded into 20 new commits.

Ghabry:
thx!

src/scene_equip.cpp
fmatthew5876: Half SP?

If you mean the half SP cost flag on equipment, that doesn't change the equip numbers.

If you mean half / double stats from states, this is RPG_RT compatible. The equip window doesn't show the effect of those states in the arrows, they only appear after you change equipment.

Ghabry: Yes I mean the half/double skill stuff.
The problem is that this is a RPG2k3 feature, so my educated guess is: They forgot to update the scene code. ^^'

fmatthew5876: I could see a use for this. Imagine a cursed equipment with apparently good stats which ends up halving all your stats after you equip it and can't remove it!

fmatthew5876:
FYI I haven't really tested this split here. Not sure if there would be regressions while we're in between this and #1785

PR #1736, opened 2019-03-31
Fix battle events page triggering

### LINKED ISSUE #1652: Battle events refresh and update logic is wrong
Create a 2k battle with 2 enemies. Give each enemy an action which enables switch 1.

Now create 2 copies of a battle event enabled by switch 1 with this code:
```
Msg: Hello
Switch 1 off
```

In RPG_RT, we get: Enemy attacks, "Hello", Other enemy attacks, "Hello"
In Player, we get: Enemy attacks, "Hello", "Hello", Other enemy attacks

This happens because in `Game_Battle::RefreshEvents()` we mark both pages as runnable and then later in `Game_Battle::UpdateEvents()` we run both.

Instead we need to have a single iteration which checks the page conditions and runs each page at the same time, so that we don't enable both events.

fmatthew5876:
- [x] Battle 2k
- [x] Battle 2k3

Fix #1652

### DIFF
--- a/src/game_battle.cpp
+++ b/src/game_battle.cpp
@@ -77,7 +77,9 @@ void Game_Battle::Init() {
 	// troop_id is guaranteed to be valid
 	troop = ReaderUtil::GetElement(Data::troops, Game_Temp::battle_troop_id);
 	page_executed.resize(troop->pages.size());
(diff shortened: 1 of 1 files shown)


fmatthew5876:
This is ready now. Instead of some deep refactor I just fixed the bug in the current setup.

Issue #757, opened 2016-02-08
When loading a savegame, Player might skip a command in each interpreter


Zegeri:
https://github.com/EasyRPG/Player/blob/126265e206b1/src/game_interpreter_map.cpp#L118 increments the current_command index of every top interpreter. This index is not decremented at loading, so it might skip a command like ProceedWithMovement (from now: PWM) or KeyInputProc (from now: KIP) and mess things up. The increment is done to match the savegames produced by RPG_RT.

There's a catch. In Player: while PWM or KIP are running, the interpreter index points to its line. While Wait is running, the index points to the next line. In our savegames, the current_command will point to the next line of PWM and KIP and the second next line of Wait.

In RPG_RT's savegames, the index points to the next line in all three cases. We can guess that RPG_RT isn't parsing the command each frame like we do, but it uses the local variables we identified as SaveEventData::unknown_0d_move_waiting, keyinput_*, time_left for PWM, waiting KIP and Wait respectively.

This must affect...a lot of games, but I can confirm it affects Monigote Fantasy.

Ghabry:
I can confirm your guess because I noticed this when writing the event tracer (by stepping though the ASM code). RPG_RT never goes in the event code again but has if-checks for these flags before and exits earlier.

Ghabry:
This is even worse. Besides skipping lines it also ignores all flow changes.

Ghabry:
My fix only handles the really bad bug: skipping of an event command.

The save will now point at the "KeyInput" line. Which is wrong (when loading in RPG_RT). The proper fix here is not in the savegame code but altering KeyInputProc to go to the next line and handle it as zegeri suggested it.

fmatthew5876:
Here are concrete test cases:

Create this parallel event:
```
Wait 0.0s
Key Input Processing, Shift Key, Wait
Text: Hello
```

Start a new game in RPG_RT, save your game from the menu.
Saved chunks include:

```
current_command=2
keyinput_wait=1
```

Load the game from `RPG_RT`, continues waiting.
Load the game from `Player`, "Hello" message is displayed.

Ghabry:
You also see in the dynrpg event tracer that RPG_RT does e.g. the "wait for key input" processing outside of the event command.

fmatthew5876:
Do my 3 test cases adequately cover this issue? Did I miss anything?

fmatthew5876:
Fixed by #1740

Issue #1720, opened 2019-03-22
Test bad event code for crashes


fmatthew5876:
For every possible event command, we need to test it for bad data. Things like #1719 

For each event command, try to find a way to have it use invalid data. Player should either ignore it if its benign, or report an error using `Output::Error()`.

This will be more important when people are using Player to develop games, as bad event code should report errors instead of crashing.

Ghabry:
When I added sanity checks to all the data coming from LCF files I ignored event script on purpose because it was too much work to handle this but should happen someday.

Don't use Output::Error for this. Errors are extremely annoying for the user because the game terminates without saving. When you look through the code you will see that any recoverable Error is a Warning.
 You will also notice that some Warnings must be later changed to a Debug because RPG_RT accepts some bad values silently without giving an error.

fmatthew5876:
> Don't use Output::Error for this. Errors are extremely annoying for the user because the game terminates without saving. When you look through the code you will see that any recoverable Error is a Warning.

I'm not sure about this. Take #1719 for example. Here RPG_RT will exit with a dialog popup. This is because its a bug in the game's event code. Do we really want to just silently skip the `CallChangeName` command, log a warning (which the user will never check), and continue on?

In cases like this, we could end up in a state of bad event code where instead of the game crashing without allowing the user to save, the user saves and their save is corrupted.

It's also not our fault. The error indicates a fatal bug in the game, not Player.

I think it's definitely a case by case basis. There are some other situations where you could just warn and ignore without consequences.

Ghabry:
Check missing files, invalid event calls, invalid event page calls, invalid skills, invalid items, invalid actor references: These are all cases where RPG_RT terminates but we handle all of them with a warning.
What makes this "invalid actor ID" warning different to all the others?

fmatthew5876:
Lets imagine some scenarios:

1. A simple and optional action event with a bug. This event is provides an optional feature and is not critical for game progression. The player goes and talks to the event, the buggy command is skipped and the event just doesn't work. Then the player saves their game and continues on. This is *maybe* ok.

2. A main story event with a bug. The player talks to a key NPC, chooses some options which triggers the bug. The bug is ignored and we finish the event without setting all the proper conditions to continue the game, but we do  manage to enable a switch to disable talking to this event again.  The player saves, and then find out they are stuck and cannot continue the game. They are forced to start over from beginning.

[...]

PR #1740, opened 2019-04-06
Refactor Interpreter command execution

fmatthew5876:
This PR modifies the interpreter to use `SaveEventExecState` and `SaveEventExecFrame` chunks so that it executes more accurately and allows correct saving and loading of LSD chunks while events are running.

Changes include
* Refactoring call stack to use frame array instead of child interpreters
* Fixing wait commands
* Fixing key input proc
* Fixing wait for all movement
* Other cleanups

Control flow refactors using `subcommand_path` chunks will come in the next PR.

~~Depends: #1717~~
~~Depends: #1739~~
~~Depends: https://github.com/EasyRPG/liblcf/pull/324~~

Fix: #757
Fix: #1777

Test Cases from #1707 addressed by this PR
- [x] 9
- [x] 10
- [x] 11
- [x] 12
- [x] 13 - commands work in rm2k3e mode
- [ ] 13 - commands crash otherwise (We currently print a warning)
- [x] 14
- [x] 15
- [x] 16
- [x] 17
- [x] 18

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -2596,7 +2596,14 @@ bool Game_Interpreter::CommandKeyInputProc(RPG::EventCommand const& com) { // co
 	assert(frame);
 	auto& index = frame->current_command;
 
+	int var_id = com.parameters[0];
 	bool wait = com.parameters[1] != 0;
+
+	if (wait) {
+		// While waiting the variable is reset to 0 each frame.
+		Game_Variables.Set(var_id, 0);
+		Game_Map::SetNeedRefresh(Game_Map::Refresh_Map);
+	}
 
 	// FIXME: Is this valid?
 	if (wait && Game_Message::visible)
@@ -2604,7 +2611,7 @@ bool Game_Interpreter::CommandKeyInputProc(RPG::EventCommand const& com) { // co
 
 	_keyinput = {};
 	_keyinput.wait = wait;
-	_keyinput.variable = com.parameters[0];
+	_keyinput.variable = var_id;
 
 	_keyinput.keys[Keys::eDecision] = com.parameters[3] != 0;
 	_keyinput.keys[Keys::eCancel] = com.parameters[4] != 0;
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -58,9 +58,6 @@ namespace {
 	static Game_Interpreter* transition_owner = nullptr;
 }
 
-// 10000 based on: https://gist.github.com/4406621
-constexpr int loop_limit = 10000;
-
 Game_Interpreter::Game_Interpreter(bool _main_flag) {
 	main_flag = _main_flag;
 	updating = false;
@@ -3047,8 +3044,9 @@ bool Game_Interpreter::CommandCallEvent(RPG::EventCommand const& com) { // code 
 	int evt_id;
 	int event_page;
 
-	//FIXME: Max call stack depth of 100??
-
+	if ((int)_state.stack.size() > call_stack_limit) {
+		Output::Error("Call Event limit has been exceeded");
+	}
 
 	RPG::SaveEventExecFrame new_frame;
 	new_frame.ID = _state.stack.size() + 1;
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -75,9 +75,9 @@ Game_Interpreter::~Game_Interpreter() {
 void Game_Interpreter::Clear() {
 	waiting_battle_anim = false;
 	continuation = NULL;			// function to execute to resume command
-	button_timer = 0;
 	wait_messages = false;			// wait if message window is visible
 	_state = {};
+	_keyinput = {};
 }
 
 // Is interpreter running.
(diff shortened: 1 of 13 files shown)


fmatthew5876:
I'm cutting this one here and will do the control flow / subcommand_path stuff in the next PR.

Test cases related to this have been listed and need to be verified.

fmatthew5876:
All test cases pass, except that for doing "ThisEvent" from common events in Player causes warnings instead of crashing. Personally I think this should be fixed, but that falls under #1720 and I don't want a huge debate about it holding up this PR.

This is ready for review and merge.

fdelapena:
Diff between PR1717 and PR1740: https://github.com/fmatthew5876/Player/compare/scene_fixes...fmatthew5876:event_state

Ghabry:
No idea where to put this but profiling the script heavy game smashbros indicated that 25% of the whole CPU time is consumed by "list = _list" (now: frame->commands = _list).
Is there any way to get rid of this copy?

fmatthew5876:
> No idea where to put this but profiling the script heavy game smashbros indicated that 25% of the whole CPU time is consumed by "list = _list" (now: frame->commands = _list).
> Is there any way to get rid of this copy?

Great find! That makes sense for games constantly loading new events.

Could you write up a github issue. From a first glance, this seems non-trivial to me. Especially since `frame->commands` in the LSD data is literally a copy.

Since event code never changes, it should be possible to just use something akin to C++20 `std::span` to just point to the desired event buffer. That will require some surgery possibly at the liblcf level too.

So lets make an issue and plan it out. I think this is worthy performance optimization.

Ghabry:
This breaks Yume2kki completely.
After fast-forwarding through this stupid text only intro the start logo will loop forever.

fmatthew5876:
Yume 2kki bug was fixed. It was a silly mistake the call stack refactor piece.

fmatthew5876:
Rebased to master.

Retested wait commands and confirmed this PR is compatible with RPG_RT.

fmatthew5876:
Peuteris Grey issue is fixed in #1779

Ghabry:
I see that you removed CommonEvent refreshing in ``Game_Map::Refresh``.
Because this is gone you could also remove (in another PR, rebase madness) all the ``RefreshType`` stuff and replace it with a ``bool needs_refresh`` as it was years ago.

This was a performance optimisation to ensure that only toggled switches refresh common events but as only map events are refreshed now this is not needed anymore. This also means that games with many common events will run much faster now because switch refreshing of common events is gone :+1:

Ghabry:
``struct event_info`` was completely removed. This was used to make debugging easier (for me): You could directly see which event was running in the debugger (and when using child interpreters the complete callstack).
Is there still a simple way to achive this with the new code?

Ghabry:
I found a regression in the game log.[in] ([Download](https://gabgab.xyz/yumenikki/Yume%20Nikki%20Fan%20Game%20Archive%20%28October%2022%2C%202018%29/Log.%5Bin%5D/Original/Ver.%201.3/Log.%5Bin%5D.rar))

Go through the intro, start the computer, select "Enhanced Reality". Once entered you can't leave the ER menu anymore. That "Please turn on ER headset" doesn't always appear and that cancelling sometimes closes the whole PC window are game bugs.

Ghabry:
This branch seems to have a significant performance regression.
The game Frozen Triggers (http://viprpg17kh.php.xdomain.jp/main/04_game.php?id=48 click ドウンロード and extract with unar) runs in the Release (!) build with 1 FPS.

The 0.6.1 Debug build runs the game with 25 FPS.

Ghabry:
I'm not a fan of all these code duplications:
```cpp
auto* frame = GetFrame();
assert(frame);
const auto& list = frame->commands;
```

Maybe add a ``GetCommands()`` function which returns the list and does all these checks.

As with other stuff I accept this as part of later PRs when rebasing is too ugly.

src/game_interpreter_battle.cpp
```diff
@@ -91,6 +91,10 @@ bool Game_Interpreter_Battle::CommandCallCommonEvent(RPG::EventCommand const& co
 		return true;
 	}
 
+	if ((int)_state.stack.size() > call_stack_limit) {
+		Output::Error("Call Event limit has been exceeded");
```
Ghabry: Before the message was: ``Interpreter: Maximum callstack depth (100) exceeded``, the 100 could be mentioned somewhere.

src/game_interpreter.cpp
```diff
@@ -1139,19 +1126,20 @@ std::vector<Game_Actor*> Game_Interpreter::GetActors(int mode, int id) {
...
-Game_Character* Game_Interpreter::GetCharacter(int character_id) const {
-	if (!event_info.page && character_id == Game_Character::CharThisEvent) {
+Game_Character* Game_Interpreter::GetCharacter(int event_id) const {
+	if (event_id == Game_Character::CharThisEvent) {
+		event_id = GetThisEventId();
 		// Is a common event
 		if (event_id == 0) {
 			// With no map parent
-			Output::Warning("Can't use ThisEvent in common event %d: Not called from a map event", event_info.x);
+			Output::Warning("Can't use ThisEvent in common event: Not called from a map event");
```
Ghabry: is there no way anymore to get the common event id for better debug output?

src/game_interpreter.cpp
```diff
@@ -103,38 +90,27 @@ bool Game_Interpreter::IsRunning() const {
...
+	frame->ID = 1;
+	frame->commands = _list;
+	frame->current_command = 0;
+	frame->triggered_by_decision_key = started_by_decision_key;
+	frame->event_id = event_id;
 
-	CancelMenuCall();
-
-	if (main_flag && depth == 0) {
+	if (main_flag) {
```
Ghabry: > Refactor event call stack 

did you check that this is really not dependent on "depth == 0"?

src/game_interpreter.cpp
```diff
@@ -2422,125 +2551,108 @@ bool Game_Interpreter::CommandPlayMemorizedBGM(RPG::EventCommand const& /* com *
...
+				return 20 + i;
 			}
 		}
 	}
 
-	if (var_id > 0) {
-		Game_Variables.Set(var_id, result);
-		Game_Map::SetNeedRefresh(Game_Map::Refresh_Map);
-	}
+	return 0;
```
Ghabry: > Refactor keyinputproc to use LSD chunks properly 

In the old code when holding e.g. Shift + Num0 the result would be Num0, in the new code it would be shift. Is this correct?

Ghabry: yeah this change is wrong, when I just create an event which changes the graphic based on Confirm or Shift in RPG\_RT it shows shift (7) when holding both, in Player is now confirm (5)

Ghabry:
Found another regression.
Create an event with two pages, page 1 is empty, page 2 is:

```
Change Faceset: Actor 1
Message: Bla
CallEvent: ThisEvent, Page 1
Message: Blab
```

Blab is now shown. I found this by accident, actually wanted to put Blab on Page1 xD

src/game_interpreter.cpp
Ghabry: ah nvm, you don't create new interpreters anymore, so this looks fine.

src/game_interpreter.cpp
fmatthew5876: For multiple keys pressed, the logic seems to be greatest key value wins. I totally overlooked this situation. This was a really great find!

src/game_interpreter_battle.cpp
fmatthew5876: Fixed

src/game_interpreter.cpp
fmatthew5876: I took at a look at this and it's not easy. Mostly because we would have to augment the call stack to store the id for each stack frame. I could easily store the id of the running event but I wouldn't be able to fix it when I restore it.

This would need some refactor to not use LSD data directory but implement the call stack in a similar way to keyinput with translation to/from LSD data.

I agree we should have a good interpreter debugging mechanism, with first class logging and other things.

Can we make an issue and punt on that for now? Don't want to add another complex commit to this which will rebase hell everything.

fmatthew5876:
> I'm not a fan of all these code duplications:
> 
> ```c++
> auto* frame = GetFrame();
> assert(frame);
> const auto& list = frame->commands;
> ```
> 
> Maybe add a `GetCommands()` function which returns the list and does all these checks.
> 
> As with other stuff I accept this as part of later PRs when rebasing is too ugly.

I did this initially to try to minimize the amount of diffs. But I can come up with a better solution. I agree the asserts are ugly.

fmatthew5876:
> Found another regression.
> Create an event with two pages, page 1 is empty, page 2 is:
> 
> ```
> Change Faceset: Actor 1
> Message: Bla
> CallEvent: ThisEvent, Page 1
> Message: Blab
> ```
> 
> Blab is **not** shown. I found this by accident, actually wanted to put Blab on Page1 xD
> 
> Edit:
> When Page1 contains a command Blab is shown.

I had a bug in how I wasn't correctly handling empty events in my new call event code. I added a fix which fixes this and Peuteris Grey issue reported by @fdelapena

fmatthew5876:
> This branch seems to have a significant performance regression.
> The game Frozen Triggers (http://viprpg17kh.php.xdomain.jp/main/04_game.php?id=48 click ドウンロード and extract with unar) runs in the Release (!) build with 1 FPS.
> 
> The 0.6.1 Debug build runs the game with 25 FPS.

Are you sure?

I just tried this game with master in get 0 fps on linux.

fmatthew5876:
> I found a regression in the game log.[in] ([Download](https://gabgab.xyz/yumenikki/Yume%20Nikki%20Fan%20Game%20Archive%20%28October%2022%2C%202018%29/Log.%5Bin%5D/Original/Ver.%201.3/Log.%5Bin%5D.rar))
> 
> Go through the intro, start the computer, select "Enhanced Reality". Once entered you can't leave the ER menu anymore. That "Please turn on ER headset" doesn't always appear and that cancelling sometimes closes the whole PC window are game bugs.

This one is fixed now.

I have addressed all review comments and game regressions. Also rebased to master.

Ghabry:
Well the stuff I found before my vacation is fixed, so is fine for me now.
Will just merge this one now to get this finally in. But all the other PRs have to wait until next Sunday when I'm back home.

Issue #1706, opened 2019-03-17
Messages need to be refactored.


fmatthew5876:
Our messaging system needs a refactor. We need to determine frame accuracy for messages and what is / is not prohibited when they are up.

Timing from Cherry:
```
Main update methods for common events run (they just check if they are parallel process and the condition is met and then invoke the parallel process worker for them)
Main update methods for map events run (they do a bunch of things, described in my previous post - but actually even more, because after the check for whether there is any active event page, they will run their parallel process worker and then check once more for active)
Message window updates
Timer updates
Screen effects update
Pictures update
Foreground worker runs
Panorama updates
```

fmatthew5876:
Test Cases 1 - What can run while messages are up?
-----------------------

1.
EV01 parallel `Message \v[1] \v[2]`
EV02 parallel `Var1+1`
EV03 autostart `Var2+1`

| Msg | V1 | V2 |
| -- | -- | -- |
| 1 | 0 | 0 |
| 2 | 75 | 10 |
| 3 | 134 | 20 |
| 4 | 215 | 30 |

Result:
Auto start event advances by 10 frames between each message. Parallel runs during the message. The longer you wait to close the message box, the higher var1.

2. Change EV01 to 
```
Wait 0.0s
Loop:
  Msg: \v[1] \v[2]
```

| Msg | V1 | V2 |
| -- | -- | -- |
| 1 | 1 | 0 |
| 2 | 471 | 9 |
| 3 | 567 | 18 |

Result: Var2 advances by 9 frames

fmatthew5876:
Test Cases 2 - Message box animations
---------------------------------------------

1. parallel event `Message: Hello`

Result: Open, "hello", Close, Open, "hello", etc..

2. parallel event `Loop: Message: Hello`

Result: Open, "hello", Close, Open, "hello", etc..

3. autostart event `Message: Hello`

Result: Open, "hello", "hello", "hello", etc..

4. autostart event `Loop: Message: Hello`

Result: Open, "hello", "hello", "hello", etc..

5. autostart event 
```
Loop: 
  Message: Hello
  Wait 0.0s (x4)
```

Result: Open, "hello", Closes halfway, Opens from halfway position, "hello", ...

Conclusion:
--------------

* Messages from autostart events don't wait for message box animations to complete.
* Message box shink and grow animations can be "interupted" and reversed before they complete.

fmatthew5876:
Test Cases 3 - Message box timings
-----------------------------------

[...]

PR #1742, opened 2019-04-07
Refactor interpreter control flow

fmatthew5876:
~~Depends on #1740~~

The PR refactors control flow to use `subcommand_path` chunks. The goal is to ensure that every event command is either executed or skipped the same as `RPG_RT`. The practical results of this are:
* Events which run 10k commands will abort the interpreter at the right time
* Save games that are saved and loaded during executing conditionals will load correctly and be compatible between RPG_RT and Player.

Fixes most of #1707

Retested: #1613

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -405,6 +405,7 @@ void Game_Interpreter::Update(bool reset_loop_count) {
 		// Save the frame index before we call events.
 		int current_frame_idx = _state.stack.size() - 1;
 
+		const int index_before_exec = frame->current_command;
 		if (!ExecuteCommand()) {
 			break;
 		}
@@ -419,14 +420,15 @@ void Game_Interpreter::Update(bool reset_loop_count) {
 			continue;
 		}
 
-		// FIXME?
-		// After calling SkipTo this index++ will skip execution of e.g. END.
-		// This causes a different timing because loop_count reaches 10000
-		// faster then Player does.
-		// No idea if any game depends on this special case.
 		// Note: In the case we executed a CallEvent command, be sure to
 		// increment the old frame and not the new one we just pushed.
-		_state.stack[current_frame_idx].current_command++;
+		frame = &_state.stack[current_frame_idx];
+
+		// Only do auto increment if the command didn't manually 
+		// change the index.
+		if (index_before_exec == frame->current_command) {
+			frame->current_command++;
+		}
 	} // for
 
 	if (loop_count > loop_limit - 1) {
@@ -735,9 +737,9 @@ bool Game_Interpreter::ExecuteCommand() {
 		case Cmd::ChangeBattleCommands:
 			return CommandChangeBattleCommands(com);
 		case Cmd::ElseBranch:
-			return SkipTo(Cmd::EndBranch);
+			return CommandElseBranch(com);
 		case Cmd::EndBranch:
-			return true;
+			return CommandEndBranch(com);
 		case Cmd::ExitGame:
 			return CommandExitGame(com);
 		case Cmd::ToggleFullscreen:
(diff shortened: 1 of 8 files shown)


fdelapena:
Diff between PR1740 and PR1742 https://github.com/fmatthew5876/Player/compare/event_state...fmatthew5876:event_cond

Ghabry:
is the subcommand_index used anywhere for control flow or just for RPG\_RT compatibility?

fmatthew5876:
Good question. I just rebased the commits here to be more clean. I need to fully retest all of this as I haven't looked at it in months and don't remember how it is is supposed to work.

fmatthew5876:
I thought you were referring to some LSD chunk called `subcommand_index` but there is none so what is your question exactly?

`GetSubcommandIndex()` is used in `CommandOptionGeneric()` to select between different control flow options. I did this in a generic way so all option type commands end up calling this one.

The value is usually set to 0, 1, 3, etc.. to indicate which option to choose. Then the interpreter jumps to each option and checks if the option value matches the index.

It has a default sentinel value of 255 which generally means "skip to the end".

There is also a quirk where when a option command is started, the `subcommand_path` array is resized, but the values stored in it at this time are uninitialized. This is an RPG_RT bug. I initialized them to 255. ❗️

fmatthew5876:
This is ready for review and merge now.

Nearly all test cases of #1707 pass except:
* Test 4 - Requires inn refactoring #1869 
* Test 6 - LSD chunks are ok but saves too many times due to message timing issues #1706 
* Tests 22, 24, 25 - Continuations need to be moved out of game interpreter and into scenes. Continuations don't trigger if a scene is interrupted by an event. #1867

src/game_interpreter_map.cpp
```diff
@@ -338,7 +339,8 @@ bool Game_Interpreter_Map::CommandEndShop(RPG::EventCommand const& com) { // cod
 bool Game_Interpreter_Map::CommandShowInn(RPG::EventCommand const& com) { // code 10730
 	int inn_type = com.parameters[0];
 	Game_Temp::inn_price = com.parameters[1];
-	Game_Temp::inn_handlers = com.parameters[2] != 0;
+	// Not used, but left here for documentation purposes
+	bool has_inn_handlers = com.parameters[2] != 0;
```
fdelapena: This could be commented out as in `has_shop_handlers` for consistency.

fmatthew5876: Fixed

src/game_interpreter_map.cpp
```diff
@@ -212,57 +225,54 @@ bool Game_Interpreter_Map::ContinuationEnemyEncounter(RPG::EventCommand const& c
...
+bool Game_Interpreter_Map::CommandVictoryHandler(RPG::EventCommand const& com) { // code 20710
+	return CommandOptionGeneric(com, 0, {Cmd::EscapeHandler, Cmd::DefeatHandler, Cmd::EndBattle});
+}
+
+bool Game_Interpreter_Map::CommandEscapeHandler(RPG::EventCommand const& com) { // code 20711
+	return CommandOptionGeneric(com, 1, {Cmd::DefeatHandler, Cmd::EndBattle});
+}
+
+bool Game_Interpreter_Map::CommandDefeatHandler(RPG::EventCommand const& com) { // code 20712
+	return CommandOptionGeneric(com, 2, {Cmd::EndBattle});
```
Ghabry: Where do these numbers that are the 2nd parameter for CommandOptionGeneric come from?

src/game_interpreter.cpp
```diff
@@ -388,14 +421,15 @@ void Game_Interpreter::Update(bool reset_loop_count) {
...
-		// This causes a different timing because loop_count reaches 10000
-		// faster then Player does.
-		// No idea if any game depends on this special case.
 		// Note: In the case we executed a CallEvent command, be sure to
 		// increment the old frame and not the new one we just pushed.
-		_state.stack[current_frame_idx].current_command++;
+		frame = &_state.stack[current_frame_idx];
+
+		// Only do auto increment if the command didn't manually 
+		// change the index.
```
Ghabry: Under which cases does this happen?

src/game_interpreter_map.cpp
fmatthew5876: These are the values that `subcommand_path` has to be to trigger this action. 

There are example frame dumps in #1706

This deserves an enum to help make the code more self documenting.

src/game_interpreter.cpp
fmatthew5876: Any commands that need to manipulate the `index` internally. The show message is one complex example.

Loop commands are another

Issue #1744, opened 2019-04-09
VD2 healitems are not consumed after usage in menu


Tondorian:
Please fill in the following fields before submitting an issue:
#### Name of the game:
Vampires Dawn 2 
#### Player platform:
Windows, Linux, Android, ...
Windows and Switch
#### Attach files (as a .zip archive or link them)
 - A savegame next to the problem (if it is not at the beginning of the game)
 - The `easyrpg_log.txt` log file (you can find it in the game directory)
[SaveLogItems.zip](https://github.com/EasyRPG/Player/files/3059805/SaveLogItems.zip)

#### Describe the issue in detail and how to reproduce it:
If you use a potion or mana reg item it heal the player but it does not get consumed.
So right now you can heal your party for free from the menu.
In battles the item still get consumed

Issue #1743, opened 2019-04-09
VD2 screen never stop shaking after cutscene


Tondorian:
Please fill in the following fields before submitting an issue:
#### Name of the game:
Vampires Dawn 2 - Ancient Blood
#### Player platform:
Windows, Linux, Android, ...
Tested on switch and windows x64

windows works fine
switch never stop shaking

#### Attach files (as a .zip archive or link them)
 - A savegame next to the problem (if it is not at the beginning of the game)
 - The `easyrpg_log.txt` log file (you can find it in the game directory)
[saveLog.zip](https://github.com/EasyRPG/Player/files/3059764/saveLog.zip)

#### Describe the issue in detail and how to reproduce it:
The screen start to shake in the cutscene but on my switch this effect never stops.
On the windows x64 build it works fine. 

Save 3 is dumped before the cutscene started yust walk down a bit and skip the dialogues
Save 2 is the one taken from switch after the cutscene is finished

Tondorian:
I was not able to reproduce it on my switch twice. i guess it was some wired race conditon.
But save 2 is still broken

fmatthew5876:
Could this be some kind of frame skipping bug? Do we do that on low power platforms like switch?

Tondorian:
yes this bug happend on my switch but only once, maybe power was low but i do not remember if this might be the reason, i could not reproduce this with 3x speed so i guess it was an unlucky event.

Ghabry:
Frame skipping only affects ``Draw`` but maybe you see something in one of the Draw functions that looks suspicious and should be in update instead.

fmatthew5876:
Well one thing I see already is that game_screen Reset doesn't clear the shake chunks.

fmatthew5876:
@Tondorian 

At which point in the cutscene does the continuous shake start? I played all the way through, walking down, up into the castle, and then back down the hills again when the hero tries to escape.

fmatthew5876:
I found it, this is caused by the ShakeOnce command at the beginning of the cutscene.

Normally a 2k3 shake once looks like this:
```
0000: code=ShakeScreen(11050) indent=0 str="" params=[5 5 2 1 0]
```

But in this game, params size is only 4.
```
0084: code=ShakeScreen(11050) indent=0 str="" params=[5 5 2 0]
```

The bug is caused by an uninitialized read of `params[4]` which explains why it occurs randomly.

Maybe this game was originally 2k and then ported to 2k3?

[...]

PR #1745, opened 2019-04-11
Item + Shake fixes

fmatthew5876:
Fix: #1743
Fix: #1744


src/game_interpreter.cpp
```diff
@@ -2023,14 +2023,10 @@ bool Game_Interpreter::CommandShakeScreen(RPG::EventCommand const& com) { // cod
 	int speed = com.parameters[1];
 	int tenths = com.parameters[2];
 	bool wait = com.parameters[3] != 0;
+	int shake_cmd = com.parameters.size() > 4 ?
```
carstene1ns: I would like a comment here referring to 2k.

fmatthew5876: Good idea, added.

PR #1746, opened 2019-04-14
CMake: Add libsamplerate + maintanance

Ghabry:
cherry-picking stuff out of the libretro branch

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -363,7 +363,7 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/builds/cmake/Modules"
 include(ConfigureWindows)
 include(PlayerFindPackage)
 
-if(CMAKE_BUILD_TYPE MATCHES "Debug")
+if(CMAKE_BUILD_TYPE STREQUAL "Debug")
 	target_compile_definitions(${PROJECT_NAME} PUBLIC _DEBUG=1)
 endif()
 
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -497,12 +497,38 @@ else()
...
-	option(PLAYER_WITH_SPEEXDSP "Resample audio with speexdsp." ON)
-	player_find_package(NAME speexdsp
-		CONDITION PLAYER_WITH_SPEEXDSP
-		DEFINITION HAVE_LIBSPEEXDSP
-		TARGET speexdsp::speexdsp)
+	set(PLAYER_AUDIO_RESAMPLER "Auto" CACHE STRING "Audio resampler to use. Options: Auto speexdsp samplerate OFF")
+	set_property(CACHE PLAYER_AUDIO_RESAMPLER PROPERTY STRINGS Auto speexdsp samplerate OFF)
+
+	if(${PLAYER_AUDIO_RESAMPLER} MATCHES "Auto")
+		option(PLAYER_WITH_SPEEXDSP "Use speexdsp as resampler when available." ON)
```
carstene1ns: Should these `option`s not be moved up, outside of the condition?

Ghabry: They are correctly placed there because they are used to configure the options for the autodetection but I will change all these dependent options to CMAKE_DEPENDENT_OPTIONs. This way they are also properly conditionaly hidden in ccmake and cmake-gui. (try disabling PLAYER_WITH_FREETYPE in these tools, PLAYER_WITH_HARFBUZZ will be appear/dissapear)

Ghabry:
Short story about the
```
if(${PLAYER_AUDIO_BACKEND} STREQUAL "Auto")
	set(PLAYER_AUDIO_BACKEND_IS_AUTO ON)
endif()
```

In theory "${PLAYER_AUDIO_BACKEND} STREQUAL \"Auto\"" could be passed to CMAKE_DEPENDENT_OPTION because argument 4 is simply inserted into a if-block but due to strange if-variable expansion in CMake this doesn't work. (also would result in more code duplication because of the if-check below)

Also no idea how to use these dependencies when they are strings/properties and not bools, at least the macro doesn't support this :/.

Ghabry:
7e8d750 is also bogus, another failed rebase conflict resolution, waiting for there slow buildbots...

Ghabry:
this is correct now

PR #1749, opened 2019-04-17
fixes #1748. Reviving items does not revive.

### LINKED ISSUE #1748: VD2 reviving items in battle does not work

---
#### Name of the game:
VD2 

#### Player platform:
Windows, Linux, Android, ...

[easyrpg_log.zip](https://github.com/EasyRPG/Player/files/3087273/easyrpg_log.zip)
Ignore the logfile.

Reviving items in battle heals 0 hp and the actor is still dead.
i am not sure if it is vd2 that is that much special but reviving items did not heal the dead actor.
Maybe vd2 is a very special case because vampires are not dead and only in "starre" while humans get a dead status.  
The item is called "Trank der Ewigkeit"

Tondorian:
fixes #1748.
maybe this solution only fixes the revive issue. I am not aware if condition healing items in general are broken in 2k3 maybe it can be refactored as a common condition heal fix.

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -1263,6 +1263,7 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 			if (IsRevived() && !skill.affect_hp) {
 				this->hp = std::max<int>(0, std::min<int>(GetTarget()->GetMaxHp() - GetTarget()->GetHp(),
 							GetTarget()->GetMaxHp() * effect / 10));
+				this->healed_conditions.push_back(1);
 				this->success = true;
 			}
 		}


fmatthew5876:
Here is my feedback:

First, I should have mentioned this before. But please retest this with #1726. I refactored states in battle algorithm for both battle systems, despite it being a 2k refactor. Maybe the bug is already fixed? I should have mentioned this earlier, sorry about that.

1. This is probably the worst commit message. I know you're unsure and looking for feedback, but the commit message should be descriptive of what you're intending to do with this change, details about it that aren't clear from the code, and nothing more. Remember in the history its only your commit message the code diff. How you felt about the change is irrelevant to someone looking at this 6 months from now and trying to figure out what the change was for. So describe what you want to do in the commit message, and put the "I'm not sure about this" details in the PR description.

2. The `healed_conditions` variable is actually used for conditions healed by physical attacks such as confuse. Take a look at `Normal::Execute()` to see how this is used. The name is bad and so I don't blame you for making this mistake. If you look at #1726 you'll see I actually rename this one.

3. Take a look later on in `Skill::Execute()`. This is where we apply states in the `condition` variable. There is just one list of conditions which are removals for healing skills and additions for offensive skills.

I already see a bug here where we unconditionally set lethal flag for both positive and negative skills if `kDeathId` is in the states vector. It should only be lethal if the skill is killing the target, not reviving.

Essentially you should verify `condition` vector has the death state when `Execute()` finishes, and then when `Apply()` runs it should actually remove that state from the battler. Something is broken within that sequence of logic, and you'll need to debug to find out why and then propose a proper fix.

Or maybe my battle pr already fixes this?

Tondorian:
Thank you for your detailed feedback pr #1726 fixes this issue already. I was not aware of this pull requst.
So we can close this one.

Issue #1361, opened 2018-05-07
Incorrect text color behavior for choices


vgperson:
#### Name of the game:
Standstill Girl (http://vgperson.com/games/standstillgirl.htm)

#### Player platform:
Windows

#### Attach files (as a .zip archive or link them)
[StandstillGirlChoiceColor.zip](https://github.com/EasyRPG/Player/files/1978499/StandstillGirlChoiceColor.zip)

#### Describe the issue in detail and how to reproduce it:
Minor but simple: In RPG_RT, text color is not reset between choice options, so color changes can carry over between lines. Choices in Standstill Girl make use of this, only putting a \C[2] in the first option and letting it color all the others too. Walk forward and get to the "Nope / Nope nope" choice for an example.

Window_Message::InsertNewLine() says that choices reset font color, but this seems to be completely false - RPG_RT even has color spill over from the preceding message lines. So that line should just be removed. (Although some extra work is necessary to support the color change for disabled choices.)

PR #1751, opened 2019-04-19
Fix Message Color is reset on multiple choices.

Tondorian:
In RPG_RT text color is not reset between multiple choices, this pr will apply this behavior to the player.
New Message boxes resets the color to default in rpg_rt and player.

Fix: #1361

### DIFF
--- a/src/window_message.cpp
+++ b/src/window_message.cpp
@@ -274,9 +274,6 @@ void Window_Message::InsertNewLine() {
 	++line_count;
 
 	if (line_count >= Game_Message::choice_start && Game_Message::choice_max > 0) {
-		// A choice resets the font color
-		text_color = Font::ColorDefault;
-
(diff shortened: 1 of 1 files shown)


Tondorian:
>  When the previous message changed color and choices are on a new window.

New Windows resets colors to default on RPG_RT and Player. 

 > When message text is in the same window as choices, and message changes color?

This will print the message and the choices in this color
   
>  When choices change color, followed by a message?

This is also reset by Window_Message::InsertNewPage() which will reset to the default color
 
> 2 sets of Choices, when first changes color?

The second choice is in a new window so it will be the default color  

>   How about number input?

Number inputs are printed in the default color in both cases

> Any other edge cases you can think of?

I think most cases are cought by Window_Message::InsertNewPage().

Ghabry:
So in general choices are just like normal messages, the only difference is, that a cursor is overlayed

Issue #1728, opened 2019-03-26
Title scene background should be black if title graphic is undefined in RPG_RT.ldb


fdelapena:
Currently, Player uses the system default background pixel color instead of rendering a black texture. This is a corner case, but there is some chance some game look different in this scene due to this difference. The current background color used with `HideTitle` is correct, so the fix should also preserve this behavior.

Test case in the title scene of:
https://github.com/EasyRPG/Player/files/3006572/midi-loop-test.zip
Online test:
https://easyrpg.org/play/master/?game=issue-834

Player screenshot:
![Captura de pantalla de 2019-03-25 23-30-49](https://user-images.githubusercontent.com/174885/54973558-0e776c00-4f56-11e9-8b37-671f42cf6317.png)

RPG_RT screenshot:
![Captura de pantalla de 2019-03-25 23-30-00](https://user-images.githubusercontent.com/174885/54973552-028baa00-4f56-11e9-8e8e-69ae78920de4.png)

PR #1752, opened 2019-04-19
Replace Default Color with Black if there is no title images

Tondorian:
This PR replaces the default Color of the title screen to black if there is no background image available.
If the player is started with --hide-title the behavior remains unchanged.

Fixes: #1728

### DIFF
--- a/src/scene_title.cpp
+++ b/src/scene_title.cpp
@@ -35,6 +35,7 @@
 #include "scene_load.h"
 #include "scene_map.h"
 #include "window_command.h"
+#include "baseui.h"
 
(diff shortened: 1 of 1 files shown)


Tondorian:
do we expect --hide-title to be also black if there is no image present?

Ghabry:
The supported way to do this is by overloading DrawBackground.

Your change modifies the global state of the background graphic -> The menu gets a black background.

Tondorian:
for some reason overwriting DrawBackground always break transitionOut so I decided to create a black colored Titlesprite to resolve this issue.

PR #1754, opened 2019-04-22
Refactor RTP handling

### LINKED ISSUE #802: Write smarter RTP translation table logic
Maybe use [`std::unordered_multimap`](http://www.cplusplus.com/reference/unordered_map/unordered_multimap/).

### LINKED ISSUE #1709: Files not listed in the RTP replacement tables are not found
This issue affects modified or unknown RTPs, e.g. games using Don's RTP extras (there are quite a few) and Don's RTP 1.32.

How to reproduce:

Add some file to the RTP with a custom name not in RTP tables, then use this asset from a game.

A simpler way to reproduce this is by installing Don's RTP and run run Don's Adventures game.

Possibly related with #1630.

Ghabry:
Installed RTP selection algorithm:
All assets in a directory are mapped against all RTP and the hit/max ratio is checked. The RTPs with the highest ratio in the folder are considered as the candidates.

Game RTP selection algorithm:
We don't know which RTP the game uses, so the strategy is: Take the requested game asset (that was not found in the game folder) and map it against all RTP in the table -> Results:
- 0 RTP matches: Not a RTP asset -> do direct search in all RTP paths
- 1 RTP match: Perfect, we know the RTP now (the Game RTP selection algorithm is finished)
- 2+ RTP matches: When this was the first asset -> write all candidates in a list
- 2+ RTP matches and this was not the first asset: Compare the new candidates against the candidates of the previous asset and only keep the RTPs that contain all assets (this will reduce the list to 1 RTP after 2 or 3 assets). In rare cases (totally messed up game or rtp folders) the can result in 0 RTPs, in this case the algorithm repeats from the beginning.

Lookup strategy:
All game RTP against all detected RTP until the FileFinder finds a file.
This is the new Any-To-Any lookup, it doesn't matter anymore which RTP(s) are installed.
Fix #802 
(while typing this I noticed that I forgot the "game RTP = detected RTP" case for the lookup)

When all this detection magic fails it directly searches the file in all RTP search paths (Fix #1709). This is after all the magic to prevent undesired hits in the folder (because of name overlaps between the different RTPs)

I changed the RTP structure to a table and do linear searching in it. Some cases can be still optimized, e.g. finding the correct asset category can be a binary search because this column is sorted.
Also prints diagnostics about the detected RTPs, great to blame the Android users when they send us bogus bug reports :P.

Design decision: Don's RTP extras are included in the table, this is not really necessary because there is mapping possible but it is great for the startup diagnostics (maybe could add an extra warning that the game uses such an asset)


Ghabry:
Test cases:

- [ ] Folder contains one RTP
- [ ] Folder contains two RTPs (e.g. Don + DonExtra)
- [ ] Folder contains partial RTP
- [ ] Folder contains nonsense or is empty (hello, Android users)

- [ ] Game uses same RTP as installed
- [ ] Game uses other RTP as installed
- [ ] Game uses asset that is not in the game folder and not a RTP asset (passthrough test)

- [ ] Game RTP detection logic test: The 1st asset yields 2 possible RTP

carstene1ns:
>Ghabry> wow RPG advocate reencoded all sound effects to mp3 m(

As said on IRC (which you do not read), they did it to decrease download size and shipped a decompression script: https://paste.xinu.at/Fsl8Aj/ `RPG_RT` does not play mp3 SE.

Ghabry:
I would prefer keeping this mp3 detection because this mitigates a user error (though this is only cosmetic, even with no sounds detected Advocate will still win in a Advocate-only folder).

Ghabry:
Some lookup test cases, e.g. the Title BGM is "Suspicion" (Official English) and the installed RTP is Don+extras:

```
Debug: RTP is "Don Miguel English Translation" (500/500)
Debug: RTP is "Don Miguel RTP Addon" (503/503)

system
Candidate: Official English
Candidate: Don Miguel English Translation
---
title1
Candidate: Official English
Candidate: Don Miguel English Translation
---
suspicion
Candidate: Official English
---
Debug: Game uses RTP "Official English"
```

The Title is skipped and installed RTP2k3 is Advocate:

```
Debug: RTP is "RPG Advocate English Translation" (675/675)

system
Candidate: Official English
Candidate: Vlad Russian Translation
Candidate: RPG Universe Spanish/Portuguese Translation
---
ship
Candidate: Official English
Candidate: RPG Advocate English Translation
Candidate: Vlad Russian Translation
---
actor1
Candidate: Official English
---
Debug: Game uses RTP "Official English"
```

To RTP are installed (2kjp and 2ken) in different folders, 2kjp lacks the Music files for some reason:
Result: The song still plays.

```
Debug: RTP is "Official Japanese" (368/465)
Debug: RTP is "Official English" (465/465)

system
Candidate: Official English
Candidate: Don Miguel English Translation
---
title1
Candidate: Official English
Candidate: Don Miguel English Translation
---
suspicion
Candidate: Official English
---
Debug: Game uses RTP "Official English"
```

Ghabry:
I observed something strange with the 64bit Wine prefix: Some keys are not redirected to Wow6432Node, as you can see in this picture:

Screenshot_20190423_211018-fs8

RTP and 2003 are 32bit executables, so I'm not sure what is causing this. I changed our wine registry now to check both Wow6432Node and the not redirected case. Now my RTP installs in Wine are actually found again :)

carstene1ns:
> (HKCU redirection is wrong btw, Windows doesn't redirect them)

What does this mean? Is it wrongly implemented? Is it wrong in the docs? :-1:
Because that is exactly how my patch should work: https://github.com/EasyRPG/Player/commit/317c70ccc8780a916d97345a0599341105941130#diff-059176e8293a79703c0e08f26fad62aeR64

Also, `wine` may have bugs as well?

Ghabry:
I mean MSDN says that HKCU is shared and not redirected but Wine does (but not always? look at my screenshot), so our implementation was wrong.
Could be a Wine bug, but I can't search through thousand registry bugs in Wine.
The new solution is to look now in Wow6432Node and in the not redirected folder.

carstene1ns:
I am not really against testing both, but this looks like a bug, and we should not emulate "not an emulator" bugs.

src/registry_wine.cpp
```diff
@@ -25,32 +25,41 @@
...
 			getenv("WINEPREFIX")? getenv("WINEPREFIX"):
 			getenv("HOME")? std::string(getenv("HOME")).append("/.wine"):
 			std::string();
 
-	std::string registry_file = "";
-	std::string string_value = "";
-	std::string formatted_key = key;
+	std::string registry_file;
+	std::string string_value;
+	std::string formated_key = key;
```
carstene1ns: 1 or 2? https://en.wiktionary.org/wiki/formated :8ball:

Ghabry:
When I install the RTP before the Editor all registry values appear in Wow6432Node, must be some Wine bug, but can't figure it out.

Keys created through a small test program always appear in Wow6432Node and the RTP also always appears always in Wow6432Node, therefore I will only check this node.

(Our current code is incorrect, as we check HKCU\Software\Kadokawa and not HKCU\Software\Wow6432Node\Kadokawa, I wonder if this worked years before and Wine changed this behaviour)

carstene1ns:
Then again: HKLM/Software is redirected, HKCU/Software is shared. Either wine or the MSDN docs are wrong here.
https://docs.microsoft.com/en-us/windows/desktop/WinProg64/shared-registry-keys

Ghabry:
Wine doesn't follow the MSDN docs, they redirect HKCU, Windows shares it.

Ghabry:
Okay I finally found the code, this proofs that currently Wine only shares Software\Classes

https://github.com/wine-mirror/wine/blob/master/server/registry.c#L1853-L1863

src/filefinder.cpp
```diff
@@ -169,80 +181,113 @@ namespace {
 		return "";
 	}
 
-	bool is_not_ascii_char(uint8_t c) { return c > 0x80; }
+	// returns empty string when the file is not belonging to an RTP
+	const std::string rtp_lookup(const std::string& dir, const std::string& name, const char* exts[], bool& is_rtp_asset) {
+		int version = Player::IsRPG2k() ? 2000 : 2003;
```
carstene1ns: `Player::EngineVersion()` ?

src/filefinder.cpp
```diff
@@ -447,9 +492,26 @@ std::string FileFinder::FindFont(const std::string& name) {
 static void add_rtp_path(const std::string& p) {
 	using namespace FileFinder;
 	std::shared_ptr<DirectoryTree> tree(CreateDirectoryTree(p));
-	if(tree) {
+	if (tree) {
 		Output::Debug("Adding %s to RTP path", p.c_str());
-		search_paths.push_back(tree);
+		rtp_state.search_paths.push_back(tree);
+
+		auto hit_info = RTP::Detect(tree, atoi(Player::GetEngineVersion().c_str()));
```
carstene1ns: `Player::EngineVersion` ?

Ghabry: hm, yes. That's exactly the function I needed, thx :D

Ghabry:
(just rebased to master)

src/rtp_table.h
```diff
@@ -1,43 +0,0 @@
...
-#include <cstring>
-
-namespace RTP {
-	struct rtp_table_cmp {
-		bool operator()(char const* a, char const* b) const {
-			return std::strcmp(a, b) < 0;
-		}
-	};
-
-	using rtp_table_type = std::map<const char*, std::map<const char*, const char*, rtp_table_cmp>, rtp_table_cmp>;
```
fmatthew5876: Since you're constructing this all at startup, you'd get better performance using `std::lower_bound` on sorted vectors than you would with `std::map`.

The problem with `map` is that it's a node based container, which means a separate heap allocation for each node. This uses more memory and even worse has terribly cache locality.

To pull this off cleanly, you' have to make your `rtp_table_type` an actual class with these semantics embedded inside. Boost provides something called a `flat_map` which does this, but we don't want to use boost and implementing a simplified version of this for rtp table is trivial.

If there is lack of time for 6.1, we could punt this optimization for the next PR.

Ghabry: The std::map RTP code is deleted as part of this PR. The new data structure is 2D-char*-array / table.

src/rtp.cpp
```diff
@@ -0,0 +1,163 @@
...
+#include <cassert>
+#include <cstring>
+#include "rtp.h"
+
+namespace RTP {
+    extern const char* rtp_table_2k[][5];
+    extern const char* rtp_table_2k3[][7];
+
+	constexpr int num_2k_rtps = 4;
+	constexpr int num_2k3_rtps = 6;
```
fmatthew5876: You could declare the second array bound with `num_2k_rtps + 1` etc..

src/rtp.cpp
```diff
@@ -0,0 +1,163 @@
...
+static void detect_helper(const FileFinder::DirectoryTree& tree, std::vector<struct RTP::RtpHitInfo>& hit_list,
+		T rtp_table, int num_rtps, int offset) {
+	for (int i = 0; rtp_table[i][0] != nullptr; ++i) {
+		const char* category = rtp_table[i][0];
+		for (int j = 1; j <= num_rtps; ++j) {
+			const char* name = rtp_table[i][j];
+			if (name != nullptr) {
+				std::string ret;
+				// TODO: Filefinder refactor should provide FindImage etc. for non-project trees
+				if (!strcmp("sound", category)) {
```
fmatthew5876: This logic would be faster if we used enums for the categories.

src/rtp_table.cpp
```diff
@@ -1044,6 +1042,42 @@ const char* rtp_table_2k[][5] = {
...
+	521,
+	522,
+	613,
+	614,
+	741,
+	755,
+	756,
+	978,
+	1002,
+	1006
```
fmatthew5876: What you're doing makes sense from a performance standpoint, but maintaining these tables is very brittle. Possible to write any unit test or `static_assert()` to ensure correctness here?

fmatthew5876:
Looks good to me, my review comments are nit-picky so up to you if its worth addressing them.

This one is a bit tough to review as there are many evolutions in design. I think thorough testing is going to be more effective here than code review.

src/rtp_table.cpp
Ghabry: the numbers are calculated by the bundled Python script, will try to add a check to ensure the mapping is up-to-date.

src/rtp.cpp
Ghabry: good idea, this way it generates a compiler error when these numbers are not updated :)

Ghabry:
Optimized the RTP in folder detection and added some unit tests (also contains a test if the generated tables are up-to-date)

Ghabry:
Good news: Wine seems to have fixed the incorrect Wow6432Node redirection in HKCU. So this workaround can be removed :confetti_ball: .
At least in a clean Prefix it works. In my older prefix it still writes in Wow6432Node, maybe backwards compatible to not break old prefixes.

Issue #1177, opened 2017-05-22
Memory usage. Vampires Dawn: creating Image Error


Klauserus:
#### SaveFile: 
sorry, i belive cant save the game before, but it takes 4 minutes to the point of error

#### Describe the issue in detail and how to reproduce it:

When i fly first time to Asgars castle/palace, the screen get red and a error message come out :

Couldn´t create 288x256 Image. EasyRPG Player will close now.

Thank you and greetings

carstene1ns:
This is likely a memory related problem, so we cannot really do much at this point.

Klauserus:
ok. You mean it is a RAM problem? Oh man. This is really a Problem. We can´t upgrade the RAM of a Wii :D 
Are there any solution approaches? This makes some games unplayable
thx

carstene1ns:
Well, we currently do not really try to have a small memory footprint. Further explanation here:
https://community.easyrpg.org/t/wii-build-test-reports/197/69 (other game, same issue).

Klauserus:
Can one say then that the Wii version has reached its limits and there will be no 100% game compatibility in the future? ...i will donate for more development Power :)

Ghabry:
The irony is somehow that some of these cachings were invented to make the experience on slow systems (Wii & co.) more smooth but now they break it :/.

We have some ideas to reduce the memory usage they are just not high priority.

Ghabry:
Using this as a meta issue for memory footprint stuff.
One place where we waste some memory is that we allocate all Sprites on the map upon map load even if they are off-screen. This uses significant amounts of memory in some games. The disadvantage of loading a sprite on first appearance are micro stutters (especially bad on the 3DS I guess...) but better than OOM I guess?

Comparison for The Looking Glass (discussed here): https://community.easyrpg.org/t/wii-build-test-reports/197/69?u=ghabry

Before: 
![worse_mem](https://user-images.githubusercontent.com/1331889/27153699-c48da54e-5152-11e7-9546-bdc695e8d675.png)

After:
![better_mem](https://user-images.githubusercontent.com/1331889/27153706-c98a5326-5152-11e7-8781-afda2797706c.png)

Ghabry:
More potential memory saving: Cache::Tile could return a Rect to the tilemap instead of a new Tile. Saves ~1 Kbyte per tile (not really worth it).

Autotiles are already created on demand as far as I can see, no memory savings possible.

[...]

PR #1760, opened 2019-05-06
Cache improvements and effect caching

Ghabry:
not ready yet, have strange flickering.

Improved the caching strategy, before our Cache was unlimited and never cleared any recently (<5s) used assets. Now the clear is forced when the memory usage is 10 MB (+3 MB SE).

Which means the memory foodprint is in simple games (havn't used the MSVC memory profiler yet) ~20 MB plus Database. 

Besides reduced memory usage the effect cache should also help in some situations to make the effect faster because tone & flash are only calculated once now per same charset.

Related #1177

### DIFF
--- a/src/audio_secache.cpp
+++ b/src/audio_secache.cpp
@@ -31,7 +31,7 @@ namespace {
 
 	cache_type cache;
 
-	const int cache_limit = 1 * 1024 * 1024;
+	constexpr int cache_limit = 3 * 1024 * 1024;
 	int cache_size = 0;
 
 	void FreeCacheMemory() {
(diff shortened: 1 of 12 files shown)


Ghabry:
Comparison images: (already making them now, as the "flicker fix" won't increase the memory usage)
The biggest impact has "Fix the out of bounds check code" but combined with "Support caching" it helps to prevent increase of memory usage over time (when more parts of the map are visited).

_Note: This are manual recordingings, so the time axis won't match, look at the memory usage._

**The Looking Glass**:

Master:
glass_before-fs8

This branch up to "Fix out of bounds":
glass_before2-fs8

This branch all commits:
glass_after-fs8


**Yume2kki**

Master: (32bit limit -> crash)
yume2kki_before-fs8

This branch up to "Fix out of bounds":

yume2kki_before2-fs8

This branch all commits:
yume2kki_after-fs8

Ghabry:
The flicker fix also will contain yet another minor improvement: It saves 258 KB per CharSet that is affected by a screen effect. Saves a non measurable amount for "The Looking Glass" and 2 MB in Yume2kki.

Ghabry:
> Does this require constexpr? It could be useful as a tunable parameter in RPG_RT.ini

I can make it non-const.
Though I don't see that much use in making this tunable. You will always need at least a small cache for reducing lag on small systems (e.g. without the SE cache Yume Nikki was unplayable on older systems because the IO and the decode took too long).
And even on newer systems not caching e.g. the system graphic has a noticable performance penalty, the Game Browser startup speed is now instant instead of delayed (and I always thought the cause is the FileFinder...)

fmatthew5876:
> > Does this require constexpr? It could be useful as a tunable parameter in RPG_RT.ini
> 
> I can make it non-const.

Unless it's going to be exposed as a runtime config parameter I would leave it constexpr, at least then you can maybe eke out some additional compile time optimizations.

Issue #1761, opened 2019-05-07
Open menu black frequency


s9060666:
Thank you for supporting EasyRPG Player.
Before creating an issue, please verify that your bug still occurs in the "continuous builds" of the Player.
You can get them at the bottom of our download section: https://easyrpg.org/player/downloads

---
Please fill in the following fields before submitting an issue:
#### Player platform:Android 
Windows, Linux, Android, ...

#### Attach files (as a .zip archive or link them)
 - A savegame next to the problem (if it is not at the beginning of the game)
 - The `easyrpg_log.txt` log file (you can find it in the game directory)

#### Describe the issue in detail and how to reproduce it:
After turning on the menu, it will be directly black. Nothing can be done.

fdelapena:
Thank you for the bug report.

Some notes:

- Game website: https://sites.google.com/site/mrjxrmplan/game_mf2
- Game download (ver. 1.15): https://www.dropbox.com/s/mnm1eav9qa27cif/mf2.rar?dl=1
- This game uses custom menus: https://i.imgur.com/feeKLDE.jpg

Ghabry:
This is a post 0.6 regression!
In 0.6 the menu works and doesn't blackscreen.

fmatthew5876:
Here is a savegame after the intro. Press ESC and see a black screen

[Save01.zip](https://github.com/EasyRPG/Player/files/3219828/Save01.zip)

fmatthew5876:
This looks like some interpreter bug

Common event 5 is parallel has the key input proc to trigger off of ESC. When this event is triggered, you get teleported to map 5.

On Map5, an autostart map event is supposed to take over. But it looks like common event 10 somehow keep running over and over in the foreground, preventing it from executing. 

Digging deeper...

fmatthew5876:
Test Cases
==========

1. Setup EV01 autostart
```
Wait 0.0s
```

EV02 autostart
```
Shake Screen
```

Result: Screen shakes

2. Delete EV01,

Setup CE01 as autostart
```
Wait 0.0s
```

Result: Screen Shakes

fmatthew5876:
Test Cases 2 - Messages
=================

1. EV01 autostart `Msg: EV01`
EV02 autostart `Msg: EV02`
CE01 autostart `Msg: CE01`

Result: EV01, CE01, EV01, CE01, ...

fmatthew5876:
Test Cases 3 - Exec Order
=====================

[...]

PR #1771, opened 2019-05-26
Fix foreground exec order (temporary workaround)

fmatthew5876:
Map and Common events take turns executing
in the foreground.

Fixes the game bug in #1761 and most of the test cases 

Proper solution is in #1779

### DIFF
--- a/src/game_interpreter.h
+++ b/src/game_interpreter.h
@@ -55,6 +55,7 @@ public:
 	int GetLoopCount() const;
 	bool ReachedLoopLimit() const;
 	void Update(bool reset_loop_count=true);
+	bool IsRunningMapEvent() const;
 
 	void Setup(
 			const std::vector<RPG::EventCommand>& _list,
@@ -242,6 +243,10 @@ protected:
 
 inline int Game_Interpreter::GetLoopCount() const {
 	return loop_count;
+}
+
+inline bool Game_Interpreter::IsRunningMapEvent() const {
+	return event_id != 0;
 }
 
 #endif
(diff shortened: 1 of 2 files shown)


fmatthew5876:
jenkins: test this please

fmatthew5876:
So it looks like this PR is incomplete. I don't correctly emulate the teleport test cases of #1761 

If you have autostart CE1 'Msg: CE01' and autostart EV01 `OpenSaveMenu` you get something like this in LSD chunks:

```
0x71 (size: 50, pos: 0x1e8e): foreground_event_execstate
0x01 (size: 47, pos: 0x1e90): stack
0x01 (size: 1, pos: 0x1e94):
  13
0x02 (size: 13, pos: 0x1e97): commands
0x16 (size: 0, pos: 0x1ea6): subcommand_path

0x01 (size: 1, pos: 0x1eaa):
  9
0x02 (size: 9, pos: 0x1ead): commands
0x0b (size: 1, pos: 0x1eb8): current_command
  1
0x0c (size: 1, pos: 0x1ebb): event_id
  1
0x16 (size: 0, pos: 0x1ebe): subcommand_path

```

So if both a common event and a map event can run. It'll actually push the common event onto the stack, then push the map event as if the common event had called it!

The proper fix for this will have to wait until after #1740. But the change I have no is a stop gap to fix the game bug in #1761 for 6.1 release.

Ghabry:
One interesting test is #1247 was depending on the correct Common/Map event order and it still works :+1:

PR #1772, opened 2019-05-26
Refactor character flash

### LINKED ISSUE #1768: Flash Character commands are not frame accurate
Use an event flash command (31,31,31,31), the LSD chunk values player uses are wrong.

- [x] RPG_RT uses [0,31] for colors, player uses [0,255]
- [x] RPG_RT defaults to 0 for colors, player default is 100
- [x] Player doesn't write `flash_current_level` or `flash_time_left` chunks.
- [x] When does RPG_RT update event flash in the main loop? Needs to be discovered and emulated..

So player is not savegame compatible with event flash.

fmatthew5876:
Use event flash LSD chunks to save and load flash state.
Animate flash from characters update routine.

Fix #1768 

All test cases pass, except `wait_time` chunk which is handled in #1740

### DIFF
--- a/src/battle_animation.cpp
+++ b/src/battle_animation.cpp
@@ -167,10 +167,10 @@ void BattleAnimation::ProcessAnimationTiming(const RPG::AnimationTiming& timing)
 
 	// Flash.
 	if (timing.flash_scope == RPG::AnimationTiming::FlashScope_target) {
-		SetFlash(Color(timing.flash_red * 255 / 31,
-			timing.flash_green * 255 / 31,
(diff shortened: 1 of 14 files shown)


carstene1ns:
Need to be careful here, you messed up `g` and `b` all over the place.

fmatthew5876:
> Need to be careful here, you messed up `g` and `b` all over the place.

Fixed, Thanks!

fmatthew5876:
I added one more change to fix flash coloring.

See #1768 test case 4. We now match RPG_RT patched with true color support.

PR #1774, opened 2019-05-26
Battle animation sprites affect by screen flash

### LINKED ISSUE #985: HH3: some sounds, animations, actions (events?) not appearing or appearing incorrectly

fmatthew5876:
All this flash sprite stuff is due for a larger refactor, we need to remove that quad and flash the sprites individually. 

That will be a 6.2 thing. I'm putting this one in to fix #985 for 6.1

~~Depends on #1772~~
Fix #985

### DIFF
--- a/src/battle_animation.cpp
+++ b/src/battle_animation.cpp
@@ -59,6 +59,13 @@ DrawableType BattleAnimation::GetType() const {
 }
 
 void BattleAnimation::Update() {
+	auto flash_color = Main_Data::game_screen->GetFlashColor();
+	if (flash_color.alpha > 0) {
+		Sprite::Flash(flash_color, 0);
+	} else {
+		Sprite::Flash(Color(), 0);
+	}
+
 	Sprite::Update();
 
 	if (frame_update) {


Ghabry:
If I see this correctly 3 out of 4 commits are already merged -> Rebase needed

fmatthew5876:
rebased

Ghabry:
@fmatthew5876 to ensure I understand this correctly. This fixes "Shake screen / flash on damage doesn't affect animations. When the enemy uses his physical attack, you can see the last few frames are off as the fake background in the attack animation is offset with the real background"  in the linked issue?

fmatthew5876:
Yes it fixes that.

Shake screen already appears to work for battle animations.

PR #1775, opened 2019-05-26
Render flying events (airship) above all others

### LINKED ISSUE #1683: Above tile events should be drawn below non-above tile events
Issue summary:

- [x] Draw order of above and non-above tile events is wrong
- [x] Airship and shadow are drawn under above events while flying

Create EV01 as `layer=below` with a passable but not above tile graphic.
Create EV02 as `layer=below` with an above tile graphic (passable or not)

Setup another event to place EV01 and EV02 on the same tile.

In RPG_RT, EV01 is drawn ontop of EV02
In Player, EV02 is drawn ontop of EV01

fmatthew5876:
Fix #1683

### DIFF
--- a/src/drawable.cpp
+++ b/src/drawable.cpp
@@ -31,7 +31,7 @@ int Drawable::GetPriorityForMapLayer(int which) {
 		case RPG::SavePicture::MapLayer_tilemap_above:
 			return Priority_TilesetAbove;
 		case RPG::SavePicture::MapLayer_events_above:
-			return Priority_EventsAbove;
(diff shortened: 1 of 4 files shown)


fmatthew5876:
> Not tested but I think this will break rendering of pictures with setting "Above Events above Hero", they should render above the airship.
> Drawable::GetPriorityForMapLayer was not updated.

Good catch. 

* 4: Above tileset above player <- picture below flying airship, above landed airship
* 5: Above events above player <- picture above flying airship

Fixed this.

PR #1776, opened 2019-05-26
Amendment for "Cache improvements and effect caching"

Ghabry:
I failed to implement operator "<" correctly... this breaks the cache lookup and picks incorrect assets in some cases resulting in wrong colors. It uses now std::tie from tuple which is constexpr and offloads the proper operator implementation to the std.

Noticed this in Vampires Dawn, while testing the memory usage :/.
Didn't happen in Yume2kki, hmpf.

Removed system.h include from these files, requires including memory_management now in some.

### DIFF
--- a/src/color.h
+++ b/src/color.h
@@ -18,8 +18,8 @@
 #ifndef EP_COLOR_H
 #define EP_COLOR_H
 
-// Headers
-#include "system.h"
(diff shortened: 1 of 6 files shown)


fmatthew5876:
Good catch, and solution

PR #1779, opened 2019-05-27
 Correct autostart events execution order

fmatthew5876:
If both a common event and a map event are autostart and can
run on the current frame, RPG_RT will push both of them onto
the stack and execute the interpreter. This is very similar to
as if the common event called the map event.

* Rename Setup() to Push()
* Explicitly call Clear() before pushing
* Change CommandEnd() to pop the
  stack instead of just clearing commands.
* Change IsRunning() to just check for empty stack.

~~Depends on: #1740~~
Fix: #1761 all test cases

### DIFF
--- a/src/game_battle.cpp
+++ b/src/game_battle.cpp
@@ -385,7 +385,8 @@ bool Game_Battle::UpdateEvents() {
 
 	for (const auto& page : troop->pages) {
 		if (page_can_run[page.ID - 1]) {
-			interpreter->Setup(page.event_commands, 0);
+			interpreter->Clear();
+			interpreter->Push(page.event_commands, 0);
 			page_can_run[page.ID - 1] = false;
 			page_executed[page.ID - 1] = true;
 			return false;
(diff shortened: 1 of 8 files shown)


carstene1ns:
https://github.com/fmatthew5876/Player/compare/event_state...event_order

fmatthew5876:
rebased to master

Ghabry:
could you add a short explanation to the two "revert" commits message explaining why they were reverted?

fmatthew5876:
They were from the other PR which was a temp fix for 0.6.1 release

fmatthew5876:
I just folded the revert commits into "Correct autostart..."

Hopefully makes the intention more clear.

Ghabry:
This patch breaks the game "TMG v1.2" completely.
I'm still preparing the other issue that was reported via Mail, then more info follows.

fmatthew5876:
Can you give a specific example of breakage? Or I can just start a new game and see everything failing right away?

Ghabry:
oh forgot to continue here :)
Just load the Savegame I linked in the other issue, walk anywhere, you will see the problem, talk to the inn dude, you see the next problem, leave the map to the left, you will see black.

fmatthew5876:
That was due to a rebase error. Fixed Thanks!

Ghabry:
this results in transitions playing twice per custom message box in COLORS but is fixed by #1784 again, so I will let that pass.

Issue #1783, opened 2019-05-29
Preupdate teleport transition test cases


fmatthew5876:
Setup these maps:

Map1

EV01 action=trigger
```
Teleport Map2
```

Ev02 autostart
```
Set EscapePoint Map2
Set TeleportPoint Map2
EraseEvent
```

Map2

EV01 parallel
```
Show Picture
Teleport Map3
```

Map3

No events

1. Start a new game on map 1 and speak to EV01

Result: Transition out, teleport transition in onto **Map3**

![test1_1](https://user-images.githubusercontent.com/1198466/58601363-7304bf00-8256-11e9-8781-29712db7518f.gif)

2. Start a new game on map 1 and cast escape

Result:

Fade Transition into map 2 with picture visible, Transition out, Transition into map 3

![test1_2](https://user-images.githubusercontent.com/1198466/58601368-75ffaf80-8256-11e9-899c-e7fd30755282.gif)

3. Start a new game on map 1 and cast teleport to map 2

Result: Same as (2)

4. Start a new game on map 2

Default transition into **Map3**

![test1_3](https://user-images.githubusercontent.com/1198466/58601371-7ac46380-8256-11e9-8e66-dae0fbc1367b.gif)

Conclusion
=========

Using escape or teleport skill always forces a transition in, unlike teleport commands or starting a new game.

We don't emulate (2) or (3) in player. Doing so is actually very complicated as it adds yet another unique suspend and resume point for async transition code..

fmatthew5876:
Test Cases 2 - Event Transition
===================

Setup these maps:

Map1

EV01 action=trigger
```
Teleport Map2
```

Ev02 autostart
```
Set EscapePoint Map2
Set TeleportPoint Map2
EraseEvent
```

Map2

EV01 parallel
```
Show Screen Blinds
Teleport Map3
```

Map3

No events

1. Start a new game on map 1 and speak to EV01

Result: Blinds transition Into Map2, Transition Into Map3 (with no trans out!), 


2. Start a new game on map 1 and cast escape

Result:

Blinds transition into map 2, transition out, default transition into map 3

3. Start a new game on map 1 and cast teleport to map 2

Result: Same as (2)

4. Start a new game on map 2

Result: Blinds transition into map 2, Instant (!) transition to map 3

fmatthew5876:
Test Cases 3 - Event Transition with 4 maps
===================

Setup these maps:

Map1

EV01 action=trigger
```
Teleport Map2
```

[...]

PR #1782, opened 2019-05-29
Improved support for SetVehicleLocation() teleport..

fmatthew5876:
Fix: #1778

~~Depends on: #1740~~

Test case 4 requires #1779

Issue #1261, opened 2017-09-10
Mondschein: Player speed too high in cutscenes


carstene1ns:
**Name of the game**: Mondschein ([download](http://www.rpg2000.4players.de/index.php?inhalt=download&id=234))

#### Describe the issue in detail and how to reproduce it:

Start the game and watch the intro for one minute, then you run away from your mother... 🏃

![test](https://user-images.githubusercontent.com/4691314/30253019-f3358438-967c-11e7-864c-9f25b6d351f7.gif)
(left is RPG_RT, right is Player)

The same seems to happen later in the game, when you again run away from 3 girls (not verified myself, was reported on google play)

Relevant Map: `5 (erzählung-weg)`. Both the Player and the `[mother]` event have a custom move route containing just "Move Left" commands at frequency 8, nothing special.

Ghabry:
I guess this is caused by a page switch. In that case the moveroute speed is overwritten due to a bug.

carstene1ns:
Another turbo bug happens when stepping on the slide in Grimps - Level 4:
https://youtu.be/-R_NyRr6Sz0?t=9m32s

In Player, the slide does not work correctly and you stop midway while the actor speed is too high afterwards.

Nothing too fancy here about the "ICE" Events:
![screenshot](https://p.f4ke.de/XjhgFvvzpj.png)

Ghabry:
The Mondschein bug is caused by the last move command executed in the event on the map "Schwarzer Bildschirm" which is "Increase Move Speed, Increase Move Speed" followed by a teleport. The move command is not executed on RPG_RT and the move speed stays low. When you put a "Wait for All Movement" before the teleport the fast movement issue is also in RPG_RT.

Ghabry:
via Event Tracer plugin I observed that no Move commands are executed during the first frame on the new map.
Even when not a single event runs on the new map the movement will be one frame late.

Ghabry:
Fixed the problems in Grimps @carstene1ns but that was a completely different bug ^^'

fmatthew5876:
This can be distilled down to a very simple test case

Create 2 maps. 

Map 1 has autostart event
```
SetMoveRoute: Player, Increase Speed, Increase Speed
Teleport to Map 2
```

Map 2 has autostart event
```
SetMoveRoute: Player, Right, Right, Right, Right, Right
```

In RPG_RT, the player walks at normal speed. In Player, the player walks fast.

fmatthew5876:
[...]

PR #1784, opened 2019-05-29
Implement fully async transitions

fmatthew5876:
~~Depends on #1782~~
~~Depends on #1779~~

Fix: #1783
Fix: #1787 
Fix: #1819 
Part of fixing #1261 

This PR implements fully asynchronous transitions which correctly suspend and resume the `Game_Map::Update()` loop. Supports suspending and resuming common events, map events, and foreground events.

Also support the various graphical and timing behaviors of transitions done during pre-update.

### DIFF
--- a/src/game_map.cpp
+++ b/src/game_map.cpp
@@ -921,14 +921,18 @@ void Game_Map::Update(MapUpdateAsyncContext& actx, bool is_preupdate) {
 		UpdateProcessedFlags(is_preupdate);
 	}
 
-	if (!UpdateCommonEvents(actx)) {
-		// Suspend due to common event async op ...
-		return;
+	if (!actx.IsActive() || actx.IsParallelCommonEvent()) {
+		if (!UpdateCommonEvents(actx)) {
+			// Suspend due to common event async op ...
+			return;
+		}
 	}
 
-	if (!UpdateMapEvents(actx)) {
-		// Suspend due to map event async op ...
-		return;
+	if (!actx.IsActive() || actx.IsParallelMapEvent()) {
+		if (!UpdateMapEvents(actx)) {
+			// Suspend due to map event async op ...
+			return;
+		}
 	}
 
 	if (is_preupdate) {
@@ -950,12 +954,16 @@ void Game_Map::Update(MapUpdateAsyncContext& actx, bool is_preupdate) {
 		Main_Data::game_screen->Update();
 	}
 
-	if (!UpdateForegroundEvents(actx)) {
-		// Suspend due to foreground event async op ...
-		return;
+	if (!actx.IsActive() || actx.IsForegroundEvent()) {
+		if (!UpdateForegroundEvents(actx)) {
+			// Suspend due to foreground event async op ...
+			return;
+		}
 	}
 
 	Parallax::Update();
+
+	actx = {};
 }
 
 void Game_Map::UpdateProcessedFlags(bool is_preupdate) {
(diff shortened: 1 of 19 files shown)


fmatthew5876:
The PR is ready for review. I am only fixing transition timing and async logic.

We need additional graphical fixes for transitions, but that can be done in a later PR.

Ghabry:
but here "TMG v1.2" works again but that PR is not rebased yet

fmatthew5876:
Rebased to #1779

src/game_map.cpp
```diff
@@ -926,41 +970,71 @@ void Game_Map::Update(bool is_preupdate) {
...
-	UpdatePan();
+bool Game_Map::UpdateMapEvents(MapUpdateAsyncContext& actx) {
+	int resume_ev = actx.GetMapEvent();
 
-	for (auto& vehicle: vehicles) {
-		if (vehicle->GetMapId() == location.map_id) {
-			vehicle->Update();
+	for (Game_Event& ev : events) {
+		if (resume_ev != 0) {
+			// If resuming, skip all unit the event to resume from ..
```
Ghabry: minor: unit -> until (same in line 981)

Ghabry:
What I don't fully understand is this MapUpdateAsyncContext:
I understand that the purpose is to continue the interpreter exactly where the suspend happened (don't rerun previous events) but lets say I have this situation:

CE1 (Parallel): V1 += 1
EV1 (Parallel): Wait 1.0, Hide Screen, Teleport

Hide Screen will result in a AsyncSuspend. When the execution continues all Common Events will run because MapUpdateAsyncContext common_event is 0. The only reason why they don't execute is that the ``transition_owner`` is set to the event that did the suspend and not due to the MapUpdateAsyncContext. Is that intended? Because then I don't really understand the invention of this AsyncContext when the ``transition_owner`` already prevents any execution of previous events.

fmatthew5876:
https://github.com/fmatthew5876/Player/blob/458ba373eced0731996451fa3fcd0b7f3e021daf/src/game_map.cpp#L982


`Common_Event::GetIndex()`, not the interpreter I'd is used. So it should be suspending and resuming regardless of `transition_owner`. Also we have other async use cases outside of transition. Exit game and to title use it. Later, inn will also use it.

I have not looked at `transition_owner`. Maybe it is not needed anymore.

fmatthew5876:
Checked and `transition_owner` is not needed anymore. Thanks for catching this.

I pushed a commit to remove it.

Ghabry:
With transition_owner removed it exposes now the bug in the AsyncContext I was talking about:

[asynctest.zip](https://github.com/EasyRPG/Player/files/3565027/asynctest.zip)

When the map parallel event does a asyncsuspend the context is "CE 0, Event 3", then when it returns from the async operation it will rerun all common events (which is wrong, they already executed) resulting in a wrong value for the variable that is displayed after the transition.

Ghabry:
Just noticed that the TestGame softlocks when talking to anybody: https://easyrpg.org/play/pr1784/

fmatthew5876:
Thanks for catching this.

The bug was that in each stage (par CE, par ME, FG) I wasn't explicitly checking that the async operation happened at that stage.

Essentially when `GetCommonEvent()` returns 0. It either means there was no async operation at all, or there was an async operation for map or fg events.

I pushed a fix for this.

> EDIT: Where there cases when parallel map events run before parallel common events?


This never happens, common events always run first.

fmatthew5876:
Retested #1819 . Still broken in master, and fixed by this PR.

As we discussed in irc, the emscripten thing is a mystery and we're not sure if it's a bug in the PR or a bug in our emscripten setup.

src/game_map.cpp

fmatthew5876:
I did some other tests to try to make a minimally intrusive change to try to debug the emscripten issue. I had a suspicion about potential timing issues, but even that appears to be not the case. 

So I agree this is probably due to some emscripten issue. We can revisit this in master with latest version of emscripten to see if such issues persist.

So with that, I don't think there is anything else for this PR?

Issue #1770, opened 2019-05-25
Equipment inflict states test cases


fmatthew5876:
Test Cases 1 - Equip and remove + status counter
=============

1. Have an event inflict poison on an actor
2. Save your game - status value is 1
3. Finish a battle
4. Save your game - status value is >= 2
5. Equip an armor which inflicts point
6. Save your game - status value is 1
7. Finish a battle
8. Save your game - status value is >= 2
9. Unequip the armor
10. Poison is removed
11. Save your game - status value is 0

Conclusions
===========

* Equipping armor always resets status counter to 1
* Unequipping always removes the state, giving free infinite healing of states!
* State counter still grows during battle turns

fmatthew5876:
Test Cases 2
==========

Create poison with priority 50, stun with priority 30, both persisting after battle

1. Armor removes states

* Have an event inflict stun
* Save your game -> stun status is 1
* Equip an armor which inflicts poison
* Save game -> Poison status is 1, stun is 0
* Remove poison armor
* Save game -> all states gone

2. Armor has low state, event / armor tries to remove

* Equip armor inflict stun
* Save game -> stun status is 1
* Have an event inflict poison
* Save game -> stun and poison status is 1
* Equip poison armor
* Save game -> stun and poison status is 1

fmatthew5876:
Test Cases 3
=============
Setup confuse as battle only state

1. Battle States by equip

* Equip armor that inflicts confuse
* Save game -> No states
* Start a battle - actor is confused
* Actor attacks self -> stays confused
* Finish the battle
* Save game -> no states

2. Battle states by event

* Have event add confuse
* Save game -> actor is confused
* Start battle - actor is confused
* Finish Battle
* Save game -> no states

Conclusion
==========

* Equipment that inflicts states will not inflict battle states on the map
* Equipment battle states cleared after battle end
* Events can inflict battle states - they cleared after battle ends

fmatthew5876:
Test Cases 4 - Healing Items and Skills
=================

1. Items

* Equip poison armor
* Try to use an antidote
* Menu doesn't allow usage

2. Skills

* Equip poison armor
* Try to use a skill which does not except cure poison 
* Skill is used, sound effect, mp consumed, but poison not healed

fmatthew5876:
Test Cases 5 - events healing states
=============================

1. Events heal states

* Equip poison armor
* Have event heal poison
* Poison still inflicted

2. Events heal battle states

* Have event inflict confuse
[...]

PR #1785, opened 2019-05-29
Refactor Battle2k message timings + state fixes 2 / 2

fmatthew5876:
~~Depends on #1726~~

Fix: #1725 
Fix: #1677
Fix: #1748
Fix: #1770
Fix: #821
Fix: #1490

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -1285,8 +1285,9 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 			}
 
 			if (heals_states) {
+				// RPG_RT 2k3 skills which fail due to permanent states don't "miss"
+				this->success = true;
 				if (State::Remove(state_id, target_states, target_perm_states)) {
-					this->success = true;
 					states.push_back({state_id, StateEffect::Healed});
 				}
 			} else if (Utils::PercentChance(GetTarget()->GetStateProbability(state_id))) {
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -77,6 +77,32 @@ static inline int ToHitPhysical(Game_Battler *source, Game_Battler *target, int 
 	return to_hit;
 }
 
+static void BattlePhysicalStateHeal(Game_Battler* target, int physical_rate, std::vector<Game_BattleAlgorithm::StateEffect>& states) {
+	if (physical_rate <= 0) {
+		return;
+	}
+
+	for (auto state_id: target->GetInflictedStates()) {
+		auto* state = ReaderUtil::GetElement(Data::states, state_id);
+		if (state == nullptr) {
+			continue;
+		}
+		if (state->release_by_damage > 0) {
+			int release_chance = state->release_by_damage * physical_rate / 100;
+
+			if (!Utils::ChanceOf(release_chance, 100)) {
+				continue;
+			}
+
+			if (target->RemoveState(state_id)) {
+				states.push_back(Game_BattleAlgorithm::StateEffect(state_id, Game_BattleAlgorithm::StateEffect::HealedByAttack));
+			}
+		}
+	}
+}
+
+
+
 Game_BattleAlgorithm::AlgorithmBase::AlgorithmBase(Type ty, Game_Battler* source) :
 	type(ty), source(source), no_target(true), first_attack(true),
 	source_restriction(RPG::State::Restriction(source->GetSignificantRestriction()))
@@ -934,11 +960,7 @@ bool Game_BattleAlgorithm::Normal::Execute() {
 			auto target_copy = target->Clone();
 
 			// Conditions healed by physical attack:
-			for (auto state_id: target_copy->BattlePhysicalStateHeal(GetPhysicalDamageRate())) {
-				if (target_copy->RemoveState(state_id)) {
-					states.push_back(StateEffect(state_id, StateEffect::HealedByAttack));
-				}
-			}
+			BattlePhysicalStateHeal(target_copy.get(), GetPhysicalDamageRate(), states);
 
 			// Conditions caused / healed by weapon.
 			if (source->GetType() == Game_Battler::Type_Ally) {
@@ -1230,11 +1252,7 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 
 		// Conditions healed by physical attack:
 		if (!IsPositive() && skill.affect_hp) {
-			for (auto state_id: target_copy->BattlePhysicalStateHeal(GetPhysicalDamageRate())) {
-				if (target_copy->RemoveState(state_id)) {
-					states.push_back(StateEffect(state_id, StateEffect::HealedByAttack));
-				}
-			}
+			BattlePhysicalStateHeal(target_copy.get(), GetPhysicalDamageRate(), states);
 		}
 
 		// Conditions:
@@ -1765,9 +1783,8 @@ bool Game_BattleAlgorithm::SelfDestruct::Execute() {
 		killed_by_dmg = true;
 	}
 
-	for (auto state_id: target->BattlePhysicalStateHeal(GetPhysicalDamageRate())) {
-		states.push_back({state_id, StateEffect::HealedByAttack});
-	}
+	//FIXME: Do we need a clone here?
+	BattlePhysicalStateHeal(target, GetPhysicalDamageRate(), states);
 
 	success = true;
 
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -121,9 +121,7 @@ void Game_BattleAlgorithm::AlgorithmBase::Reset() {
 	absorb = false;
 	revived = false;
 	reflect = -1;
-	conditions.clear();
-	phys_healed_conditions.clear();
-	shift_attributes.clear();
+	states.clear();
 
 	if (!IsFirstAttack()) {
 		switch_on.clear();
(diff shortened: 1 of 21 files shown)


Ghabry:
Good test:
Wadanohara http://vgperson.com/games/wadanohara.htm

[Save04.zip](https://github.com/EasyRPG/Player/files/3244754/Save04.zip)

The 2nd char in the party has a skill called flock which inflicts up to 9 states.

Ghabry:
Review note: The first commit here is "Fixes for physically healed condition"

src/scene_battle_rpg2k3.cpp
```diff
@@ -740,8 +740,6 @@ bool Scene_Battle_Rpg2k3::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBas
 							action->IsPositive() ? Font::ColorHeal : Font::ColorDefault,
 							std::to_string(action->GetAffectedHp()));
 					}
-
-					target->BattlePhysicalStateHeal(action->GetPhysicalDamageRate());
```
Ghabry: > Refactor BattlePhysicalStateHeal 

"Remove call from battle 2k3 that did nothing"

Did this break now autostatehealing in 2k3 or where is this handled?

src/state.cpp
```diff
@@ -0,0 +1,184 @@
+#include "state.h"
```
Ghabry: > Add state functions for manipulating states 

lacks copyright header

src/state.cpp
```diff
@@ -0,0 +1,184 @@
...
+	for (int i = 0; i < (int)states.size(); ++i) {
+		auto state_id = i + 1;
+
+		if (!Has(state_id, states)) {
+			continue;
+		}
+
+		auto* state = ReaderUtil::GetElement(Data::states, state_id);
+		if (state == nullptr) {
+			Output::Warning("State::GetSignificantState: Can't delete state with invalid ID %d", state_id);
```
Ghabry: > Add state functions for manipulating states 

Can't delete state. Msg looks wrong.

src/game_battlealgorithm.cpp
```diff
@@ -1549,15 +1555,15 @@ bool Game_BattleAlgorithm::Item::Execute() {
 			this->sp = std::max<int>(0, std::min<int>(item.recover_sp_rate * GetTarget()->GetMaxSp() / 100 + item.recover_sp, GetTarget()->GetMaxSp() - GetTarget()->GetSp()));
 		}
 
-		// We allocate a copy of the target in order to check if we can apply states..
-		auto target_copy = target->Clone();
+		// Make a copy of the target's state set and see what we can apply.
+		auto target_states = target->GetStates();
```
Ghabry: > Remove Game_Battler::Clone() 

How does this copy? GetStates returns a reference.

src/game_battlealgorithm.cpp
```diff
@@ -1776,8 +1782,11 @@ bool Game_BattleAlgorithm::SelfDestruct::Execute() {
 		killed_by_dmg = true;
 	}
 
-	//FIXME: Do we need a clone here?
-	BattlePhysicalStateHeal(target, GetPhysicalDamageRate(), states);
+	// Make a copy of the target's state set and see what we can apply.
+	auto target_states = target->GetStates();
```
Ghabry: (same here)

src/game_battlealgorithm.cpp
```diff
@@ -78,13 +78,13 @@ static inline int ToHitPhysical(Game_Battler *source, Game_Battler *target, int
 	return to_hit;
 }
 
-static void BattlePhysicalStateHeal(int physical_rate, std::vector<int16_t>& target_states, std::vector<Game_BattleAlgorithm::StateEffect>& states) {
+static void BattlePhysicalStateHeal(int physical_rate, std::vector<int16_t>& target_states, const PermanentStates& ps, std::vector<Game_BattleAlgorithm::StateEffect>& states) {
```
Ghabry: >  Add PermanentStates concept 

(Flagging the whole commit). Is there any sanity check for PermanentStates? What happens when a weapon inflicts a state that is out-of-bounds?
Looks filtered in next commit "Support equipment inflicting states "?

src/game_battler.cpp
```diff
@@ -531,7 +533,7 @@ void Game_Battler::ChangeHp(int hp) {
 
 		// Death
 		if (GetHp() <= 0) {
-			AddState(RPG::State::kDeathID);
+			AddState(RPG::State::kDeathID, true);
```
Ghabry: > AddState - add allow_battle_states

Why does this recursive call use ``true```instead of ``allow_battle_states``

src/game_battler.cpp
```diff
@@ -427,8 +427,16 @@ bool Game_Battler::AddState(int state_id, bool allow_battle_states) {
 	return was_added;
 }
 
-bool Game_Battler::RemoveState(int state_id) {
-	auto was_removed = State::Remove(state_id, GetStates(), GetPermanentStates());
+bool Game_Battler::RemoveState(int state_id, bool always_battle_states) {
```
Ghabry: >  RemoveState event commands remove battle equip states 

Could you give the var a better name, like ``[always_]remove_battle_states``

src/game_battlealgorithm.cpp
fmatthew5876: I say `auto x = ...` not `auto& x =` I am making a copy locally.

Ghabry: ah, don't know how auto deduces the type, thought it just takes the return type.

fmatthew5876: `auto` drops references. If you use `decltype(auto)` it will retain references.

src/game_battler.cpp
carstene1ns: Wrong function :see_no_evil:
(Easy to mess up with the omission arrows)

src/game_battler.cpp
fmatthew5876: Fixed

src/state.cpp
fmatthew5876: Fixed

src/state.cpp
fmatthew5876: Fixed

src/scene_battle_rpg2k3.cpp
fmatthew5876: This method returns a vector of conditions to heal from a physical attack. I have no idea why it's here in 2k3 battle, as it just computes this vector and does nothing with it.

Physical states healing is all setup properly in `Game_BattleAlgorithm`.

Auto state healing is a different concern. I haven't checked where that is in 2k3, but this change doesn't affect that either way.

src/game_battlealgorithm.cpp
fmatthew5876: The warnings will show up when the equipment tries to add the states.

I added a new commit to prevent adding these invalid states to the `PermanentStates` object.

fmatthew5876:
Rebased and addressed review comments

Ghabry:
@fmatthew5876 
I noticed in RPG\_RT that "Full Recovery" inflicts battle equipment states even on the map.
Player doesn't do this and I consider the RPG\_RT behaviour a bug, your opinion?

fmatthew5876:
> @fmatthew5876
> I noticed in RPG_RT that "Full Recovery" inflicts battle equipment states even on the map.
> Player doesn't do this and I consider the RPG_RT behaviour a bug, your opinion?

I guess we need to emualte this for completeness? If I get around to it, I'll add it in another PR. Don't want to delay this one for such an edge case.

Ghabry:
Also happens when using the Inn.
Nice so I go to bed and wake up with Poison and Silence.
Not sure if that should be even emulated, to get rid of these bogus states you can just reequip the equipment in the equipment scene... Because they are battle states they are not reeinflicted :)

fmatthew5876:
Well technically the effect should be benign, since they are battle states they aren't supposed to affect the map. Once you start a battle, you'll get them regardless.

I'm in favor of emulating this. The feature in general is pretty broken, but maybe some game is relying on these edge case behaviors?

Ghabry:
```AddressSanitizer:DEADLYSIGNAL
=================================================================
==16125==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000064 (pc 0x560fc702240d bp 0x7ffeaa073dd0 sp 0x7ffeaa073a00 T0)
==16125==The signal is caused by a READ memory access.
==16125==Hint: address points to the zero page.
    #0 0x560fc702240c in Scene_Equip::UpdateStatusWindow() /home/gabriel/Programmierung/easyrpg/easyrpg-player/src/scene_equip.cpp
    #1 0x560fc7021804 in Scene_Equip::Update() /home/gabriel/Programmierung/easyrpg/easyrpg-player/src/scene_equip.cpp:62:2
    #2 0x560fc6ba0ec2 in Player::Update(bool) /home/gabriel/Programmierung/easyrpg/easyrpg-player/src/player.cpp:309:21
    #3 0x560fc6befd66 in Scene::MainFunction() /home/gabriel/Programmierung/easyrpg/easyrpg-player/src/scene.cpp:119:4
    #4 0x560fc6ba04b4 in Player::MainLoop() /home/gabriel/Programmierung/easyrpg/easyrpg-player/src/player.cpp:225:19
    #5 0x560fc6ba0222 in Player::Run() /home/gabriel/Programmierung/easyrpg/easyrpg-player/src/player.cpp:219:3
    #6 0x560fc6b9c996 in main /home/gabriel/Programmierung/easyrpg/easyrpg-player/src/main.cpp:28:2
    #7 0x7f71ec73ace2 in __libc_start_main (/usr/lib/libc.so.6+0x23ce2)
    #8 0x560fc6a8402d in _start (/tmp/easyrpg/cmake-debug/easyrpg-player+0x1d802d)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /home/gabriel/Programmierung/easyrpg/easyrpg-player/src/scene_equip.cpp in Scene_Equip::UpdateStatusWindow()
==16125==ABORTING
```

No idea if this is a regression caused by this PR, probably not.

Have one equippable item in the inventory. Scroll the cursor to the right (on the empty unqeuip slot) and it crashes when highlighted.

Ghabry:
I think this is caused by this PR, is in new code and the problem is ``current_item->two_handed``. current_item is NULL.

fmatthew5876:
Nice catch, added a fix and rebased

fmatthew5876:
Added the full recover and inn equipment states stuff.

Issue #1791, opened 2019-06-02
RPG2k: Rendering order of enemies is wrong


Ghabry:
Wadanohara http://vgperson.com/games/wadanohara.htm

Go through the dialog and start the battle (should be ID 13). THe Z order of the enemies is wrong:

![Screenshot_20190602_135110-fs8](https://user-images.githubusercontent.com/1331889/58760838-9624bc80-853d-11e9-988c-12db9c34cd72.png)
![Screenshot_20190602_135120-fs8](https://user-images.githubusercontent.com/1331889/58760839-9624bc80-853d-11e9-9cf9-ea13cf9c3ca4.png)

[Save04.zip](https://github.com/EasyRPG/Player/files/3244754/Save04.zip)

Here is a save earlier in the dialog which works in RPG_RT:

[Save02.zip](https://github.com/EasyRPG/Player/files/3244758/Save02.zip)

**Save02 doesn't load in RPG\_RT**

fmatthew5876:
So the Z order priority (front to back) is this:

1. Enemy Y coordinate - lower is in front
2. Enemy ID - lower id is in front
3. Actor Y coordinate - lower is in front (2k3)
4. Actor ID - lower id is in front (2k3)

PR #1792, opened 2019-06-03
Fix Battler Z ordering

fmatthew5876:
Fix: #1791

### DIFF
--- a/src/sprite_battler.cpp
+++ b/src/sprite_battler.cpp
@@ -26,8 +26,8 @@
 #include "reader_util.h"
 #include "output.h"
 
-Sprite_Battler::Sprite_Battler(Game_Battler* battler) :
-	battler(battler) {
+Sprite_Battler::Sprite_Battler(Game_Battler* battler, int index) :
+	battler(battler), battle_index(index) {
 }
 
 Sprite_Battler::~Sprite_Battler() {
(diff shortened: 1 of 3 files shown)


Ghabry:
There is already a SetZ call here: https://github.com/EasyRPG/Player/blob/51c2598d0d7139532519120727d547f7368ad4f4/src/sprite_battler.cpp#L310

Can Z change during the battle, in the old code it is only set once on creation.

src/game_battler.cpp
```diff
@@ -785,6 +785,21 @@ int Game_Battler::GetDisplayY() const {
...
 }
 
+int Game_Battler::GetDisplayZ() const {
+	static_assert(Game_Battler::Type_Ally < Game_Battler::Type_Enemy, "Game_Battler enums re-ordered! Fix Z order logic here!");
+
+	constexpr int id_limit = 128;
+
+	int z = GetType();
+	z *= SCREEN_TARGET_HEIGHT * 2;
+	z += GetDisplayY();
```
carstene1ns: Might be better to use `GetBattleY()` here, since it is unaffected from flying.

fmatthew5876:
The use of `GetId()` is wrong. For enemies it returns the enemy id, not the troop id.

For the algorithm, enemies use troop id, but actors always use actor id, even if the part is ordered differently.

fmatthew5876:
Tested this more thoroughly and pushed new fixes

* For enemies, Y position considered is actually the bottom edge of the sprite, not the centered y position
* Flying offset is ignored
* Actors used actor id, regardless of party order
* Enemies use troop id

For actors, I ignored graphic size, since all actors have same size graphics and the graphic isn't properly initialized at the time of the call.

Issue #1722, opened 2019-03-22
Battle 2k3: Weapons don't invoke skills


s9060666:
Thank you for supporting EasyRPG Player.
Before creating an issue, please verify that your bug still occurs in the "continuous builds" of the Player.
You can get them at the bottom of our download section: https://easyrpg.org/player/downloads

---
Please fill in the following fields before submitting an issue:
#### Player platform:Android
Windows, Linux, Android, ...

#### Attach files (as a .zip archive or link them)
 - A savegame next to the problem (if it is not at the beginning of the game)
 - The `easyrpg_log.txt` log file (you can find it in the game directory)

#### Describe the issue in detail and how to reproduce it:
Use item：白虎杖
Normal use magic：狂風IV
After using the item, nothing happens. Calculate the round directly.

Ghabry:
Download link: http://www.cooltey.org/ping/game_page.php?id=mbl1

Ghabry:
This is basicly "Weapons that invoke skills (When you use them as an item) don't work"

fmatthew5876:
Weapons invoking skills is already fixed in master

Ghabry:
doesn't work for me

fmatthew5876:
Ok I'll have to check this one out. I tested a weapon which invokes healing magic in a test game and it worked. I didn't test this game specifically.

Ghabry:
I also didn't test the game, as I couldn't find the item ^^'.

I used a test game and just changed Club to do Poison.

fmatthew5876:
Ok I'm really confused now.

![pswd](https://user-images.githubusercontent.com/1198466/58375470-b9d67a00-7f21-11e9-8aa4-ed158e6061ed.png)

This works for me. I can use the item and do poison attack. I also tried to change poison attack to not do damage and only inflict poison, still works.

How is yours different?

Ghabry:
I found the problem.
It only fails when the item is not in the inventory but only equipped (when equipped and has a skill it will still appear in the item list as x1 for this actor)

fmatthew5876:
Ok, so I did that. Equipped my "POISON SWD". The actor with it equipped can use it, the other actor it doesn't appear.

So still I say it works for me!

What else is different in your case?

s9060666:
@fmatthew5876 
I don't know, I uploaded my log file to give you a try.
[...]

PR #1797, opened 2019-06-06
Fix Weapons invoking skills when only 1 of weapon and it's equipped

fmatthew5876:
Also includes a cleanup of item skill logic in scene_battle

Fix: #1722

### DIFF
--- a/src/scene_battle.cpp
+++ b/src/scene_battle.cpp
@@ -42,8 +42,7 @@
 
 Scene_Battle::Scene_Battle() :
 	actor_index(0),
-	active_actor(NULL),
-	skill_item(NULL)
+	active_actor(NULL)
 {
 	Scene::type = Scene::Battle;
 	Game_Temp::battle_result = Game_Temp::BattleAbort;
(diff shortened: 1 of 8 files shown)


src/scene_battle.cpp
```diff
@@ -264,19 +271,27 @@ void Scene_Battle::EnemySelected() {
...
+				|| (item->use_skill && (item->type == RPG::Item::Type_weapon
+						|| item->type == RPG::Item::Type_shield
+						|| item->type == RPG::Item::Type_armor
+						|| item->type == RPG::Item::Type_helmet
+						|| item->type == RPG::Item::Type_accessory)))
+		{
+			const RPG::Skill* skill = ReaderUtil::GetElement(Data::skills, item->skill_id);
 			if (!skill) {
-				Output::Warning("AllySelected: Item %d references invalid skill %d", skill_item->ID, skill_item->skill_id);
+				Output::Warning("EnemySelected: Item %d references invalid skill %d", item->ID, item->skill_id);
```
Ghabry: AllySelected

fmatthew5876: Fixed, thanks

Issue #1799, opened 2019-06-06
Inventory is not sorted


fmatthew5876:
- [ ] Ensure inventory items are added and removed in sorted order
- [ ] Ensure battle items are sorted, including equipment added to the list due to being equipped but not in inventory
- [ ] Handle broken Player saves that aren't sorted (in RPG_RT compatible way??)

Create this event
```
Add Item: LongSword x1
Add Item: Potion x1
```

In RPG_RT, the inventory will be `Potion`, `Long Sword`
In Player, the inventory will be `Long Sword`, `Potion`

The lsd chunks for `item_ids` etc.. will also reflect this ordering.

A good example of this is in #1722 

The save game was created in player, and so the inventory in the LSD chunks is not sorted.

If you load this save in Player, the menu inventory won't be sorted. When you start a battle, the weapon described in the bug will be the last item.

If you load this save in RPG_RT, the menu inventory also won't be sorted! That's probably because RPG_RT just loads the LSD directly. It remains to be seen what happens when you add items to an unsorted inventory. Where do they get added? Will RPG_RT do a broken binary search or something else? :exclamation: 

When you start a battle in RPG_RT, all of the sudden the items in battle are now sorted. This includes the weapon in question, which appears in it's proper sorted location.

PR #1800, opened 2019-06-06
Sort the inventory

fmatthew5876:
Fix: #1799 

Depends: #1798 
Depends: #1797

### DIFF
--- a/src/game_party.cpp
+++ b/src/game_party.cpp
@@ -87,17 +87,12 @@ void Game_Party::GetItems(std::vector<int>& item_list) {
 		item_list.push_back(*it);
 }
 
-int Game_Party::GetItemCount(int item_id) {
-	if (item_id > 0) {
-		for (int i = 0; i < (int) data().item_ids.size(); i++)
-			if (data().item_ids[i] == item_id)
-				return data().item_counts[i];
-	}
-
-	return 0;
+int Game_Party::GetItemCount(int item_id) const {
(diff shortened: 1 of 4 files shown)


fmatthew5876:
Rebased to maste

src/game_party.cpp
```diff
@@ -707,3 +700,9 @@ Game_Actor* Game_Party::GetHighestLeveledActorWhoCanUse(const RPG::Item* item) c
 	}
 	return best;
 }
+
+std::pair<int,bool> Game_Party::GetItemIndex(int item_id) const {
+	auto& ids = data().item_ids;
+	auto iter = std::lower_bound(ids.begin(), ids.end(), item_id);
+	return std::make_pair(iter - ids.begin(), (iter != ids.end() && *iter == item_id));
```
Ghabry: Why do you use std::lower_bound and not std::find here when the goal is to find the item with the specific id.

fmatthew5876: lower_bound does a binary search, which find the item in lg(N) time instead of linear time for find.

Ghabry: and std::binary_search returns bool instead of an iterator, hmpf.

fmatthew5876: Oh yeah, that one is kind of useless.. Unless all you want to do is ask "Does this thing exist in the set"? Which is rare.

Issue #1802, opened 2019-06-07
Alter A.I.L.A. Genesis: Game stays black after transition (Regression)


Ghabry:
**Game:** Alter A.I.L.A. Genesis https://rpgmaker.net/games/272/downloads/2215/

**Platform:** Windows/Android, ...

I'm pretty sure this worked before.

This seems to be fixed by #1784 but I would like to retest this when the PR is rebased as a recent change could have broken it.

Just go through the intro, afterwards the screen is black, open the menu and return to the map to fix it. Happens everytime the map is changed.

PR #1803, opened 2019-06-08
Fix erased by event logic on show screen during preupdate

fmatthew5876:
We don't yet emulate transitions that happen on preupdate,
but we should disable the erased by screen flag so that normal
fade in and out still works.

Fix: #1802


fmatthew5876:
Fixed the commit comment

fmatthew5876:
jenkins: test this please

Issue #1804, opened 2019-06-08
States that heal while walking not supported (inflict damage instead)


Ghabry:
This is a 2k3 feature.

When for a state HP/MP is set to "[*] Increase" then the _map movement increases_ the HP/MP, 
not decreases, also there is _no red screen_ flash.

Currently Player ignores the flag and just decreases. _In battle the flag is correctly implemented._

PR #1805, opened 2019-06-08
Map Stepping +/- HP/SP states fixes

fmatthew5876:
Fix: #1804
Fix: #1806


fmatthew5876:
Changed it to `else if`

fmatthew5876:
I added some more commits to fix #1806

Issue #1750, opened 2019-04-18
Battle 2k - Handle enemies with no actions


fmatthew5876:
Enemies with an empty action list don't process their states correctly and don't animate or have correct message delays in #1726. I haven't verified, but I think they also don't naturally heal states like they should.

For enemies with empty action list:
- [x] If no states - the enemy will not flash and there are no delays
- [x] If has states with no message - no flash and no delays
- [x] If has a state with message - enemy flashes and state message printed
- [x] If healed a state - enemy flashes and message printed

More detailed samples in #1677 Test cases 6 & 7

PR #1810, opened 2019-06-09
Battle2k: Handle enemies with no actions + Death timing fixes

fmatthew5876:
Fix: #1750


fmatthew5876:
This has been tested and is now ready for review and merge.

fmatthew5876:
I didn't realize it at the time, but this also fixes state processing for enemies that only have skill actions but don't have enough SP to use any of those skills.

PR #1811, opened 2019-06-10
Allow escaping in 2k3 battles

fmatthew5876:
* Allowed if event allows and is not surprise or pincer


fmatthew5876:
Looks like this is not quite correct. I am going to close this one for now.

Issue #1813, opened 2019-06-12
Request: Adding maps to Debug Mode


Kupo91:
Ye the title says it all

Just an idea, dont know if its possible

The PSone game Grandia has this

fdelapena:
For reference:

https://grandia.fandom.com/wiki/Debug_Mode_(Grandia)

Ghabry:
Ah you mean a map selector. Well, technically this is possible.

PR #1817, opened 2019-06-13
Debug menu improvements + Teleport to maps

fmatthew5876:
Fix: #1813 

Also adds some usability improvements to remember indices.

This code is nasty and due for a refactor. I would advise just testing it instead of trying to review some of these diffs...

### DIFF
--- a/src/scene_map.cpp
+++ b/src/scene_map.cpp
@@ -50,6 +50,7 @@ Scene_Map::~Scene_Map() {
 }
 
 void Scene_Map::Start() {
+	Scene_Debug::ResetPrevIndicies();
 	spriteset.reset(new Spriteset_Map());
 	message_window.reset(new Window_Message(0, SCREEN_TARGET_HEIGHT - 80, SCREEN_TARGET_WIDTH, 80));
 
--- a/src/scene_map.cpp
+++ b/src/scene_map.cpp
@@ -50,7 +50,7 @@ Scene_Map::~Scene_Map() {
 }
 
 void Scene_Map::Start() {
-	Scene_Debug::ResetPrevIndicies();
+	Scene_Debug::ResetPrevIndices();
 	spriteset.reset(new Spriteset_Map());
 	message_window.reset(new Window_Message(0, SCREEN_TARGET_HEIGHT - 80, SCREEN_TARGET_WIDTH, 80));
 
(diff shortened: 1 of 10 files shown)


Ghabry:
> This code is nasty and due for a refactor.

Haha. I wanted to add something to that scene one week ago and thought the same. The code is really hard to follow/extend. :/

Ghabry:
guess this doesn't work in emscripten? In the web player I only get a blackscreen while testing. ( https://easyrpg.org/play/pr1817/?test-play )

Trying it again locally later.

fmatthew5876:
Gaps are supported. You'll see an empty entry in the list and it'll do nothing when you select it.

Area maps appear in the list but are disabled

fmatthew5876:
Added support for Full Heal to the Debug Menu.

I want to add call common event, but that's going to be in 6.2 as it comes after #1779 .

This PR is finished.

fmatthew5876:
Regarding this black screen on emscripten, I see this too.

Maybe it's an async bug related to enabling teleports from scenes other than the map? Do we have any test game setup with quick access to an Escape or Teleport spell?

I suspect this might be a more general issue and if so, they would exhibit the same problem.

fmatthew5876:
emscripten issue is fixed for debug menu and escape/teleport skills.

fmatthew5876:
In chat @fdelapena commented that this PR doesn't handle map bounds. Unfortunately I don't think that can be done easily as we actually have to load the lmu file to know the bounds, which would incredibly slow on normal platforms and basically impossible on emscripten.

It's a debug feature, so I guess not a huge problem for now.

Ghabry:
Minor suggestion: When Full Heal is invoked play SFX_UseItem

fmatthew5876:
> 
> 
> Minor suggestion: When Full Heal is invoked play SFX_UseItem

Good idea, added this

src/scene_map.cpp
```diff
@@ -50,6 +50,7 @@ Scene_Map::~Scene_Map() {
 }
 
 void Scene_Map::Start() {
+	Scene_Debug::ResetPrevIndicies();
```
Ghabry: typo: Indices

fmatthew5876: Fixed thanks

PR #1829, opened 2019-07-07
Remove emscripten virtual hack from Game_Character destructor

fmatthew5876:


fmatthew5876:
Guess this workaround was old and no longer needed?

fmatthew5876:
Folded this into #1828

Ghabry:
that was some compiler bug which seems to be fixed by now.

PR #1832, opened 2019-07-08
Add support for calling events from debug scene + Code Refactor

fmatthew5876:
~~Depends on: #1784~~

TBD:
- [x] Run common events on FG map interpreter
- [x] Run common events on FG battle interpreter

TBD later Pr's:

- [ ] Run common events on Parallel interpreter in map
- [ ] Run map events on FG map interpreter
- [ ] Run map events on Parallel interpreter in map
- [ ] Figure out save game incompatibility with forced parallel execution

When an event is forced onto the foreground interpreter, its like a call event was executed. This works as expected and survives saving and loading since the `forground_event_execstate.stack` will have the event commands pushed to it.

Saving and loading will break if an event is forced to run in parallel when it normally shouldn't. While we can push the event commands onto `parallel_event_execstate.stack`, there is no LSD chunk to save the fact that we want to force execution to happen even if the controlling conditions are not met (or the event trigger isn't parallel).

We could add a custom chunk for this, or just let save games break here since this is a debug feature.

### DIFF
--- a/src/scene_debug.cpp
+++ b/src/scene_debug.cpp
@@ -56,6 +56,7 @@ struct Scene_Debug::PrevIndex {
 	IndexSet item;
 	IndexSet troop;
 	IndexSet map;
+	IndexSet event;
 };
 
 Scene_Debug::PrevIndex Scene_Debug::prev = {};
@@ -148,6 +149,12 @@ void Scene_Debug::Update() {
 			case eFullHeal:
 				ReturnToMain(8);
 				break;
+			case eCallEvent:
+				CancelListOption(prev.event, 9);
+				break;
+			case eCallEventSelect:
+				CancelListOptionSelect(eCallEvent, prev.event);
+				break;
 			}
 	} else if (Input::IsTriggered(Input::DECISION)) {
 		switch (mode) {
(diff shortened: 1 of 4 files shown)


Ghabry:
ah I guess the MsgBox call event deadlock in emscripten is the same issue as with the other PR...
need to test this locally ^^'

fmatthew5876:
> Full Heal is broken, shows text from Call Event

Fixed

> Call Event softlocks for me. https://easyrpg.org/play/pr1832/?test-play Use event "Call Test" (should show a messagebox).

Can not reproduce outside of emscripten

> Call Event does not "PopUntil" the Map. WHen in TestPlay mode there is a Debug option in the regular menu, in this case you only return to the Menu after an event call, not to the map.

Fixed

Thanks!

fmatthew5876:
I am cutting this one here. There is more powerful we can do with calling events but I'll make a new PR for that when I find time to do it later.

This one is useful on its own and can get merged to not let the code get stale.

fmatthew5876:
The `scene_debug` refactor causes a lot of rebase pain. Would be good to get this in if we can since it's approved already.

Ghabry:
tested everything again, works for me

Issue #1753, opened 2019-04-21
0.6.0 source tarball doesn't build, missing files


akien-mga:
The EasyRPG Player 0.6.0 source tarball provided on the website cannot be used to build from source: https://easyrpg.org/downloads/player/0.6.0/easyrpg-player-0.6.0.tar.gz

Console-specific files have been excluded when generating it, but they're still listed in CMakeLists.txt, so the build fails:

```
$ cmake .
-- The CXX compiler identification is GNU 8.3.1
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Check if the system is big endian
-- Searching 16 bit integer
-- Looking for C++ include sys/types.h
-- Looking for C++ include sys/types.h - found
-- Looking for C++ include stdint.h
-- Looking for C++ include stdint.h - found
-- Looking for C++ include stddef.h
-- Looking for C++ include stddef.h - found
-- Check size of unsigned short
-- Check size of unsigned short - done
-- Using unsigned short
-- Check if the system is big endian - little endian
-- Found liblcf: /usr/lib64/liblcf.so (found version "0.6.0") 
-- Found ZLIB: /usr/lib64/libz.so (found version "1.2.11") 
-- Found PNG: /usr/lib64/libpng.so (found version "1.6.36") 
-- Found Pixman: /usr/lib64/libpixman-1.so  
-- Found SDL2: /usr/lib64/libSDL2.so  
-- Found Freetype: /usr/lib64/libfreetype.so (found version "2.9.1") 
-- Found Harfbuzz: /usr/lib64/libharfbuzz.so  
-- Found SDL2_mixer: /usr/lib64/libSDL2_mixer.so  
-- Could NOT find speexdsp (missing: SPEEXDSP_LIBRARY SPEEXDSP_INCLUDE_DIR) 
-- Could NOT find mpg123 (missing: MPG123_LIBRARY MPG123_INCLUDE_DIR) 
-- Could NOT find LibSndFile (missing: LIBSNDFILE_LIBRARY LIBSNDFILE_INCLUDE_DIR) 
-- Found Vorbisfile: /usr/lib64/libvorbisfile.so  
-- Found Opusfile: /usr/lib64/libopusfile.so  
[...]

PR #1835, opened 2019-07-14
Add missing dist files

fdelapena:
Towards a proper fix for #1753, workaround for now with the traditional way so far, add guard headers to keep compiler building properly for other ports.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -7,26 +7,34 @@ bin_PROGRAMS = easyrpg-player
 
 noinst_LIBRARIES = libeasyrpg-player.a
 libeasyrpg_player_a_SOURCES = \
+	src/3ds_ui.cpp \
+	src/3ds_ui.h \
 	src/async_handler.cpp \
 	src/async_handler.h \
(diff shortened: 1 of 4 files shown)


Ghabry:
Don't think you need extra header guards in the header.
The header is never included so the compiler won't complain about missing symbols.

fmatthew5876:
> Don't think you need extra header guards in the header.
> The header is never included so the compiler won't complain about missing symbols.

Is it? If you include `3ds_ui.cpp` in the automake sources for example then I think `3ds_ui.h` gets included.

fdelapena:
That cpp file has the include guarded, so it's safe in that particular case.

I saw the guard at `audio_psp2.h` (also at `audio_libretro.h`), so it also tried to keep some consistency, but I could drop the commit anyways.

Ghabry:
> I saw the guard at

Hm, you are right, we are just not consistent there. Keep them.

Ghabry:
oh completely forgot this one, works for me with CMake after make dist :+1:

PR #1842, opened 2019-07-17
Refactor game system

fmatthew5876:
* RPG_RT compatible LSD chunks
* If graphics_name.empty(), message_stretch and font_id are 0
  and we fallback to database
* If !graphics_name.empty(), all 3 parameters are taken literally

* Refactor the Game_System Set() interface to uphold these invariants
* Refactor the Game_System Get() interface to fallback

* Encapsulate graphic change logic within the Cache to simplify users of
system graphics.

* Reset system graphic when returning the to the title screen.

Depends on: https://github.com/EasyRPG/liblcf/pull/333

Tests:

- [ ] Test async cases with emscripten
- [ ] Test multiple change system commands on same frame behavior.
- [x] No System Graphic - Title Screen
- [x] No System Graphic - Show Message
- [x] No System Graphic - Timer
- [x] No System Graphic - Airship Shadow
- [x] No System Graphic - Shop
- [x] No System Graphic - Main Menu
- [x] No System2 Graphic - Battle types A, B, and C

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -57,6 +57,8 @@ namespace {
 	std::string system_name;
 	BitmapRef default_system;
 
+	std::string system2_name;
+
 	constexpr int cache_limit = 10 * 1024 * 1024;
 	size_t cache_size = 0;
 
@@ -498,6 +500,10 @@ void Cache::SetSystemName(std::string const& filename) {
 	system_name = filename;
 }
 
+void Cache::SetSystem2Name(std::string const& filename) {
+	system2_name = filename;
+}
+
 BitmapRef Cache::System() {
 	if (!system_name.empty()) {
 		return Cache::System(system_name);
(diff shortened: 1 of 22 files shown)


Ghabry:
> If I do multiple async file requests, are they guaranteed to callback in order they were requested?

It is guaranteed that only the last registered callback is called but you broke the callback logic.

As you can see in the code you deleted ``request_id`` is an instance variable and not a local variable. The ID must stay valid, as this is implemented with shared and weak ptr.

See the documentation of Bind:
```
	 * Binds a member function as request-finished event handler.
	 * The event handler is only invoked once (one-shot).
	 * The caller must hold a reference to the returned object, otherwise the
	 * handler is unbound.
```

fmatthew5876:
rebased to master and liblcf #333

Ghabry:
Well this was supposed to be merged right after the lcf build.

PR #1843, opened 2019-07-17
Include player version in game browser

fmatthew5876:


Ghabry:
test case (PR builder doesn't set version info): https://easyrpg.org/play/pr1843/?game=none

PR #1844, opened 2019-07-17
Fixes for Empty Shops - Fix crash and graphical errors.

fmatthew5876:


fmatthew5876:
Added a fix for https://community.easyrpg.org/t/monigote-fantasy-several-issues/638/16

PR #1846, opened 2019-07-19
EasyRPG Player 0.6.1 - Surprise Attack

Ghabry:
output


Ghabry:
Okay nothing release blocking bug-wise. We can release :+1:

PR #1847, opened 2019-07-19
Minor fixes before release

Ghabry:
~~Don't merge yet,~~ just some minor things that don't hurt to add.


src/game_map.cpp
```diff
@@ -193,7 +193,7 @@ void Game_Map::Setup(int _id) {
 			break;
 		}
 		if (parent_index < 0) {
-			Output::Warning("Map %d has invalid parent id %d!", Data::treemap.maps[current_index].parent_map);
+			Output::Warning("Map %d has invalid parent id %d!", current_index, Data::treemap.maps[current_index].parent_map);
```
Ghabry: @fmatthew5876 
I'm not 100% sure if the variable I added here is the correct one.

Ghabry:
Thats now ready for review but I still need to check if the ExFont log message is really fixed on Android.

src/game_map.cpp
fmatthew5876: You want the id of the map, not the index since maps can have gaps.

`Data::treemap.maps[current_index].ID`

Ghabry: thx, changed it

Ghabry:
found a way to trigger a assert in my RTP code, also fixing this

Ghabry:
somehow logging is unreliable on my Android, sometimes the file isn't written o.O
But I was able to test it, works as expected now. Also tested RTP detection.

Ghabry:
also fixed the UBSan stuff now, nothing big there.

fmatthew5876:
All looks good to me. Lets merge this sucker and get the party started!

Issue #1854, opened 2019-07-27
Multi-Game Save Importing


sorlok:
Hey all,

Some RPG Maker games allowed importing save files from previous versions (e.g., Legion Saga 1 -> Legion Saga 2) or were episodic (e.g., "The Way", Episodes 1->6). They accomplished this by asking players to copy save files from the old game's folder to the new game's folder, using a specific Map ID to synchronize the load.

I think it would be nice to have built-in support for this, since copying files around is not trivial on some platforms. I wrote some sample code to accomplish this:
https://github.com/sorlok/Player/commit/be26cd56b21d15ccd073bdc00a2a305937b6fbdf

This requires that a metadata file be added to the game folder with (a) the name of the folder containing the game to import from and (b) the ID of the map to use as a pivot point. If that file exists, it adds an "Import Save" option to the title screen:
https://i.postimg.cc/rpZ4fykx/easy-rpg-import-1.png

This menu option opens the regular "Continue Game" window on the import folder, and filters save files based on the map ID:
https://i.postimg.cc/HxJyRGYK/easy-rpg-import-2.png

The code functions, but is not release-quality; it's meant more as a discussion point on Importing. I was wondering if the Player developers were interested in supporting multi-game save importing (or were even aware that games did this), and if this approach seems like the right way or not.

Ghabry:
Hello, currently is summer vacation time so don't expect too much feedback during the next weeks. :)

I only scrolled quickly through your code yet without giving any feedback on it but in general we accept enhancements when they can't break existing games.
At least the "Import Savegame" function is not visible in games that lack the special file, so it won't annoy in any other games.

I also wonder if instead of the Title scene maybe it fits better in the game browser (but extra work for platforms which have a custom game browser like Android e.g.)

sorlok:
Thanks for the feedback! I'm thinking about your question regarding who provides the file; will get back to you on this.

What's the "Game Browser" you mentioned? I've only used the EasyRPG Player from the game folder directly...

Finally, two more questions: 
1) Is there a preferred data format for config files? I.e., XML or JSON or CSV, etc.
2) Is there any other per-game metadata that already exists in EasyRPG-land? (In other words, do I really need a dedicated file for this, or is there some other file I can add this to?)

Thanks!

Ghabry:
> What's the "Game Browser" you mentioned
When you put the Player one level above the game folder a game listing will appear.

> Is there any other per-game metadata that already exists in EasyRPG-land?

[...]

Issue #1210, opened 2017-06-28
Feature Request: Spiel erkennung für passende Einstellungen für patches etc.


rygos:
In some cases it is good to know which game has just started. 
For example, specific settings can be set which might be necessary for some specific patches. This way, the player would not have to guess which is the right approach for the game. This was done, for example, with the Amiga Emulator Frontend Lemonade (http://www.lemonamiga.com/lemonade/). He has recognized the game and set the appropriate settings for the emulator.

In tests, it turned out to be useful to hash the rpg_rt. * And map files, and to determine a "global" hash.
Thanks to Demetrius for this suggestion.

This would allow you to query an optional external or internal database and get the appropriate information.

Further thoughts: The player could display title images or description texts in the game selection screen if several games are in the player directory.

Ghabry:
Now with #1860 in the queue we should decide about some database format for games.

I talked about this with @rygos before, the author of rmarchiv.tk. The website was perfect as a base for testing this because it has hundreds of machine readable RPG Maker 2000/2003 game archives.

We decided on using a hash function for this.
The fastest approach is probably hashing the LMT first because it is much smaller but I can't remember how many conflicts this had. Maybe ryg can find the statistics again :).
And when there is a LMT hash conflict also hash the LDB.
(hash function TBD, should be the fastest we can find, no security characteristics required, but should have small amount of collisions).

When looking at different emulation projects the file format is usually INI based:
https://github.com/project64/project64/blob/master/Config/Project64.rdb
https://github.com/dolphin-emu/dolphin/tree/master/Data/Sys/GameSettings (Dolphin uses the game ID as the filename, we could use the LMT hash)

As we need to ship some kind of database for existing games and to make accessing it fast (not too many files in one directory) the folder structure could be like e.g. the pip or firefox cache directory way:

The first two folders are the first two values of the hash and then the filename is the full hash, e.g. game has hash ab12351155136a1f16, the the path is a/b/ab12351155136a1f16.ini. (directory depth could be decided after hashing thousands of games, just two folders deep will result in many files in one folder, more than 10000 games are known).

Then the structure could be something like:

The [12abcdef] is for the case that it is in the global database, there could be also a easyrpg.ini provided in the game directory which contains [Game] instead of the hash.

```
[EasyRPG]
Version=1 <- File format version

[12abcdef]/[Game] <- hash of LDB (LMT is already in filename)
Name=game name
Engine=RPG2k1.05
[...]

Issue #167, opened 2013-11-12
license changing


take-cheeze:
Since all the code copyright holder agreed to move to non-GPL new license we need to prepare for it.
The issues for it will be
1. which license to move on
-> strong candidate would be MIT license.
2. when to move on
-> I'm thinking of moving to new license when the mruby branch is merged.
3. how do we treat binary distribution linking GPL library
-> maybe we need to remove those libraries or stop distributing it.
4. license text change of source file
-> we'll use some text replace program like sed

----

- [x] 20kdc
- [x] Albeleon
- [x] BlisterB
- [x] carstene1ns
- [x] ChristianBreitweiser
- [x] Desdaemon
- [x] drxgb
- [x] ell1e
- [x] eneway (Narcodis)
- [x] fdelapena
- [x] florianessl
- [x] fmatthew (mateofio)
- [x] Ghabry
- [x] glynnc
- [x] jetrotal
- [x] johannschopplich
- [x] JorgeMaker8000
- [x] kakurasan
- [x] MackValentine
- [x] MakoInfused
- [x] MarianoGNU
- [x] PrimeKick
- [x] Rikku
- [x] Rinnegatamante
- [x] rohkea
- [x] Rueta
- [x] scurest
- [x] sorlok
- [x] take-cheeze
- [x] Toolman2k
- [x] Tondorian
- [x] tyrone-sudeium
- [x] vgvgf
- [x] Zegeri
- [x] Zhek

----

If you cannot find your name in the list above:

- This happens when your change was _trivial_, this applies to stuff like fixing of typos (or other single character changes), only deleting code and reordering of existing code (without inventing new code). 
- You are also not asked when the contribution was WQY font related (this part stays GPL due to the font license).
- You are also not asked when your change affects only homebrew platforms and other exotic platforms (often dependencies used here are under GPL so this would change nothing)
- You contributed only to the Android User Interface: This is already under MIT
- You contributed only (Android) translations: They are already under CC0

When you think that you made a significant contribution and cannot find your name in the list above: We are sorry. Please contact us.

fdelapena:
This issue is quite outdated, the current situation is:

- Already switched to SDL2 from SDL1.2.
- Already switched to ICU from iconv.
- Already switched to mpg123 from libmad. There are no GPL deps anymore.
[...]

PR #1860, opened 2019-08-01
Multi-Game save Import support and Metadata file

sorlok:
This PR adds support for "Multi-Game Saves", which are games like Legion Saga and The Way that allow you to import a previous game's save file when continuing into the next game. This requires some amount of metadata to be added to the game, which I created a class for (that uses the existing json parser). 

A larger discussion of features and motiviations is in Issue https://github.com/EasyRPG/Player/issues/1854

Tested building/running on Windows, OSX, Linux, and PS-Classic.

### DIFF
--- a/resources/shared/easyrpg.ini
+++ b/resources/shared/easyrpg.ini
@@ -1,14 +1,14 @@
 [EasyRPG]
 FileFormatVersion=1
 
-[783aa77b/*]
+[ff3d8719/*]
 Name=The Way: Episode 1
 Crc32LDB=59ea95a5
 
 [23fa0fb7/55257547]
 Name=The Way: Episode 2
 Vocab_ImportSave=Import journey
-ImportSaveParent=783aa77b/*
+ImportSaveParent=ff3d8719/*
 ImportSavePivotMap=5
 
 [1954d31a/*]
@@ -32,7 +32,7 @@ Vocab_ImportSave=Import journey
 ImportSaveParent=cec21ae7/*
 ImportSavePivotMap=5
 
-[799c58da/*]
+[e868be31/*]
 Name=The Way: Episode 6
 Crc32LDB=1518161c
 Vocab_ImportSave=Import journey
@@ -48,4 +48,9 @@ Name=Legion Saga II
 Crc32LDB=473b0b53
 ImportSaveParent=ec9de7f8/*
 ImportSavePivotMap=133
+
+[999e9fb4/*]
+Name=Legion Saga III
+Crc32LDB=28ea6edc
+
 
--- a/resources/shared/easyrpg.ini
+++ b/resources/shared/easyrpg.ini
@@ -1,0 +1,51 @@
+[EasyRPG]
+FileFormatVersion=1
+
+[783aa77b/*]
+Name=The Way: Episode 1
+Crc32LDB=59ea95a5
+
+[23fa0fb7/55257547]
+Name=The Way: Episode 2
+Vocab_ImportSave=Import journey
+ImportSaveParent=783aa77b/*
+ImportSavePivotMap=5
+
+[1954d31a/*]
+Name=The Way: Episode 3
+Crc32LDB=16a7b30b
+Vocab_ImportSave=Import journey
+ImportSaveParent=23fa0fb7/55257547
+ImportSavePivotMap=5
+
+[cec21ae7/*]
+Name=The Way: Episode 4
+Crc32LDB=5503cdf7
+Vocab_ImportSave=Import journey
+ImportSaveParent=1954d31a/*
+ImportSavePivotMap=5
+
+[81b7345c/*]
+Name=The Way: Episode 5
+Crc32LDB=b7094417
+Vocab_ImportSave=Import journey
+ImportSaveParent=cec21ae7/*
+ImportSavePivotMap=5
+
+[799c58da/*]
+Name=The Way: Episode 6
+Crc32LDB=1518161c
+Vocab_ImportSave=Import journey
+ImportSaveParent=81b7345c/*
+ImportSavePivotMap=5
+
+[ec9de7f8/*]
+Name=Legion Saga
+Crc32LDB=71ad9378
+
+[a4627550/*]
+Name=Legion Saga II
+Crc32LDB=473b0b53
+ImportSaveParent=ec9de7f8/*
+ImportSavePivotMap=133
+
--- a/src/meta.cpp
+++ b/src/meta.cpp
@@ -72,7 +72,7 @@ Meta::Meta(const std::string& meta_file) {
 		if (version == "1") {
 			IdentifyCanonName();
 		} else {
-			Output::Warning("Metadata error in %s, format property %s:%s is missing or invalid: '%s'", MTINI_EASY_RPG_SECTION, MTINI_FILE_FORMAT_VERSION, version.c_str());
+			Output::Warning("Metadata error in %s, format property %s:%s is missing or invalid: '%s'", meta_file.c_str(), MTINI_EASY_RPG_SECTION, MTINI_FILE_FORMAT_VERSION, version.c_str());
 		}
 	}
 
--- a/src/meta.cpp
+++ b/src/meta.cpp
@@ -1,0 +1,223 @@
+/*
+ * This file is part of EasyRPG Player.
+ *
+ * EasyRPG Player is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * EasyRPG Player is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <iostream>
+
+// Headers
+#include <fstream>
+#include <iomanip>
+#include <zlib.h>
+#include "data.h"
+#include "filefinder.h"
+#include "lmt_reader.h"
+#include "lsd_reader.h"
+#include "main_data.h"
+#include "meta.h"
+#include "output.h"
+#include "player.h"
+#include "reader_util.h"
+
+
+// Constants used for identifying fields in the easyrpg.ini file.
+#define MTINI_NAME "Name"
+#define MTINI_IMPORT_SAVE_PIVOT_MAP "ImportSavePivotMap"
+#define MTINI_IMPORT_SAVE_PARENT "ImportSaveParent"
+#define MTINI_EASY_RPG_SECTION "EasyRPG"
+#define MTINI_FILE_FORMAT_VERSION "FileFormatVersion"
+
+// Extended vocab key/value pairs
+#define MTINI_EXVOCAB_IMPORT_SAVE_HELP_KEY "Vocab_ImportSaveHelp"
+#define MTINI_EXVOCAB_IMPORT_SAVE_HELP_VALUE "Import Existing Save (Multi-Games Only)"
+#define MTINI_EXVOCAB_IMPORT_SAVE_TITLE_KEY "Vocab_ImportSave"
+#define MTINI_EXVOCAB_IMPORT_SAVE_TITLE_VALUE "Import Save"
+
+
+// Helper: Get the CRC32 of a given file as a hex string
+std::string crc32file(std::string fileName) {
+	if (!fileName.empty()) {
+		std::ifstream in(fileName.c_str(), std::ios::binary);
+		if (in.is_open()) {
+			std::string buffer((std::istreambuf_iterator<char>(in)), std::istreambuf_iterator<char>());
+			unsigned long crc = ::crc32(0, reinterpret_cast<const unsigned char*>(buffer.c_str()), buffer.length());
+
+			std::stringstream res;
+			res <<std::hex <<std::setfill('0') <<std::setw(8) <<crc;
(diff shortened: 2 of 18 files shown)


Ghabry:
Thanks for your contribution.
Will review it when I'm back home in a week.

Ghabry:
Jenkins: test this please

sorlok:
Note: The latest commit fixes a bug (me not understanding how map IDs were stored).

Ghabry:
Wow, this Picojson stuff reminds me again how ugly there API is.

Before merging this we need to decide on the file format of this config file. The problem is: When we accept the format and ever change it we must still support the old format **but I will review the other stuff ASAP**.

Maybe this can be combined with #1210 (don't know if there is a more appropriate issue for this).
I was discussing this with ryg months ago, what we wanted to do is

This question is if this could be also possible through a simple INI format as just suggested in #1210.

```
[Game]
(...) See example in other issue
SaveImport=12345678,abcdef12,fedcba54,3210abcd
^-- List of importable games grouped by LDB,LMT,LDB,LMT,...
```

sorlok:
Yeah, the picojson API is pretty bad. I think merging this with #1210 is probably the right way to go, and an INI format sounds fine to me. 

The idea of simply hashing the LMT file (+ a few other things) is clever, and I like that a lot better than my approach.

Since I need to identify the current game anyway, do you want me to update this PR to implement #1210's detection?

Ghabry:
Concerning the hashing, could you give me until the weekend?

I want to check different (fast enough) hash functions and how to reduce the conflicts, as ryg already told me, there are, especially same game in different versions, with the same hash for LDB and LMT.

Ghabry:
Well I made some quick test against 6000 LDB and LMT.

We can't use the simple djbhash we use for the diagnostic whether ExFont was replaced, has too many collisions.

Good old CRC32 seems to work, same amount of hashes as md5, so collision free (and shorter than md5) but you have to find a non-GPL implementation of the algorithm first :/

sorlok:
Sure thing, I'll get started on it. 2 small questions:

1) Why is a non-GPL CRC32 needed? I thought EasyRPG was as GPL-licensed code base? 
2) Would it make more sense to hash RPG_RT.ini and LMT (instead of LMT and LDB)? I expect the "Game Name" field would be a very good differentiator.

Ghabry:
> Would it make more sense to hash RPG_RT.ini and LMT. Expect the "Game Name" field would be a very good differentiator.

The LMT already contains a copy of the game name at the beginning. Is likely the reason why the LMT is more unique than the LDB, so adding INI won't give anything new, already part of the LMT.

Main purpose of both LMT+LDB is to increase the change that minor game version changes are detectable. For LMT+INI it becomes more unlikely, INI is constant and LMT requires a new map.

> Why is a non-GPL CRC32 needed

We plan to lift some restrictions of the GPLv3 someday (see #167), so GPL code would be in the way of this goal.
Main motivation is the ability to sell games on platforms whose license is GPL-incompatible, e.g. Steamworks (Steam features like Achivements), most video game consoles like the Switch.

sorlok:
Ah, I see. Didn't realize the LMT file has the game name too, thanks!

Understood on the license; I'll find an appropriate MIT or similar CRC32 library.

fdelapena:
Hi, thanks for this great patch!

About the LMT title: it's not 100% reliable if the user changes the game title from the .ini, but I guess most games will work properly because it gets synced once the game gets opened and saved from the map editor. Time will tell if there are too much exceptions (or doing an automated mass game check).

About CRC32-like library, zlib provides Adler32 and we already use this library as a dependency for XYZ (and indirectly for libpng).

sorlok:
Thanks, I'll check out Adler32!

Ghabry:
adler32 has the same collision behaviour as CRC32 in my tests (= no collisions), so it could be used, at least the code is much simpler. Wikipedia provides an implementation:

```cpp
const uint32_t MOD_ADLER = 65521;

uint32_t adler32(unsigned char *data, size_t len) 
/* 
    where data is the location of the data in physical memory and 
    len is the length of the data in bytes 
*/
{
    uint32_t a = 1, b = 0;
    size_t index;
    
    // Process each byte of the data in order
    for (index = 0; index < len; ++index)
    {
        a = (a + data[index]) % MOD_ADLER;
        b = (b + a) % MOD_ADLER;
    }
    
    return (b << 16) | a;
}
```

Ghabry:
oh this is really exported via zlib :D.

https://www.zlib.net/manual.html#Checksum

```ZEXTERN uLong ZEXPORT adler32 OF((uLong adler, const Bytef *buf, uInt len));```

first argument ``adler`` looks confusing but can be 0 when you pass the entire buffer in one go.

**Comment by Ghabry** [comment]
oh now that I look in zlib:
In the same link two lines below is also a crc32 function xD.

``ZEXTERN uLong ZEXPORT crc32 OF((uLong crc, const Bytef *buf, uInt len));``

So basicly the more well-known crc32 algorithm could still be used. ^^

**Comment by sorlok** [comment]
I pushed a commit that switches to using ini + crc32. The code is *much* simpler and shorter using INI instead of JSON.

Some notes:
1) The file is named easyrpg.ini, and it's only searched for inside the game folder.
   A) Ultimately the sub-folder strategy outlined in #1210 should be used; for now I'm trying to keep it simple.

2) I chose to represent the section title as LMTcrc/LDBcrc, where LMT is required, and LDB can be "*" to mean "any". The LDB's crc can still be listed; e.g.:

```
[783aa77b/*]
Name=The Way: Episode 1
Crc32LDB=59ea95a5

[23fa0fb7/55257547]
Name=The Way: Episode 2
ImportSaveParent=783aa77b/*
ImportSavePivotMap=5
```

  A) Right now, if a game matches "123/\*", then it will never search for its LDB equivalent. I.e., "LMT/*" wins over "LMT/LDB" in all cases.
  B) I decided against using [Game], since it is ambiguous unless it is the *only* entry  --and even in that case it makes multi-game save importing difficult.

3) The INI reader doesn't allow one to search for section titles, so I had to designate at least one property as required. I chose to make "Name" a required field, since it's the most useful field.

4) The INI file is kept in memory for later searching (i.e., for Vocabulary extensions). It is accessible via the Meta class, which is now stored in an instance in Player::meta.

5) Sample file in resources/shared/easyrpg.ini that works with The Way and Legion Saga.

**Comment by Ghabry** [comment]
As a first reference implementation this looks good to me and I would accept the INI format this way for now. So I can go over now to the code review and testing :). Really much better with INI instead of JSON.


For more complex use cases a "GetSections" method in the ini reader seems to be useful but this can be postboned to later revisions when I ever find time to implement this global game database feature :D.

**Comment by Ghabry** [comment]
btw what is the purpose of this pivot map?

**Comment by sorlok** [comment]
Thanks for checking. I pushed a minor fix for OSX builds that I found today.

The idea behind the "pivot map" is to act as a safety when importing save files. Essentially, the "pivot map" is the map ID that you end on in "Game 1" and start on in "Game 2". If the save file you're examining doesn't have the party on this specific map, then it's definitely not an end-game save file. 

Typically, "Game 2" will have an auto-start event on the pivot map that resets all variables and shuffles around the party, etc.

File: src/meta.cpp [hunk 1]
```diff
@@ -0,0 +1,223 @@
...
+
+// Extended vocab key/value pairs
+#define MTINI_EXVOCAB_IMPORT_SAVE_HELP_KEY "Vocab_ImportSaveHelp"
+#define MTINI_EXVOCAB_IMPORT_SAVE_HELP_VALUE "Import Existing Save (Multi-Games Only)"
+#define MTINI_EXVOCAB_IMPORT_SAVE_TITLE_KEY "Vocab_ImportSave"
+#define MTINI_EXVOCAB_IMPORT_SAVE_TITLE_VALUE "Import Save"
+
+
+// Helper: Get the CRC32 of a given file as a hex string
+std::string crc32file(std::string fileName) {
```
Ghabry: style: ``file_name`` (and use ``const std::string&``)

resources/shared/easyrpg.ini
```diff
@@ -0,0 +1,51 @@
+[EasyRPG]
+FileFormatVersion=1
+
+[783aa77b/*]
```
Ghabry: Could you recheck all the checksums? The LDB matches for me but the LMT is wrong, I get "FF3D8719".
It is very easy to change the LMT by accident: Just opening the editor and leaving with a different Map selected alters the LMT.

src/meta.cpp
Ghabry: this applies in general to some of the arguments of other functions

src/meta.cpp
```diff
@@ -0,0 +1,223 @@
...
+}
+
+
+std::vector<Meta::FileItem> Meta::BuildImportCandidateList(const FileFinder::DirectoryTree& parentTree, const std::string& child_path, const std::string& parentGameName, int pivotMapId) const {
+	// Scan each folder, looking for an ini file
+	// For now, this only works with "standard" folder layouts, since we need Game files + Save files
+	std::vector<Meta::FileItem> res;
+
+	// Try to read the game name. Note that we assume the games all have the same encoding (and use Player::encoding)
+	auto childPath = FileFinder::MakePath(parentTree.directory_path, child_path);
```
Ghabry: code style: ``child_path`` (also applies to other vars)

src/meta.cpp
```diff
@@ -0,0 +1,223 @@
...
+	return !canon_ini_lookup.empty();
+}
+
+bool Meta::Empty() const {
+	return ini == nullptr || ini->ParseError() == -1;
+}
+
+
+
+
```
Ghabry: lots of whitespace here

src/meta.h
```diff
@@ -0,0 +1,87 @@
...
+
+
+/**
+ * The Meta class loads from a file (easyrpg.ini) and contains 
+ * additional information on a per-game basis.
+ */
+
+class Meta {
+public:
+	// @ini_file - The path to the ini file to load
```
Ghabry: When making doc-comments please follow the javadoc-style we use:

```
/**
 * brief description
 * @param ini_file The path to the ini file to load
 * @return some return
 */
```
see other files for examples

btw, the argument is called meta\_file, not ini\_file

src/meta.h
```diff
@@ -0,0 +1,87 @@
...
+
+	// Returns true if no metadata (file) exists
+	bool Empty() const;
+
+	// Root of the INIReader
+	std::unique_ptr<INIReader> ini;
+
+	// Current canonical lookup string in easyrpg.ini
+	// Can be "Crc32LMT/Crc32LDB" or "Crc32LMT/*"
+	std::string canon_ini_lookup;
```
Ghabry: (also turn everything else in this file into doc comments)

src/scene_title.cpp
```diff
@@ -29,15 +30,17 @@
...
 #include "player.h"
 #include "scene_battle.h"
+#include "scene_import.h"
 #include "scene_load.h"
 #include "scene_map.h"
 #include "window_command.h"
 #include "baseui.h"
 
-Scene_Title::Scene_Title() {
+Scene_Title::Scene_Title() : new_game_index(0), continue_index(1), exit_index(2), import_index(-1) {
```
Ghabry: because we use C++11 you can assign them directly in the Header instaed of using this initializer

src/window_savefile.h
```diff
@@ -79,8 +89,11 @@ class Window_SaveFile :	public Window_Base {
 
 protected:
 	void UpdateCursorRect();
+	std::string GetSaveFileName() const;
 
 	int index;
+	std::string overrideName;
+	int overrideIndex;
```
Ghabry: also naming here

sorlok:
Commit pushed that fixes (I think) all the requested changes. If I missed anything, let me know.

I also tidied up a few C++11 style initializers in existing classes like scene_file, since my change touched those files anyway.

src/meta.cpp
sorlok: Should be done across all files now.

resources/shared/easyrpg.ini
sorlok: You're correct! Re-downloaded and checked on the command line (no opening in the editor).

sorlok:
(Forgot to mention: if you want me to squash + rebase this, just let me know.)

src/meta.cpp
```diff
@@ -0,0 +1,215 @@
...
+Meta::Meta(const std::string& meta_file) {
+	ini.reset(new INIReader(meta_file));
+
+	// Cache per-game lookups
+	if (!Empty()) {
+		std::string version = ini->GetString(MTINI_EASY_RPG_SECTION, MTINI_FILE_FORMAT_VERSION, "");
+		if (version == "1") {
+			IdentifyCanonName();
+		} else {
+			Output::Warning("Metadata error in %s, format property %s:%s is missing or invalid: '%s'", MTINI_EASY_RPG_SECTION, MTINI_FILE_FORMAT_VERSION, version.c_str());
```
Ghabry: 4 %s but only 3 arguments passed

Ghabry:
@sorlok 
Yeah, could you Squash the first 4 commits from  "Mult-Save support and Metadata file" to "Some compiler warning/error fixes on OSX." into one.

As they just contain the PicoJSON -> INI migration it is not necessary to keep them.

src/meta.cpp
sorlok: Thanks, I missed that. Fixed in the latest commit.

sorlok:
Branch rebased and squashed, and issue found in code review was dealt with.

sorlok:
Whoops, forgot to fetch upstream. Ok, *now* it's rebased.

Ghabry:
This is basically ready since 18 days. If anybody else wants to comment on this I give you an additional 48h grace period before pressing the merge button.

Ghabry:
okay, no objections. Thanks for your contribution :)
And sorry for the long delay.

sorlok:
Thanks for following up and approving it. :) No worries on the delay --it's a pretty big feature to review. Glad you like it!

Issue #1865, opened 2019-08-13
TMG (Three Magic Girl) v1.2: Distributing EXP in the Inn fails with "Call Event: Can't call page 0 of event 2"


Ghabry:
#### Name of the game:
TMG (Three Magic Girl) v1.2 (Warning: Hentai game)

https://vipkohaku.x.fc2.com/2015/entry86.html

To get the game click [ダウンロード(83.3MB)], then on the yahoobox page ensure all scripts are enabled and at the top-left is another [ダウンロード] button.

Extract the game with unar or any other application that supports shift-jis encoded zip files.

#### Attach files (as a .zip archive or link them)
[Save03.zip](https://github.com/EasyRPG/Player/files/3498654/Save03.zip)

#### Describe the issue in detail and how to reproduce it:
Walk upwards and talk to the owner of the Inn.

Select:
やどやにとまる → けいけんちをわける → はい → わてり → [Input 1]

The game reports "CallEvent: Can't call non-existant page 0 of event 2".
In RPG_RT with the same savegame it shows a pornographic picture and continues.

Ghabry:
Noticable is that the game has the switch and variable list truncated to 1 entry.

The problematic call is on map ``event -> 001`` event 1, page 3, ``@> Call Event: Variable [0016][0017]`` on line 234

Ghabry:
Likely a clamping problem.

Earlier code does

```
@> Change Parameters: [], Max MP - 999
@> Change Parameters: [], Max MP - 999
@> Change Parameters: [], Max MP + 999
```

The purpose is obviously to reset the value to 0.
But this behaviour should be verified for other Makers and versions. (game uses 2000)

fmatthew5876:
So it looks like we should be clamping the `stat_mod` chunks and we don't. Just testing rm2k3 now I see it should clamp at `[-9999,9999]`. In player it goes on forever.

PR #1866, opened 2019-08-13
Clamp stat mod values to match RPG_RT

fmatthew5876:
Fix: #1865

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -1212,46 +1212,56 @@ std::string Game_Actor::GetClassName() const {
     return GetClass()->name;
 }
 
+static int ClampMaxHpMod(int hp, const Game_Actor* actor) {
+	auto limit = actor->MaxHpValue();
+	return Utils::Clamp(hp, -limit, limit);
+}
+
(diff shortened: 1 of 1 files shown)


src/game_actor.cpp
```diff
@@ -1212,46 +1212,56 @@ std::string Game_Actor::GetClassName() const {
     return GetClass()->name;
 }
 
+static int ClampMaxHpMod(int hp, const Game_Actor* actor) {
+	auto limit = actor->MaxHpValue();
+	return Utils::Clamp(hp, -limit, limit);
+}
+
+static int ClampStatMod(int hp, const Game_Actor* actor) {
```
Ghabry: still called hp -> stat/mod

Ghabry:
Thanks for creating a patch based on my investigation :)

Game works properly ( ͡° ͜ʖ ͡° )

Now I only need to censor the picture for the blog video...

src/game_actor.cpp
fmatthew5876: Fixed

PR #1202, opened 2017-06-11
Android: update translations for 0.5.2

carstene1ns:

PR #1867, opened 2019-08-16
Refactor scene initialization and continuations - Game_Temp 1 of 2

fmatthew5876:
This PR refactors the scene request / call mechanism. 

Fix: #1707

Objectives:
- [x] Fix edge cases in multiple scene calls
- [x] #1707 Test 22
- [x] #1707 Test 24
- [x] #1707 Test 25


fmatthew5876:
This PR refactors the scene request API and then removes game temps for inn, shop, and name scene.

The last thing for `Game_Temp` is battle, but this is complicated and will require a lot of testing. These are pretty straightforward.

I'm putting the battle stuff in a follow up PR.

This has been tested and is ready for review and merge.

src/game_battle.h
```diff
@@ -189,6 +192,13 @@ namespace Game_Battle {
 
 	/** @return death teleport handler if one is installed, otherwise an inactive target */
 	TeleportTarget GetDeathHandlerTeleport();
+
+	/** Don't reference this, use IsBattleRunning()! */
+	extern bool battle_running;
```
Ghabry: Can't you move this in an anonymous namespace to prevent the access?

src/game_battle.h
```diff
@@ -31,6 +31,9 @@ namespace RPG {
 }
 
 namespace Game_Battle {
+    /** @return true if a battle is currently running */
+    bool IsBattleRunning();
```
Ghabry: wrong whitespace. Also considering that "Init" is the "constructor" it should be somewhere below Init

src/scene_shop.cpp
```diff
@@ -16,25 +16,46 @@
...
+		std::vector<int> goods,
+		int shop_type,
+		bool allow_buy,
+		bool allow_sell,
+		Continuation on_finish)
+	: on_finish(std::move(on_finish))
+	  , goods(std::move(goods))
+	  , shop_type(shop_type)
+	  , allow_buy(allow_buy)
+	  , allow_sell(allow_sell)
```
Ghabry: the indent looks weird. 2 Tabs for the args and then "," in front for the init list

src/scene_shop.h
```diff
@@ -37,12 +37,19 @@
...
 	/**
 	 * Constructor.
 	 */
-	Scene_Shop();
+	Scene_Shop(
+			std::vector<int> goods,
+			int shop_type,
+			bool allow_buy,
+			bool allow_sell,
+			Continuation on_finish);
```
Ghabry: same here

Ghabry:
> Refactor requested scene API

Here it should be carefully checked that there are no side-effects by code that runs in the constructors. Before these happened in Push, now they happen when the scene is requested.

src/game_battle.h
fmatthew5876: Unfortunately not because it's in the header. I share your concerns, but I wanted to make sure access gets inlined.

This could be changed from namespace to struct and then it made private. But I did not want to make such a big change at this time.

src/scene_shop.cpp
fmatthew5876: Fixed

src/scene_shop.h
fmatthew5876: Fixed this one too

fmatthew5876:
Addressed the review comments

> Here it should be carefully checked that there are no side-effects by code that runs in the constructors. Before these happened in Push, now they happen when the scene is requested.

Shop, Name, Save, Load, Gameover, and Menu are all affected in this PR and none of them do anything non-trivial in their constructors.

Battle does do actual game logic in it's constructor, but this is actually desired. 

As shown in #1642 when the battle scene is requested: RPG_RT plays battle start SFX, does battle transition, plays battle bgm, and clears the face. All of this happens even if you immediately cancel the battle scene by calling another scene in another parallel event on the same frame.

The clear face logic is implemented in #2027. The music/sfx logic is in #1202 

I have not implemented the special transition effect for cancelled scenes. This happens for both battle and gameover scene. Since it's only visual and such a wierd edge case (bug?) I'm not sure we care.

Ghabry:
Jenkins: Test this please

src/game_battle.cpp
```diff
@@ -68,7 +69,7 @@ void Game_Battle::Init() {
 	spriteset->Update();
 	animation.reset();
 
-	Game_Temp::battle_running = true;
+	Game_Battle::battle_running = true;
```
carstene1ns: Why using the prefix here?

fmatthew5876: More obvious it's a global variable

PR #1870, opened 2019-08-18
Fix Tile Memory Leak w.r.t the cache_effects map

sorlok:
There is an issue with RPG Maker 2003 games (and likely others) when played in Easy RPG wherein the cache_effects map will hold on to Tile resources for events that use Tiles for their visual elements. You can observe this with the minimal example below:
https://drive.google.com/file/d/1hBFYKPiB0LaKrjAgkmxVj8hy3Qsh_Zjf/view?usp=sharing

Run the game and hit F12 to restart it; the Debug console will show:
`  Debug: possible leak in cached tilemap Snow/0`

This bug is caused by the "Snow/0" tile being cached in the '''key''' of the cache_effects map. I changed that key to use a weak_ptr, a.k.a.:
` using effect_key_type = std::tuple<std::weak_ptr<Bitmap>, Rect, bool, bool, Tone, Color>; `

The change required some additional structuring to use std::owner_less() to compare these two. This should have been pulled in automatically, but since the weak_ptr is inside a tuple, it's possible the compiler got a little confused. (Using the struct is fine, so I'm not worried about this.)

Using a weak_ptr as the key (instead of a shared_ptr) makes sense to me ---the key should not participate in ownership, and the code does not rely on this. FYI, the sample map has a single "Tint Screen" in an event ---this is enough to trigger the bug.

In case you're curious, this also seems to fix the issue of some games (like the test game) crashing with a segfault on OSX  --at least for me.

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -481,6 +481,7 @@ BitmapRef Cache::SpriteEffect(const BitmapRef& src_bitmap, const Rect& rect, boo
 }
 
 void Cache::Clear() {
+	cache_effects.clear();
 	cache.clear();
 	cache_size = 0;
 


Ghabry:
Good find.

This is not the correct way to implement a ``<`` operator. You always need to check for ``<``, then for ``==``, then the next for ``<`` etc. see https://stackoverflow.com/questions/3882467/defining-operator-for-a-struct (I know this because I made the same mistake when I wrote the operator for Color and Tone :/).

Isn't it enough to call ``cache_effects.clear()`` before the for-loop? I forgot to put the call there.

I understand what you mean but in this case I don't think changing this to ``weak_ptr`` is necessary due to how the Sprite class (the only user of SpriteEffects) handles effects: They always build a pair. When a Sprite has a Bitmap and an effect and the Sprite is destroyed the Effect is also destroyed, so the Bitmap has no references anymore.

Same applies when the bitmap of the sprite is changed, then the effect is destroyed (and the src bitmap is destroyed because the effect_cache only holds a weak ref).

sorlok:
Good catch --that comparator is indeed busted! I pushed a fix for it. The ` if (a != b) { return a < b; } ` should be equivalent to the stackoverflow post. And std::tie is really handy; didn't know about it!

Ok, regarding the broader question you had:
1) Calling cache_effects.clear() will indeed fix the problem. I was thinking that there might be some kind of Sprite graphic that persists across reloads  --but on second thought that doesn't make much sense.

So given (1), do you want to (a) keep the weak_ptr implementation, or (b) change it to 
` cache_effects.clear() ` ? Either one's fine with me.

Ghabry:
Well if you don't mind you can just go with the ```cache_effects.clear()``` solution. Or do both.

sorlok:
Let's just do ` cache_effects.clear() ` for now. I'll rebase this.

sorlok:
I force-pushed the update to keep the history clean. Tested and confirmed that cache_effects.clear() works. This is ready for merge (on my side anyway).

Ghabry:
Oh excellent, you already know how git history rewriting works. :)

Issue #1873, opened 2019-08-25
Handle embedded control characters in text


fmatthew5876:
It is possible for game authors to embed control characters like newlines into messages. RPG_RT completely ignores these.

Our text processing code takes multi line sources and smashes them all into one string. Using `\f` to split pages and `\n` to split lines.

Because of the way we do this, we don't distinguish between our control characters and the copy and pasted ones. We should ignore any control characters in the original string.

Issue #1781, opened 2019-05-28
Text overlaps with face on teleport


pulsejet:
Please fill in the following fields before submitting an issue:
#### Name of the game:
Ahriman's Prophecy

#### Player platform:
Windows, Android (probably all)

#### Describe the issue in detail and how to reproduce it:
On displaying a message with a face to the left after teleporting from inside a common event, the face is displayed but the offset for the text gets unset for all except the first line. So the first line shows up with the proper offset but the rest of the text overlaps with the face.

pulsejet:
Ugh, for some reason the game defines a bunch of parallel process common events the first statement for all of which is to reset the face graphic to nothing (along with other things below in a conditional). The fact that this does work with RPG_RT does indicate some deviation of behavior from the original implementation though.

Ghabry:
No need to close this yet, as we behave different to RPG_RT we consider this a bug in our implementation, even if RPG_RTs behaviour is strange.

fmatthew5876:
This appears more complicated to reproduce

@pulsejet Do you have a save game near where it happens in the example game?

pulsejet:
I've isolated a minimal test case [here](https://drive.google.com/open?id=12EqMjC9BzjQrBkeVzmUMCnez-w6my_pb)  (RTP dependent, let me know if this is not okay)

Basically, all you need is is one event that the player can interact which has
```
SetFace (1)
ShowMessage (1)
Teleport (to another map)
SetFace (2)
ShowMessage (2)
```
and one parallel process common event with just 
```
SetFace (Erase)
```
The second line of message 2 should overlap with the face. Note that it's necessary to telepoprt to a _different_ map, not the one you're on right now.

PR #1876, opened 2019-08-25
Refactor messaging system

fmatthew5876:
Refactor message window to get correct frame timings, interpreter behavior, and avoid lockups

Depends on #1742 

Fix: #1873 
Fix: #1781

TBD:
- [ ] All of #1706 
- [ ] Figure out exact timing of speed 1 half width characters when combined with full width.


fmatthew5876:
All of this now has been subsumed into other message PRs

PR #1877, opened 2019-08-27
Emulate RPG_RT crash: vehicle teleport from parallel map event

fmatthew5876:


Ghabry:
But doesn't the currently running parallel interpreter survive for one more frame when the map changes (until they yield with a Wait e.g.)?

Issue #1875, opened 2019-08-25
DQ2: Battle doesn't end after defeating the Bomb Rock


Ghabry:
#### Name of the game:
DQ2　悪霊の勇者・製品版ver1.50

#### Attach files (as a .zip archive or link them)
Download: http://www1.axfc.net/uploader/so/1768371 (NFSW. full of hentai ads!)

1. Click "[ 1768371.zipをダウンロードします。 ]"
2. ＜ダウンロードする | click here to start download. ＞
3. When the download doesn't start click on  こちら

Savegame: 
[DQ2-Save05.zip](https://github.com/EasyRPG/Player/files/3538453/DQ2-Save05.zip)

#### Describe the issue in detail and how to reproduce it:
Kill the enemy in front of you using fast forward and normal attacks.
When he dies the battle continues and selecting "Attack" again will crash the engine because the enemy list is empty.

PR #1879, opened 2019-08-27
Fix compatibility with older save games

fmatthew5876:
Fixes bug where save in #1875 would hang when loaded

This is not needed if we merge #1884


fmatthew5876:
Since we have #1884 which is the proper fix I close this. Can be reopened if we think we want the short term hack sooner.

Ghabry:
fine for me, considered also to propose this, just to save some reviewing time

PR #1882, opened 2019-08-28
Migrate to C++14 / Gcc5

fmatthew5876:
Depends on: https://github.com/EasyRPG/liblcf/pull/339


configure.ac
```diff
@@ -112,7 +112,7 @@ AC_SUBST([BASHCOMPLETION_DIR])
 AM_CONDITIONAL([HAVE_BASHCOMPLETION], [test "x$with_bash_completion_dir" != "xno"])
 
 # C++11 is mandatory
```
carstene1ns: 14 ;)

fmatthew5876:
Added to all the make files, including secret android ones.

configure.ac
Zegeri: It should be "# C++14 is mandatory"

fmatthew5876: Fixed, thanks!

Issue #1871, opened 2019-08-18
Verify how stack is cleared on event finish


fmatthew5876:
Study how exactly RPG_RT behaves when an event stack frame finishes executing.

Test Cases 1
---------------

EV01 autostart
```
Var1 + 1
```

EV02 parallel
```
Loop:
  OpenSaveMenu
```

| Save | EV01 current command | EV02 current command | Var1 |
| -- | -- | -- | -- |
| 1 | | 2 | 0 |
| 2 | 1 | 2 | 1 |
| 3 | | 2 | 1 |
| 4 | 1 | 2 | 2 |
| 5 | | 2 | 2 |

Conclusion
--------------

Foreground stack is cleared each frame when events finish.

fmatthew5876:
Test Cases 2
---------------

EV01 parallel
```
Var1 + 1
```

EV02 parallel
```
Loop:
  OpenSaveMenu
```

| Save | EV01 current command | EV02 current command |
| -- | -- | -- |
| 1 | 0 | 2 |
| 2 | 0 | 2 |
| 3 | 0 | 2 |

Conclusion
--------------

Parallel frame is not cleared but index reset to 0?

fmatthew5876:
Test Cases 3
---------------

EV01 parallel
```
Var1 + 1
EraseEvent
```

EV02 parallel
```
Loop:
  OpenSaveMenu
```

| Save | EV01 current command | EV02 current command |
| -- | -- | -- |
| 1 | 2 | 2 |
| 2 | 2 | 2 |
| 3 | 2 | 2 |

Conclusion
--------------

EraseEvent causes stack frame to freeze

fmatthew5876:
Test Cases 4
---------------

EV01 parallel
```
Var1 + 1
EndEventProcessing
Var1 +1
```

EV02 parallel
```
Loop:
  OpenSaveMenu
```

| Save | EV01 current command | EV02 current command |
| -- | -- | -- |
| 1 | 0 | 2 |
| 2 | 0 | 2 |
| 3 | 0 | 2 |

Conclusion
--------------

EndEventProcessing behaves the same as the event ending normally.

fmatthew5876:
Test Cases 5 - empty parallel events
--------------------

EV01 parallel
```
Loop:
  OpenSaveMenu
```

EV02 parallel
```
Var1 + 1
```

EV03 parallel
```
```

[...]

Issue #1764, opened 2019-05-08
Performance: Improve interpreter setup


Ghabry:
via profiling I observed that in the very script heavy game "smashbros" 25% of all CPU time is consumed by ``list = _list`` (soon: ``frame->commands = _list``).

The only operations happening on this list are read and clear, therefore it is feasible to work on a reference instead of a copy but this change is not trivial.

by fmatthew:

> Since event code never changes, it should be possible to just use something akin to C++20 std::span to just point to the desired event buffer. That will require some surgery possibly at the liblcf level too.

fmatthew5876:
One very stupid thing we do is reload completely when common events repeat themselves:

https://github.com/EasyRPG/Player/blob/master/src/game_commonevent.cpp#L50

So I think one quick win here could be to check if we're reloading the same event and if so, just "rewind" instead of copying the code again.

Ghabry:
iirc the stacktrace of the Profiler was indeed from Common events, so this sounds like a feasible short-term solution.

fmatthew5876:
Another thing:

https://github.com/EasyRPG/Player/blob/master/src/game_interpreter.cpp#L101

We'll probably need to change `IsRunning()` to `index >= list.size()`

Issue #1883, opened 2019-08-29
Performance Regressions in script heavy games


fmatthew5876:
Something has caused a performance regression in script heavy games.

- [x] もしもスマブラだったら (Wannabe Smash Bros.) http://viprpgsum15.webcrow.jp/sakuhin.html?71

- [ ] Frozen Triggers http://viprpg17kh.php.xdomain.jp/main/04_game.php?id=48

These games are currently unplayable in master with debug builds.

Looking at the graphs, these are performance bottlenecks in Frozen Triggers.

- [x] https://github.com/EasyRPG/Player/blob/0e05729f1c170e65800bfee487800c4e12e29683/src/cache.cpp#L128
    * Suggestion: Call `GetTicks()` once each frame and cache the value. Replace calls to `DisplayUI::GetTicks()` to the cached one
- [ ] https://github.com/EasyRPG/Player/blob/0e05729f1c170e65800bfee487800c4e12e29683/src/game_picture.cpp#L256
    * Suggestion: Cache / Reuse Picture Sprites
- [x] https://github.com/EasyRPG/Player/blob/0e05729f1c170e65800bfee487800c4e12e29683/src/game_picture.cpp#L253
    * Suggestion: Cache / Reuse Picture Sprites
- [ ] https://github.com/EasyRPG/Player/blob/aca489b85761d64d9d111589fe1848b16a18bc74/src/game_picture.cpp#L244
    * Calling `Bind()` allocates a `shared_ptr<int>`. In the case where the asset is already in the cache this is wasted. Provide an optimized code path which avoids allocating and throwing away these objects when the asset is already loaded in cache.

fmatthew5876:
I just tested this commit https://github.com/EasyRPG/Player/commit/00ad8ab137261fff23559365ad50397a2d98ffcc and still see the issue.

This must have made it into 0.6.1?

carstene1ns:
I have not followed development for some time, but built from master today and do not observe this.

fmatthew5876:
What platfrom?

I'm testing this on linux. Ghabry also observed this while testing #1740.

carstene1ns:
Hmm, might be my setup then.

Ghabry:
When I use the debug builds of our Linux releases (The ones published on our website) I get:

0.5.4: 18 FPS - 34 FPS (has random lags)
0.6.0: 12 FPS - 20 FPS
0.6.1: 21 FPS - 30 FPS
[...]

PR #1884, opened 2019-08-29
Refactor Interpreter Stack Frame Completion 

fmatthew5876:
Fix: #1871 

This should also fix: #1764  once we solve #1883

Supercedes #1879 If we merge this we don't need that one.

* Rename CommandEnd() to OnFinishStackFrame()

This name and it's nearby comment was very misleading. The
interpreter has actually always ignored `CommandEnd` commands.
This function was only called when the index ran past end of the stack
frame.

* Do OnFinishStackFrame() before execution loop.

Call this in one place and handle multiple stack pops. Also,
don't hang on empty events.

* Optimize parallel events

Reset the index and never pop the base stack frame. RPG_RT
does this and we avoid copying parallel event data every frame.

RPG_RT will always create a stack frame for parallel events,
even if they are empty.

### DIFF
--- a/src/game_commonevent.cpp
+++ b/src/game_commonevent.cpp
@@ -22,27 +22,36 @@
 #include "game_interpreter_map.h"
 #include "main_data.h"
 #include "reader_util.h"
+#include <cassert>
 
 Game_CommonEvent::Game_CommonEvent(int common_event_id) :
-	common_event_id(common_event_id) {
+	common_event_id(common_event_id)
+{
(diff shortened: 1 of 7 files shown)


fmatthew5876:
jenkins: test this please

fdelapena:
test this please

fmatthew5876:
jenkins: test this please

fmatthew5876:
Rebased and added a very minor optimization when we check if parallel event can execute.

Issue #1840, opened 2019-07-16
Lufia V: For the Savior softlock almost immediately after starting game


M1CR0H4CK3R:
#### Name of the game:
Lufia V: For the Savior Part 1

#### Player platform:
PC & 3DS

#### Attach files (as a .zip archive or link them)
[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/3397906/easyrpg_log.txt)

#### Describe the issue in detail and how to reproduce it:
After all of the intro videos, once you gain control of your character, going downstairs will cause the game to softlock with the music still playing. This issue does not happen when regularly launching the RPG_RT.exe, since I have completed the whole game this way already. I just wanted to replay this game on my 3DS, but this program isn't compatible with the game as it is now. The issue happens on both PC and my 3DS. I was told by the game developer that this game uses a few 'quick hacks' to run properly, so maybe tweaking this program a little can make it work properly. Hopefully this can be fixed soon.

fmatthew5876:
Game Link: https://rpgmaker.net/games/4184/

Confirmed this is still broken in master.

When you walk down the stairs, the foreground interpreter gets stuck in a do nothing infinite loop on `Game_Message::message_waiting`.

Attached a save game created by debug menu.

[Save01.zip](https://github.com/EasyRPG/Player/files/3399214/Save01.zip)

fmatthew5876:
Test Cases 1 - Message command delays
================================

EV01 parallel
```
ShowMessage: "\v[1] \v[2]"
```

EV02 parallel
```
Var[1] += 1
```

EV03 autostart
```
Var[2] += 1
```

Result:
"0 0"
"84 10"
"149 20"
"212 30"

Ghabry:
There is only one plugin in the DynPlugins folder:
"Also if you see in the dynrpg folder you will see a strange plugin called NMBattlers.dll, it was a plugin made by my friend Dahrkael from spanish community, that disable the auto-reverse graphics when you start a battle using front side battle. " (Source: https://rpgmaker.net/forums/topics/12520/)

gadesx:
NMBattlers.dll, of dynrpg disable the flip function on battle, because front battle view doesn't need the flip.
Delete the dll and try if work with player

Issue #1819, opened 2019-06-14
Amillusion - Game Freeze


fmatthew5876:
From an email report:

> The game is Amillusion 0.01. works fine on Windows. The bug occurs at the very beginning. After > launching new game intro plays fine, game starts and you can walk in first location as usual, but if > you interact with anything the game freezes. and if you relaunch it the screen is black, only sounds of footsteps are playing.

https://compoteteam.wixsite.com/teamcompote/amillusion

Confirmed this bug exists. Start a new game and talk to the stack of papers in front of you. The game freezes.

fmatthew5876:
Somehow this is fixed by #1784. I haven't looked into the game code to figure out why.

PR #1887, opened 2019-09-03
Async improvements + Async Inn Sequence

fmatthew5876:
~~Depends on #1742~~

The PR has 2 goals:
* Refactor the async framework:
    * Don't use globals
    * Kill `transition` game temp vars
    * Support parameters for async ops
* Change Inn Fade Out, BGM, Fade In sequence to be asynchronous

Test Cases of #1689 
- [x] Test 1
- [ ] Test 2 - Requires message fixes - to be done in a later PR after message refactor commits get in.
- [x] Tests 3 - 5
- [ ] Test 6 - not sure we even want to emulate this..?

### DIFF
--- a/src/scene_map.cpp
+++ b/src/scene_map.cpp
@@ -197,6 +197,10 @@ void Scene_Map::PreUpdateForegroundEvents(MapUpdateAsyncContext& actx) {
 }
 
 void Scene_Map::Update() {
+	if (activate_inn) {
+		UpdateInn();
+		return;
+	}
 	MapUpdateAsyncContext actx;
 	UpdateStage1(actx);
 }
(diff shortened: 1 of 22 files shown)


fmatthew5876:
I retested #1819 and Ghabry's async test case from #1784 

I rebased a local copy of this onto #1881 and retested both #1261 and #1840 

This is as much as I can do for the Inn until some of the message work is done.

This PR is done and ready for review.

Ghabry:
RFC: Is there any way to combine this with the Async code we have for emscripten or do you consider them too different?

fmatthew5876:
> RFC: Is there any way to combine this with the Async code we have for emscripten or do you consider them too different?

Which async code are you referring to specifically?

This one is a sort of poor mans way of turning our game loop into a resumable co-routine.

Ghabry:
I mean the Async code for downloading files that emscripten calls (async_handler.cpp)

fmatthew5876:
They already kind of work together since file IO works the same as animating transitions.

Do we have any case where we would need to suspend and resume the game loop at a specific point for a file load? I could add an `AsyncOp::eFileLoad` to do that.

It would likely cause the game to stutter though, as everything would have to freeze in the middle of the frame until the IO finishes.

As far as I'm aware, most things in the game logic don't depend on file IO to be done and are fine waiting until the end of the frame.

Ghabry:
No, we already suspend via "ImportantFilePending" when the game loop returns and for non-important files which don't suspend there is code which delays rendering operations on these files (e.g. when Width/height are necessary), but update code still runs as expected. 
Just thought that this can be maybe simplified by merging it with the new AsyncOp code, but not necessary.

src/scene_map.cpp
```diff
@@ -392,6 +396,58 @@ void Scene_Map::OnAsyncSuspend(F&& f, AsyncOp aop, bool is_preupdate) {
...
+	}
+
 	AsyncNext(std::forward<F>(f));
 }
 
+void Scene_Map::StartInn() {
+	const RPG::Music& bgm_inn = Game_System::GetSystemBGM(Game_System::BGM_Inn);
+	if (bgm_inn.name.empty() ||
+		bgm_inn.name == "(OFF)" ||
+		bgm_inn.name == "(Brak)") {
```
Ghabry: I wanted to complain now that this is wrong but this is even in the legacy code.

The new way to detect if a string is (OFF) or similiar is "isStopFilename" in Game_System but the function is static and not exported yet.

src/scene_map.cpp
```diff
@@ -392,6 +396,58 @@ void Scene_Map::OnAsyncSuspend(F&& f, AsyncOp aop, bool is_preupdate) {
...
+		FinishInn();
+		return;
+	}
+
+	Game_System::BgmStop();
```
Ghabry: It's not necessary to call BgmStop before BgmPlay

fmatthew5876: Fixed

src/scene_map.cpp
fmatthew5876: Fixed

fmatthew5876:
I pushed another change which creates a `Scene_Inn` for the inn logic. Now the inn logic is separated from `Scene_Map`. This change allows us to use push and pop scenes within async operations.

This will enable #1891 to be implemented using this same machinery.

fmatthew5876:
Redid the inn test cases.

This is ready again

fmatthew5876:
I took out the async `Scene_Inn`. I'm still on the fence about whether it's good idea to support asynchronous scene calling.

Issue #1869, opened 2019-08-17
Inn frame timing is not accurate


fmatthew5876:
When the inn is called, some transitions occur. We need to figure out the timing of this. Maybe it's async like screen transitions? Or something else?

Test case 4 in #1707 fails until we figure this out.

fmatthew5876:
Test Cases 1 - Async behavior / blocking
----------------------

EV01 parallel
```
Wait 0.0s
Inn Cost 0
OpenSaveMenu
```

EV02 parallel
```
Var2 + 1
```

| Save | Var 1 | Player stop count |
| -- | -- | -- |
| 1 | 2 | 1 |

Result: Ev02 never executes while Inn is running

Conclusion
----------

Inn either blocks events or is async (probably async)

fmatthew5876:
Test Cases 2 - Inn with message
-----------------

EV01 autostart
```
Var[Var10] +1
```

EV02 parallel
```
Gold+9999
Var10+1
Inn Cost 1
Var10+1
OpenSaveMenu
Var10+1
```

| Save | Var1 | Var2 | Var10 | Player stop count |
| -- | -- | -- | -- | -- |
| 1 | 2 | 1 | 2 | 2 |


```

**Comment by fmatthew5876** [comment]
Test Cases 3 - BGM Behavior
----------------------

Setup Map01 with some music


1. Talk to EV01 trigger
```
ChangeSystemBGM Inn: (OFF)
Inn Cost 0
```

Result: Music fades, erase screen, immediately show screen, music plays again from beginning

2. Talk to EV02 trigger
```
ChangeSystemBGM Inn: (Some Very long BGM)
Inn Cost 0
```

Result: Music fades, erase screen, entire inn bgm plays to finish, show screen, map music plays again from beginning

3. Talk to EV03 trigger
```
ChangeSystemBGM Inn: (Map BGM)
Inn Cost 0
```

Result: Music fades, erase screen, same bgm plays to finish, show screen, map music plays again from beginning

**Comment by fmatthew5876** [comment]
Test Cases 4 - Inn during preupdate
-------------------------------------

EV01 parallel
```
ChangeSystemBGM Inn: (OFF)
Inn Cost 0
```

Result: Screen fades in like normal

**Comment by fmatthew5876** [comment]
Test Cases 5 - Inn and erase screen
-------------------

EV01 autostart
```
HideScreen
Inn Cost 0
EraseEvent
```

Result: Fade out, inn, Fade in, event disappears

[...]

PR #1890, opened 2019-09-04
Fix Message Blocking rules - Messages 1 of N

fmatthew5876:
Depends on #1905
Fix: #1781 

I took the commits from #1876 which relate to the conditions which cause the interpreter to block. This is arguably one of the more important pieces of the message refactor. Also includes the LSD `show_message` chunk.

Related Test cases:
* https://github.com/EasyRPG/Player/issues/1706#issuecomment-525077134

There is an additional case where autostart events will wait for the message box to close if they run 1 frame after the previous message finishes. While I added the variable for this, it is not yet implemented. Implementing it correctly requires a lot of frame timing surgery in `Window_Message` and is deferred until later.

### DIFF
--- a/tests/utf.cpp
+++ b/tests/utf.cpp
@@ -1,0 +1,87 @@
+#include <cassert>
+#include <cstdlib>
+#include "utils.h"
+
+#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
+#include "doctest.h"
+
+// Correct Tests
+
+struct TestSet {
+	std::string u8;
+	std::u16string u16;
+	std::u32string u32;
+};
+
+#define CONCAT2(x, y) x ## y
+#define CONCAT(x, y) CONCAT2(x, y)
+#define STRINGIZE(x) #x
+//#define TS(X) { CONCAT(u8, STRINGIZE(X)), CONCAT(u, STRINGIZE(X)), CONCAT(U, STRINGIZE(X)) }
+#define TS(X) { CONCAT(u8, X), CONCAT(u, X), CONCAT(U, X) }
+
+TestSet tests[] = {
+	//Valid Strings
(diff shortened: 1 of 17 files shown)


fmatthew5876:
I retested everything in #1706 Test 19 and all pass except the 2 inn cases.

The Inn sequence has some very strange interplay with a closing message box. I've noted this in #1869

Ghabry:
If possible without headache you can cherry-pick "Filter ASCII control chars from Game_Message" here to get this bugfix in faster.

fmatthew5876:
Unfortunately that one the way it's implemented has dependencies not included here.

fmatthew5876:
I added 2 more standalone message commits. These should be non-controversal but if any issues during review I can just back them out.

Ghabry:
The utf-8 unit tests are not provided for autoconf.

see https://github.com/EasyRPG/Player/blob/master/Makefile.am#L439 and following

tests/utf.cpp
```diff
@@ -0,0 +1,87 @@
...
+TestSet tests[] = {
+	//Valid Strings
+	TS("κόσμε"),
+	//First Char tests
+	TS("\U00000000"),
+	TS("\U00000080"),
+	TS("\U00000800"),
+	TS("\U00010000"),
+//	TS("\U00200000"),
+//	TS("\U04000000"),
```
Ghabry: What are these commented values? Invalid codepoints?

fmatthew5876: They are codepoints > 4 bytes. The compiler doesn't like them so they are commented out.

fmatthew5876:
Added the tests to autoconf and rebased to master

fmatthew5876:
jenkins: test this please

fmatthew5876:
Rebased to master

PR #1896, opened 2019-09-19
Emscripten: Support non-canonical paths

Ghabry:
Emscripten makes now the path canonical before doing a lookup and the FileFinder does not flood the log with "Path adjusted" and "File not found bla.xyz", "File not Found bla.png" (when probing for the adjusted file) anymore.

This adds some minor enhancements: Picojson not included anymore outside of emscripten and added a async testing macro.

### DIFF
--- a/src/async_handler.cpp
+++ b/src/async_handler.cpp
@@ -34,6 +34,10 @@
 #include "utils.h"
 #include "graphics.h"
 
+// When this option is enabled async requests are randomly delayed.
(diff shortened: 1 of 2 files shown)


Ghabry:
Bit slow but seems to work: https://easyrpg.org/play/pr1896/?game=frozen_triggers&engine=rpg2k3e :+1:

fmatthew5876:
Very cool. Thanks!

PR #1897, opened 2019-09-19
Fix Player getting stuck at 0 FPS.

Ghabry:
When the Player runs at 0 FPS the next_frame will fall behind start_time which means the Player will continue running at 0 FPS even when the current screen could run faster.
next_frame and start_time are now properly reset when 0 FPS are reached.

For testing wait for ~10 sec in Frozen Triggers, then leave the screen to the left to enter the shop. With my patch the FPS go up to 60 again, in master it stays at 0 FPS for a long time until it recovers.

This should also fix some Android bug reports where the logfile was spammed with 0 FPS.

### DIFF
--- a/src/graphics.cpp
+++ b/src/graphics.cpp
@@ -88,7 +88,7 @@ void Graphics::Update() {
 		if (fps_overlay->GetFps() == 0) {
 			Output::Debug("Framerate is 0 FPS!");
 			Draw();
-			Player::FrameReset();
+			Player::FrameReset(current_time);
 		} else {
 			next_fps_time = current_time + 1000;
 			fps_overlay->ResetCounter();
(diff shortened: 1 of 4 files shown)


src/graphics.cpp
```diff
@@ -88,10 +88,12 @@ void Graphics::Update() {
 		if (fps_overlay->GetFps() == 0) {
 			Output::Debug("Framerate is 0 FPS!");
 			Draw();
+			Player::FrameReset();
```
fmatthew5876: This ends up calling `DisplayUi->GetTicks()` 3 times (4?) which results in 3(4?) system calls. Even worse, it happens when your fps is 0, causing you to slow down even more when your game is already lagging.

I would change both `Player::FrameReset()` and `Graphics::FrameReset()` to optionally accept an `int32_t = 0` for the ui ticks. If it's 0, fetch call `DisplayUi->GetTicks()` to set it. 

Call `DisplayUi->GetTicks()` once at the top of `Graphics::Update()` and use it in both places it's used there. Then pass it down into the 2 `FrameReset()` calls.

fmatthew5876:
Other than my `GetTicks()` comment, LGTM

src/graphics.cpp
Ghabry: fixed this for the 0 FPS case.
There are other cases where FrameReset is called but there is no tick-variable around, so didn't change them.

PR #1898, opened 2019-09-19
Fix FastBlit edge cases

Ghabry:
see commit messages for more details.

### DIFF
--- a/src/tilemap_layer.cpp
+++ b/src/tilemap_layer.cpp
@@ -164,16 +164,18 @@ TilemapLayer::TilemapLayer(int ilayer) :
 	sublayers.push_back(std::make_shared<TilemapSubLayer>(this, Priority_TilesetBelow + layer));
 }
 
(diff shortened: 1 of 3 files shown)


Ghabry:
About all this strange fast blit probing: I benchmarked this years ago on the old 3DS and it is really noticable faster to do these checks instead of paying for the alpha channel aware lower layer blit.

src/tilemap_layer.cpp
```diff
@@ -232,6 +234,7 @@ void TilemapLayer::Draw(int z_order) {
 			if (z_order == tile.z) {
 				if (layer == 0) {
 					// If lower layer
+					bool allow_fast_blit = tile.z == Priority_TilesetBelow;
```
fmatthew5876: Style nit `bool allow_fast_blit = (tile.z == Priority_TilesetBelow);`

PR #1901, opened 2019-09-20
Game_Event: Always clear interpreter on page change

fmatthew5876:
Whenever we change the event page, we should clear
the interpreter if it was created.

Remove asserts that can be violated by the interpreter being
created from another page.

This is a regression from #1884


Ghabry:
is there any test case to show what this fixes?

fmatthew5876:
Make an event with 2 parallel pages. The first page should have code and the second page parallel but no code.

Run the game with the first page running, then set the conditions for page 2 to run.

The event will switch to page 2 but not clear the interpreter stack, and the assertion will file

Ghabry:
okay confirmed the crash.

Issue #1913, opened 2019-09-23
Dhux Scar: Problem with transition in parallel process


Ghabry:
Thank you for supporting EasyRPG Player.
Before creating an issue, please verify that your bug still occurs in the "continuous builds" of the Player.
You can get them at the bottom of our download section: https://easyrpg.org/player/downloads

---
Please fill in the following fields before submitting an issue:
#### Name of the game:
Dhux Scar: https://rpgmaker.net/games/1492/downloads/1508/

#### Attach files (as a .zip archive or link them)
[Save14_dhux.zip](https://github.com/EasyRPG/Player/files/3643749/Save14_dhux.zip)

#### Describe the issue in detail and how to reproduce it:
"Leave the map to the right. Select Jehridan on the world map and press enter"

#### Simplified test case:
[Transition.zip](https://github.com/EasyRPG/Player/files/3643759/Transition.zip)

Expected: After toggling the switch you are moved to a new map.
What happens: Screen stays black, same map (_toggling the menu will fade in, is this another bug?_)

#### Online test:
https://easyrpg.org/play/master/?game=issue-1913

Ghabry:
The relevant code is the parallel process with run condition "0001 ON"
```
Switch 0001 OFF
Hide Screen
Transfer Player
```

As we know the disabling of the switch will keep the parallel event running until the next yield, then it is destroyed. Because a Screen Hide is a blocking operation without return in RPG\_RT it won't yield the interpreter and Transfer will execute.

fmatthew5876:
Test Cases 1 - parallel map events
-----------------------------

1. EV01 parallel page 1
```
Wait 0.0s
Sw1 on
Sw2 on
```

EV01 page 2 trigger activated by SW01
```
```

Result: Sw1 is on, Sw2 is off

fmatthew5876:
Test Cases 2 - parallel common events
----------------------------

1. CE01 parallel Sw01
```
Sw1 off
Sw2 on
```

Start a new game and enable switch 1

Result: Sw1 off, Sw2 on

fmatthew5876:
Test Cases 3 - parallel events 10k loop limit
----------------------------

1. EV01 parallel
```
Wait 0.0s
Loop:
  Var1 += 1
  If Var1 == 400:
    BreakLoop:
HideScreen
ShowScreen
Loop:
  Var2+=1
```

EV02 parallel
```
Wait 0.0s
Msg: \v[1] \v[2]
```

Result: 400, 2666

[...]

PR #1914, opened 2019-09-23
Fixes for async operations and parallel events

fmatthew5876:
* If a parallel common event disables it's switch and
  then does an async operation, we need to force it to
  rerun on resume, despite the switch being off.
* If a parallel common or map event resumes, we should not
  reset the loop counter.

Fix: #1913

### DIFF
--- a/src/game_commonevent.cpp
+++ b/src/game_commonevent.cpp
@@ -49,10 +49,10 @@ void Game_CommonEvent::SetSaveData(const RPG::SaveEventExecState& data) {
 	}
 }
 
-AsyncOp Game_CommonEvent::Update() {
-	if (interpreter && IsWaitingBackgroundExecution()) {
+AsyncOp Game_CommonEvent::Update(bool resume_async) {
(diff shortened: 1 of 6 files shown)


Ghabry:
Test case: https://easyrpg.org/play/pr1913/?game=issue-1913

Ghabry:
Doesn't look fixed in the web app, testing later locally.

fmatthew5876:
No you're right, my fix is incomplete. There is something else going on with teleports here. Will investigate.

Ghabry:
ah. I confused this when I wrote my text. Okay that is new to me, sounds a bit like a bug in the RPG_RT interpreter when the behaviour is so inconsistent.
Nice corner case again.

fmatthew5876:
Fixed the issue, but I discovered yet another edge case in #1913 test 6.

When you change pages, the interpreter yields in RPG_RT. In player we don't yield and continue executing.

fmatthew5876:
This is now passing all extended tests of #1913.

Issue #1903, opened 2019-09-21
RTP search in Wine registry is inefficient


Zegeri:
Every call of `read_rtp_registry` in `FileFinder::InitRtpPaths` makes the engine to open `$WINEPREFIX/system.reg` and `$WINEPREFIX/user.reg`, make lowercase the entire text and create a std::map with each key. During the initialization, this is done 3 times with both `system.reg` and `user.reg`. This can become the most expensive part in the engine initialization.

![flamegraph initialization](https://user-images.githubusercontent.com/4390703/65378014-1c167000-dcb3-11e9-81c7-414059eb0637.png)

This could be fixed by caching the registry files or doing a single lookup.

Ghabry:
Okay to solve this either need to write a custom INI parser or use a custom read function for the liblcf INI reader that prefilters the lines.
First option sounds easier to implement.

Ghabry:
I hacked together a naive parser which skips all lines except the requested one. This will still not cache the file which means both files are read 3 times. But even this naive implementation is already twice as fast as the current ini reader. I also tested a cached INI reader, time measurement:

master: 2300ms
cached ini: 1600ms
my simple parser: 1050ms

By changing the API so the simple parser tests for all keys at once an even better speedup should be possible.

Ghabry:
is the current solution good enough to close this?

carstene1ns:
I think this should be ok for now.
To only open the files once, we can still optimize the wine registry lookup, but this might have a different api than the windows lookup then (e.g. passing an array of keys to search for).

PR #1916, opened 2019-09-25
Speedup Startup + RTP detection

Ghabry:
Two improvements:

- The RTP is now initialized the first time a file is missing. This way there is no startup delay, instead the delay is now when the first RTP asset is requested. Yes this only moves the problem but for some games this means 0 delay.
- lcf INI parser replaced with a simple parser that just reads all lines and skips until the Key/Value pair is found. No extra caching, that's why I don't consider this a fix for #1903, likely can be improved more.

### DIFF
--- a/src/filefinder.cpp
+++ b/src/filefinder.cpp
@@ -73,17 +73,17 @@ namespace {
 	std::string fonts_path;
 
 	struct {
-		// all RTP search paths
+		/** all RTP search paths */
 		search_path_list search_paths;
-		// RTP was disabled with --disable-rtp
+		/** RTP was disabled with --disable-rtp */
 		bool disable_rtp = true;
-		// Game has FullPackageFlag=1, RTP will still be used as RPG_RT does
+		/** Game has FullPackageFlag=1, RTP will still be used as RPG_RT does */
 		bool game_has_full_package_flag = false;
-		// warning about "game has FullPackageFlag=1 but needs RTP" shown
+		/** warning about "game has FullPackageFlag=1 but needs RTP" shown */
 		bool warning_broken_rtp_game_shown = false;
-		// RTP candidates per search_pathwarning_broken_rtp_game_shown
+		/** RTP candidates per search_path */
 		std::vector<RTP::RtpHitInfo> detected_rtp;
-		// the RTP the game uses, when only one left the RTP of the game is known
+		/** the RTP the game uses, when only one left the RTP of the game is known */
 		std::vector<RTP::Type> game_rtp;
 	} rtp_state;
 
--- a/src/registry_wine.cpp
+++ b/src/registry_wine.cpp
@@ -85,8 +85,8 @@ std::string Registry::ReadStrValue(HKEY hkey, const std::string& key, const std:
 
 	// Custom, simple INI parser because liblcf ini is not efficient enough
 	// (lcf ini stores all keys/values but we only need one)
-	std::string formatted_key_search = "[" + Utils::LowerCase(formatted_key) + "]";
-	formatted_val = Utils::LowerCase(formatted_val);
+	std::string formatted_key_search = "[" + Utils::LowerCaseInPlace(formatted_key) + "]";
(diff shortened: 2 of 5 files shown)


fmatthew5876:
> Delay the RTP detection until a file is missing.

How long does it take to do this routine the first time? 

I'm not sure I like this approach because it'll cause random stutter in the middle of game play the first time an asset is missing. That could be very jarring for the player.

Many games which don't use rtp still have missing assets. It's very common, and all those games will still pay the cost of initializing rtp.

I'd rather just pay the cost up front and deterministically, than have a random slowdown occur during game player.

> Registry Wine: Use custom INI parser (faster)

This looks like a step in the right direction. Ultimately we could just refactor it further to be most optimal. 

* Keep a set of all the keys we'll ever need in some array, tag each one with an enum. 
* Change `Registry` to an object, in it's constructor read the file and load the full set of keys once into an array of string values.
* Provide get methods to just return the key values by enum as an index into the array.

The result would be exactly one time reading the registry file and simple array indexing to retrieve values.

Ghabry:
thanks for the feedback. Will work on a better solution.

When the RTP detect is fast enough there is no reason to use the "delay rtp detection" code.

fmatthew5876:
With the delays added in #1940, maybe there is a possibility to use them to our advantage.

In most cases, I would expect the title screen, system graphic, and title music to be included in the game assets and not be RTP. The face graphics in the load game menu are likely to be custom for most serious games also.

Given that, how about we do this:

1. Load all assets needed for title and load screen only. 
    * If none of them in RTP, skip RTP detection. 
    * If any are RTP, then we have to wait for RTP detection on startup.
2. Upon loading the title screen, immediately kick off a background thread to start detecting RTP. 
3. Once we new game / load game and wait for transition out and the 1 second delay, block on black screen if RTP detection still hasn't finished.

With this approach we could achieve parallel and deterministic loading of RTP.

fmatthew5876:
Can we push back the "delay rtp detection" piece to a follow up PR? The custom parser your wrote for wine results in a significant speedup and should go in sooner

Master
```
--------------------------------------------------                                                                                                                           
Benchmark           Time           CPU Iterations                                  
--------------------------------------------------                                                                                                                           
BM_InitRtp   31469504 ns   31464948 ns         22   
```

With new parser
```
--------------------------------------------------                                                                                                                            
Benchmark           Time           CPU Iterations                                                                                                                             
--------------------------------------------------                                                                                                                            
BM_InitRtp    5514361 ns    5512936 ns        137  
```

Ghabry:
Using a Lowercase func that works in-place gives a 25% speedup and completely removing the Lowercasing (incompatible with Windows registry behaviour) gives another 75% speedup.

```
With normal LowerCase:
-----------------------------------------------------
Benchmark           Time             CPU   Iterations
-----------------------------------------------------
BM_InitRtp   38521503 ns     38474798 ns           18

With in-place Lowercase func
-----------------------------------------------------
Benchmark           Time             CPU   Iterations
-----------------------------------------------------
BM_InitRtp   25587533 ns     25436706 ns           27

Without Lowercase:
-----------------------------------------------------
Benchmark           Time             CPU   Iterations
-----------------------------------------------------
BM_InitRtp    7270756 ns      7258756 ns           95
```

carstene1ns:
The last commit "lowercasing all in place" is wrong.
You lowercase the actual path. Also, you do it twice.

```
Debug: RTP path c:\\users\\carsten\\application data\\kadokawa\\common\\rpg maker 2003 rtp is invalid, not adding
```

```ini
[Software\\Wow6432Node\\KADOKAWA\\rpg2003] 1576949727
#time=1d5b82508a24236
"RuntimePackagePath"="C:\\users\\carsten\\Application Data\\KADOKAWA\\Common\\RPG Maker 2003 RTP"

```

Ghabry:
ups. Removed the in-place lowercasing when inside of the section, thx. Won't hurt the parsing speed much as the section-searching is the slow part.

PR #1917, opened 2019-09-26
Preparations for new emscripten backend

Ghabry:
New build line is:
```
emcc -g0 -O2 -v --llvm-lto 1 -s ASSERTIONS=0 -s NO_EXIT_RUNTIME=1 \
    -s ALLOW_MEMORY_GROWTH=1 -s DEMANGLE_SUPPORT=1 \
    -s ASYNCIFY \
    -lidbfs.js \
    easyrpg-player.bc \
    -o player-js/index.html \
    --pre-js resources/emscripten/emscripten-pre.js \
    --post-js resources/emscripten/emscripten-post.js \
    --shell-file resources/emscripten/emscripten-shell.html
```

Needed to split the argument handling because the generated code changes:
Arguments must be prepared after the pre-js-file was executed.

Currently not added because of problems with the instrumentation (?): "-s ASYNCIFY_IGNORE_INDIRECT"

sleep_for: ``_ZNSt11this_thread9sleep_forIlSt5ratioILl1ELl1EEEEvRKNSt6chrono8durationIT_T0_EE``

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -203,11 +203,8 @@ void Player::Run() {
 	FrameReset();
 
 	// Main loop
-#ifdef EMSCRIPTEN
-	emscripten_set_main_loop(Player::MainLoop, 0, 0);
-#elif defined(USE_LIBRETRO)
-	// Do nothing
-#else
+	// libretro invokes the MainLoop through a retro_run-callback
+#ifndef USE_LIBRETRO
 	while (Graphics::IsTransitionPending() || Scene::instance->type != Scene::Null) {
 #  if defined(_3DS)
 		if (!aptMainLoop())
@@ -332,9 +329,6 @@ void Player::Update(bool update_scene) {
 
 	BitmapRef disp = DisplayUi->GetDisplaySurface();
 
-#ifdef EMSCRIPTEN
-	Graphics::Draw(*disp);
-#else
 	cur_time = (double)DisplayUi->GetTicks();
 	if (cur_time < next_frame) {
 		Graphics::Draw(*disp);
@@ -343,11 +337,10 @@ void Player::Update(bool update_scene) {
 #if !defined(USE_LIBRETRO)
 		// Still time after graphic update? Yield until it's time for next one.
 		if (cur_time < next_frame) {
-			DisplayUi->Sleep((uint32_t)(next_frame - cur_time));
+			DisplayUi->Sleep(static_cast<uint32_t>(next_frame - cur_time));
 		}
 #endif
 	}
-#endif
 }
 
 void Player::IncFrame() {
@@ -376,8 +369,6 @@ int Player::GetFrames() {
 
 void Player::Exit() {
 #ifdef EMSCRIPTEN
-	emscripten_cancel_main_loop();
-
 	BitmapRef surface = DisplayUi->GetDisplaySurface();
 	std::string message = "It's now safe to turn off\n      your browser.";
 
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -168,10 +168,6 @@ void Player::Init(int argc, char *argv[]) {
 	// Create initial directory structure in our private area
 	// Retrieve save directory from persistent storage
 	EM_ASM(({
-
-		FS.mkdir("easyrpg");
-		FS.chdir("easyrpg");
-
 		var dirs = ['Backdrop', 'Battle', 'Battle2', 'BattleCharSet', 'BattleWeapon', 'CharSet', 'ChipSet', 'FaceSet', 'Frame', 'GameOver', 'Monster', 'Movie', 'Music', 'Panorama', 'Picture', 'Sound', 'System', 'System2', 'Title', 'Save'];
 		dirs.forEach(function(dir) { FS.mkdir(dir) });
 
--- a/src/scene.h
+++ b/src/scene.h
@@ -76,7 +76,7 @@ public:
 	 * executes the scene that is currently on the top of
 	 * the stack.
 	 */
-	virtual void MainFunction();
+	void MainFunction();
 
 	/**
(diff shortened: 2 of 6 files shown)


Ghabry:
obviously can't be tested yet because it needs the new emscripten backend installed :)

Ghabry:
What I'm kinda missing yet is something like -s EMTERPRETER_ADVISE which tells me which functions are asynchified.

src/scene.h
```diff
@@ -72,7 +72,7 @@ class Scene {
 	 * executes the scene that is currently on the top of
 	 * the stack.
 	 */
-	virtual void MainFunction();
+	void MainFunction();
```
Ghabry: made that change because it is on the stack when emscripten_sleep is called, so there shouldn't be multiple implementations of it -> simpler instrumentation

src/sdl2_ui.cpp
```diff
@@ -153,7 +153,9 @@ uint32_t Sdl2Ui::GetTicks() const {
 }
 
 void Sdl2Ui::Sleep(uint32_t time) {
-#ifndef EMSCRIPTEN
+#ifdef EMSCRIPTEN
+	emscripten_sleep(time);
+#else
 	SDL_Delay(time);
```
Ghabry: the docs say SDL_Delay is supposed to work, but it freezes the browser when I use.

Ghabry:
added "-s BYSYNCIFY_IGNORE_INDIRECT" to the compiler line. This tells the asyncifier that no indirect calls will be on the stack when sleep is called.

Ghabry:
Added -lidbfs.js and BISYNCHIFY is now ASYNCHIFY again (was fastcomp removed?).

Current status:
Fails with

```
Uncaught RuntimeError: unreachable
    at Sdl2Ui::RefreshDisplayMode() (wasm-function[2883]:0x1ac512)
    at Sdl2Ui::EndDisplayModeChange() (wasm-function[2882]:0x1a9fb4)
    at Sdl2Ui::Sdl2Ui(long, long, bool, int) (wasm-function[2876]:0x1a8f19)
    at BaseUi::CreateUi(long, long, bool, int) (wasm-function[283]:0x95db)
    at Player::Init(int, char**) (wasm-function[1884]:0x105140)
```

src/sdl2_ui.cpp
fmatthew5876: How about `std::this_thread::sleep_for`? The clock PR switches to that. We could add it earlier if needed for emscripten.

Ghabry:
The problem is that main() is invoked twice and I have no idea why yet.

src/sdl2_ui.cpp
Ghabry: Possible but you have to tell me the mangled name of this function. Must be added to the asyncify list.

Ghabry: Somehow it works now with SDL_Delay so I keep this untouched for now and just removed the ifdef.
sleep_for also works, tested it but you already make this change in your PR.

Ghabry:
"Ready to Review". Requires https://github.com/EasyRPG/buildscripts/pull/52 to be merged and the toolchain recompiled.
I would prefer to keep this PR this minimal as the Game_Clock stuff touches all the Update stuff again.

I found the reason: "-s ASYNCIFY_IGNORE_INDIRECT" doesn't work properly. Guess not providing this gives a small speed penalty as more functions need to test if the callstack unwinds/rewinds. No idea why this worked months ago and now not anymore. Maybe this fixes itself with a later emscripten release but I don't consider this a blocker, still better than what we currently have.

src/player.cpp
```diff
@@ -207,9 +203,7 @@ void Player::Run() {
 	FrameReset();
 
 	// Main loop
-#ifdef EMSCRIPTEN
-	emscripten_set_main_loop(Player::MainLoop, 0, 0);
-#elif defined(USE_LIBRETRO)
+#if defined(USE_LIBRETRO)
```
carstene1ns: Vacuumize this to `#ifndef` and a better comment

Ghabry:
Not part of this PR but the obligatory CMake stuff:
https://github.com/emscripten-core/emscripten/pull/10184

Fun fact: Initially I contributed this to vitasdk but somebody else took the same code, even with my PR message and created another PR with it: https://github.com/leetal/ios-cmake/pull/36  xD

carstene1ns:
The final build command needs to be fixed yet for `player-js` and `player-js-pr`.
Also the asm.js stuff should be removed.

Ghabry:
Updated the jobs, just waiting for jenkins now.
Also asm.js is currently still in because I need this to verify that frame-skipping works xD

Ghabry:
Everything updated:
Firefox is much slower than Chromium for me but frameskip works in both.

Test cases: (asmjs for testing frameskip)
https://easyrpg.org/play/pr2028/?game=frozen_triggers&engine=rpg2k3e
https://easyrpg.org/play/pr2028/asmjs/?game=frozen_triggers&engine=rpg2k3e

asmjs became in general much faster. This is with "enable memory growth" so it is actually not asm.js code but normal js code that is JITed :o

fmatthew5876:
In the old setup we used `emscripten_set_main_loop` where the docs recommend you set the frame rate to 0 to auto detect the browser refresh rate and use that.

Now that we're using asyncify, how do you configure optimal refresh rate? Do you use vsync in sdl?

The new main loop code actually makes this stuff pretty easy, even with the old emscripten setup we used to have.

Ghabry:
This was a complete waste of time.
#2017 is faster with ``emscripten_set_main_loop`` than with ASYNCIFY.

Additionally frameskipping, which was deemed as "not possible with emscripten" after months of experimenting, is suddenly working.

PR #1918, opened 2019-10-02
window_equipstatus: Draw fullwidth arrow in Japanese games

kakurasan:
See screenshots.

RPG_RT on Windows:
equip_arrow_rpgrt
EasyRPG Player (Unpatched):
equip_arrow_unpatched
EasyRPG Player (Patched):
equip_arrow_patched

### DIFF
--- a/src/window_equipstatus.cpp
+++ b/src/window_equipstatus.cpp
@@ -22,6 +22,7 @@
 #include "window_equipstatus.h"
 #include "bitmap.h"
 #include "font.h"
+#include "player.h"
 
 Window_EquipStatus::Window_EquipStatus(int ix, int iy, int iwidth, int iheight, int actor_id) :
 	Window_Base(ix, iy, iwidth, iheight),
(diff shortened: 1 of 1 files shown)


kakurasan:
Official Japanese version of RPG Maker 2k (called "RPG Tkool 2000") generates the `RPG_RT.exe`. This means all Japanese rpg2k games contain the version of `RPG_RT.exe`.

Ghabry:
@kakurasan 
The item count uses a : instead of X in your screenshot. Is that also Jp rpgrt specific?

kakurasan:
I don't know much about non-Japanese rm2k/2k3 games, but screenshots in [this article](https://rpgmaker.net/articles/1307/) show that English games, too, draw ":".

I think fullwidth "：" is more suitable than halfwidth ":".
Fullwidth:
item_num_fullwidth_colon
Halfwidth:
item_num_halfwidth_colon

Ghabry:
hm, this : vs x should be checked. but not here. merging

Issue #1929, opened 2019-11-03
Home crash


DuendeInexistente:
Easyrpg-Player: f5230a176ccc71392979b7e17bb8146a753da538
LibLCF: https://github.com/EasyRPG/liblcf/commit/47db42b6977e33591dc5ef1f7644844096d5a7bd

---
#### Player platform:
Ubuntu
#### Attach files (as a .zip archive or link them)
[save.zip](https://github.com/EasyRPG/Player/files/3801187/save.zip)

#### Describe the issue in detail and how to reproduce it:
Upon beginning a fight (ID 103) the game crashes. I tried running it with --no-audio, but it's having a segmentation fault no mater what I do. Triggering the battle manually in the debug menu does not cause this crash.
You can play from the uploaded savegame going up and evading the chasers, or simply use the teleport map to teleport to X30 Y10 (Or the other way around?) in map 145 and using the ghosts around the batter to trigger the battle.

fmatthew5876:
That `save.zip` contains a Map file and a log file. Did you mean to upload saved game?

Please upload a save if you can, it will make it much easier to debug and fix this issue.

DuendeInexistente:
[save.zip](https://github.com/EasyRPG/Player/files/3827748/save.zip)
Huh, weird, but here it is.

fmatthew5876:
The crash is here:

```
#2  0x0000555555a2ad1d in Spriteset_Battle::FindBattler (this=0x0, battler=0x60d000093620) at /home/matt/EasyRPG/hh3-rm2k/dev/easyrpg/Player/src/spriteset_battle.cpp:105  
#3  0x0000555555a652e1 in BattleAnimationBattle::FlashTargets (this=0x612000272140, r=0, g=0, b=0, p=0)
    at /home/matt/EasyRPG/hh3-rm2k/dev/easyrpg/Player/src/battle_animation.cpp:328   
#4  0x0000555555a64346 in BattleAnimation::UpdateTargetFlash (this=0x612000272140) at
/home/matt/EasyRPG/hh3-rm2k/dev/easyrpg/Player/src/battle_animation.cpp:223          
#5  0x0000555555a62372 in BattleAnimation::Update (this=0x612000272140) at /home/matt/EasyRPG/hh3-rm2k/dev/easyrpg/Player/src/battle_animation.cpp:77                      
[...]

PR #1933, opened 2019-11-10
Fix battle event battle animation crash

fmatthew5876:
When a battle event does an animation as the first command, which
also triggers a flash effect on the first frame the game crashes.

This happens because SpriteSet_Battle::Update() is called within
it's constructor before the spriteset unique_ptr is set within
Game_Battle.

Fix: #1929


fmatthew5876:
Calling `Update()` during construction like this probably causes frame timing issues with battle animations. That issue is *not* addressed here.

Issue #1939, opened 2019-11-11
Loading time


Br4ssman:
This is in general, using Player 0.6.1 for windows 10.

I note RPG_RT makes fade screen to black when press slot to load and fade out when the game is loaded (the black screen  remains between 1 - 2 sec), whereas Player makes load game immediately (sharply)

2 examples RPG_RT and Player.

[Player-load.zip](https://github.com/EasyRPG/Player/files/3832100/Player-load.zip)
[RPG_RT-load.zip](https://github.com/EasyRPG/Player/files/3832101/RPG_RT-load.zip)

fmatthew5876:
The issue doesn't appear to be with the transitions, but the lack of the black screen delay between transitions. I don't know how many frames it lasts.

I also observe this with:

- [ ] New Game
- [ ] Load Game
- [ ] Return to Title Screen

Ghabry:
you can try calling Player::FrameReset after the long taking function call.

fmatthew5876:
Well I did a screen capture with simplevideorecorder on linux. It looks like it *might* be a delay of around 32 frames.

PR #1940, opened 2019-11-17
Add RPG_RT compatible scene delays

fmatthew5876:
* New Game -> 60 frames
* Load Game -> 60 frames
* Return to Title -> 20 frames

Fix: #1939

### DIFF
--- a/src/scene.cpp
+++ b/src/scene.cpp
@@ -34,6 +34,8 @@
 #endif
 
 
+constexpr int Scene::kStartGameDelayFrames;
+constexpr int Scene::kReturnTitleDelayFrames;
 std::shared_ptr<Scene> Scene::instance;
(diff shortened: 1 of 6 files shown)


fmatthew5876:
@Br4ssman Could you please confirm this fixes the issues in your game

Br4ssman:
It works, 100/10.

src/scene.cpp
```diff
@@ -34,6 +34,8 @@
 #endif
 
 
+constexpr int Scene::kStartGameDelayFrames;
+constexpr int Scene::kReturnTitleDelayFrames;
```
carstene1ns: Why are these needed? These should get pulled in from the header.

fmatthew5876: static data members need definitions. Without them you'll likely get linker errors in debug builds as unoptimized code is trying to find the address of these constants.

PR #1947, opened 2019-11-21
Transition Random blocks up/down were down/up

Ghabry:
Found a while ago (iirc it was smivan) in a Yume Nikki comparison video.

Not sure if other transitions have the same problem, but here it was obvious because the menu items were named "Random blocks from above/below" but we did "Random block Up/down" (Up/Down as in "upwards/downwards").

Tested with:
 - [x] Transition predefined in the database and teleporting
 - [x] Changing transition through an event command and teleporting
 - [x] Show/Hide Screen event command


fmatthew5876:
Nice catch

Issue #1946, opened 2019-11-19
Picture wave must be more accurate


Br4ssman:
**Name of the game**: Super Crystal Hunter
**Player platform**: Player 0.6.1, Win10 x64
**Describe the issue in detail and how to reproduce it**:

I noted (from my last issue https://github.com/EasyRPG/Player/issues/1945) the wave doesn't make the same effect in Player, maybe for the same reason (new features from 2k3).

Check out the results in video:
https://app.box.com/s/dt7pwwyz4tlfkam534l1qf0ezdzoludt

Also I send you a small a test game with the picture and custom snowfall.
[wave_picture.zip](https://github.com/EasyRPG/Player/files/3862702/wave_picture.zip)

fmatthew5876:
As shown in Player #1703 Test case 7,

The wave effect does a full cycle in 32 frames. From a simple test it looks like player is slower than RPG_RT.

PR #1949, opened 2019-11-23
Fix picture waver effect

fmatthew5876:
* Animate LSD chunks accurately
* Fix visual effect speed, period, and displacement

Fix: #1946 

@Br4ssman please confirm

A good way to test this is with this event:
```
ShowPicture: Wave effect w/ some power
Loop:
  HideScreen Fade Out
  ShowScreen Fade In
  Wait 0.5s
  Wait 0.0s
  Wait 0.0s
```

The screen fades will happen between 1 complete cycle (32 frames), regardless of wave power.

Here are some good images to test the fact that the wave effect period is 32 pixels:
waver32
waver64

I tested the following scenarios:

- [x] Wave effect with different powers
- [x] Wave effect with different starting x and y positions
- [x] Wave effect with zoom
- [x] Wave effect during movement

### DIFF
--- a/src/bitmap.cpp
+++ b/src/bitmap.cpp
@@ -16,6 +16,7 @@
  */
 
 // Headers
+#define _USE_MATH_DEFINES
 #include <cmath>
 #include <cstdlib>
 #include <cstring>
@@ -739,8 +740,8 @@ void Bitmap::WaverBlit(int x, int y, double zoom_x, double zoom_y, Bitmap const&
 			continue;
 		if (dy >= this->height())
 			break;
-		int sy = static_cast<int>(std::floor((i+0.5) / zoom_y));
-		int offset = (int) (2 * zoom_x * depth * sin((phase + (src_rect.y + sy) * 11.2) * 3.14159 / 180));
+		double sy = i * 360.0 / (32.0 * zoom_y);
+		int offset = 2 * zoom_x * depth * std::sin((phase + sy) * M_PI / 180);
 
 		pixman_image_composite32(src.GetOperator(mask),
 								 src.bitmap, mask, bitmap,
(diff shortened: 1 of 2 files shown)


fmatthew5876:
jenkins: test this please

Br4ssman:
It seems the picture wave has improved slightly. But I note the picture wave in RPG_RT is more curved (or maybe the effect is stronger even if use wave:1) in Player the wave is softer.

fmatthew5876:
Maybe floating point or rounding related? I'll take a further look

fmatthew5876:
Here is my wave64 image with 2x zoom. They look identical to me?

wave

fmatthew5876:
Actually using your test game, it looks related to picture movement, something I didn't catch. If you disable the movement, the effects match.

fmatthew5876:
There looks to be some weird effect when the picture is off the screen. This is a 320x240 cropped version of your rain with vertical down movement. Notice how the effect changes halfway through.

rain

fmatthew5876:
As indicated in irc, the discrepancy is due to the fact that if an image crosses the top of the screen, the phase in y is computed relative to the Y coordinate at the top of the screen instead of being computed relative to the Y coordinate of the image.

The effect in the Crystal Hunter is actually an RPG_RT bug from this calculation that makes images which wave and move down from the top of the screen wave faster than they normally would. You can clearly see the speed change in my previous gif that occurs as soon as the top edge of the image comes into view.

Since some games may depend on this behavior, I added a commit to emulate it.

This should be good to go now. @Br4ssman

Br4ssman:
It may be too much work to emulate certain strange behaviors of RPG_RT and maybe it should not be done.

Users simply must assume that moving from RPG Maker to EasyRPG brings improvements, such as this case that you have been tackling. Emulate in detail something that is a bad behavior is a waste of time.

Anyway, thanks for the hard work!

Ghabry:
Interesting find!
matthew commented how to disable this bug emulation, so it's okay for me to keep this in and we can simply opt-out later when all editor stuff is ever ready ^^'

fmatthew5876:
I am leaning towards developing a feature flags system where one can configure between compatible behavior or not.  This would drive customizations as well as bug fixes.

PR #1955, opened 2019-11-28
Fix Vehicle Teleport transition regression

fmatthew5876:
Previous async refactor caused a regression where screen fade in would
occur for vehicle teleports. Add an extra flag to handle it.


fmatthew5876:
Test Cases:

```
VehicleTeleport Other map
```

Result: Instant transition

```
HideScreen
VehicleTeleport Other map
```

Result: Screen stays hidden

PR #1909, opened 2019-09-22
Refactor Game_Message / Window_Message interface - Messages 3 of N

fmatthew5876:
Depends on: #1900 
Fix: #1908 
Fix: #1873 
Fix: #1909

This PR refactors the messaging interface to better separate concerns between Game_Message and Window_Message. It also fixes more frame timing edge cases from #1706 

Tests from #1706
- [x] Tests 1-11 - All pass
- [x] Tests 13-16 - All pass
- [x] Tests 19-20 - All pass
- [x] Tests 21-25 - All Pass
- [x] Tests 30-32 - All Pass
- [x] Test 33 - Passes with #1955
- [x] Tests 34-38 - All Pass

Failing tests #1706:
- [ ] Test 12 - Not 100% compatible
    * Doesn't crash when RPG_RT crashes
    * Embedded variables supported in Player but not RPG_RT
    * Embedded variables in actor name itself also supported in Player but not RPG_RT
- [ ] Tests 17 - Doesn't accurately emulate the glitch in test 1, everything else works
- [ ] Tests 26 - Inn cost with wait 0.5s
- [ ] Tests 28 - Multiple level up messages don't work at all
- [ ] Tests 29 - Doesn't behave right with message
- [ ] Tests 32 - Fails because we don't render the transitions properly. Message timing is accurate.

Passing with/without minor glitches #1706:
- [ ] Tests 18 - We don't emulate the glitch
- [ ] Tests 27 - Same behavior, but doesn't emulate glitches
    * No Wait - Message box doesn't appear briefly
    * Wait 0.5s - Box doesn't close when game over scene happens
- [ ] Tests 31 - We don't emulate corruption, and probably don't care

PR #1956, opened 2019-11-29
Refactor param parsing methods + use utf8 + level up msg fix - Messages 4 of N

fmatthew5876:
Depends on #1909 

Moves param parsing methods to `Game_Message` so they can be used generically. Adds unit tests. It also changes `Window_Message::text` to be utf8 and uses `UTF8Next` to iterate the text. 

This optimizes our message processing by removing roundtrips between utf8 -> utf32 -> utf8. It reduces the memory footprint of the text and also spreads the cost of utf8 parsing over the frames of message rendering instead of doing it upfront.

In later PR's we'll eliminate `Window_Message::text` entirely and do everything from `PendingMessage`, removing the additional memory copies.

Fixes test cases of #1706:
- [x] Test cases 12 - includes supporting recursive variables
- [x] Test cases 39
- [x] Tests 28 & 29 for level up

### DIFF
--- a/src/pending_message.cpp
+++ b/src/pending_message.cpp
@@ -110,7 +110,7 @@ std::string PendingMessage::ApplyTextInsertingCommands(std::string input, uint32
 
 			const auto* actor = ReaderUtil::GetElement(Data::actors, value);
 			if (!actor) {
-				Output::Error("Invalid Actor Id %d in message text", value);
+				Output::Warning("Invalid Actor Id %d in message text", value);
 			} else{
 				output.append(actor->name);
 			}
--- a/src/pending_message.cpp
+++ b/src/pending_message.cpp
@@ -1,4 +1,12 @@
 #include "pending_message.h"
+#include "game_variables.h"
+#include "game_message.h"
+#include "data.h"
+#include "reader_util.h"
+#include "output.h"
+#include "utils.h"
+#include "player.h"
+#include "main_data.h"
 #include <cassert>
 #include <cctype>
 #include <algorithm>
@@ -11,6 +19,7 @@ static void RemoveControlChars(std::string& s) {
 
 int PendingMessage::PushLineImpl(std::string msg) {
 	RemoveControlChars(msg);
+	msg = ApplyTextInsertingCommands(std::move(msg), Player::escape_char);
 	texts.push_back(std::move(msg));
 	return texts.size();
 }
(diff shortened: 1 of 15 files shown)


fmatthew5876:
I retested the message timing test cases and they pass. This is ready.

fmatthew5876:
Also retested word wrap code in battle, still works.

fmatthew5876:
Updated the parse test to support test_runner build.

fmatthew5876:
I was under the mistaken impression that level up / class change messages required some complicated append message mechanism. The truth is far simpler, so I added the fix for level up messages to this PR so they go in sooner.

Currently the level up text in battles is broken until this PR goes in.

tests/parse.cpp
```diff
@@ -0,0 +1,468 @@
+#include "game_message.h"
```
Ghabry: Please add the test to Makefile.am

src/pending_message.cpp
```diff
@@ -60,3 +69,73 @@ void PendingMessage::SetChoiceResetColors(bool value) {
...
+		++iter;
+
+		if (ch == 'N' || ch == 'n') {
+			auto parse_ret = Game_Message::ParseActor(iter, end, escape_char, true);
+			iter = parse_ret.next;
+			int value = parse_ret.value;
+
+			const auto* actor = ReaderUtil::GetElement(Data::actors, value);
+			if (!actor) {
+				Output::Error("Invalid Actor Id %d in message text", value);
```
Ghabry: Why "Error"?
This has no negative side-effects.
Years ago this showed a warning but was somehow lost o.O
https://github.com/EasyRPG/Player/blob/6b12808f1a701eebde5882413932a263181c4ea6/src/window_message.cpp#L535

fmatthew5876: RPG_RT crashes in this case. Do we prefer silent warning?

tests/parse.cpp
fmatthew5876: Added

src/pending_message.cpp
fmatthew5876: Thought about it and changed this to warning. After thinking about it, it seems pretty disruptive to crash the game and make the player lose all their progress because of 1 malformed message.

Ghabry: Funny for invalid read would be some DOS-like error handler: Retry / Abort / Continue but "warn and continue" is what I consider appropriate here.

PR #1959, opened 2019-12-01
Flatten Bitmap formats_map into an array

fmatthew5876:
Put most common formats first and do a linear scan.

Shown to be faster than map, unordered_map, and
lower_bound on sorted array.

I have a google bench setup in another branch which is part of the message PR queue. I wanted to present this one earlier since it's standalone.

```
static void BM_FindFormatSingle(benchmark::State& state) {
    const auto fmt = format_R8G8B8A8_a().format();
    for (auto _: state) {
        BitmapAccess::find_format(fmt);
    }
}

BENCHMARK(BM_FindFormatSingle);

static void BM_FindFormat(benchmark::State& state) {
    for (auto _: state) {
        BitmapAccess::find_format(format_R8G8B8A8_a().format());
        BitmapAccess::find_format(format_B8G8R8A8_a().format());
        BitmapAccess::find_format(format_A8B8G8R8_n().format());
        BitmapAccess::find_format(format_B8G8R8A8_n().format());
    }
}

BENCHMARK(BM_FindFormat);
```

Master:
```
BM_FindFormatSingle         19 ns         18 ns   38761196
BM_FindFormat               59 ns         58 ns   11620732
```

This PR:
```
BM_FindFormatSingle          3 ns          3 ns  227203497
BM_FindFormat               13 ns         13 ns   54225434
```

PR #1960, opened 2019-12-01
Make DynamicFormat and Bitmap formats_map constexpr

fmatthew5876:
Depends on #1959 

Make this a separate PR because it likely breaks GCW0 build despite being C++14 compliant.

This allows us to remove lazy initialization. It also removes startup runtime code to construct the mapping.


fmatthew5876:
jenkins test this please

PR #1962, opened 2019-12-02
Add micro benchmark framework

fmatthew5876:
https://github.com/google/benchmark

Enabled by `-DPLAYER_ENABLE_BENCHMARKS=ON` in cmake. Turned off by default.

Only when build is enabled, we pull in the libbenchmark dependency.

I used this to get the numbers in #1959

Also adds a bunch of smaller changes to make it easier to write automated tests

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -55,7 +55,6 @@ namespace {
 	cache_effect_type cache_effects;
 
 	std::string system_name;
-	BitmapRef default_system;
 
 	std::string system2_name;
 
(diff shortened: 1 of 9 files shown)


src/cache.cpp
```diff
@@ -55,30 +59,29 @@ namespace {
...
+		auto cur_ticks = clock::now();
 
 		for (auto& i : cache) {
 			if (i.second.bitmap.use_count() != 1) {
 				// Bitmap is referenced
 				continue;
 			}
 
-			if (cache_size <= cache_limit && cur_ticks - i.second.last_access < 3000) {
+			if (cache_size <= cache_limit && cur_ticks - i.second.last_access < 3000s) {
```
Ghabry: is this correct? The original check is for 3000ms/3s

Ghabry:
The benchmark stuff is not invasive and disabled by default, so I'm fine with this change.

Only problem I see is that contrary to unit tests the benchmarks are not part of PR reviews, so they will likely break when the API changes.

src/cache.cpp
carstene1ns: yeah, he force-pushed this away. seems like an oversight.

fmatthew5876: Was a bug on my part, fixed.

fmatthew5876:
> Only problem I see is that contrary to unit tests the benchmarks are not part of PR reviews, so they will likely break when the API changes.

We always have the option to enable them in PR builds if we want. Maybe only for a subset if the library dependency becomes onerous.

PR #1963, opened 2019-12-02
Make bitmapfont objects constexpr

fmatthew5876:
* Move to headers
* Add test

### DIFF
--- a/src/bitmapfont_wqy.h
+++ b/src/bitmapfont_wqy.h
@@ -1,6 +1,10 @@
-#include "bitmapfont.h"
+#ifndef EP_BITMAPFONT_WQY_H
+#define EP_BITMAPFONT_WQY_H
 
-BitmapFontGlyph const BITMAPFONT_WQY[20941] = {
+#include <array>
+#include "bitmapfont_glyph.h"
+
+constexpr const std::array<BitmapFontGlyph,20941> BITMAPFONT_WQY = {{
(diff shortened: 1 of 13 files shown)


fmatthew5876:
Unfortunately this may be impossible with MSVC 2017. I haven't found any supporting articles, but if you go to `bitmapfont_wqy.h` and `#if 0 .. #endif` all of the initializers after 24110 it compiles.

It looks like a limitation related to the size of the initializer expression because smaller data sets do compile as constexpr.

fmatthew5876:
This appears to build in visual studio 2019.

fmatthew5876:
jenkins test this please

carstene1ns:
This should involve a generator update: https://github.com/EasyRPG/Player/blob/master/resources/shinonome/generate_cxx_font.rb

fmatthew5876:
Updated the ruby script and regenerated the headers

src/bitmapfont_wqy.h
```diff
@@ -1,6 +1,10 @@
-#include "bitmapfont.h"
+#ifndef EP_BITMAPFONT_WQY_H
```
carstene1ns: seems like this one is not updated.
(missing `GENERATED FILE` stuff)

fmatthew5876: For some reason, this file is not in the ruby script so it's not a generated file??

fmatthew5876: I found the python script, updated  and reran it.

Ghabry:
Jenkins: test this please <3

PR #1966, opened 2019-12-02
Fix compiler warnings

fmatthew5876:
Compiles cleanly with gcc 8.3 and clang 8.0 with `-Wall -Wextra -Werror` on Ubuntu for me.

All except for a warning about unsigned `BGM_GetTicks()`. I didn't silence this warning yet as it indicates something we should actually revisit and maybe change this method to return something else.

### DIFF
--- a/bench/bitmap.cpp
+++ b/bench/bitmap.cpp
@@ -257,7 +257,6 @@ static void BM_MaskedBlit(benchmark::State& state) {
 	auto dst_rect = dest->GetRect();
 	auto src = Bitmap::Create(320, 240);
 	auto mask = Bitmap::Create(320, 240);
-	auto color = Color(255, 255, 255, 255);
 	for (auto _: state) {
 		dest->MaskedBlit(dst_rect, *mask, 0, 0, *src, 0, 0);
 	}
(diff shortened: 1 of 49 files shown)


fmatthew5876:
I added some more warning fixes found in the jenkins builds. This is ready now.

Ghabry:
Do you intend to also fix the warnings reported by MSVC?

fmatthew5876:
Not at this time, fixing those would mean littering the code with casts, and that scares me. If you add a cast and then change the types later but forget to change the casts, the compiler won't warn you.

Ghabry:
jenkins: test this please

fmatthew5876:
Added some new warnings fixes from recent PR's that went in. Once this gets merged, we can do all future development with `-Wall -Werror` to avoid adding warnings to the code.

fmatthew5876:
I added the 3ds build fix here

fmatthew5876:
Rebased to fix warnings in latest PRs. Also found a regression in the message PR where wordwrap would be enable in 2k battles, even for old versions of the engine.

fmatthew5876:
Removed it

Issue #1741, opened 2019-04-07
Destined: Boss picture moves upwards when damaged


Ghabry:
#### Name of the game:
Destined https://rmarchiv.tk/games/1922

#### Attach files (as a .zip archive or link them)

The official save that doesn't load in RPG_RT:
[Save12.zip](https://github.com/EasyRPG/Player/files/3051959/Save12.zip)

Custom save by me obtained via "--start-map-id 193 TestPlay Window --start-party 2" that loads in RPG_RT:
[Save14.zip](https://github.com/EasyRPG/Player/files/3051967/Save14.zip)
(Hero doesn't receive damage because it is Don and not Alex, not a bug)

Load the save with RPG_RT (note: For some reason RPG_RT is RPG_RT2.exe because the normal one is EasyRPG :/).

#### Describe the issue in detail and how to reproduce it:

Walk upwards, after the cutscene attack the small slimes by mashing DECISION, then the big slime shrinks and moves upwards.

Because the event code is only move picture (always at the same coordinate!) I assume this is a X/Y miscalculation of Move picture when combined with zoom changes.

fmatthew5876:
Whats the bug here? When I play this with RPG_RT and attack slimes, the big slime also moves up.

Ghabry:
did you see my note? "Load the save with RPG_RT (note: For some reason RPG_RT is RPG_RT2.exe because the normal one is EasyRPG :/)."

fmatthew5876:
I must have gotten this from a different place than you, because mine doesn't have `RPG_RT2.exe`

Anyone have a proper download link for this game? It's name is hard to google.

Ghabry:
Oh didnt notice that the DL link is missing. added

fmatthew5876:
Ok so for this one, the `RPG_RT2.exe` works as expected. The big slime stays put and just gets smaller. If I upgrade `RPG_RT.exe` to rm2ke, the bug we see in player happens.

So I guess there is something about these move picture commands where the position doesn't change in old versions of `RPG_RT.exe`

```
0000: code=MovePicture(11120) indent=0 str="" params=[1 0 160 -15 0 185 0 0 80 80 120 100 0 15325932 10 1]
0000: code=MovePicture(11120) indent=0 str="" params=[1 0 160 -15 0 160 0 0 80 80 120 100 0 5497976 10 1]
```

fmatthew5876:
[...]

PR #1969, opened 2019-12-03
Emulate legacy rm2k/3 move picture behavior + refactor engine accessors

fmatthew5876:
MovePicture command doesn't change position of pictures
with fixed_to_map chunk set.

Used in Destined game See: #1741

Confirmed with Don's rm2k and Japanese rm2k3 RPG_RT 1.0.3

### DIFF
--- a/src/game_screen.cpp
+++ b/src/game_screen.cpp
@@ -41,7 +41,6 @@ static int GetDefaultNumberOfPictures() {
 		return 50;
 	}
 	if (Player::IsRPG2k3()) {
-		// FIXME: Needs confirmation with 2k3 RPG_RT <= 1.0.4
 		return 40;
 	}
 	return 20;
--- a/src/game_screen.cpp
+++ b/src/game_screen.cpp
@@ -37,13 +37,16 @@ static int GetDefaultNumberOfPictures() {
 	if (Player::IsEnglish()) {
 		return 1000;
 	}
-	if (Player::IsMajorUpdatedVersion()) {
+	else if (Player::IsMajorUpdatedVersion()) {
 		return 50;
 	}
-	if (Player::IsRPG2k3()) {
+	else if (Player::IsRPG2k3Legacy()) {
 		return 40;
 	}
-	return 20;
+	else if (Player::IsRPG2kLegacy()) {
+		return 20;
+	}
+	return 0;
 }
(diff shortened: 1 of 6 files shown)


fmatthew5876:
I found a japanese RPG_RT 1.0.3 and discovered it also has the same move picture behavior, so I redid this to be more general.

Needs to be reviewed again.

src/game_screen.cpp
```diff
@@ -37,14 +37,16 @@ static int GetDefaultNumberOfPictures() {
 	if (Player::IsEnglish()) {
 		return 1000;
 	}
-	if (Player::IsMajorUpdatedVersion()) {
+	else if (Player::IsMajorUpdatedVersion()) {
 		return 50;
 	}
-	if (Player::IsRPG2k3()) {
-		// FIXME: Needs confirmation with 2k3 RPG_RT <= 1.0.4
+    else if (Player::IsRPG2k3Legacy()) {
```
carstene1ns: messed up spacing.

fmatthew5876: Fixed

src/player.h
```diff
@@ -148,37 +148,39 @@ namespace Player {
...
 	bool IsRPG2kE();
 
+	/** @return If engine is the official English RM2k3 release (v1.10) or newer. */
+	bool IsRPG2k3E();
+
+	/** @return If engine is RPG2kLegacy() or RPG2k3Legacy() */
+	bool IsLegacy();
+
+	/** @return If engine is RPG2kUpdated() or RPG2k3Updated() */
+	bool IsMajorUpdatedVersion();
```
Ghabry: You changed the range of the functions, e.g. before "IsMajorUpdated" was "Everything not Legacy" and now it is "Every not Legacy and not English/Steam" or are the comments misleading?
When this is intended we need a better naming here. Maybe split between 2kE and 2kJ and for J add additional split 2kJLegacy and 2kJUpdated?

fmatthew5876: I can change the comments, updated should just be `>=`

We do need better naming. And maybe even finer grained control. Because there are major features differences between these lines. For this PR I just wanted to clean it up a little and support legacy rm2k3, but not completely redo it as I am not yet sure what it should look like.

PR #1973, opened 2019-12-05
Refactor Time and Sleep - MainLoop 2 of N

fmatthew5876:
Depends on #2021 

This adds `Game_Clock` for getting the current time and and sleeping. 

- [x] Remove dependency on Ui for time keeping
- [x] Use highest precision time available on system
- [x] Prefer standard time functions over platform specific ones wherever possible

TBD:
- [x] Check if we can use `std::this_thread::sleep_for()` on 3ds
- [x] Check if we can use `std::this_thread::sleep_for()` on switch
- [x] Check if we can use `std::this_thread::sleep_for()` on vita

### DIFF
--- a/src/3ds_ui.cpp
+++ b/src/3ds_ui.cpp
@@ -147,11 +147,6 @@ CtrUi::~CtrUi() {
 	}
 }
 
-void CtrUi::Sleep(uint32_t time) {
-	u64 nsecs = time * 1000000;
-	svcSleepThread(nsecs);
-}
-
 static inline double u64_to_double(u64 value) {
 	return (((double)(u32)(value >> 32)) * 0x100000000ULL + (u32)value);
 }
--- a/src/3ds_ui.cpp
+++ b/src/3ds_ui.cpp
@@ -147,10 +147,6 @@ CtrUi::~CtrUi() {
 	}
 }
 
-static inline double u64_to_double(u64 value) {
-	return (((double)(u32)(value >> 32)) * 0x100000000ULL + (u32)value);
-}
-
 void CtrUi::BeginDisplayModeChange() {
 	// no-op
 }
--- a/src/3ds_ui.cpp
+++ b/src/3ds_ui.cpp
@@ -156,11 +156,6 @@ static inline double u64_to_double(u64 value) {
 	return (((double)(u32)(value >> 32)) * 0x100000000ULL + (u32)value);
 }
 
-uint32_t CtrUi::GetTicks() const {
-	double ticks = u64_to_double(svcGetSystemTick());
-	u64 usecs = (u64)(ticks / ticks_per_msec);
-	return usecs;
-}
 void CtrUi::BeginDisplayModeChange() {
 	// no-op
 }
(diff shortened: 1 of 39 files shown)


fmatthew5876:
Is there any known platform where `std::this_thread::sleep_for()` doesn't work? I just added a commit here to get rid of our `Sleep()` functions which I hope we can keep.

Ghabry:
> Is there any known platform where std::this_thread::sleep_for() doesn't work?

Hard to answer, would need a test on Wii, Vita, Switch and 3DS. No idea how this is implemented when using newlib, could be a stub or working.

fmatthew5876:
This is getting folded into #2017

fmatthew5876:
I moved the clock code from #2017 here. Rebased everything and now this is ready.

For now I have kept the platform specific sleep functions for 3ds, switch, and psp2. Would love to know if we can get rid of these but that can also be done later.

src/game_clock.cpp
```diff
@@ -17,5 +17,28 @@
...
+void Game_Clock::SleepFor(std::chrono::nanoseconds ns) {
+#if defined(_3DS) || defined(__SWITCH__)
+	// FIXME: Can we use std::this_thread::sleep_for() for 3ds and switch?
+	svcSleepThread(ns.count());
+#elif defined(PSP2)
+	// FIXME: Can we use std::this_thread::sleep_for() for psp?
+	auto us = std::chrono::duration_cast<std::chrono::microseconds>(ns);
+	sceKernelDelayThread(us.count());
+#else
+	std::this_thread::sleep_for(ns);
```
Ghabry: based on the define of _GLIBCXX_USE_NANOSLEEP (they all use libstdc++) except for Vita all support ::nanosleep (which is called by sleep_for)

src/switch_ui.cpp
```diff
@@ -326,21 +326,10 @@ NxUi::~NxUi() {
 	deinitEgl();
 }
 
-void NxUi::Sleep(uint32_t time) {
-	u64 nsecs = time * 1000000;
-	svcSleepThread(nsecs);
-}
-
 static inline double u64_to_double(u64 value) {
```
carstene1ns: This one is useless now.

src/3ds_ui.cpp
```diff
@@ -147,20 +147,10 @@ CtrUi::~CtrUi() {
 	}
 }
 
-void CtrUi::Sleep(uint32_t time) {
-	u64 nsecs = time * 1000000;
-	svcSleepThread(nsecs);
-}
-
 static inline double u64_to_double(u64 value) {
```
carstene1ns: same here

src/game_clock.h
```diff
@@ -60,4 +69,18 @@ inline uint32_t Game_Clock::GetTicks() {
 	return ms.count();
 }
 
+constexpr int Game_Clock::GetSimulationFps() {
+	return DEFAULT_FPS;
+}
+
+/** Get the amount of time each logical frame should take */
```
carstene1ns: duplicated comment

fmatthew5876:
Rebased and addressed review comments

PR #1976, opened 2019-12-07
Move unit tests in a single executable (test_runner, lcf-like) and support FileFinder test

Ghabry:
Instead of building one executable per test file we create now a single "test_runner.exe". Code taken from liblcf.

Advantage:
- Much faster linking speed (especially on Windows)
- Smaller files (the single test executables are huge)
- The "check" target allows enabling unit tests on the buildbot for Windows and OSX via "cmake --build . --target check"

For FileFinder unit tests the build system provides now the path to the test directory.
I added folders with empty files that represent a project directory for suitable testing.


fmatthew5876:
Should double check all test files define their local functions and classes in an anonymous namespace, otherwise we could get symbol collisions with common names.

fmatthew5876:
This is going to be tricky with tests that exercise singletons. Now we have to be careful of singleton state across test suits interfering with each other.

Ghabry:
I don't consider fixing the Singleton problems a blocker for this PR but we must really get rid of most of the global state.
At least everything Player related (except maybe DisplayUi and Audio as they are global by Operation system design, but they are not critical) should be resettable by deleting some "Player instance" object.

carstene1ns:
I dislike the additional header file. A simple preprocessor definition should be sufficient.
Also, the tests will not run by default with autotools currently.

Ghabry:
oh, you mean passing via "-D". Good point why didn't think of this 😅😅😅

Ghabry:
@carstene1ns 
When I add
```
test_runner_LDFLAGS = -no-install

check-local:
        $(AM_V_at)./test_runner
```
to solve "Also, the tests will not run by default with autotools currently."
as in liblcf it doesn't work.
no-install reports "unrecognized command" by gcc

and check-local gives an unreadable sed error:
```
config.status:949: creating Makefile
config.status:949: creating resources/Doxyfile
config.status:949: creating config.h
config.status:1115: config.h is unchanged
config.status:1163: executing depfiles commands
config.status:1240: cd .       && sed -e '/# am--include-marker/d' Makefile         | make -f - am--depfiles
/tmp/Gmpvw9GM:5014: *** missing separator.  Stop.
config.status:1245: $? = 2
config.status:1249: error: in `/home/gabriel/Programmierung/easyrpg/easyrpg-player/a':
config.status:1251: error: Something went wrong bootstrapping makefile fragments
    for automatic dependency tracking.  Try re-running configure with the
    '--disable-dependency-tracking' option to at least be able to build
    the package (albeit without support for automatic dependency tracking).
See `config.log' for more details
```

carstene1ns:
Makefile targets use tabs, not spaces.

Ghabry:
wow. This is some incredible shitty error reporting then.
Thanks!

Ghabry:
Fixed the mixed spaces (are now tabs).
"test_runner_LDFLAGS = -no-install" is still not working though, is that important?

carstene1ns:
That is only needed for libtool, which Player does not use (anymore). Works for me.
(more info: https://autotools.io/libtool/wrappers.html)

Issue #1944, opened 2019-11-18
Picture settings (tint) doesn't work properly


Br4ssman:
**Name of the game**: Super Crystal Hunter
**Player platform**: Player 0.6.1, Win10 x64
**Describe the issue in detail and how to reproduce it**:

In my case, I use a picture in a spritesheet + loop
`@> Show Picture: 2, 'SCH-ente-estandar[1,4]', (103,50), 1x4, Anim. @10, Loop, 100%, 0%, M1`

Then, I apply a flash in this way:
```
@> Move Picture: 2, (103,50), 100%, 0%, @0.1s, Wait
@> Move Picture: 2, (103,50), 100%, 0%, @0.1s, Wait
```

The 1st move picture has picture settings > tint: RGB 180, and Sat 100.
The 2nd move picture has picture settings > tint: RGB 100 and sat 100.

If you compare this, you'll see in RPG_RT the "flash" is stronger, in Player is weaker.

fmatthew5876:
I just tried to reproduce this and I don't see a difference. I even used the same picture from your example game.

Br4ssman:
I don't know what else to add except a video-comparison where I still see a more powerful flash in RPG_RT and weaker in Player.

[Flash-via-picture.zip](https://github.com/EasyRPG/Player/files/3915865/Flash-via-picture.zip)

fmatthew5876:
Here is a tint screen of (200, 100, 100, 100) on of the RM2k Cloud.png picture

![tint](https://user-images.githubusercontent.com/1198466/70382049-94401980-1923-11ea-84f1-98821401c6ca.png)

Left is Player, right is RPG_RT.

We're definitely off.

Br4ssman:
I knew it wasn't just a problem in my eyes!

Ghabry:
Not merged yet -> Reopening

PR #1979, opened 2019-12-08
Fix tone Blending calculation

fmatthew5876:
Fix: #1944

@Br4ssman please confirm


Br4ssman:
Super fixed, thanks matt!

PR #1980, opened 2019-12-08
Window_ShopParty: render custom charset image sizes correctly

### LINKED ISSUE #1315: Shop loads CharSet images as if they would be prefixed with $ (dynamic size)
#### Name of the game: Final Tear 3
https://rpgmaker.net/games/3983/

[Final Tear 3.zip](https://github.com/EasyRPG/Player/files/1536844/Final.Tear.3.zip)

Is either Save12 or 13. The correct one is the one in the fields.

Walk down until you reach a city, go to the left and enter the shop.
When you want to buy something the character preview is wrong.
This is caused by the window buy code which assumes that the CharSet width is correct but the used file only contains one char. Imo the best way to solve this would be to configure custom Game_Characters and to simply render them into that field. Avoids code duplication.

fmatthew5876:
Fixes #1315.

### DIFF
--- a/src/sprite_character.cpp
+++ b/src/sprite_character.cpp
@@ -116,24 +116,32 @@ void Sprite_Character::OnTileSpriteReady(FileRequestResult*) {
 	Update();
 }
 
-void Sprite_Character::OnCharSpriteReady(FileRequestResult*) {
-	SetBitmap(Cache::Charset(character_name));
+Rect Sprite_Character::GetCharacterRect(const std::string& name, int index, const Rect bitmap_rect) {
+	Rect rect;
 	// Allow large 4x2 spriteset of 3x4 sprites
 	// when the character name starts with a $ sign.
 	// This is not exactly the VX Ace way because
 	// VX Ace uses a single 1x1 spriteset of 3x4 sprites.
-	if (character_name.size() && character_name.at(0) == '$') {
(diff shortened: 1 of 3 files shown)


src/sprite_character.cpp
```diff
@@ -120,26 +120,32 @@ void Sprite_Character::OnTileSpriteReady(FileRequestResult*) {
...
 	// Allow large 4x2 spriteset of 3x4 sprites
 	// when the character name starts with a $ sign.
 	// This is not exactly the VX Ace way because
 	// VX Ace uses a single 1x1 spriteset of 3x4 sprites.
-	if (character_name.size() && character_name.at(0) == '$') {
-		chara_width = GetBitmap()->GetWidth() / 4 / 3 * (TILE_SIZE / 16);
-		chara_height = GetBitmap()->GetHeight() / 2 / 4 * (TILE_SIZE / 16);
+	if (!name.empty() && name.front() == '$') {
+		rect.width = bitmap_rect.width * (TILE_SIZE / 16) / 4;
+		rect.height = bitmap_rect.width * (TILE_SIZE / 16) / 2;
```
Ghabry: that should be .height

fmatthew5876: Nice catch, fixed!

Ghabry:
but needs a rebase ;)

PR #1982, opened 2019-12-09
Refactor DrawableList + Battle Animation Crash Fix - Drawables 3 of N

fmatthew5876:
Depends on: #1977 

Fix: #1997 

This PR creates `DrawableList` and `DrawableMgr` classes to manage the draw list and adds a full suite of unit tests.


fmatthew5876:
Added the fix for #1997 here

fmatthew5876:
Added license to cpp files

PR #1983, opened 2019-12-10
Introduce PoolAllocator for Graphics Stack - Drawables 4 of N

fmatthew5876:
This PR adds a pool based memory allocator to improve the memory performance of small objects, of which many are allocated and they are often used together.

The benefits are:
* Improved cache locality because the objects are stored closer together in physical memory
* Extremely fast alloc()
* Extremely fast free()

TBD:
- [ ] Detailed Benchmarks showing performance differences
- [ ] Ability to tune the page size using INI file for games with different workloads.
- [ ] Logging and metrics on number of objects allocated, max memory usage, page size fit, etc..
- [ ] Figure out licensing. I would like a more liberal license for this allocator class so it can be reused more freely.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,7 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 add_library(${PROJECT_NAME}
 	src/3ds_ui.cpp
 	src/3ds_ui.h
+	src/align.h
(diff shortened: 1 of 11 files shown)


Ghabry:
Closing this due to lack of interest and progress. Please reopen when there are any updates.

PR #1985, opened 2019-12-15
Optimize SDL2 Renderer

fmatthew5876:
~~In our SDL rendering code we call `SDL_UpdateTexture` which is documented to be a very slow function and not intended for streaming textures.~~

~~https://wiki.libsdl.org/SDL_UpdateTexture~~

~~This PR uses the recommended Lock/Unlock methods instead.~~

Further testing shows at least on my machine despite what the docs say, `SDL_UpdateTexture` is faster on my machine.

### DIFF
--- a/src/sdl2_ui.cpp
+++ b/src/sdl2_ui.cpp
@@ -59,6 +59,69 @@ AudioInterface& Sdl2Ui::GetAudio() {
 }
 #endif
 
+static uint32_t GetDefaultFormat() {
+	return SDL_PIXELFORMAT_RGBA32;
+}
+
+/**
+ * Return preference for the given sdl format.
+ * Higher numbers are better, -1 means unsupported.
+ * We prefer formats which have fast paths in pixman.
+ */
+static int GetFormatRank(uint32_t fmt) {
+	switch (fmt) {
+		case SDL_PIXELFORMAT_RGBA32:
+			return 2;
+		case SDL_PIXELFORMAT_BGRA32:
+			return 2;
+		case SDL_PIXELFORMAT_ARGB32:
+			return 1;
+		case SDL_PIXELFORMAT_ABGR32:
+			return 0;
+		default:
+			return -1;
+	}
+}
+
+static DynamicFormat GetDynamicFormat(uint32_t fmt) {
+	switch (fmt) {
+		case SDL_PIXELFORMAT_RGBA32:
+			return format_R8G8B8A8_n().format();
+		case SDL_PIXELFORMAT_BGRA32:
+			return format_B8G8R8A8_n().format();
+		case SDL_PIXELFORMAT_ARGB32:
+			return format_A8R8G8B8_n().format();
+		case SDL_PIXELFORMAT_ABGR32:
+			return format_A8B8G8R8_n().format();
+		default:
+			return DynamicFormat();
+	}
+}
+
+static uint32_t SelectFormat(const SDL_RendererInfo& rinfo, bool print_all) {
+	uint32_t current_fmt = SDL_PIXELFORMAT_UNKNOWN;
+	int current_rank = -1;
+
+	for (int i = 0; i < static_cast<int>(rinfo.num_texture_formats); ++i) {
+		const auto fmt = rinfo.texture_formats[i];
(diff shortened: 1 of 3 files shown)


fmatthew5876:
Here are some accurate results from Frozen Triggers. Our old SDL2 logic was choosing a pixel format which was not in the list of formats supported by my GPU (Nvidia GTX 1080). Adding the auto detection logic resulted in a significant speedup.

Master
sdlold

This PR
sdlnew

fmatthew5876:
I added a pixel format benchmark to see what formats are fastest for blitting

```
-----------------------------------------------------
Benchmark              Time           CPU Iterations
-----------------------------------------------------
BM_BlitBGRA_a      14002 ns      14002 ns      49880
BM_BlitRGBA_a      14096 ns      14095 ns      50315
BM_BlitABGR_a     288853 ns     288850 ns       2410
BM_BlitARGB_a     132282 ns     132271 ns       5428
BM_BlitBGRA_n      10030 ns      10030 ns      72699
BM_BlitRGBA_n       9996 ns       9995 ns      68567
BM_BlitABGR_n     198156 ns     198147 ns       3335
BM_BlitARGB_n      10216 ns      10216 ns      68363
```

Ghabry:
Do you have a raspi or some other ARM device around for testing?

We made many tests with non-x86 platforms before and we noticed:
When you use RGBA instead of ABGR on little endian it becomes slow (no joke...).
And on Big endian (Wii/PowerPC) you must use RGBA because pixman swaps the byte orders... and otherwise it is slow....

fmatthew5876:
> Do you have a raspi or some other ARM device around for testing?
> 
> We made many tests with non-x86 platforms before and we noticed:
> When you use RGBA instead of ABGR on little endian it becomes slow (no joke...).
> And on Big endian (Wii/PowerPC) you must use RGBA because pixman swaps the byte orders... and otherwise it is slow....

The SDL_PIXELFORMAT_XXXX32 macros I use in this PR are actually endian safe wrappers.

https://wiki.libsdl.org/SDL_PixelFormatEnum?highlight=%28%5CbCategoryEnum%5Cb%29%7C%28CategoryPixels%29

On my intel machine (little endian), it uses SDL_PIXELFORMAT_ARGB888 (not endian aware) and I get the massive speed up noted above.


I squinted and this pretty hard and I think it's right, but it does need a test on

fmatthew5876:
I put this apk on my Pixel 2 android phone. The FPS for Frozen Triggers in the snow scene went from 35fps to 43fps.

Ghabry:
And why does ARGB get a ranking of 2 but in your code the ranking is 1?

fmatthew5876:
> And why does ARGB get a ranking of 2 but in your code the ranking is 1?

Rank 2 is for SDL_PIXELFORMAT_RGBA32

SDL_PIXELFORMAT_RGBA32 is an alias for SDL_PIXELFORMAT_RGBA8888 on big endian machines and for SDL_PIXELFORMAT_ABGR8888 on little endian machines

src/sdl2_ui.cpp
```diff
@@ -59,6 +59,73 @@ AudioInterface& Sdl2Ui::GetAudio() {
...
+					i, SDL_GetPixelFormatName(fmt), rank);
+		} else {
+			Output::Debug("SDL2: Detected format (%d) %s : Not Supported",
+					i, SDL_GetPixelFormatName(fmt));
+		}
+	}
+	if (current_rank >= 0) {
+		Output::Debug("SDL2: Selected format=%s", SDL_GetPixelFormatName(current_fmt));
+	} else {
+		Output::Debug("SDL2: None of the detected formats were supported! Falling back to %s. This will likely cause performance degregation ...",
```
fdelapena: s/degregation/degradation/ ?

fmatthew5876: Fixed, thanks!

fmatthew5876:
> I find the "Not supported" lines spammy, maybe remove them as they are not that useful.

I agree and put `#ifdef` around it.

fmatthew5876:
Cleaned this up a bit. Also changed it so that it prints all the formats when detection fails to find anything suitable.

PR #1986, opened 2019-12-15
Bitmap Improvements: PixmanImagePtr and Opacity - Bitmaps 1 of N

fmatthew5876:
This creates a simple RAII wrapper around pixman images which uses pixman's reference counting and provides a `shared_ptr` like interface.

No material change other than code cleanup. But this will be used later to optimize Bitmap and remove the need for `BitmapRef`

Tested this by playing Frozen triggers and everything works.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -170,6 +170,7 @@ add_library(${PROJECT_NAME}
 	src/message_overlay.h
 	src/meta.cpp
 	src/meta.h
+	src/opacity.h
 	src/options.h
 	src/output.cpp
 	src/output.h
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -177,6 +177,7 @@ add_library(${PROJECT_NAME}
 	src/pending_message.cpp
 	src/picojson.h
 	src/pixel_format.h
+	src/pixman_image_ptr.h
 	src/plane.cpp
 	src/plane.h
 	src/player.cpp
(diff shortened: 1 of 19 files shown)


Ghabry:
Assuming later somebody wants to introduce a "GlBitmap" that uses the GPU for blitting (or e.g. SDLBitmap to handle some stuff in SDL instead of Pixman) how would be the design here when BitmapRef is removed?
In the current implementation I would expect a conditional in the "Create" function that instantiates different Bitmap child classes and put them in the smart ptr. Will the Create functions stay? How will this all work then?

fmatthew5876:
> Assuming later somebody wants to introduce a "GlBitmap" that uses the GPU for blitting (or e.g. SDLBitmap to handle some stuff in SDL instead of Pixman) how would be the design here when BitmapRef is removed?
> In the current implementation I would expect a conditional in the "Create" function that instantiates different Bitmap child classes and put them in the smart ptr. Will the Create functions stay? How will this all work then?

I haven't completely thought this through yet. I am leaning more towards there just being one `Bitmap` class and maybe we use PIMPL idiom with some hidden data inside to switch on the backends, but I'm not committed to anything. One thing we want to avoid is putting too many virtual function calls and heap allocations in the drawing hot path.

I think the first step is more optimizations like #1989 , removing as many extraneous bitmaps as possible. Then we can revisit where the bottlenecks are and how a multi-renderer system would look like. It's also possible that the allocator approach in #1983 mitigates the overhead of dynamically allocating bitmaps, we'll see.

For now, this PR just adds RAII and cleans up the code. So it's a step we want to take regardless.

fmatthew5876:
This has been tested quite extensively on all different kinds of picture blitting while developing #1989

PR #1988, opened 2019-12-16
Fix spritesheet regression

fmatthew5876:
* spritesheet_frame from command starts with 1, but save data starts 0
* Optimize case for invalid frames, which can be caused by code
  which uses variables to set the frame id

The bug can be noticed in Frozen Triggers. Right at the beginning the are visual glitches with the weapons.

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -2433,7 +2433,8 @@ bool Game_Interpreter::CommandShowPicture(RPG::EventCommand const& com) { // cod
 		if (com.parameters[24] == 2) {
 			params.spritesheet_speed = com.parameters[25];
 		} else {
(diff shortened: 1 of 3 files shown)


Ghabry:
In general there is something broken with Frozen Triggers:

https://easyrpg.org/play/pr1988/?game=frozen_triggers

The weapon glitch is gone but there is some white box, also already in master. So not sure was broke it :/

Ghabry:
Ignore that white box thing. That has another cause: RPG2k3e must be the engine.

proper test link: https://easyrpg.org/play/pr1988/?game=frozen_triggers&engine=rpg2k3e

You seem to have accidently partially activated spritesheet support even when the engine is not rpg2k3e. Before spritesheets were completely disabled as they are not a RPG2k3J feature.

fmatthew5876:
I think that happened in the picture PR.

Added back the 2k3e check.

PR #1989, opened 2019-12-17
Refactor and optimize Game_Picture + Wave fix - Bitmaps 2 of N

fmatthew5876:
Fix #1093
Fix #1883 

 * Remove sheet bitmap
 * Fix Effect's blit to properly handle src_rect for rotations in pixman

Tests with normal picture

For these tests I created several pictures with different sprite sheet frames of the actor1 charset. I also created a picture without a spritesheet.

For the 75% / 25% test, you have to hack player code because top/bottom trans are not supported in 2k3e.

Test Cases
- [x] No effects
- [x]  50% trans
- [x] 100% trans
- [x] 75% / 25% trans
- [x] 50% zoom
- [x] 50% zoom, 50% trans
- [x] 50% zoom, 100% trans
- [x] 50% zoom, 75% / 25% trans
- [x] 200% zoom
- [x] 200% zoom, 50% trans
- [x] 200% zoom, 100% trans
- [x] 200% zoom, 75% / 25% trans
- [x] Rotate
- [x] Rotate, 50% trans
- [x] Rotate, 100% trans
- [x] Rotate, 75% / 25% trans
- [x] Rotate, 50% zoom
- [x] Rotate, 50% zoom, 50% trans
- [x] Rotate, 50% zoom, 100% trans
- [x] Rotate, 50% zoom, 75% / 25% trans
- [x] Rotate, 200% zoom
- [x] Rotate, 200% zoom, 50% trans
- [x] Rotate, 200% zoom, 100% trans
- [x] Rotate, 200% zoom, 75% / 25% trans
- [x] Wave
- [x] Wave, 50% trans
- [x] Wave, 100% trans
- [x] Wave, 75% / 25% trans
- [x] Wave, 50% zoom
- [x] Wave, 50% zoom, 50% trans
- [x] Wave, 50% zoom, 100% trans
- [x] Wave, 50% zoom, 75% / 25% trans
- [x] Wave, 200% zoom
- [x] Wave, 200% zoom, 50% trans
- [x] Wave, 200% zoom, 100% trans
- [x] Wave, 200% zoom, 75% / 25% trans

PR #1990, opened 2019-12-17
Minor Blitting optimizations - Bitmaps 3 of N

fmatthew5876:
Depends on #1989

### DIFF
--- a/bench/bitmap.cpp
+++ b/bench/bitmap.cpp
@@ -266,9 +266,8 @@ BENCHMARK(BM_BlendBlit);
 static void BM_Flip(benchmark::State& state) {
 	Bitmap::SetFormat(format);
 	auto dest = Bitmap::Create(320, 240);
-	auto rect = dest->GetRect();
 	for (auto _: state) {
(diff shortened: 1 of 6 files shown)


fmatthew5876:
Performance differences in flip blitting:

Master
```
---------------------------------------------------
Benchmark            Time           CPU Iterations
---------------------------------------------------
BM_FlipBlit      71557 ns      71555 ns       9793
BM_Flip          93533 ns      93530 ns       7500
```

This PR
```
---------------------------------------------------
Benchmark            Time           CPU Iterations
---------------------------------------------------
BM_FlipBlit      71263 ns      69876 ns      10061
BM_Flip          59174 ns      57976 ns      12220
```

fmatthew5876:
These are small improvements to some of our blitting routines. Some of them I looked into pixman code and found that the functions we were calling were just wrappers around other functions and simplified them. 

The only one that improves by a lot is `Flip()`. The rest barely register on the benchmark because all the time is dominated by the actual composite operation.

PR #1991, opened 2019-12-18
Optimize weather rendering - Drawables 3' of N

fmatthew5876:
Depends on #1977


fmatthew5876:
I made a simple map with one event which causes strong snow, the measured 10 seconds of idle playtime in it.

#1985 
snowmaster

This + #1985 
snowpr

Issue #1945, opened 2019-11-19
Picture affected by tint + weather commands doesn't work properly


Br4ssman:
**Name of the game**: Super Crystal Hunter
**Player platform**: Player 0.6.1, Win10 x64
**Describe the issue in detail and how to reproduce it**:

I don't know if this issue was fixed o simply this function is not supported yet (being a new featured from rm 2k3 https://github.com/EasyRPG/Player/issues/1263).

In any case, I have this code:
```
@> Tint Screen: (100,90,120,100), @0.0s
@> Show Picture: 6, 'SCH-nevada', (0,-1), 100%, 15%, M6
@> Move Picture: 6, (320,240), 100%, 15%, @2.5s, Wait
```
"SCH-nevada" is my custom snowfall. But sometimes, there's an snowstorm and my snowfall is faster, besides, I add `@> Weather Effect Settings: Fog, Medium` for more realistic effect, and this is the result:

RPG_RT:
![Captura](https://user-images.githubusercontent.com/18174790/69128613-02c05300-0aad-11ea-8fcf-3e8519ebc52d.PNG)

Player:
![Captura2](https://user-images.githubusercontent.com/18174790/69128633-0bb12480-0aad-11ea-9757-eac6ceb03673.PNG)

fmatthew5876:
Part (or maybe all??) Of this is due to the fact that we don't render fog and sandstorm effects accurately.

We just tint the screen whereas RPG_RT has a texture. It looks like for some levels of fog it even renders the texture over itself scrolling in different directions. So it's not easy to figure out exactly what is it and reproduce the effect.

Maybe someone knows how to hack RPG_RT and get the fog effect details out?

Br4ssman:
Maybe @CherryDT can shed light on this issue.

fmatthew5876:
Started on this here: https://github.com/EasyRPG/Player/pull/1992

It's never going to be perfect, but could use your help as you have a good eye for discrepancies.

Issue #1737, opened 2019-03-31
Unimplemented 2k3 Fog and Sandstorm weather effects


fmatthew5876:
We have not implemented 2k3 Fog and Sandstorm effects.

Overall effects
- [x] Fog Overlay effect
- [x] Sand Overlay effect
- [x] Sand particle effect

Fog & Sand Overlay features:
- [x] Implement a noise function to generate Fog and Sand texture
- [x] Correct colors for Fog
- [x] Correct colors for Sand
- [x] Set correct opacities for strengths
- [x] Verify correct speed for upper layer left movement
- [x] Verify speed for lower layer right? movement
- [x] Correct timing for slow up/down movement (appears random)
- [x] Correct up / down random selection for Y movement.
- [x] Affected by Tone

Sand Particle features:
- [x] Implement a "good enough" version
- [x] Correct rate of particle generation
- [x] Correct angle
- [x] Correct speed
- [x] Correct colors
- [x] Correct size (1 pixel wide, 2 pixels tall)
- [x] Affected by Tone

fmatthew5876:
@CherryDT

Do you have the particle math for these? Also how they interact with `SaveScreen_pan_x` and `y`?

Even for rain and snow, it would be good to verify we're correct.

fmatthew5876:
Here are some truecolor screenshots

![mist1](https://user-images.githubusercontent.com/1198466/58384954-95c57800-7fb7-11e9-9879-698917d52d4b.png)
![mist2](https://user-images.githubusercontent.com/1198466/58384955-965e0e80-7fb7-11e9-9eb0-64e7956abed4.png)
![mist3](https://user-images.githubusercontent.com/1198466/58384956-965e0e80-7fb7-11e9-81e7-13f3de5ffda5.png)


These are not enough to rip the texture. In medium or weak mist, you can see pretty obviously that they are 2 layers of texture going in opposite directions. I'm not sure how we can tease them out of RPG_RT.

elsemieni:
I don't know if it helps but I tweaked the intensity value modifing hex values at copypasting Weather Options in clipboard. The last byte in the command is the intensity (where 0x0, 0x1 and 0x2 is weak, medium strong). I set fog intensity at the max possible value 0x7F (more than that RPG Maker refuses me to paste the command with a "Error loading from stream" error... delphi things for sure). This is the result:

[...]

PR #1992, opened 2019-12-18
Implement Fog and Sandstorm overlay effect

fmatthew5876:
Fix: #1945

Depends on #1991 

Notes about how this works in #1737 

Obviously not accurate, and not sure how much accuracy is possible. But already looks pretty close in spirit.

Done:
- [x] Implement noise function 
- [x] Correct colors for Fog
- [x] Correct colors for Sand
- [x] Set correct opacities for strengths
- [x] Approximates fog movement speed - not perfect

TBD here or in later PR
- [ ] Improve noise function and verify which one best matches RPG_RT.
- [ ] Verify correct speed for upper layer left movement
- [ ] Verify speed for lower layer right? movement
- [ ] Correct timing for slow up/down movement (appears random)
- [ ] Correct up / down random selection for Y movement.

This is the image I used for the texture loading version.
fog
sand


fmatthew5876:
Heres a very interesting bug, Apply a screen shake and the fog layers aren't affected, but you can see the offsets.

shake_sand

fmatthew5876:
Fog layers, using true color RPG_RT

truelayer

fmatthew5876:
Here is another discovery, the fog appears not be a texture, but instead some procedurally generated noise.

I have a save game with strength=3 hacked in. I loaded the save 3 times and got 3 results:

pat1
pat2
pat3

fmatthew5876:
So far I've got what looks like accurate opacity values in this PR. More to come.

fmatthew5876:
Sand particle colors
-------------------------

There are 4 sand particles: white, red, yellow, and orange

The pixel values however appear to vary.

Here are some sand particle colors with strength=3

| Name | Color1 | Color2 | Color3 | Color 4 |
| -- | -- | -- | -- | -- |
| White | (252,252,233) | (251,251,233) | (249,249,222)  | (249,249,225) |
| Red | (250,75,62) | (252,77,64) | (251, 93, 76) | (249, 84, 68) |
| Yellow | (248, 248, 72) | (245, 245, 69) | (251, 251, 74) | (248, 248, 72) |
| Orange | (247, 190, 64) | (252, 195, 69) | (246, 191, 65) | (247, 190, 64) |

The particles get slightly darker at lower strength levels. Here are some samples

| Name | Weak | Mid | Strong | 3 |
| -- | -- | -- | -- | -- |
| White | (212,212,196) | (235, 235, 218 ) | (237, 237, 214) | (249,249,225) |
| Red | (207, 46, 40) | (221, 53, 45) | (239, 62, 52) | (249, 84, 68) |
| Yellow | (205, 205, 40) | (218, 218, 45) | (236, 236, 57) | (248, 248, 72) |
| Orange | (204, 151, 39) | (221, 164, 44) | (234, 178, 55) | (247, 190, 64) |

fmatthew5876:
I added procedurally generated noise. I used the simplest algorithm imaginable, and it looks similar to RPG_RT.

This is not complete but it is pretty close and should go into 6.2.

@Br4ssman Could you please check if this satisfies #1945

Further refinements I'll either add to this if it's still open or do it in followup prs.

src/weather.cpp
```diff
@@ -175,37 +166,71 @@ void Weather::DrawSnow(Bitmap& dst) {
...
+	sand_bitmap = Bitmap::Create(w, h);
+
+	auto* fog_img = reinterpret_cast<uint32_t*>(fog_bitmap->pixels());
+	auto* sand_img = reinterpret_cast<uint32_t*>(sand_bitmap->pixels());
+
+	for (int i = 0; i < w * h; ++i) {
+		// FIMXE: How well does this match RPG_RT textures?
+		int px = Utils::GetRandomNumber(0, fog_overlay_num_colors - 1);
+		fog_img[i] = fog_pixels[px];
+		sand_img[i] = sand_pixels[px];
```
fmatthew5876: @Ghabry do we need to support non 32bit rgba pixel formats? This one assumes each pixel fits into a uint32_t.

src/weather.cpp
```diff
@@ -175,37 +166,71 @@ void Weather::DrawSnow(Bitmap& dst) {
...
+	const auto w = fog_overlay_tile_width;
+
+	fog_bitmap = Bitmap::Create(w, h);
+	sand_bitmap = Bitmap::Create(w, h);
+
+	auto* fog_img = reinterpret_cast<uint32_t*>(fog_bitmap->pixels());
+	auto* sand_img = reinterpret_cast<uint32_t*>(sand_bitmap->pixels());
+
+	for (int i = 0; i < w * h; ++i) {
+		// FIMXE: How well does this match RPG_RT textures?
```
Ghabry: FIMXE

src/weather.cpp
```diff
@@ -175,37 +166,71 @@ void Weather::DrawSnow(Bitmap& dst) {
...
-void Weather::DrawFogOverlay(Bitmap& dst, const Bitmap& src) {
+void Weather::CreateFogOverlay() {
+	uint32_t fog_pixels[fog_overlay_num_colors];
+	uint32_t sand_pixels[fog_overlay_num_colors];
+
+	for (int i = 0; i < fog_overlay_num_colors; ++i) {
+		auto fc = fog_overlay_colors[i];
+		auto sc = sand_overlay_colors[i];
+		fog_pixels[i] = Bitmap::pixel_format.rgba_to_uint32_t(fc.red, fc.green, fc.blue, fc.alpha);
+		sand_pixels[i] = Bitmap::pixel_format.rgba_to_uint32_t(sc.red, sc.green, sc.blue, sc.alpha);
```
Ghabry: You must be the first who uses this function instead of manually shifting the buffer.
PixelFormat is also veryyyyyyyy old legacy code almost nobody understands.

src/weather.cpp
Ghabry: This is fine.
16bit support is nice to have (memory usage) but due to how pixman works this is a performance killer, so currently all tested code paths are 32bit only.

src/weather.cpp
fmatthew5876: fixed

src/weather.cpp
fmatthew5876: I added a fixme note, in case we do decide to get more fancy with pixel formats later one.

Yeah, someone was trying to get really clever with this stuff, especial the compile time formats. It looks like most of that code is not used.

src/weather.cpp
fmatthew5876: I added a fixme note, in case we do decide to get more fancy with pixel formats later on.

Br4ssman:
I can't try it.

After a few sec. Player is closed.
I tried with x64 and x86 Windows builds.

fmatthew5876:
> I can't try it.
> 
> After a few sec. Player is closed.
> I tried with x64 and x86 Windows builds.

Sorry about that, we had a crashing regression in master. That's been fixed now.

@Br4ssman could you try again?

Br4ssman:
Tested and nice! it's practically the same that RM ^^

Thanks.

PR #1993, opened 2019-12-20
Optimize Cache lookup

fmatthew5876:


fmatthew5876:
This one also makes a measurable improvement in Frozen triggers. 2% of frame time is spent in `std::map` lookups.

Without this PR
cacheold

With this PR
cachenew

fmatthew5876:
Adding constexpr for Spec yields more small improvements.

constexprcache

fmatthew5876:
After removing the async check.

cache-nocheck

src/cache.cpp
```diff
@@ -262,13 +275,16 @@ namespace {
 			return LoadDummyBitmap<T>(s.directory, f);
 		}
 
+#ifndef NDEBUG
 		// Test if the file was requested asynchronously before.
 		// If not the file can't be expected to exist -> bug.
+		// This text is expensive and turned off in release builds.
```
carstene1ns: > test

src/cache.cpp
```diff
@@ -274,13 +275,16 @@ namespace {
...
 		FileRequestAsync* request = AsyncHandler::RequestFile(s.directory, f);
 		if (!request->IsReady()) {
-			Output::Debug("BUG: File Not Requested: %s/%s", s.directory, f.c_str());
+			Output::Error("BUG: File Not Requested: %s/%s", s.directory, f.c_str());
```
carstene1ns: Error will currently close Player. Might be better to use Warning until #858.

src/cache.cpp
fmatthew5876: FIxed

src/cache.cpp
fmatthew5876: There are a lot of places where they just assume the cache will return non-null and then immediately start using the object. I'd rather we crash here with a big error message than get a mysterious segfault later on.

This error indicates a bug in Player code, and not game event code.

Open to suggestions though. What do you guys prefer?

Ghabry: my original intention when writing this (with return nullptr) was to segfault right after to have a stacktrace.

fmatthew5876: Agree let's use assert

Ghabry:
Could you discuss the advantage of "hash key via string concat" vs "hash key as tuple"?

fmatthew5876:
The real win here is switching from tree based `map` to hash based `unordered_map`. 

As for the difference between building a composite string vs hashing tuple elements, that would need to be profiled to get a definite answer. 

Unfortunately both idioms require string copies, but it's possible the tuple version uses shorter strings more and so we could trigger the small buffer optimization more often and avoid calls to malloc().

The standard library doesn't support arbitrary hashing of tuples, for that we would need something like `boost::hash_combine`

fmatthew5876:
jenkins test this please

PR #1995, opened 2019-12-21
Switch to ctest and fix conflicts with liblcf tests

fmatthew5876:
This fixes build errors with tests with liblcf and Player when compiled with -DPLAYER_BUILD_LIBLCF=ON

Also fixes errors in tests recently merged from other PRs


Ghabry:
sorry for not testing this use case :/
Will propose a solution

Ghabry:
more research: There is doctest_discover_tests for CMake. This ensures that CTest will show all individual test runs but it swallows the assertion/test failure message including the failing line so I consider this pretty useless.

PR #1999, opened 2019-12-21
Switch + macOS fix

### LINKED ISSUE #1994: Update Switch build for new libnx
At least `threadCreate()` needs additional `NULL` argument.

carstene1ns:
Fixes #1994.

### DIFF
--- a/src/audio_switch.cpp
+++ b/src/audio_switch.cpp
@@ -84,7 +84,7 @@ NxAudio::NxAudio() :
 	
 	mutexInit(&audio_mutex);
 
-	threadCreate(&audio_thread, switch_audio_thread, NULL, 0x10000, 0x2B, -2);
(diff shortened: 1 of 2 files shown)


src/audio_switch.cpp
```diff
@@ -84,7 +84,7 @@ NxAudio::NxAudio() :
 	
 	mutexInit(&audio_mutex);
 
-	threadCreate(&audio_thread, switch_audio_thread, NULL, 0x10000, 0x2B, -2);
+	threadCreate(&audio_thread, switch_audio_thread, NULL, NULL, 0x10000, 0x2B, -2);
```
fmatthew5876: Should use `nullptr`

carstene1ns: Yeah, should be used in other places as well. Just using it for consistency.

Ghabry:
reran OSX builder: Works :+1: 

The NULL stuff is unfortunate but this goes through all of the DisplayUi stuff, so changing it at one line will not help much.

Issue #1998, opened 2019-12-21
Eternal Legends: Boss battle ends after 1st phase


Ghabry:
Can't be tested. Blocked by #1997

#### Name of the game: Eternal Legends
https://rmarchiv.tk/games/595 (use 2.27a)

The original bug report via IRC (thanks!) is about 2.26. Also available on rmarchiv.tk. Maybe test both versions if not reprocuable.

#### Attach files (as a .zip archive or link them)
[Save14_el.zip](https://github.com/EasyRPG/Player/files/3991393/Save14_el.zip)

#### Describe the issue in detail and how to reproduce it:

Walk to the right into the next room, defeat the first phase.
In RPG_RT the 2nd boss phase starts, in Player the fight is over :tada: 

There are also two other things noticable, not verified if they are in the original game:
 - There is some "piece of dirt" in the middle of the screen
 - One of the fades to the boss shows the old map.

carstene1ns:
Maybe should be also tried against version 2.26 like in the original bug report.

Ghabry:
More context: ID 241

The boss order is:
Phase 1: 3: Gott der Zerstörung
Phase 2: 2: Gott der Zerstörung
Phase 3: 1: Gott der Zerstörung

Relevant is only Page 1. When the enemy HP is 0-50% it will do a full heal "Selbstgeheilt" once, then the second time 0-50% is reached it executes:

```
@> Change Enemy State: [3:Gott der Zerstung], + [$o Tod] <-- All enemies dead here
@> Tint Screen: (120,0,0,100), @5.0s
@> Show Animation: [Bio], [1:Gott der Zerstung]
@> Play SE: 'Monster2', 100, 50, 50
@> Tint Screen: (200,200,0,100), @4.5s, Wait
@> Flash Screen: Once, (22,0,31,31), @1.0s, Wait
@> Shake Screen: Once, 5, 5, @0.2s, Wait
@> Show Animation: [Feuer], [3:Gott der Zerstung], Wait
@> Enemy Encounter: [2:Gott der Zerstung] <-- Next enemy spawns here
@> Tint Screen: (100,100,100,100), @1.8s, Wait
@> Control Switches: [4191:Phase1] = OFF
@> Control Switches: [4192:Phase2] = ON
@> Control Switches: [4190:Selbstgeheilt] = OFF
```

fmatthew5876:
Things to verify:

- [ ] When all party dead and interpreter running
[...]

PR #2001, opened 2019-12-22
Fix interaction of battle end conditions with running events

fmatthew5876:
* 2k wait for interpreter to finish
* 2k3 interupt and end immediately

Fix: #1998

There are some visual things in 2k3 battles observed in the tests that we don't implement properly. Punting on that until 2k3 battle system is done proper.

### DIFF
--- a/src/game_battle.cpp
+++ b/src/game_battle.cpp
@@ -368,12 +368,20 @@ bool Game_Battle::AreConditionsMet(const RPG::TroopPageCondition& condition) {
 }
 
 bool Game_Battle::UpdateEvents() {
-	if (Game_Battle::CheckWin() || Game_Battle::CheckLose()) {
+	const auto battle_end = Game_Battle::CheckWin() || Game_Battle::CheckLose();
+
+	// 2k3 battle interupts events immediately when battle end conditions occur.
+	if (Player::IsRPG2k3() && battle_end) {
 		return true;
 	}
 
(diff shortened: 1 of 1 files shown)


Ghabry:
Recommend a retest with #1129

src/game_battle.cpp
```diff
@@ -368,14 +368,22 @@ bool Game_Battle::AreConditionsMet(const RPG::TroopPageCondition& condition) {
 }
 
 bool Game_Battle::UpdateEvents() {
-	if (Game_Battle::CheckWin() || Game_Battle::CheckLose()) {
+	const auto battle_end = Game_Battle::CheckWin() || Game_Battle::CheckLose();
+
+	// 2k3 battle interupts events immediately when battle end conditions occur.
+	if (Player::IsRPG2k3() && battle_end) {
```
Ghabry: Not sure if this check is needed because we already do a check here in the interpreter after each command:

https://github.com/EasyRPG/Player/blob/master/src/game_interpreter.cpp#L423-L426

fmatthew5876: It is, because the return value of this function is used to determine whether we process battle actions. The interpreter aborts if the win conditions are set, but `IsRunning()` would still be true and we'd end up in an infinite loop.

fmatthew5876:
Retested #1129 it works.

Issue #1488, opened 2018-11-06
Picture commands should work in battle too (even though their effect may not immediately be visible)


CherryDT:
Currently, Player will do nothing if Show/Move/Erase Picture is called in battle:

https://github.com/EasyRPG/Player/blob/4c589a4b2bda92c777d0bf8847928cc104e7bb36/src/game_interpreter.cpp#L2133-L2136

The actual RPG_RT behavior however is to modify the pictures nonetheless, although they are never displayed in battle in RM2k and RM2k3 pre-1.12. (Picture move operations would also be initiated but basically "paused" until battle ends.)

Issue #1263, opened 2017-09-14
Support RPG Maker 2003 v1.12


Ghabry:
Released on 14th of September: https://steamcommunity.com/games/362870/announcements/detail/3214864483356355770

- [x] The “This Event” target may now also be used in common events and will refer to the last map event in the call stack.
- [x] The “restart game” hotkey is now Alt+F12 instead of F12. This change was done to avoid frustrating accidental restarts, especially when the user is used to pressing F12 in order to create a screenshot (like it’s done in most Steam games). When F12 is pressed, a message is now displayed instead. (**countdown since 0.6.2**)
- [x] If multiple events were stacked on top of each other on the same tile, their drawing order would change in unpredictable ways based on the position of other events (including the player). Now, they are sorted by their IDs (highest ID drawn last). (**since 0.6.1**)
- [x] The current position of the parallax background was not saved in the save file.

Pictures:
- [x] Variables can now be used instead of fixed values for picture ID, filename suffix, transparency and zoom.
- [x] The Z position (layer) in which a picture is drawn can now be changed (for example, pictures can be shown between the tileset and the events, or even above message windows, etc.).
- [x] Pictures can now be used in battle as well. (**since 0.6.2**)
- [x] For each picture, it’s now possible to select whether it is affected by screen shake
- [x] ... or affected by tint
- [x] ... or affected by flash
- [x] For each picture, it’s now possible to select whether it should be automatically cleared on map change
- [x] ... cleared on battle end. (**since 0.6.2**)
- [x] Spritesheet feature: Pictures may get their graphics from a spritesheet with variable size. It’s possible to select a fixed sprite to display or set it through a variable.
- [X] Automatic animation: A spritesheet with individual animation frames can be used for automatically animating a picture with a user-defined speed. The animation can be run in a loop or once. In the latter case, the picture is automatically erased once its animation is finished. This feature can be used to show picture animations in battle, where no parallel process event for manual animation may be used.

Not applicable:
- Engine understands PicPointerPatch now (already supported)
- “Show Picture” used in a loop with the same picture file will no longer cause high CPU and disk load (which led to lags) (already supported)
- Displaying a picture with 100% transparency will now skip the drawing step entirely (already supported)
- The game volume can now be controlled from the video options dialog. - Not applicable
- When using a gamepad, the D-Pad may now be used as alternative to the analog stick. - Not applicable, already in?
[...]

Issue #1954, opened 2019-11-27
Remove global save data


fmatthew5876:
The global save game data `Main_Data::game_data` causes a lot of problems in our code. We have several examples like `Game_Character` and `Game_Picture` where game objects reference this data via pointers. Then we have others like the interpreter where it actually operates on a copy.

This makes it impossible to tie the lifetime of the save data to the game objects, and is a source of bugs and confusion.

We should eliminate `Main_Data::game_data` and have each game object own it's own state. During saving we collect the states of all the game objects and construct save data, during loading we can take a save object and construct game objects.

Essentially instead of storing the data in a global variable and referencing it there, we serialize and de-serialize it in effect.

fmatthew5876:
Removing this and wrapping up the init and teardown of all singletons behind a single interface will greatly aid the writing of tests.

fmatthew5876:
With #2208 , #2320 and similar PRs the various test suites are getting creating mocks that setup a subset of the game singletons. All these mocks need to be consolidated so that all tests have a uniform way to (1) Setup LDB/LMT/LMU data needed and (2) create a mock to setup minimal game state required for the test.

PR #2003, opened 2019-12-22
Implement Pictures and Weather in Battle

fmatthew5876:
Depends on: #1989

- [x] Implement erase on battle end flag
- [x] Implement displaying pictures in battle
- [x] Implement displaying weather effects in battle
- [x] Rebase after #2004 merged
- [x] Rebase after #2012 merged and fix `Game_Screen` for #1954 

Fix: #1488 
Fix: #1263
Progress on: #1954

### DIFF
--- a/src/game_picture.h
+++ b/src/game_picture.h
@@ -78,10 +78,10 @@ public:
 	void Erase();
 
 	void Update(bool is_battle);
-	void UpdateSprite();
+	void UpdateSprite(bool is_battle);
 
 	static void Update(std::vector<Game_Picture>& pictures, bool is_battle);
-	static void UpdateSprite(std::vector<Game_Picture>& pictures);
+	static void UpdateSprite(std::vector<Game_Picture>& pictures, bool is_battle);
 
 	void OnMapChange();
 	void OnBattleStart(Scene* map_scene, Scene& battle_scene);
--- a/src/game_picture.h
+++ b/src/game_picture.h
@@ -80,9 +80,12 @@ public:
 	static void UpdateSprite(std::vector<Game_Picture>& pictures);
 
 	void OnMapChange();
+	void OnBattleEnd();
 
 	// FIXME: Use C++20 span
 	static void OnMapChange(std::vector<Game_Picture>& pictures);
+	// FIXME: Use C++20 span
+	static void OnBattleEnd(std::vector<Game_Picture>& pictures);
 
 private:
 	RPG::SavePicture data;
--- a/src/game_picture.h
+++ b/src/game_picture.h
@@ -90,9 +90,9 @@ public:
 	// FIXME: Use C++20 span
 	static void OnMapChange(std::vector<Game_Picture>& pictures);
 	// FIXME: Use C++20 span
-	static void OnBattleStart(std::vector<Game_Picture>& pictures);
+	static void OnBattleStart(std::vector<Game_Picture>& pictures, Scene* map_scene, Scene& battle_scene);
 	// FIXME: Use C++20 span
-	static void OnBattleEnd(std::vector<Game_Picture>& pictures);
+	static void OnBattleEnd(std::vector<Game_Picture>& pictures, Scene* map_scene);
 
 private:
 	RPG::SavePicture data;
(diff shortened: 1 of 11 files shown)


fmatthew5876:
This is easy to test, create an event which makes these pictures:

| PicId | OnMap | OnBattle | EraseOnMapChange | EraseOnBattleEnd |
| -- | -- | -- | -- | -- |
| 1 | 1 | 0 | 0 | 0 |
| 2 | 1 | 0 | 1 | 0 |
| 3 | 1 | 0 | 0 | 1 |
| 4 | 1 | 0 | 1 | 1 |
| | | | |
| 5 | 0 | 1 | 0 | 0 |
| 6 | 0 | 1 | 1 | 0 |
| 7 | 0 | 1 | 0 | 1 |
| 8 | 0 | 1 | 1 | 1 |
| | | | |
| 9 | 1 | 1 | 0 | 0 |
| 10 | 1 | 1 | 1 | 0 |
| 11 | 1 | 1 | 0 | 1 |
| 12 | 1 | 1 | 1 | 1 |

1. Run the event on a map, start and finish a battle, the start another battle

Result: 
* Pics 1-4, 9-12 visible on map
* pics 5-12 visible in battle
* pics 1,2,9,10 visible on map after
* pics 5,6,9,10 visible in next battle

2. Run the event in a battle, finish the battle, start a new battle

Result:
* pics 5-12 visible in battle
* pics 1,2,9,10 visible on map after
* pics 5,6,9,10 visible in next battle

3. Run the event on a map, and teleport to the same map
* Pics 1-4, 9-12 visible on map
* Pics 1-4, 9-12 still visible after teleport

4. Run the event on a map, and teleport to a different map
* Pics 1-4, 9-12 visible on map
* Pics 1,3,9,11 visible on other map

fmatthew5876:
Would prefer #2004 goes in before this one

fmatthew5876:
Rebased and now was finally able to add the commit removing `Main_Data::game_data.screen`.

I did redid the battle and map change picture tests in my test game. I also tested saving and loading screen state.

This is ready once #1989 goes in.

src/game_picture.h
```diff
@@ -69,15 +70,29 @@ class Game_Picture {
...
 	static void UpdateSprite(std::vector<Game_Picture>& pictures);
 
-	void Update();
-	void UpdateSprite();
+	void OnMapChange();
+	void OnBattleStart(Scene* map_scene, Scene& battle_scene);
+	void OnBattleEnd(Scene* map_scene);
+
+	// FIXME: Use C++20 span
+	static void OnMapChange(std::vector<Game_Picture>& pictures);
```
Ghabry: these span fixme again

fmatthew5876:
> Question: When ending battle you don't call Take before Append. Can this have any negative side effects? Double free of Drawables e.g.

The `DrawableList` doesn't own the drawables, and since battle is ending the `Scene_Battle` list is about to be destroyed, so we don't need to remove them.

> 
> Problem:
> 
>     1. Create a picture on the map that is not visible in battle
> 
>     2. Start battle
> 
>     3. Call show picture with the same picture ID but this time visible in battle. _Currently the pic is not shown_

Great catch, I put in a small fix for this.


Overall I don't like this model we have with drawable list. It's hard to reason about which list each drawable is on and easy to have dangling pointers.

Also calling `Take()` is O(n) for each picture, making the whole battle start sequence O(map list size * num pictures)  which is obviously sub-optimal.

I have plans to refactor and clean this up somehow, but I haven't yet decided on the best approach. Anyway, that's for a later PR.

src/game_picture.h
fmatthew5876: Removed

fmatthew5876:
Retested this on windows.

Ghabry:
The drawable list is also a more historical concept.
Initially we did the scene handling like RPG XP and only had one scene without any Push/Pop concept though this was very inflexible because when returning from menus the menu had to know which scene to go to and you had to serialize all scene state.
VX Ace switched to a scene stack, guess they noticed independently from us that there solution was bad :)
With the Push/Pop added we started to encounter problems with other scenes drawing each other so the pushable graphic stack was the solution for it but I agree that drawable transfer between scenes is tricky, that was not part of the design because until RPG2k3E this wasn't really needed :/.

fmatthew5876:
Scene stack is absolutely the way to go.

 I am thinking maybe drawables should be `unique_ptr` owned directly by the list. That would cleanup the confusion and enforce correctness.

PR #2008, opened 2019-12-23
Filefinder: Move macro hell into new platform class, fix PSVita build

Ghabry:
This is an attempt to make the FileFinder and all this #define mess more clean.

TODO:
- [x] Windows is untested yet because I have to install VS2019 locally and this will take half a day with a German internet connection.
- [x] PSVita untested due to lack of hardware :/

Now some FileFinder functions (e.g. Exists & IsDirectory) look like simple forwarders but I want to keep them this way for now because when I ever find time to do the VFS stuff this will be refactored, so I want to avoid an in-between refactor of this.

Design decision for Directory::Read() -> This does not return name & type because I wanted to avoid transforming name/type when not necessary. In the FileFinder case though both is always invoked, consider it a "thinking in the future" feature..


Ghabry:
Windows unit tests passed and was able to boot Yume2kki (Japanese filenames).

Ghabry:
@NootBoot64

You reported that the Vita build is missing.
Could you test this VPK and tell us if it is working?

https://ci.easyrpg.org/job/player-vita-pr/35/artifact/builds/psvita/EasyRPG.vpk

Nootboot64:
@Ghabry Now it's saying "ux0:/data/easyrpg-player/ is not a valid location."

Ghabry:
@Nootboot64 
Are you 100% sure that you have the folder /data/easyrpg-player on your Vita? (= does the normal continuous build https://ci.easyrpg.org/job/player-vita/495/artifact/builds/psvita/EasyRPG.vpk work?)

Nootboot64:
Yes, I put games in there and they worked before.

Ghabry:
okay thanks for testing. Will try to find the bug in my code then but not on christmas ;). Guess I will write you again in 2 days.

Nootboot64:
Np, I'm pretty sure that besides the path bug it works fine, but as you know I can't test without being able to actually play the game on my Vita.

Ghabry:
@Nootboot64 
Well one try ;) :
https://ci.easyrpg.org/view/Pull%20Requests/job/player-vita-pr/lastSuccessfulBuild/artifact/builds/psvita/EasyRPG.vpk

Nootboot64:
I'm not sure if you wanted me to, but I tested the commit you made and build just now and it didn't fix it. Just saying

Ghabry:
@Nootboot64 
Well there was a typo in the code, the check for "IsDirectory" was the same as for "IsFile".

Try again:
https://ci.easyrpg.org/job/player-vita-pr/lastSuccessfulBuild/artifact/builds/psvita/EasyRPG.vpk

Nootboot64:
It boots this time, but the 'Games' tab can't find any games. The debug log says that ux0:/data/easyrpg-player is not a valid directory though.

Ghabry:
Could you paste the last log messages? When I have the full text it helps me to track down where it failed.

Nootboot64:
======================================================
[2019-12-25 15:42:07] Debug: Seeded the RNG with 1577288527.
[2019-12-25 15:42:07] Debug: CLI: 
[2019-12-25 15:42:07] Info: Shader set to None.
[2019-12-25 15:42:07] Debug: Error opening dir ux0:/data/easyrpg-player: No such file or directory
[2019-12-25 15:42:08] Debug: Error opening dir ux0:/data/easyrpg-player: Success
[2019-12-25 15:42:37] Warning: Removing invalid party member 1

Ghabry:
@Nootboot64 
Found another silly bug in my code: The dir handle was never assigned...
You can get the VPK in a few minutes via: https://ci.easyrpg.org/job/player-vita-pr/40/

Nootboot64:
It works fine now 👍 Good work

src/platform.h
```diff
@@ -0,0 +1,157 @@
...
+		/** @return true if opening the directory was successful */
+		explicit operator bool() const noexcept;
+
+	private:
+#if defined(_WIN32)
+		_WDIR* dir_handle = nullptr;
+		struct _wdirent* entry = nullptr;
+#elif defined(PSP2)
+		int dir_handle = -1;
+		struct SceIoDirent entry;
```
fmatthew5876: Should we say `= {};` here so this gets zero initialized?

tests/platform.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include "platform.h"
+#include "doctest.h"
+
+TEST_SUITE_BEGIN("Platform");
+
+namespace {
+	const std::string onekb = EP_TEST_PATH "/platform/1kb";
+	const std::string empty = EP_TEST_PATH "/platform/empty";
+	const std::string folder = EP_TEST_PATH "/platform/folder";
+	const std::string bad = EP_TEST_PATH "/platform/!!!inexistant!!!";
```
fmatthew5876: nonexistant

src/filefinder.cpp
```diff
@@ -779,42 +750,11 @@ std::string FileFinder::FindSound(const std::string& name) {
...
-#elif (defined(GEKKO) || defined(_3DS) || defined(__SWITCH__))
-	struct stat sb;
-	return ::stat(filename.c_str(), &sb) == 0;
-#elif defined(PSP2)
-	struct SceIoStat sb;
-	return (sceIoGetstat(filename.c_str(), &sb) >= 0);
-#else
-	return ::access(filename.c_str(), F_OK) != -1;
-#endif
+	return Platform::File(filename).Exists();
```
fmatthew5876: This approach ends up creating a string copy to construct the `File` object, just to call `stat()` and then throw away the copy. So it will be slightly slower than the old code. Not by much though, as the cost of a system call is much more than allocating a string.

However it is possible the compiler may be smart enough to see this and optimize away the copy.

src/filefinder.cpp
```diff
@@ -920,28 +826,15 @@ FileFinder::Directory FileFinder::GetDirectoryMembers(const std::string& path, F
...
-	StatBuf sb;
-	int result = GetStat(file.c_str(), &sb);
-	return (result == 0) ? sb.st_size : -1;
+int64_t FileFinder::GetFileSize(const std::string& file) {
+	return Platform::File(file).GetSize();
 }
 
 bool FileFinder::IsMajorUpdatedTree() {
-	EasyRPG_Offset size;
+	int64_t size;
```
fmatthew5876: This variable is uninitialized, and also seems to only be used in some conditional blocks. I would move it inside the blocks.

`auto size = GetFileSize(...)`

src/platform.h
```diff
@@ -0,0 +1,157 @@
...
+		FileType GetType(bool follow_symlinks) const;
+
+		/** @return Filesize or -1 on error */
+		int64_t GetSize() const;
+
+	private:
+		const std::string filename;
+#ifdef _WIN32
+		std::wstring filename_w;
+#endif
```
fmatthew5876: Do we need to store both `filename` and `filename_w` on windows?

src/platform.cpp
```diff
@@ -0,0 +1,196 @@
...
+#	endif
+#endif
+}
+
+void Platform::Directory::Close() {
+#if defined(PSP2)
+	bool valid_dir_handle = dir_handle >= 0;
+#else
+	bool valid_dir_handle = dir_handle != nullptr;
+#endif
```
fmatthew5876: You could reuse `operator bool()` here

src/platform.h
```diff
@@ -0,0 +1,157 @@
...
+	};
+
+	/** Wrapper around directory reading */
+	class Directory {
+	public:
+		explicit Directory() = delete;
```
fmatthew5876: Delete copy constructor and copy assignment operator.

Can implement move operations if you want, but we probably don't need them now.

fmatthew5876:
LGTM, just some minor nits.

Ghabry:
Got rid of the additional string copy (Except on Windows due to UTF-16) by using a reference and requiring the string to stay valid.
Removed the unused filename member on Windows.
+ the other mentioned stuff, thx!

src/platform.h
```diff
@@ -92,16 +96,19 @@ namespace Platform {
 		int64_t GetSize() const;
 
 	private:
-		const std::string filename;
 #ifdef _WIN32
-		std::wstring filename_w;
+		const std::wstring filename;
+#else
+		const std::string& filename;
```
fmatthew5876: This is really easy to get wrong. I can write the following:

```
File f("Hello World");
```

It will compile but cause a dangling reference error at runtime. The `const std::string&` constructor accepts temporaries which will get destroyed immediately.

Unfortunately there isn't a good way to do this that gives you both an easy to use interface and maximum efficiency (Unless you make a static `Exists()` method which just takes in a string, but then you also need to do windows `w_string` conversion).

Given this is only used internally for FileFinder, I'm fine with either approach for now.

src/platform.cpp
```diff
@@ -173,13 +177,7 @@ Platform::FileType Platform::Directory::GetEntryType() const {
...
 
 void Platform::Directory::Close() {
-#if defined(PSP2)
-	bool valid_dir_handle = dir_handle >= 0;
-#else
-	bool valid_dir_handle = dir_handle != nullptr;
-#endif
-
-	if (valid_dir_handle) {
+	if (operator bool()) {
```
fmatthew5876: I think it's nicer to say `if (*this)` but I don't care enough either way.

src/platform.h
Ghabry: Okay didn't knew about this corner case. Will keep the copy then. As "Exists" is not really in a hot codepath (just for rtp init and project init).

fmatthew5876: Agree, for now I'd rather keep the copy than have an interface like this.

fmatthew5876:
All looks good to me. I am ok with merging this,

src/platform.cpp
```diff
@@ -0,0 +1,195 @@
...
+		return FileType::File;
+	} else if ((attribs & (FILE_ATTRIBUTE_DIRECTORY | FILE_ATTRIBUTE_REPARSE_POINT)) == FILE_ATTRIBUTE_DIRECTORY) {
+		return FileType::Directory;
+	}
+	return FileType::Other;
+#elif defined(PSP2)
+	(void)follow_symlinks;
+	struct SceIoStat sb = {};
+	if (::sceIoGetstat(filename.c_str(), &sb) >= 0) {
+		return SCE_S_ISREG(sb.st_mode) != 0 ? FileType::File :
```
carstene1ns: This one is ugly. the `_S_IS*` macros are meant to be used as bools. Also to me looks a lot cleaner when not written with ternary operators, but that is just preference.

src/platform.cpp
```diff
@@ -0,0 +1,195 @@
...
+#else
+	struct stat sb = {};
+#  if (defined(GEKKO) || defined(_3DS) || defined(__SWITCH__))
+	(void)follow_symlinks;
+	auto fn = ::stat;
+#  else
+	auto fn = follow_symlinks ? ::stat : ::lstat;
+#  endif
+	if (fn(filename.c_str(), &sb) == 0) {
+		return S_ISREG(sb.st_mode) != 0 ? FileType::File :
```
carstene1ns: same here.

src/platform.cpp
Ghabry: Imo ternary is readable, but that's really a matter of taste.
About the != 0: I removed it. Clang tidy complained that I use integers in a boolean context (maybe false positive but doesn't really matter).

PR #2010, opened 2019-12-24
Add F12 Quit window.

fmatthew5876:
When you hold F12 a window appears and counts down. Once
countdown is complete, we reset to title screen.

Replaces 2k3e Alt+F12 feature. Enabled for all versions of the engine.

Also fixes a crash with F12, we capture old_scene before
checking for reset.

### DIFF
--- a/src/game_quit.h
+++ b/src/game_quit.h
@@ -15,43 +15,26 @@
  * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
  */
 
-#ifndef EP_WINDOW_HELP_H
-#define EP_WINDOW_HELP_H
+#ifndef EP_GAME_QUIT_H
+#define EP_GAME_QUIT_H
 
-// Headers
-#include "window_base.h"
-#include "text.h"
+#include "window_help.h"
 
-/**
- * Window_Help class.
- * Shows skill and item explanations.
- */
-class Window_Help : public Window_Base {
-
+class Game_Quit {
 public:
-	/**
-	 * Constructor.
-	 */
-	Window_Help(int ix, int iy, int iwidth, int iheight, bool is_global = false);
+	Game_Quit();
+	void Update();
 
-	/**
-	 * Sets the text that will be shown.
-	 *
-	 * @param text text to show.
-	 * @param align text alignment.
-	 */
-	void SetText(std::string text, Text::Alignment align = Text::AlignLeft);
-
-	/**
-	 * Clears the window
-	 */
-	void Clear();
-
+	bool ShouldQuit() const;
 private:
-	/** Text to draw. */
-	std::string text;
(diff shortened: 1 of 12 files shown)


src/game_quit.h
```diff
@@ -0,0 +1,47 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef EP_GAME_QUIT_H
+#define EP_GAME_QUIT_H
+
+#include "window_help.h"
+
+class Game_Quit {
```
Ghabry: Indent for public/private must be one level to the left

src/game_quit.h
```diff
@@ -0,0 +1,47 @@
...
+	public:
+		Game_Quit();
+		void Update();
+
+		bool ShouldQuit() const;
+
+		void Enable(bool value);
+	private:
+		void Reset();
+	private:
```
Ghabry: private twice

fmatthew5876:
> Also makes now Steam more happy, as F12 is the screenshot key.

Except now anytime you try to create a stream screenshot you'll have this window

> An observable difference is that reset is now not working on Title and Load (from title) anymore but that's fine for me.

I was back and forth on this. But since it's supported in 2k3 I added it back. Also the title load menu kind of makes sense.

I also verified it does not appear in game browser.

> How about reset 2 seconds? 3 seconds feels a bit long and when TestPlay is enabled maybe instant as I use reset alot when testing and this breaks my Workflow https://xkcd.com/1172/ ;)

Changed to 2 seconds

Ghabry:
> Except now anytime you try to create a stream screenshot you'll have this window

There is a short delay after F12 until Steam creates a screenshot, so this should work.

src/window_base.h
```diff
@@ -40,7 +40,7 @@ class Window_Base : public Window {
 	 * @param width window width.
 	 * @param height window height.
 	 */
```
carstene1ns: dox.

fmatthew5876: Fixed

fmatthew5876:
I see what you mean about it being handy to not have this feature be too slow. In later PRs I changed it so fast forward affects the speed of the F12 window.

Ghabry:
I just noticed what you wrote. Which PR made it affected by FF?

fmatthew5876:
#2017 the main loop refactor makes keeping this out of ff tricky. Also, after using it a lot now I think it's better that it does get sped up. It's still not too fast to make a mistake and makes testing easier.

Ghabry:
yeah I WANT it in FF because I'm already annoyed xD

carstene1ns:
Good thing we mostly run in test play mode 🤣

PR #2004, opened 2019-12-22
Rain, Snow, Screen, Bitmap, and Animations improvements

fmatthew5876:
- [x] Improve rain effect to match RPG_RT better
- [x] Improve snow effect to match RPG_RT better
- [x] Tone needs to be recalculated if weather effect changes
- [x] Encapsulate all access to `game_data.screen` from within `Game_Screen`
- [x] Apply shake effect to weather
- [x] Correct support for drawing shake_y offset, even though it's currently impossible to activate.
- [x] Add optimized routine to do edge clone blitting

Fix: #2006 
Fix: #2005
Progress on: #1954

PR #2011, opened 2019-12-24
Implement Sandstorm particles

fmatthew5876:
Depends on #2004 

This is really an approximation. No idea how to make this perfect.


fmatthew5876:
New version of this, with better colors and I think looks a little more accurate.

PR #2017, opened 2019-12-27
Refactor MainLoop and FPS - MainLoop 6 of N

fmatthew5876:
Depends on: #1973 

This refactors the player main loop to free the game update (60 fps) from rendering (any fps). It also cleans up the main loop code significantly, making it easier to understand and measure.

### DIFF
--- a/resources/unix/bash-completion/easyrpg-player
+++ b/resources/unix/bash-completion/easyrpg-player
@@ -15,9 +15,9 @@ _easyrpg-player ()
   # all possible options
   ouropts='--battle-test --disable-audio --disable-rtp --enable-mouse --enable-touch \
            --encoding --engine --fps-limit --fps-render-window --fullscreen -h --help \
-           --hide-title --load-game-id --new-game --project-path --record-input \
+           --hide-title --load-game-id --new-game --no-vsync --project-path --record-input \
            --replay-input --save-path --seed --show-fps --start-map-id --start-party \
-           --start-position --test-play --window -v --version --vsync'
+           --start-position --test-play --window -v --version'
   rpgrtopts='BattleTest battletest HideTitle hidetitle TestPlay testplay Window window'
   engines='rpg2k rpg2kv150 rpg2ke rpg2k3 rpg2k3v105 rpg2k3e'
 
--- a/resources/unix/bash-completion/easyrpg-player
+++ b/resources/unix/bash-completion/easyrpg-player
@@ -14,10 +14,10 @@ _easyrpg-player ()
 
   # all possible options
   ouropts='--battle-test --disable-audio --disable-rtp --enable-mouse --enable-touch \
-           --encoding --engine --fps-render-window --fullscreen -h --help --hide-title \
-           --load-game-id --new-game --project-path --record-input --replay-input \
-           --save-path --seed --show-fps --start-map-id --start-party --start-position \
-           --test-play --window -v --version'
+           --encoding --engine --fps-limit --fps-render-window --fullscreen -h --help \
+           --hide-title --load-game-id --new-game --project-path --record-input \
+           --replay-input --save-path --seed --show-fps --start-map-id --start-party \
+           --start-position --test-play --window -v --version'
   rpgrtopts='BattleTest battletest HideTitle hidetitle TestPlay testplay Window window'
   engines='rpg2k rpg2kv150 rpg2ke rpg2k3 rpg2k3v105 rpg2k3e'
 
@@ -51,7 +51,7 @@ _easyrpg-player ()
       return
       ;;
     # argument required but no completions available
-    --@(battle-test|encoding|seed|start-position|start-party)|BattleTest|battletest)
+    --@(battle-test|encoding|fps-limit|seed|start-position|start-party)|BattleTest|battletest)
       return
       ;;
     # these have no argument and shall be used exclusively
--- a/resources/unix/bash-completion/easyrpg-player
+++ b/resources/unix/bash-completion/easyrpg-player
@@ -17,7 +17,7 @@ _easyrpg-player ()
            --encoding --engine --fps-limit --fps-render-window --fullscreen -h --help \
            --hide-title --load-game-id --new-game --project-path --record-input \
            --replay-input --save-path --seed --show-fps --start-map-id --start-party \
-           --start-position --test-play --window -v --version'
+           --start-position --test-play --window -v --version --vsync'
   rpgrtopts='BattleTest battletest HideTitle hidetitle TestPlay testplay Window window'
   engines='rpg2k rpg2kv150 rpg2ke rpg2k3 rpg2k3v105 rpg2k3e'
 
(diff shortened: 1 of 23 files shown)


fmatthew5876:
When playing with this I already found a bug.

This logic:
https://github.com/EasyRPG/Player/blob/master/src/player.cpp#L256

It often triggers on desktop games simply because of the inaccuracy of the SDL sleep function. To fix it I added an epsilon to compare against in the refactor time keeping commit.

fmatthew5876:
With the spurious frames issue fixed, here is an example histogram vtune gives you with this instrumentation enabled:

frames

fmatthew5876:
@Ghabry Why can't we just call `emscripten_set_main_loop(one_iter, 60, 1)` and get rid of this frame skip hack?

In it's current state it'll still have bugs in emscripten and if the frame rate chosen by emscripten is < 60 fps this won't even work at all.

src/player.cpp
```diff
@@ -1039,7 +1060,7 @@ std::string Player::GetEncoding() {
 
 int Player::GetSpeedModifier() {
 	if (Input::IsPressed(Input::FAST_FORWARD)) {
-		return Input::IsPressed(Input::PLUS) ? 10 : speed_modifier;
+		return Input::IsPressed(Input::PLUS) ? 10 : 3;
```
carstene1ns: Not possible since we hook into this from the android app.

fmatthew5876: Fixed

fmatthew5876:
Just for fun, here is a new frame rate histrogram, now that the SDL improvements have been merged.

frames_sdl

fmatthew5876:
This is ready now. The new main loop implementation is based on this:

https://gafferongames.com/post/fix_your_timestep/

We now have a `--fps-limit` command line argument you can play around with to set any framerate you like.

fmatthew5876:
This PR actually contains 2 main loop refactors. 

The first one is basically a cleanup of our current main loop. Moving the sleep, fast forward, frame begin / end into `Player::MainLoop()` and essentially giving our main loop a well defined begin and end. It also makes it so emscripten/libretro and desktop platforms behave more similarly and is the real dependency of #2019 the transitions PR, since that one depends on drawing *after* cleaning up scenes so that the before/after bitmaps for transitions are correct.

The second one is a total rethink of our main loop for rendering / game logic independence and needs more careful consideration and testing.

Let me know if you'd like me to split this into 2 PRs

fmatthew5876:
> Additional command line arguments need documentation in the manual page and should be added to the bash completion file.

Added here and in #2021

release/player.sh
```diff
@@ -0,0 +1 @@
+../build/player.sh
```
fdelapena: Building symlink leftover?

fmatthew5876: Omg, I was wondering why this kept mysteriously dissapearing.. Will remove :)

src/sdl2_ui.cpp
```diff
@@ -334,6 +329,12 @@ bool Sdl2Ui::RefreshDisplayMode() {
 			Output::Debug("SDL_GetRendererInfo failed : %s", SDL_GetError());
 		}
 
+#ifdef EMSCRIPTEN
+		// Emscripten owns the main loop.
+		SetFrameRateExternallyManaged(true);
```
Ghabry: Due to ASYNCIFY magic this is not needed anymore and can use the standard logic.

src/output.cpp
```diff
@@ -194,7 +199,9 @@ static void HandleErrorOutput(const std::string& err) {
 
 	Input::ResetKeys();
 	while (!Input::IsAnyPressed()) {
-		DisplayUi->Sleep(1);
+#if !defined(EMSCRIPTEN) && !defined(USE_LIBRETRO)
+		Game_Clock::SleepFor(1ms);
```
Ghabry: Actually I can't remember anymore what the purpose of the sleep was.

carstene1ns: I guess was to loosen stress on this tight loop.

Ghabry: yeah, that should be the reason :+1:

fmatthew5876: Will revisit this when I rebased the PR and add comments

Ghabry:
Unrelated to the Async vs emscripten\_main\_loop debate:
What is really wrong in master is the code that handles low FPS situations.
Compare https://easyrpg.org/play/pr2003/asmjs/?game=frozen_triggers&engine=rpg2k3e (took the latest merged PR-build as we don't build asmjs anymore for master)
With this PR
https://easyrpg.org/play/pr2017/asmjs/?game=frozen_triggers&engine=rpg2k3e

For 2003 I get 0 FPS and 2017 I get a very "jumpy" FPS values but it is >30.

fmatthew5876:
Rebased onto #2066 and removed the old `EMSCRIPTEN` bits

fmatthew5876:
Rebased, but I need to take another look at this one.

Ghabry:
What I'm still seeing are these quickly changing FPS: https://easyrpg.org/play/pr2017/?game=frozen_triggers&engine=rpg2k3e

I have the feeling they update much more often than once per second. Maybe this is just a visual problem with the frame counter? Performance wise I see no random lags.

src/output.cpp
fmatthew5876: This is adjusted for current state, but full solution punted to #2075

fmatthew5876:
I added some tweaks and adjusted the fps counter. It appears smoother now.

From my perspective this general approach seems good. Obviously needs more testing from all of us. Unless review comments come back with major issues, I don't plan to make more large scale changes here.

So this is ready for review and testing.

fmatthew5876:
Open Question:

Should we enable vsync by default on sdl2 platforms? I'm thinking we probably should as for most people it provides a better out of the box experience. 

Thoughts?

Situations where this would be bad?

Ghabry:
Vsync by default sounds good.

carstene1ns:
Yeah, let's use vsync by default, fallback to current behavior and `-no-vsync` command line switch.

Ghabry:
FPS display looks better now - resolving my major issue. Moving on to code review.

fmatthew5876:
Rebased and made vsync the default.

resources/unix/bash-completion/easyrpg-player
```diff
@@ -14,10 +14,10 @@ _easyrpg-player ()
...
   # all possible options
   ouropts='--battle-test --disable-audio --disable-rtp --enable-mouse --enable-touch \
-           --encoding --engine --fps-render-window --fullscreen -h --help --hide-title \
-           --load-game-id --new-game --project-path --record-input --replay-input \
-           --save-path --seed --show-fps --start-map-id --start-party --start-position \
-           --test-play --window -v --version'
+           --encoding --engine --fps-limit --fps-render-window --fullscreen -h --help \
+           --hide-title --load-game-id --new-game --project-path --record-input \
+           --replay-input --save-path --seed --show-fps --start-map-id --start-party \
+           --start-position --test-play --window -v --version --vsync'
```
carstene1ns: Forgot this one.

fmatthew5876: Fixed, thanks!

PR #2022, opened 2019-12-29
Cleanup Shake and Flash logic

fmatthew5876:
This moves shake and flash animation logic to their own headers so they can be reused. It also eliminates the flash animation capability from `Sprite` and instead relies on game objects to do animation. This is important for timing as there is no guarantee of when sprites are updated vs game objects. It also makes `sizeof(Sprite)` smaller.

Finally, this fixes a bug I discovered in battle where some battle animations with flash effects would cause the flash to stick on the battler sprite and not fade away until another flash replaced it.

### DIFF
--- a/src/scene_battle.cpp
+++ b/src/scene_battle.cpp
@@ -204,6 +204,7 @@ void Scene_Battle::Update() {
 		ProcessActions();
 		ProcessInput();
 	}
+	UpdateCursors();
 
 	auto& interp = Game_Battle::GetInterpreter();
 
--- a/src/scene_battle.cpp
+++ b/src/scene_battle.cpp
@@ -168,13 +168,26 @@ void Scene_Battle::Update() {
 	item_window->Update();
 	skill_window->Update();
 	target_window->Update();
+
+	const int timer1 = Main_Data::game_party->GetTimerSeconds(Game_Party::Timer1);
+	const int timer2 = Main_Data::game_party->GetTimerSeconds(Game_Party::Timer2);
+
+	// Update Battlers
+	std::vector<Game_Battler*> battlers;
+	Main_Data::game_party->GetBattlers(battlers);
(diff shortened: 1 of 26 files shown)


fmatthew5876:
This is tested and ready.

I noticed we have a feature in 2k and 2k3 where it flashes the enemy when you're selecting them. But I don't see this happening in RPG_RT. I could of sworn I've seen this before.

Is this an RPG_RT feature or something we added?

fmatthew5876:
Found some more subtle off-by-1 timing issues with flash animations in 2k3 battle

fmatthew5876:
With this refactor #1929 is properly fixed. So I've reverted #1933

fmatthew5876:
The commits:

Fix Game_Battle update order
Update cursors after battlers, before graphics 

have now been tested pretty heavily as part of #2025

src/scene_battle.cpp
```diff
@@ -623,3 +638,8 @@ void Scene_Battle::CallDebug() {
 	}
 }
 
+void Scene_Battle::SelectionFlash(Game_Battler* battler) {
+	if (battler) {
+		battler->Flash(31, 31, 31, 24, 16);
```
carstene1ns: I do not like this "let's think like RPG_RT to do some math, so we do not need to do other math" here.
It is totally not obvious that this is meaning white color.

fmatthew5876: The idea is to streamline one API so animated flash for map events, the screen, battle sprites, battle transition effect, etc... all work exactly the same way. Similar for shake. This guarantees the pixel colors will be a perfect match one time and then the logic can be reused.

Later we could build a better flash API which works directly with `Color` and goes to and from RPG_RT format. But I don't want to tackle that now.

fmatthew5876: In other words, having a better unified flash API is blocked until #1954 is resolved. I'd like to punt on improving this for now, if that's ok.

PR #1964, opened 2019-12-02
Refactor Text Rendering stack - Messages 5 of N

fmatthew5876:
~~Depend on: #1956~~

This PR refactors the underlying `Font` and `Text` interfaces to be more streamlined and efficient.

Current Test Results
--------------------------

After "Font::Render - remove query for system"

Font:
```
------------------------------------------------------
Benchmark               Time           CPU Iterations
------------------------------------------------------
BM_FontSizeStr        637 ns        637 ns    1107660
BM_Glyph              171 ns        171 ns    4005735
BM_Render            2481 ns       2480 ns     282438
```
Text:
```
---------------------------------------------------------
Benchmark                  Time           CPU Iterations
---------------------------------------------------------
BM_TextDrawStr         94971 ns      94971 ns       7365
BM_TextDrawStrRaw      42719 ns      42718 ns      16451
```

HEAD of PR:

Font:
```
-------------------------------------------------------
Benchmark                Time           CPU Iterations
-------------------------------------------------------
BM_FontSizeStr         636 ns        632 ns    1116692
BM_FontSizeChar         21 ns         21 ns   34673138
BM_Glyph                67 ns         67 ns   10619258
BM_Render             2182 ns       2182 ns     320334
```
Text:
```
---------------------------------------------------------------
Benchmark                        Time           CPU Iterations
---------------------------------------------------------------
BM_TextDrawStrSystem         86675 ns      86676 ns       7963
BM_TextDrawStrColor          41405 ns      41405 ns      17102
BM_TextDrawCharSystem         2210 ns       2211 ns     316856
BM_TextDrawCharSystemEx       3046 ns       3046 ns     230286
BM_TextDrawCharColor           965 ns        965 ns     733548
BM_TextDrawCharColorEx        1971 ns       1971 ns     355208
```

PR #2019, opened 2019-12-28
Transition graphical fixes - MainLoop 4 of N

fmatthew5876:
Depends on #2050

Fix transition rendering

Fix: #1445
Fix: #1600
Fix: #1801 (both issues, + opacity issue in hh3 mentioned, + lufia glitch #1840 mentioned)
Fix: #1931
Fix: #1934
Fix: #1715

TBD: 
- [x] Retest issues after rebasing

PR #2026, opened 2019-12-31
Improve Window_Message timing accuracy - Messages 6 of N

fmatthew5876:
Depends on #1964 

Fixes more cases of #1706

This is yet another refactor of Window_Message. I've tried to clean up the Update loop to make it more understandable.

Unfortunately the number of edge cases here is just insane. So making this code readable is a real challenge.

There are a number of notable issues
* RPG_RT animates open/close after the main loop. Since we don't do this, I had to put in hacks to emulate the timing.
* I've finally nailed down the exact timing (and the rationale for why it works this way) for window open, close, and the interpreter blocking. 
* There are many, many subtle edge cases previously unknown with character timing. Even more-so with speed=1. Many have been discovered and added.

For this PR I have a much improved workflow to measure accuracy. To time any message, you can setup the following:

Foreground Event:
```
Msg: \^
FlashScreen 31,31,31,31 0.0s (no wait)
Var1 = 0
Msg: MESSAGE TO TEST
Msg: Time \v[1]
```

Parallel event
```
Var1+=1
```

To get individual character timings within a message, you have to do screen captures. I am doing this with `simplescreenrecorder` on linux by recording RPG_RT at 60fps. Then I use `ffmpeg -i vid.mkv frames%0004d.png` to dump the frames to png files and look at them.

The flash command in the above event helps identify exactly the frame where the message event command is executed and pushed into the messaging system.
The above construct will measure the exact number of frames the message takes to render.

Instead of writing page after page of test cases in issues. I have a test game with lots of message tests.

The test game is here. 

https://github.com/fmatthew5876/EasyRPG-InterpreterTestGame

All maps under the "Messages" Map tree have the #1706 test cases.
In particular the map `Messages / MESSAGE CRAZY` has a lot of NPCs with different text timing edge cases. Just talk to each NPC in RPG_RT and Player and compare.

#1706 Retested:
- [x] 1 - 16 Pass
- [x] 19 -24 Pass
- [x] 25 Timers retested - still working
- [x] 28-30 Pass
- [x] 31 We don't emulate system graphic corruption and shouldn't
- [x] 32-39 Pass

#1706 Newly fixed:
- [x] 26 Inn interruption tested with text, choices, and number input. 

- [x] 40 - tested and working
- [x] 41 - Timings match

#1706 Still broken:
- [ ] 17 & 18 Pushed to later PR
- [ ] 27 Test 1 - Pushed to later PR
- [ ] 27 Test 2 - box doesn't close.
- [ ] 42 - We don't interrupt the message
- [ ] 43 - Currently off by 1 frame.
- [ ] 44 - We set `show_message` chunk but it gets cleared every frame due to `Game_Map::UpdateForegroundEvents()` always calling `Game_Interpreter::Clear()` even when no events run.
- [ ] 45 Don't match

### DIFF
--- a/src/window_message.cpp
+++ b/src/window_message.cpp
@@ -125,22 +125,20 @@ Window_Message::~Window_Message() {
 }
 
 void Window_Message::StartMessageProcessing(PendingMessage pm) {
-	contents->Clear();
+	text.clear();
 	pending_message = std::move(pm);
-	SetIndex(-1);
-	kill_page = false;
 
 	if (!IsVisible()) {
 		DebugLogResetFrameCounter();
 	}
 	DebugLog("{}: MSG START");
 
-	const auto& lines = pending_message.GetLines();
-	if (!(pending_message.NumLines() > 0 || pending_message.HasNumberInput())) {
+	if (!pending_message.IsActive()) {
 		return;
 	}
 
-	text.clear();
+	const auto& lines = pending_message.GetLines();
+
 	int num_lines = 0;
 	auto append = [&](const std::string& line) {
 		bool force_page_break = (!line.empty() && line.back() == '\f');
@@ -181,6 +179,8 @@ void Window_Message::StartMessageProcessing(PendingMessage pm) {
 	item_max = min(4, pending_message.GetNumChoices());
 
 	text_index = text.data();
+
+	DebugLog("%d: MSG TEXT \n%s", text.c_str());
 
 	auto open_frames = (!IsVisible() && !Game_Battle::IsBattleRunning()) ? message_animation_frames : 0;
 	SetOpenAnimation(open_frames);
@@ -245,6 +245,18 @@ void Window_Message::InsertNewPage() {
 	face_request_ids.clear();
 
 	contents->Clear();
+	SetIndex(-1);
+	SetPause(false);
+	number_input_window->SetActive(false);
+	number_input_window->SetVisible(false);
+	kill_page = false;
+	line_count = 0;
+	text_color = Font::ColorDefault;
+	speed = 1;
+	kill_page = false;
+	instant_speed = false;
+	prev_char_printable = false;
+	prev_char_waited = true;
 
 	y = Game_Message::GetRealPosition() * 80;
 
@@ -273,13 +285,6 @@ void Window_Message::InsertNewPage() {
 	}
 
 	contents_y = 2;
-	line_count = 0;
-	text_color = Font::ColorDefault;
-	speed = 1;
-	kill_page = false;
-	instant_speed = false;
-	prev_char_printable = false;
-	prev_char_waited = true;
 
 	if (pending_message.GetNumberInputStartLine() == 0 && pending_message.HasNumberInput()) {
 		// If there is an input window on the first line
--- a/src/window_message.cpp
+++ b/src/window_message.cpp
@@ -296,7 +296,7 @@ void Window_Message::InsertNewPage() {
 		ShowGoldWindow();
 	} else {
 		// If first character is gold, the gold window appears immediately and animates open with the main window.
-		auto tret = Utils::TextNext(text_index, &*text.end(), Player::escape_char);
+		auto tret = Utils::TextNext(text_index, (text.data() + text.size()), Player::escape_char);
 		if (tret && tret.is_escape && tret.ch == '$') {
 			ShowGoldWindow();
 		}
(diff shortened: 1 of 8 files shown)


fmatthew5876:
This one is ready. There are still some minor things left. I can address them in later PRs.

This has been tested and now text timings are very accurate.

fmatthew5876:
This one just underwent rebase hell getting it over the stuff that just went in. Leaving a note for myself to do some testing of this again once dependencies are in and this is closer to being merged.

fmatthew5876:
jenkins test this please

carstene1ns:
Another test for this is PeuterisGrey. After the intro some textboxes are broken in current master.

fmatthew5876:
After #2019 goes in I'll rebase this again and retest everything.

carstene1ns:
This issue is fixed here (offset):
screenshot_3

This one is still present (name/item/variable substitution):
screenshot_0 screenshot_1 screenshot_2

fmatthew5876:
Rebased. Still needs to be retested.

fmatthew5876:
I reran my test cases and they still pass. I fixed the Peuteris Grey issue, that was a regression from when substitutions were refactored.

I've re-examined this complicated logic which animates the text box when transitions occur. The implementation I had was incomplete in several cases and not always frame accurate, so I've taken that out. It can be revisited in another PR.

fmatthew5876:
Are we still planning to get this in for 6.2 or punt to next release?

If we're punting, the "Message \n[] respect actor renames" commit should be broken out into a separate PR and included in 6.2. That's a very simple fix for a regression.

Ghabry:
I would prefer to only have the high priority commits in for 0.6.2 and the other commits for after.

fmatthew5876:
> 
> 
> I would prefer to only have the high priority commits in for 0.6.2 and the other commits for post-release.

That's the only one. Everything else here is a big nasty refactor perfecting message frame timing.

Ghabry:
This fixes a minor visual regression that didn't exist in 0.6.1: The Face graphic is not permanently shown. There is a 1 frame flicker everytime the page changes.

src/window_message.cpp
```diff
@@ -243,95 +329,116 @@ void Window_Message::InsertNewLine() {
...
+	close_started_this_frame = false;
+	close_finished_this_frame = false;
 
-		if (!Game_Message::IsMessagePending() && IsVisible() && !closing) {
-			// Start the closing animation
-			SetCloseAnimation(Game_Battle::IsBattleRunning() ? 0 : message_animation_frames);
-			// This frame a foreground event may push a new message and interupt the close animation.
-			allow_next_message = true;
-		}
+#if 0
```
Ghabry, line 381: #if 0

Ghabry:
Well with 0.6.2 out we can now get this here in and push to Android Beta (really Beta channel, not main xD) to get some feedback.

src/window_message.cpp
fmatthew5876: Removed

Ghabry:
Kinda message (choice) related:

There is a known regression about Inn continuation being broken when it costs money: https://github.com/rueter37/Player/commit/43a22be917b44837ea9fd9ba9d9aa8870b4239b1 (ugly solution)

The problem is that the async op is defined through a Choice and is never processed. Is this something that must be solved in messages, e.g. by passing in the async context:

```
Window_Message::InputChoice window_message.cpp:714
Window_Message::Update window_message.cpp:416
Game_Message::Update game_message.cpp:203
Game_Map::Update game_map.cpp:965 <- Has async ctx
```

Should thsi be solved on message or on interpreter side?

fmatthew5876:
Are we sure this is still a bug? I've tested inn quite a bit and never saw such an issue.

Interpreter continuations have been gone for quite some time now. 

What happens now is the inn continuation gets stored in the `PendingMessage` which passes it to `Window_Message`. When you select your choice, the window_message calls the inn callback immediately, and this updates the interpreter subcommand path.

This works similarly for all other continuations such as name changes, battle results, shop buy/sell, etc..

fmatthew5876:
The inn regression is completely fixed in #2212

The cross section of that PR with this one is minimal. I'm fine with rebaseding either one onto the other.

Ghabry:
Inn is in(n) so can be merged.
When 0.6.2.1 is out the current master can be pushed to Android Beta channel.
I don't want to push the beta before because the internal release number for Android is based on "commits in repo" and it is possible that stable is rejected because the number is lower 🙄🤔

Issue #759, opened 2016-02-08
CommandChangeClass not unsetting class correctly


Ghabry:
Basicly Game_Actor::SetClass needs

```
data.changed_class = class_id > 0
```

but I also dislike, that all the other logic is in Game_Interpreter and not in Game_Actor.
Have to check if something can be refactored.

Ghabry:
Skill learning based on class isn't implemented, either.

Ghabry:
"Add additional skills" option results in losing of skills that were added through event commands "Add Skill"

carstene1ns:
proposed patch for "Add additional skills":
```patch
diff --git a/src/game_interpreter.cpp b/src/game_interpreter.cpp
index fae389a9..df578fa9 100644
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -2497,25 +2497,26 @@ bool Game_Interpreter::CommandChangeClass(RPG::EventCommand const& com) { // cod
 
 	switch (skill_mode) {
 	case 0:
+		// No change
 		break;
 	case 1:
+		// Learn based on level (replace)
 		while (!actor->GetSkills().empty())
 			actor->UnlearnSkill(actor->GetSkills()[0]);
+
+		for (const RPG::Learning& learn : Data::classes[class_id - 1].skills) {
+			if (level >= learn.level)
+				actor->LearnSkill(learn.skill_id);
+		}
 		break;
 	case 2:
-	{
-		const RPG::Class& klass = Data::classes[class_id - 1];
-		while (!actor->GetSkills().empty())
-			actor->UnlearnSkill(actor->GetSkills()[0]);
-		std::vector<RPG::Learning>::const_iterator it;
-		for (it = klass.skills.begin(); it != klass.skills.end(); ++it) {
-			const RPG::Learning& learn = *it;
+		// Learn additionally (add)
+		for (const RPG::Learning& learn : Data::classes[class_id - 1].skills) {
 			if (level >= learn.level)
 				actor->LearnSkill(learn.skill_id);
 		}
 		break;
 	}
-	}
 
 	if (show && level > cur_lvl) {
 		// TODO
```

carstene1ns:
Moving the rest of this issue to 0.5.3 milestone.

Ghabry:
At least the following things are broken/wrong:

- Extra Whitespace between text and skill (only 2k3E?)
- Message shall be only printed when skill != "No change"
- Learn additional still gives skill lose in some cases :/
[...]

PR #2027, opened 2020-01-02
Fix 2k3 class change + Face graphic fixes

fmatthew5876:
Fix: #759 
Fix: #2039 ???

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -569,9 +569,9 @@ void Game_Actor::MakeExpList() {
 }
 
 std::string Game_Actor::GetExpString() const {
-	if (GetNextExp() == -1) {
-		return Player::IsRPG2k3() ? "-------" : "------";
-	}
+	// RPG_RT displays dashes for max level. As a customization
+	// we always display the amount of EXP.
+	// if (GetNextExp() == -1) { return Player::IsRPG2k3() ? "-------" : "------"; }
 	return std::to_string(GetExp());
 }
 
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -67,7 +67,7 @@ void Game_Actor::Init() {
 	const std::vector<RPG::Learning>& skills = GetActor().skills;
 	for (int i = 0; i < (int)skills.size(); i++) {
 		if (skills[i].level <= GetLevel()) {
-			LearnSkill(skills[i].skill_id);
+			LearnSkill(skills[i].skill_id, nullptr);
 		}
 	}
 
@@ -110,7 +110,7 @@ bool Game_Actor::UseItem(int item_id, const Game_Battler* source) {
 
 	if (!IsDead()) {
 		if (item->type == RPG::Item::Type_book) {
-			return LearnSkill(item->skill_id);
+			return LearnSkill(item->skill_id, nullptr);
 		}
 
 		if (item->type == RPG::Item::Type_material) {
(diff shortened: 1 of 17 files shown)


src/window_message.cpp
```diff
@@ -246,7 +246,7 @@ void Window_Message::InsertNewPage() {
 		gold_window->SetBackOpacity(GetBackOpacity());
 	}
 
-	if (!Game_Message::GetFaceName().empty()) {
+	if (IsFaceEnabled() && !Game_Message::GetFaceName().empty()) {
```
carstene1ns: double check.

fmatthew5876: Fixed, thanks!

src/game_actor.cpp
```diff
@@ -1118,31 +1119,58 @@ const RPG::Class* Game_Actor::GetClass() const {
...
-
-	GetData().class_id = _class_id;
+void Game_Actor::ChangeClass(int new_class_id,
+		int new_level,
+		ClassChangeSkillMode new_skill,
+		ClassChangeParamMode new_param,
+		PendingMessage* pm
+		)
+{
+	Output::Debug("CLASS CHANGE %d %d %d %d", new_class_id, new_level, new_skill, new_param);
```
carstene1ns: debug leftover.

fmatthew5876: Thanks for catching that, fixed.

fmatthew5876:
I added back the custom feature to always display EXP amount even at maximum level. Left a comment including correct commented out code. I personally like it better this way.

If you guys disagree we can remove it and be strictly compatible with RPG_RT.

Ghabry:
This is hard to test because RPG\_RT behaviour makes no sense :roll_eyes:

fmatthew5876:
https://github.com/fmatthew5876/EasyRPG-InterpreterTestGame

I have a Class test map which I used to test all the edge cases

Ghabry:
From a code review PoV I accept this.

---------
Bug 1:
@> Change Level: Entire Party, + 30
@> Change Actor Class: [Zack], [Soldier A] with Lv: Change to 1 Skills: Learn additional, Stats: From Level 1

Menu shows EXP: 316035 / 768

Skill menu shows MP: 0 / [NOTHING] <- rendered out of bounds
-----------------
Bug 2:

@> Change Level: Entire Party, + 30
@> Change Actor Class: [Zack], to classlessLv: No change Skills: Based on Level Stats: By level
@> Change Level: Entire Party, + 30
@> Change Actor Class: [Zack], to classlessLv: No change Skills: Learn additional Stats: Halved

This shows only "Reached LV30" in RPG_RT and Reached 30, Reached 60 in Player. But our behaviour looks more sane.

fmatthew5876:
For bug 2, we aren't strictly compatible with what level up messages we show. RPG_RT is very strange and buggy here. 

I tried to make the way we do it more sensible

Ghabry:
I just thought about BUG 2 again because I forgot some detail:
```
@> Change Level: Entire Party, + 30 [ ] Show Message
@> Change Actor Class: [Zack], to classlessLv: No change Skills: Based on Level Stats: By level [X] Show Message
@> Change Level: Entire Party, + 30 [ ] Show Message
@> Change Actor Class: [Zack], to classlessLv: No change Skills: Learn additional Stats: Halved [X] Show Message
```
So the messages for "Change Level" were disabled why does "Change class" show any level up message then?
Question is just about whether showing nothing or both makes more sense.
At least it doesn't affect the logic, so users probably won't care except for timing edge cases.

fmatthew5876:
Change class shows the level message when skills are changed. I think the reason rpgrt does this is because the level up message contains the actor name but the learning messages do not.

We mostly follow rpgrt except stupid edge cases where you change to the same class where rpgrt has a lot of bugs.

The behaviors are detailed in the test cases.

I will recheck this stuff later this evening to be sure

fmatthew5876:
Fixed the EXP issue and the menu SP text.
For menus I rechecked:
- [x] Main Menu
- [x] Skills
- [x] 2k3 battle
- [x] 2k battle
- [x] ActorTarget Window from using a healing skill / item

Also rechecked your level 30/60 example. RPG_RT is buggy here and ours is more correct so left it as is.

Issue #1417, opened 2018-09-11
Pictures behind the Tilemap are not rendered when there is no Panorama


Ghabry:
This is caused by the "BlitFast" optimisation that simply draws a non-transparent tilemap on the bottom layer when there is no panorama.

[Test%20case%20pic%20BG.zip](https://github.com/EasyRPG/Player/files/2372571/Test.20case.20pic.20BG.zip)

PR #2028, opened 2020-01-04
Optimize Tilemap

fmatthew5876:
Retested #1417 

There was a performance regression introduced into tilemap when tone support was added. Basically we never actually use the `TileOpacity` value, resulting in us rendering every tile, even when they are fully transparent.

This PR fixes that and adds several other optimizations to TilemapLayer. 

The test case is a map with all transparent tiles and no background. Each optimization commit in this PR was measured and shown to have an improvement for this test.

For `ComputeImageOpacity()`

Master:
```
---------------------------------------------------------------------
Benchmark                              Time           CPU Iterations
---------------------------------------------------------------------
BM_ComputeImageOpacity            352849 ns     352851 ns       1969
BM_ComputeImageOpacityChipset     862583 ns     862591 ns        800
```

This PR:
```
---------------------------------------------------------------------
Benchmark                              Time           CPU Iterations
---------------------------------------------------------------------
BM_ComputeImageOpacity             70558 ns      70557 ns       9301
BM_ComputeImageOpacityChipset     116219 ns     116219 ns       5241
```

Master:
tilemap_master
The calls to `operator()` are the div and mod lambdas in `TilemapLayer::Draw()`

This PR:
tilemap

Note in both profiles, the amount of time spent in `Game_picture::Update()` even though there are no pictures. This is fixed in #1989

### DIFF
--- a/src/bitmap.h
+++ b/src/bitmap.h
@@ -164,12 +164,12 @@ public:
 	 * Provides opacity information about a tile on a tilemap.
 	 * This influences the selected operator when blitting a tile.
 	 *
-	 * @param row tile row
-	 * @param col tile col
+	 * @param x tile x coordinate
+	 * @param y tile y coordinate
 	 *
 	 * @return opacity information
 	 */
-	ImageOpacity GetTileOpacity(int row, int col) const;
+	ImageOpacity GetTileOpacity(int x, int y) const;
 
 	/**
 	 * Writes PNG converted bitmap to output stream.
@@ -575,8 +575,8 @@ inline ImageOpacity Bitmap::GetImageOpacity() const {
 	return image_opacity;
 }
 
-inline ImageOpacity Bitmap::GetTileOpacity(int row, int col) const {
-	return !tile_opacity.Empty() ? tile_opacity.Get(row, col) : ImageOpacity::Partial;
+inline ImageOpacity Bitmap::GetTileOpacity(int x, int y) const {
+	return !tile_opacity.Empty() ? tile_opacity.Get(x, y) : ImageOpacity::Partial;
 }
 
 inline Color Bitmap::GetBackgroundColor() const {
--- a/src/bitmap.h
+++ b/src/bitmap.h
@@ -151,14 +151,6 @@ public:
 		Flag_ReadOnly = 1 << 16
 	};
 
-	enum TileOpacity {
-		// Image is full opaque and can be blitted fast
-		Opaque,
-		// Image has alpha and needs an alpha blit
-		Partial,
-		// Image is complately transparent
-		Transparent
-	};
 
 	/**
 	 * Provides opacity information about the image.
@@ -166,7 +158,7 @@ public:
 	 *
 	 * @return opacity information
 	 */
-	TileOpacity GetOpacity() const;
+	ImageOpacity GetImageOpacity() const;
 
 	/**
 	 * Provides opacity information about a tile on a tilemap.
@@ -177,7 +169,7 @@ public:
 	 *
 	 * @return opacity information
 	 */
-	TileOpacity GetTileOpacity(int row, int col) const;
+	ImageOpacity GetTileOpacity(int row, int col) const;
 
 	/**
 	 * Writes PNG converted bitmap to output stream.
@@ -538,13 +530,13 @@ public:
 	int bpp() const;
 	int pitch() const;
 
-protected:
-	TileOpacity CheckOpacity(Rect const& rect);
+	ImageOpacity ComputeImageOpacity(Rect rect) const;
 
+protected:
 	DynamicFormat format;
 
-	std::vector<std::vector<TileOpacity>> tile_opacity;
-	TileOpacity opacity = Partial;
+	std::vector<std::vector<ImageOpacity>> tile_opacity;
+	ImageOpacity image_opacity = ImageOpacity::Partial;
 	Color bg_color, sh_color;
 
 	friend void Text::Draw(Bitmap& dest, int x, int y, int color, FontRef font, std::string const& text, Text::Alignment align);
@@ -577,5 +569,21 @@ protected:
 	pixman_op_t GetOperator(pixman_image_t* mask = nullptr) const;
 	bool read_only = false;
 };
+
+inline ImageOpacity Bitmap::GetImageOpacity() const {
+	return image_opacity;
+}
+
+inline ImageOpacity Bitmap::GetTileOpacity(int row, int col) const {
+	return !tile_opacity.empty() ? tile_opacity[row][col] : ImageOpacity::Partial;
+}
+
+inline Color Bitmap::GetBackgroundColor() const {
+	return bg_color;
+}
+
+inline Color Bitmap::GetShadowColor() const {
+	return sh_color;
+}
 
 #endif
--- a/src/bitmap.h
+++ b/src/bitmap.h
@@ -536,8 +536,8 @@ public:
 protected:
 	DynamicFormat format;
 
-	std::vector<std::vector<ImageOpacity>> tile_opacity;
 	ImageOpacity image_opacity = ImageOpacity::Partial;
+	TileOpacity tile_opacity;
 	Color bg_color, sh_color;
 
 	friend void Text::Draw(Bitmap& dest, int x, int y, int color, FontRef font, std::string const& text, Text::Alignment align);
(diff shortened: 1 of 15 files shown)


fmatthew5876:
Here is a profile of a map with grass and water tiles
grassmap



Master:
grass-old

This PR:
grassnew

fmatthew5876:
I tested Witch's Heart which has a lot of crazy maps and a map with changing tone effects. Everything worked.

src/tilemap_layer.cpp
```diff
@@ -566,7 +578,7 @@ BitmapRef TilemapLayer::GenerateAutotiles(int count, const std::map<uint32_t, Ti
 				rect.x = (x * 2 + i) * (TILE_SIZE/2);
 				rect.y = (y * 2 + j) * (TILE_SIZE/2);
 
-				tiles->Blit((dst.x * 2 + i) * (TILE_SIZE / 2), (dst.y * 2 + j) * (TILE_SIZE / 2), *chipset, rect, 255);
+				tiles->BlitFast((dst.x * 2 + i) * (TILE_SIZE / 2), (dst.y * 2 + j) * (TILE_SIZE / 2), *chipset, rect, 255);
```
Ghabry: There is no game using transparent auto tiles but I have to flag this for an accuracy test

src/tilemap_layer.cpp
```diff
@@ -186,26 +218,35 @@ void TilemapLayer::Draw(Bitmap& dst, int z_order) {
...
-	for (int x = 0; x < tiles_x; x++) {
-		for (int y = 0; y < tiles_y; y++) {
+	const bool loop_h = Game_Map::LoopHorizontal();
+	const bool loop_v = Game_Map::LoopVertical();
+
+	auto div_rounding_down = [](int n, int m) {
+		if (n >= 0) return n / m;
+		return (n - m + 1) / m;
+	};
+	auto mod = [](int n, int m) {
```
Ghabry: For this we already have Utils::PositiveModulo.
But my benchmarks say that your version is faster than what we already have, how about replacing it?

src/tilemap_layer.cpp
```diff
@@ -141,35 +142,66 @@ TilemapLayer::TilemapLayer(int ilayer) :
...
+{
 }
 
-void TilemapLayer::DrawTile(Bitmap& dst, Bitmap& screen, int x, int y, int row, int col, bool allow_fast_blit) {
-	Bitmap::TileOpacity op = screen.GetTileOpacity(row, col);
+EP_ALWAYS_INLINE
+void TilemapLayer::DrawTile(Bitmap& dst, Bitmap& tileset, Bitmap& tone_tileset, int x, int y, int row, int col, uint32_t tone_hash, bool allow_fast_blit) {
+	auto op = tileset.GetTileOpacity(col, row);
+	if (op != ImageOpacity::Transparent) {
+		DrawTileImpl(dst, tileset, tone_tileset, x, y, row, col, tone_hash, op, allow_fast_blit);
```
Ghabry: I don't understand why there is now a DrawTileImpl function call. Besides inverting the if why isn't the cost just put in here?

src/bitmap.h
```diff
@@ -578,4 +571,20 @@ class Bitmap {
 	bool read_only = false;
 };
 
+inline ImageOpacity Bitmap::GetImageOpacity() const {
+	return image_opacity;
+}
+
+inline ImageOpacity Bitmap::GetTileOpacity(int row, int col) const {
+	return !tile_opacity.Empty() ? tile_opacity.Get(row, col) : ImageOpacity::Partial;
```
Ghabry: Why is it here Get(row, col)

src/tilemap_layer.cpp
```diff
@@ -141,35 +142,66 @@ TilemapLayer::TilemapLayer(int ilayer) :
...
+	// Its z-value should be between the z of the events in the upper layer and the hero
+	upper_layer(this, Priority_TilesetAbove + layer)
+{
 }
 
-void TilemapLayer::DrawTile(Bitmap& dst, Bitmap& screen, int x, int y, int row, int col, bool allow_fast_blit) {
-	Bitmap::TileOpacity op = screen.GetTileOpacity(row, col);
+EP_ALWAYS_INLINE
+void TilemapLayer::DrawTile(Bitmap& dst, Bitmap& tileset, Bitmap& tone_tileset, int x, int y, int row, int col, uint32_t tone_hash, bool allow_fast_blit) {
+	auto op = tileset.GetTileOpacity(col, row);
```
Ghabry: And here it was inverted to Get(col, row)

src/opacity.h
```diff
@@ -56,4 +58,73 @@ struct Opacity {
...
+		bool Empty() const;
+
+	private:
+		std::unique_ptr<uint8_t[]> _p;
+		int _w = 0;
+		int _h = 0;
+
+		/** Must be enough bits to store all valid ImageOpacity values */
+		static constexpr int tile_bit = 2;
+		static constexpr int tile_mask = (1 << tile_bit) - 1;
```
Ghabry: tile_mask looks unused.

src/tilemap_layer.cpp
Ghabry: I don't really know why but semi-transparent autotiles still render properly :thinking:

fmatthew5876: This is creating a secondary bitmap to store the generated autotiles. Since the bitmap is previously empty, we can fast blit the autotiles onto it.

src/tilemap_layer.cpp
fmatthew5876: This isn't my version, it was already here. Given this has a branch, I would expect either the utils variant to be faster or the compiler to be smart enough to optimize them both to the same thing.

If you saw utils version to be slower, it's probably because it's currently not inlined.

src/tilemap_layer.cpp
fmatthew5876: The 2 level functions here is to inline the transparency check. For the common case (especially on upper level) where we have transparent tiles avoid paying for a function call just to do the check and return.

This also makes the check duplicated in `Draw()` for every tile type instead of just existing in one place in `DrawTile()`. Now that there are separate branches, that allows the branch predictor to predict separate against each tile type, instead of having one prediction for the entire tile set. Since each class of tiles has a different probability of being transparent, this is a net win.

Both of my vtune profile examples showed a gain of 1-2% less time spent drawing tiles with this change,

src/bitmap.h
fmatthew5876: Typo: this should be `x` and `y` to match the new `TIleOpacity` class.

src/tilemap_layer.cpp
fmatthew5876: The tile opacity code is now in terms of (x,y) so this has to pass (col, row)

src/opacity.h
fmatthew5876: Removed.

src/tilemap_layer.cpp
fmatthew5876: https://gcc.godbolt.org/z/rFUMwH

They do produce different code. You trade a branch for 2 divisions.

I would want to benchmark this more carefully before deciding which to use.

fmatthew5876: Punting this one to #2042

fmatthew5876:
Addressed review comments and rebased

src/tilemap_layer.cpp
Ghabry: okay, can you add a short comment there that this is force-inline + call is intentional as it is?

To prevent others from killing this optimisation later by accident..

fmatthew5876: Sounds good, added

Ghabry:
Another candidate for the "address some warnings" commit because you just added this line:

Clang reports:
```
easyrpg-player/src/game_actor.cpp:1234:4: warning: declaration does not declare anything [-Wmissing-declarations]
                        EP_FALLTHROUGH;
                        ^
easyrpg-player/src/compiler.h:24:24: note: expanded from macro 'EP_FALLTHROUGH'
#define EP_FALLTHROUGH __attribute__((fallthrough))
                       ^
```

fmatthew5876:
Added a fix for fall through.

src/opacity.h
```diff
@@ -56,4 +56,18 @@ struct Opacity {
...
+ * Enum used to describe the opacity of an entire image.
+ * This is used to decide whether we can shortcut fast blit
+ * operations
+ */
+enum class ImageOpacity {
+	/** Image has alpha and needs an alpha blit */
+	Partial,
+	/** Image is full opaque and can be blitted fast */
+	Opaque,
+	/** Image is complately transparent and blitting can be skipped entirely */
```
carstene1ns: I know this typo was there before, but might be worth to fix… :zipper_mouth_face:

fmatthew5876:
rebased and fixed the comment

fmatthew5876:
Rebased again after mainloop1 PR.

I did a quick comparison of Frozen Triggers with vtune. Tilemap drawing is not a bottleneck in that game, but it does go from 0.5% to 0.2%.

Ghabry:
@fmatthew5876 
This is easy to explain: The game uses a 1x1 map ;). I know this because I fixed the renderer for the less than 20x15 case :D

Issue #2016, opened 2019-12-26
Aëdemphia: Image BMP needs support for BITMAPCOREHEADER


Ghabry:
#### Attach files (as a .zip archive or link them)
Problematic bitmap:
[continentest.zip](https://github.com/EasyRPG/Player/files/4002983/continentest.zip)

#### Describe the issue in detail and how to reproduce it:
This happens on the world map but the real issue is our bitmap reader (image_bmp).

1. The size of the DIP header at offset 0Eh should be checked. We only support the Win32 BITMAPINFOHEADER which has a size of 40d but there are others
2. Here the size is 12d, which means it is a BITMAPCOREHEADER: https://docs.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-bitmapcoreheader

carstene1ns:
Ok. So these really exist in the wild. :8ball:
(Good thing this game is not big already. Might bloat this up with some bitmaps)

PR #2031, opened 2020-01-04
Support BITMAPCOREHEADER for bmp format

fmatthew5876:
Fix: #2016

### DIFF
--- a/src/image_bmp.cpp
+++ b/src/image_bmp.cpp
@@ -24,13 +24,68 @@
 #include "output.h"
 #include "image_bmp.h"
 
-static uint16_t get_2(const uint8_t *p) {
-	return (uint16_t) p[0] | (p[1] << 8);
+static uint16_t get_2(const uint8_t *&p, const uint8_t* e) {
+	if (e - p < 2) {
+		p = e;
+		return 0;
+	}
+	auto value = static_cast<uint16_t>(p[0] | (p[1] << 8));
+	p += 2;
+	return value;
 }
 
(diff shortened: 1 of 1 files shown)


fmatthew5876:
The part where we do `min(256, biClrUsed)` looks very suspicious.

@fdelapena What was the context behind https://github.com/EasyRPG/Player/pull/248

What kinds of erroneous values were we seeing in this field?

fdelapena:
That std::min existed since the [initial BMP reader code by glynnc](https://github.com/EasyRPG/Player/commit/644b6bce5e04edc4be7a3dfcf244d86c03231f74#diff-f7dbf665765dd93b3b29e5f3ec088f62R126). I don't know the reason for that one.

carstene1ns:
What is the problem with it? It just caps the palette size at 256 entries.

fmatthew5876:
It's a `min` not a `max`. What if the image actually has less than 256 colors? It looks like we'll buffer overflow in that case in the next bit of code which walks the palette.

carstene1ns:
So? If the the palette has any number of colors less than 256, it will be that (e.g. 16)? If it is bigger (i.e. bogus), it will be 256.

fmatthew5876:
Hmm you're right I was getting mixed up. I guess it is ok then, although I am still not sure if we need it. Regardless it was in the previous code and I have no intention currently to change it.

carstene1ns:
Unfortunately this breaks the newlib-based platforms again (int types / long).

PR #2034, opened 2020-01-05
Fix save window when party members have no face graphic

fmatthew5876:
Save window would not display name, level, or hp when the first actor has no face graphic.

### DIFF
--- a/src/window_savefile.cpp
+++ b/src/window_savefile.cpp
@@ -49,7 +49,7 @@ void Window_SaveFile::UpdateCursorRect() {
 std::string Window_SaveFile::GetSaveFileName() const {
 	std::ostringstream out;
 	if (!override_name.empty()) {
-		if (override_name.size() > 14 && !party.empty()) {
+		if (override_name.size() > 14 && has_party) {
 			out << override_name.substr(0, 11) << "...";
 		} else {
 			out << override_name;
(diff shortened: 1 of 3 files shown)


src/window_savefile.cpp
```diff
@@ -134,12 +131,17 @@ void Window_SaveFile::Refresh() {
...
 	int hx = Font::Default()->GetSize(hp_short).width;
 	out.str("");
-	out << std::setw(Player::IsRPG2k3() ? 4 : 3) << std::setfill(' ') << hero_hp;
+	out << std::setw(Player::IsRPG2k3() ? 4 : 3) << std::setfill(' ') << data.hero_hp;
 	contents->TextDraw(46 + hx, 32 + 2, fc, out.str());
 
-	for (int i = 0; i < 4 && (size_t) i < party.size(); i++) {
-		DrawFace(party[i].second, party[i].first, 92 + i * 56, 0, false);
-	}
+	int i = 0;
```
carstene1ns: ...is this a micro optimization? :grinning:

Could even save the `++i` lines and use `i++` directly in `DrawFace`.

fmatthew5876: Removing the vector was a micro optimization of sorts, but it's more about convenience of reusing the `SaveTitle` struct.

Added `i++`, much cleaner that way thanks!

PR #2035, opened 2020-01-05
Add Scene_Settings - Config 3 of N

fmatthew5876:
Depends on #2108 

This is a WIP to add a settings menu. Early feedback is welcome on the general layout and feature set this should have.

Input:
- [ ] Add input remapping for keyboard
- [ ] Add input remapping for gamepads

Audio:
- [ ] BGM Volume
- [ ] SFX Volume
- [ ] Configure Midi backend to use
- [ ] Configure Midi soundfont to use
- [ ] Configure which system audio device to output to.
- [ ] Number of output channels
- [ ] Output sample rate
- [ ] Output bit depth
- [ ] Bit-perfect output toggle?

Video:
- [x] Show which renderer is active and leave future support for multiple renderers
- [x] Fullscreen
- [ ] Screen Resolution
- [ ] Vsync
- [ ] Frame Limiter
- [x] Show FPS

Access:
- [x] Game Browser
- [x] Title Screen
- [x] Menu Scene
- [x] OpenVideoOption event command
- [ ] Add a key to open the settings menu

Persistence:
- [ ] Save and load settings to user config file

Others:
- [ ] Refresh UI if video options changed by key press
- [x] Show title screen in background


Ghabry:
Adding anything to Scene_Title by default is a bad idea and we shouldn't do this (okay there is already Scene_Import but this opt-in and hidden by default). There are games that put the menu items in the title graphic and use a system graphic designed in such a way that it resembles a cursor:

screenshot_0_-fs8

fmatthew5876:
Most modern games let you adjust settings at the title, but I totally get your point that this will break carefully constructed title screens.

We can make it opt in with game config ini.

Ghabry:
config 2 of N is in, can you rebase this one?

fmatthew5876:
This is rebased, but I need to review all of it from scratch as it is very old. Play with it if you like, but this is not ready for review yet.

Ghabry:
sure I already started playing with it :)

Two issues I found:
- When I go from Input to Video the first line is empty
- VSync toggling does nothing (expected due to API limitation?)

fmatthew5876:
> When I go from Input to Video the first line is empty

No idea, will check 

> VSync toggling does nothing (expected due to API limitation?)

I think this is because vsync changes require recreating the sdl renderer, basically it needs some work in SDL2 ui which I haven't done yet. If not done in this PR, we should probably disable the option so as not to confuse the users.

fdelapena:
Jenkins: test this please

Ghabry:
Rebased this in https://github.com/ghabry/easyrpg-player/tree/sysmenu because I will use a submenu here to handle license attributions (#2500)

Ghabry:
I'm taken this over. Replaced by #2730

PR #2037, opened 2020-01-06
Fix newlib build

fmatthew5876:


carstene1ns:
Tested. Works. Thank you!

Issue #2038, opened 2020-01-07
Characters freeze when using hidden channels


s9060666:
Thank you for supporting EasyRPG Player.
Before creating an issue, please verify that your bug still occurs in the "continuous builds" of the Player.
You can get them at the bottom of our download section: https://easyrpg.org/player/downloads

---
Please fill in the following fields before submitting an issue:
#### Player platform:Android 
Windows, Linux, Android, ...

#### Attach files (as a .zip archive or link them)
 - A savegame next to the problem (if it is not at the beginning of the game)
 - The `easyrpg_log.txt` log file (you can find it in the game directory)

#### Describe the issue in detail and how to reproduce it:Using the record I provided, the character presses the confirmation key and a dialogue appears, and the character freezes
https://drive.google.com/file/d/15Kz2gTy4SjQJIe17JUKmMruLQVdlkqZX/view?usp=drivesdk

s9060666:
http://www.mediafire.com/download/k0969lkmqqax6hu/2003_%E5%A4%A9%E4%BD%BF%E8%9B%8B%E7%B3%95_%E6%9C%88%E8%97%8D%E5%82%B3%E5%A5%87%EF%BC%A9%EF%BC%A9%EF%BD%9E%E5%92%92%E5%8D%B0%E6%97%8F%E5%82%B3%E8%AA%AA.rar
This is the download link for the game

fmatthew5876:
Move routes bug. Somehow an event is getting stuck on this code:

```
SetMoveRoute: Player(skip), Through On, Move Up
WaitForAllMovement
```

The player steps up, but the `WaitForAllMovement` command never finishes.

fmatthew5876:
@s9060666 This looks like a bug in the game itself and not a bug in Player.

I am unable to run chinese RPG_RT.exe on my computer. Can you please confirm whether or not the bug still happens with RPG_RT.exe?

Finally, if you want to workaround the issue. You can walk down, exit the map, come back, and then open the secret door and it won't hang anymore.

fmatthew5876:
Ok this is a real Player bug, the save game is messed up because it's already after the bug occurred.

To see the actual bug, do the following:

1. Load the save game
2. Walk down and exit the map
3. Set Variable35 = 5
4. Re-enter the map
5. A cutscene will play
6. Then you will regain control of the player on the map.

In RPG_RT, the Event 5 from the cutscene is near the top of the map wandering around. If you try to enter the secret cave, it works.

[...]

Issue #1778, opened 2019-05-27
Yume2kki: Music room broken, reports "Unknown Event 10005"


Ghabry:
Name of the game: Yume2kki

Download 108c JP from here: https://rmarchiv.tk/games/1905

#### Describe the issue in detail and how to reproduce it:

Fast forward through the intro, go to your PC, select Music.
Try to play anything, the console is spammed with 10005 (This Event) warnings.

This is not game breaking because the Music room is an optional bonus.

I think this happens because the event ID is lost (ID=0) upon map transition and the ThisEvent resolving breaks.

Setting this to 0.6.2 as I expect this to be fixed by #1742, just not verified yet because this PR breaks before even reaching the room :).

fmatthew5876:
This doesn't happen in version 106, only 108

What looks like is happening here:

EV01 Page 4 in Map 620 calls it's own Page 1 multiple times.

Page1 does a `SetVehicleLocation(Party)` to hack teleport the party to Map 2, and the again back to Map620 (why???) :question: 

In Player, when this happens we do a full teleport, instead of the half broken one RPG_RT does.

This causes event_id chunk to be wiped out the first time `CallEvent(ThisEvent[1])` happens. The next time it gets called, player thinks its a common event and craps out.

I've confirmed that `CallEvent(ThisEvent)` is not supported by any version of RPG_RT after you teleport (the event id goes to 0)

Ghabry:
We concluded that this is too much bug abuse and deleting the 2 lines seems to fix the code, so they are useless.

Keeping this open, while I try to report a bug to the game devs.

Ghabry:
Maybe it's hacky enough to set the map_id of the current event to the target map at https://github.com/EasyRPG/Player/blob/master/src/game_interpreter.cpp#L1729 but I close this bug now as the discord chat decided: They don't care and will not escalate it to the Japanese devs.

fmatthew5876:
Test Cases 1 - Invalid Events
==============

1. Setup Map1 EV01 action=trigger `SetVehicleLocation: 0, -1, -1`

Setup Map2 with no events

Result: Crash with invalid event id

2. Create EV02 on Map2

Result: Warps to other map.

fmatthew5876:
Test Cases 2 - Pictures
==============

1. Setup Map1 EV01 action=trigger
```
Show Picture: Some pic
SetVehicleLocation: 0, -1, -1
```

Setup Map2 with EV01

[...]

PR #2041, opened 2020-01-10
Handle teleports run by foreground events in preupdate

fmatthew5876:
When a foreground event causes a teleport, foreground
events run during pre-update. If then during that preupdate
the fg event triggers another teleport, we need to handle it.

Fix: #2038

Leaving this as a draft as I want to do more testing around this scenario.

### DIFF
--- a/src/scene_map.cpp
+++ b/src/scene_map.cpp
@@ -377,6 +377,17 @@ void Scene_Map::FinishPendingTeleport3(MapUpdateAsyncContext actx, TeleportParam
 			OnAsyncSuspend([=] { FinishPendingTeleport3(actx, tp); }, actx.GetAsyncOp(), true);
 			return;
 		}
+
+		if (!tp.defer_recursive_teleports) {
+			// See comments about defer_recursive_teleports in FinishPendingTeleport2
+			// Deferring in this block can actually never occur, since an Escape/Teleport won't ever
+			// also trigger foreground events to run in pre-update. We leave the check in here for symmetry.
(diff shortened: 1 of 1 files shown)


Ghabry:
Is this about the "Teleport" event command or about the "Recall to Teleport place" event command?

Your PR comment is about the first but the code comment is about the second.

fmatthew5876:
It's both of those commands.

The comment actually refers to Escape/Teleport skills. If you use an escape/teleport skill and then a parallel event does another teleport, that second teleport is deferred. This is what the comment is referring to.

This code block is copied from `FinishPendingTeleport2()`

fmatthew5876:
I still want to do more tests around self teleports and interpreter behavior, but this change by itself is obviously fixing an oversight and fixes a hang bug. So I mark this as ready.

Ghabry:
The "Unknown event 10005" stuff was already resolved again it seems but this PR solves the hang after pening the music room (#1778)

fmatthew5876:
This code looked suspicious to me so I added more tests to #2038. Against my intuition, it turns out this code is correct and is exactly what RPG_RT does.

The most interesting bit is that on normal frames, if a parallel event initiates teleports, all parallel events, and the foreground interpreter still runs. 

But on pre-update if a parallel event initiates a teleport, all parallel events run, but foreground events will not run. This is because of the "chain" effect where if you have a sequence parallel events on each map teleporting you A->B->C->D, it'll appear as if you only teleported A->D.

PR #2050, opened 2020-01-15
Cleanup main loop - MainLoop 3 of N

fmatthew5876:
Depends on #1973 

This pulls the first set of commits out of #2017 which just cleans up our current main loop without doing a major refactor. The main purpose of this is to move the time keeping and loop management all into `Player::MainLoop`. 

We also do drawing *after* suspending scenes, which is necessary for visual correctness of transitions initiated by `Scene::TransitionOut()`. It also makes sure transitions are updated with the rest of the game logic, so that they animate properly even under fast forward or frameskip. With respect to transitions, this PR has all the dependencies needed for #2019 

This also ensures sleep caused by vsync happens at near the same time as mandatory sleep.

Finally, this also fixes the framerate histogram in vtune.

This should be mergable sooner, while we test #2017 further.

### DIFF
--- a/src/graphics.cpp
+++ b/src/graphics.cpp
@@ -101,7 +101,6 @@ void Graphics::Update() {
 	fps_overlay->Update();
 	fps_overlay->AddUpdate();
 	message_overlay->Update();
-	Transition::instance().Update();
 }
 
 void Graphics::UpdateTitle() {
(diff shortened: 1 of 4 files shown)


fmatthew5876:
Tested this pretty well with #2019 and #2066 . This one is ready for review.

Issue #2036, opened 2020-01-05
TDF2018 Wild demo crashes in emscripten


fmatthew5876:
https://easyrpg.org/play/master/?game=tdf2018wildcompodemo

Start a new game and the demo immediately crashes with an exception
```
uncaught exception: abort("Assertion failed: req != nullptr && req->IsReady(), at: src/cache.cpp,283,LoadBitmap")
```

Ghabry:
That's not really useful.
Before replacing it with an assert there was a debug print "BUG: Not requested Path/Filename" providing a hint.

Ghabry:
Using SAFE_HEAP + WASM debugging (really cool that this works now) this is a dangling reference to a Game_Picture when a "OnPictureReady" callback is fired. Dangling reference?

Can't test with WASM because the local server I use for testing (python http.server) does not support WASM streaming initialisation (whatever that is) and for array init the functions are too huge.

fmatthew5876:
This could be the picture vector resizing itself in between file async operations.

Ghabry:
Indeed: This must be caused by (Do)PreallocPicData. When I hardcode an array size there it doesn't crash.

fmatthew5876:
I have an idea how to fix, and also add further picture performance improvements. Will look into after #2003 goes in.

PR #2051, opened 2020-01-15
Fix crash when picture loading races with vector resize

fmatthew5876:
Fix: #2036

Temporary fix for the crash until a proper refactor is made.

https://easyrpg.org/play/pr2051/?game=tdf2018wildcompodemo


fmatthew5876:
Unfortunately this broke loading save games because the constructor and  `RequestFileSprite()` is called before the picture gets into the vector!

Will provide another fix..

PR #2052, opened 2020-01-16
Refactor Drawable transfer + visible + bitmap bug fix

fmatthew5876:
Adds a much cleaner and more efficient mechanism for battle pictures / weather. Also reverts a regression from #1990

Fix: #2058

### DIFF
--- a/src/drawable.h
+++ b/src/drawable.h
@@ -77,6 +77,8 @@ public:
 		Global = 1,
 		/** This is a shared drawable which will appear in all scenes that use shared drawables */
 		Shared = 2,
+		/** This flag indicates the drawable should not be drawn */
+		Invisible = 4,
 		/** The default flag set */
 		Default = None
 	};
@@ -101,6 +103,16 @@ public:
 
 	/* @return true if this drawable should appear in all scenes that use shared drawables */
 	bool IsShared() const;
+
+	/* @return true if the drawable is currently visible */
+	bool IsVisible() const;
+
+	/**
+	 * Set if the drawable should be visisble
+	 *
+	 * @param value whether is visible or not.
+	 */
+	void SetVisible(bool value);
 
 	/**
 	 * Converts a RPG Maker map layer value into a EasyRPG priority value.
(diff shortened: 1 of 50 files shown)


fmatthew5876:
Retested battle pictures, weather, and a few other games.

The next thing I want to do is get rid of the global drawable list, put all the global drawables in the local list, and transfer them using this mechanism every scene change. Then we wouldn't have to worry about issues like if the last item in the local list is supposed to be rendered above the first item in the global list.

However I don't want to do that until after this and #2019 as attempting it now causes a lot of issues with transitions.

carstene1ns:
`tests/drawable_mgr.cpp` needs to be updated.

Looks much cleaner now with the shared logic. :+1:

Ghabry:
and the error handling for OSX and Windows PR is broken. The test build failed there but still reported as success ^^'

fmatthew5876:
Fixed the tests

src/drawable.h
```diff
@@ -69,7 +54,21 @@ enum Priority {
...
+		/** No flags */
+		None = 0,
+		/** This is a global drawable which will appear in all scenes */
+		Global = 1,
+		/** This is a shared drawable which will appear in all scenes that use shared drawables */
+		Shared = 2,
+		/** This flag indicates the drawable should not be drawn */
+		Invisible = 4,
+		/** The default flag set */
+		Default = None
```
Ghabry: Can you document which of these Flags are "(not) affected by transitions"?

fmatthew5876: Not sure what you mean? These flags are all orthogonal to transitions.

Ghabry: I mean global drawables are never faded in or out. They stay "above" transitions (or screen show/erase stuff).

fmatthew5876: That's purely due to Z values not flags. Or at least it will be once I do the PR to kill the global list.

Globals are rendered above transitions.

Ghabry: ah okay, I was still thinking in the global vs local list scope

src/scene.cpp
```diff
@@ -95,6 +95,11 @@ void Scene::MainFunction() {
 	// it could have changed the scene.
 	if (!IsAsyncPending() && Scene::instance.get() == this) {
 		if (!init) {
+			auto prev_scene_type = prev_scene ? prev_scene->type : Null;
+			TransferDrawablesFrom(prev_scene.get());
+			// Destroy the previous scene here, before and initialization logic / transition in occurs.
```
Ghabry: and -> any I guess?

Ghabry:
I only quickly scrolled through (not a review) and I already like this change.

The flags system also makes it nicely extensible if we ever need more sophisticated Drawable transfer logic.

src/scene.cpp
fmatthew5876: Fixed

PR #2055, opened 2020-01-18
Refactor Pictures

fmatthew5876:
Depends on #2052 

This refactors the picture data. Now we have one `Game_Pictures` class which manages all the pictures and detaches them from `Game_Screen`. This fixes all the recent picture loading issues #2051 and #2053. It also cleans up initialization and provides some optimizations.

### DIFF
--- a/src/game_pictures.cpp
+++ b/src/game_pictures.cpp
@@ -32,27 +32,25 @@
 // Applied to ensure that all pictures are above "normal" objects on this layer
 constexpr int z_mask = (1 << 16);
 
-Game_Pictures::Game_Pictures() {
-	// RPG_RT Save game data always has a constant number of pictures.
-	// FIXME: Do this resize when we write save data instead of doing it here
-	// because calling Update() on all the empty pictures is slow.
-	GetPicture(GetDefaultNumberOfPictures());
+static bool IsEmpty(const RPG::SavePicture& data, int frames) {
+	RPG::SavePicture empty;
+	empty.ID = data.ID;
+	empty.frames = frames;
+
+	return data == empty;
+}
+
+static bool IsEmpty(const RPG::SavePicture& data) {
+	return IsEmpty(data, data.frames);
 }
 
 Game_Pictures::Picture::Picture(RPG::SavePicture save)
 	: data(std::move(save))
 {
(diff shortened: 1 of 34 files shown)


fmatthew5876:
I did a vtune profile of a game with no pictures. In master it spends 0.4% of time in `Game_Picture::Update()`. With this PR, the update routine doesn't even appear at all in the profile.

fmatthew5876:
Rebased and retested with Frozen Triggers. Also retested my battle pictures test game, along with saving and loading on that map.

This could use a bit more testing to be sure if you guys can help out.

src/game_pictures.cpp
```diff
@@ -255,56 +349,60 @@ void Game_Picture::Move(const MoveParams& params) {
...
 
 	FileRequestAsync* request = AsyncHandler::RequestFile("Picture", name);
 	request->SetGraphicFile(true);
 
-	int pic_id = data.ID;
+	int pic_id = pic.data.ID;
 
 	// FIXME: This lambda is here because it's possible the picture vector can be
 	// resized before the async call of onPictureSpriteReady().
 	// Pictures should be refactored so this ugly hack is not needed.
```
Ghabry: Considering that this is now refactored this lambda isn't needed anymore?

fmatthew5876: We need to capture the `pic_id` so yes. It could be done other ways, but there is nothing wrong with the lambda really.

Ghabry:
I tested smashbros, this demoscene sequence, Razas (the intro sequence where we already had bugs), "Hentai Film" intro sequence (also had pic bugs years ago) and a few random japanese games. Work for me.
Though for Spritesheet I don't really have any, Frozen Triggers is the best bet.

Save/Load also seems to be crash free.

src/game_pictures.cpp
Ghabry: Can you remove the FIXME comment at least?

(The other way would be ``request->Bind(&Game_Pictures::OnPictureSpriteReady, this, pic_id)`` as done everywhere else (bet this was added pre-C++11, looks so dated with lambda support))

fmatthew5876: Oops, that FIXME should have been removed. Took it out, thanks!

PR #2056, opened 2020-01-18
Refactor Game_Targets

fmatthew5876:
Towards #1954 

Also fixes some incompatibilities with targets save data and ordering

### DIFF
--- a/src/game_targets.cpp
+++ b/src/game_targets.cpp
@@ -21,10 +21,8 @@
 
 template <typename T>
 static auto FindTarget(T&& targets, int map_id) {
-	return std::lower_bound(targets.begin(),
-			targets.end(),
-			map_id,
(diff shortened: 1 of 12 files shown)


src/game_targets.cpp
```diff
@@ -16,91 +16,70 @@
...
-	if (!create)
-		return data.end();
-	data.resize(data.size() + 1);
-	data.back().ID = id;
-	return data.end() - 1;
+#include <algorithm>
+
+template <typename T>
+static auto FindTarget(T&& targets, int map_id) {
+	return std::lower_bound(targets.begin(),
```
Ghabry: Considering that games usually use very few targets is a binary search worth it?

fmatthew5876: Changed to linear search

PR #2060, opened 2020-02-05
Small fixes for 3DS and warnings

carstene1ns:
Needed after #1973.

### DIFF
--- a/src/decoder_oggvorbis.cpp
+++ b/src/decoder_oggvorbis.cpp
@@ -21,13 +21,6 @@
 
 // Headers
 #include <cassert>
-#ifdef HAVE_TREMOR
-#include <tremor/ivorbiscodec.h>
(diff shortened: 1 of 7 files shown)


Ghabry:
OSX does not have _DIRENT_HAVE_D_TYPE defined but d_type is properly populated. I slowly think that this is a hugher problem. Anybody with a BSD system around to check if this macro is defined there? The manpage says "this is a BSD extension" but OSX is BSD based >.>

Ghabry:
Proposed solution by LLVM (6 months ago): https://reviews.llvm.org/D64940

carstene1ns:
This solution seems ok for now.

fmatthew5876:
Confirmed on FreeBSD 12 that `_DIRENT_HAVE_D_TYPE` is not defined when you `#include <dirent.h>`

Issue #2063, opened 2020-02-06
3DS - RTP files not found (double slashes)


minusplus:
#### Name of the game:
Any that rely on RTP
#### Player platform:
3DS

#### Attach files (as a .zip archive or link them)
[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/4162870/easyrpg_log.txt)

#### Describe the issue in detail and how to reproduce it:
Install EasyRPG and RTP onto a 3DS. Launch a game that requires RTP resources.

PR #2064, opened 2020-02-06
Removed trailing slashes on 3DS RTP

minusplus:
Trailing slashes in FileFinder caused duplicate slashes on 3DS RTP path, causing RTP resources to fail to load. Fixes issue in #2063

### DIFF
--- a/src/platform.cpp
+++ b/src/platform.cpp
@@ -160,6 +160,17 @@ std::string Platform::Directory::GetEntryName() const {
 #endif
 }
 
+static inline Platform::FileType GetEntryType(...) {
+	return Platform::FileType::Unknown;
+}
+
+template <typename T, typename = decltype(std::declval<T>().d_type)>
+static inline Platform::FileType GetEntryType(T* entry) {
+	return entry->d_type == DT_REG ? Platform::FileType::File :
+		   entry->d_type == DT_DIR ? Platform::FileType::Directory :
+		   entry->d_type == DT_UNKNOWN ? Platform::FileType::Unknown : Platform::FileType::Other;
+}
+
 Platform::FileType Platform::Directory::GetEntryType() const {
 	assert(valid_entry);
 
@@ -167,14 +178,7 @@ Platform::FileType Platform::Directory::GetEntryType() const {
 	return SCE_S_ISREG(entry.d_stat.st_mode) ? FileType::File :
 			SCE_S_ISDIR(entry.d_stat.st_mode) ? FileType::Directory : FileType::Other;
 #else
-#	if defined(_DIRENT_HAVE_D_TYPE) || defined(_3DS)
-	return entry->d_type == DT_REG ? FileType::File :
-		   entry->d_type == DT_DIR ? FileType::Directory :
-		   entry->d_type == DT_UNKNOWN ? FileType::Unknown : FileType::Other;
-#	else
-	#warning Current Platform does not populate directory entry types
-	return FileType::Unknown;
-#	endif
+	return ::GetEntryType(entry);
 #endif
 }
 
--- a/src/weather.cpp
+++ b/src/weather.cpp
@@ -109,8 +109,6 @@ static constexpr int fog_opacity[2][4] = {
 	{ 64, 80, 160, 255 },
 };
 
-static constexpr int snowflake_visible = 150;
-
 const Bitmap* Weather::ApplyToneEffect(const Bitmap& bitmap, Rect rect) {
 	if (tone_effect == Tone()) {
 		return &bitmap;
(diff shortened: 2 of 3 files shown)


Ghabry:
Now I wonder if this trailing slash is wrong everywhere and just not noticed on other platforms because they fix the path?

minusplus:
I was wondering that too. It looks like redundant slashes present in the Linux build as well (along with another issue that I don't have time to look into at the moment)
[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/4165124/easyrpg_log.txt)

Ghabry:
I don't see an issue in your log-file. Just that the game uses a RTP but no RTP was found.

Okay we will check the rtp paths for other platforms.
Thx for the PR!

minusplus:
Sorry, I should have been more specific. In the above log file, on lines 36 and 42, you can see double slashes pointing to the game folder. Those particular slashes aren't related to the RTP, though, and the game folder doesn't have double slashes on the 3DS build.

The no RTP found in the above log was the other problem I was referring to. It doesn't seem to be pulling the locations from my environment variables in Manjaro Linux at all. That's a separate issue, of course, and for now I've just moved the folders to the default /data/rtp location, where it can find them.

Ghabry:
You added a trailing / to the project-path cli argument. Maybe we should strip this via the "MakeCanonical" function.

RPG2K\_RTP\_PATH and RPG2K3\_RTP\_PATH didn't work for you?

Because you use Manjaro: You can use AUR rpg2000-rtp and rpg2003-rtp.

minusplus:
Ah yes, that was the source of those.

Re: The environment variables, I am sure they are set unless I am misunderstanding something.
> ~$ printenv | grep RTP
RPG_RTP_PATH=/home/alec/ROMS/RPG Maker/rtp
RPG2K_RTP_PATH=/home/alec/ROMS/RPG Maker/rtp/2000
RPG2K3_RTP_PATH=/home/alec/ROMS/RPG Maker/rtp/2003

I could install them from the AUR, but they're working with the files in /data/rtp

carstene1ns:
The environment variables work fine here since years. You use `RPG_RTP_PATH` wrong by the way, but should not matter.

```shell
carsten@sol ~> printenv | grep RTP
RPG2K3_RTP_PATH=/usr/share/rtp/2003:/media/Daten/rtp/2k3-en
RPG2K_RTP_PATH=/usr/share/rtp/2000:/media/Daten/rtp/2k-en
```

minusplus:
I only added RPG_RTP_PATH because the other two weren't working but I went ahead and removed them all from my .bashrc for cleanliness's sake.

I don't know what to tell you. I do remember it working at some point, possibly before I switched from Mint to Manjaro, but it's not working now. I've tried moving it to different folders, both in and out of my home partition, and it simply doesn't find the files and the log makes no mention of adding a path unless it's in /data/rtp. As I've said, though, that's fine. I don't have any reason to need it somewhere else as long as it works.

carstene1ns:
I have added a debug message in #2060 now. Might be worth to see, if the path is wrong, etc.

We should remove the trailing slashes from all RTP directories, since we always add a slash when accessing files from them. 3DS is especially picky, as it does not allow opendir with trailing slash (likely Wii and Switch do the same, but not checked). On Linux and others ([POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_271)) these will be normalized to one slash.

fmatthew5876:
Agree let's fix the slashes for all platforms. Maybe it's benign in mainstream but it's better to make all the code here uniform.

fmatthew5876:
@minusplus 

Have you tried the latest continuous build of player? We recently pushed a bunch of changes, particularly ones with time keeping and sleep.

Would be great to get a confirmation everything is still working on 3ds if you can.

minusplus:
I did. Unfortunately, the current 3DS build fails to load, hanging on a black screen until I hard shutdown the device and logging no events. This is consistent between the build available in Jenkins and if I compile it locally

fmatthew5876:
If you're able to do some git bisecting in your local build and figure out which commit causes this, it would help come up with a fix. I don't have a 3ds so I can't test this unfortunately.

minusplus:
Okay. Starting at d62998ea33fa6a8c2010d7d0cdb93bb8ea79f30a, the game browser will fade in very slowly before locking up. It fails to compile from ce06a282f4d36113b2b6276f410740d5485d7cff until e323ac72ad172328621218161eb21f1db8e75b26, where it will compile but fails to load as I described above.

carstene1ns:
No need to merge the `clockfix` branch, there is a PR open for that.

PR #2065, opened 2020-02-08
Use SFINAE for d_type detection - fix OSX and BSD

fmatthew5876:
Fix #2060

### DIFF
--- a/src/platform.cpp
+++ b/src/platform.cpp
@@ -160,6 +160,17 @@ std::string Platform::Directory::GetEntryName() const {
 #endif
 }
 
+static inline Platform::FileType GetEntryType(...) {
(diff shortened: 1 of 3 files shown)


carstene1ns:
Cool

Ghabry:
Thats a nice solution!
doctest supports conditional tests. Maybe add a condition that checks whether "." is "Unknown" and in that case skip the tests?

I implemented skipping in my not-ready emscripten cmake branch:
https://github.com/Ghabry/easyrpg-player/blob/emscripten-cmake/tests/filefinder.cpp
(Basicly you multiply (ugh) a doctest::skip(bool) with a test case or suite).

Issue #87, opened 2013-02-23
Use "git describe" to display a nice rev-number


Ghabry:
After creating our first tag (0.1) "git describe" can be used to get a more readable revision number 0.1-### (where ### is the number of commits since 0.1).

The output of "git describe" should be written in some header file and then be included by player and e.g. displayed in the title bar.

This makes it easier for us to figure out which version somebody uses (Maybe also write it in the PNG-tEXT chunk under "Software" if he used our screenshot function and we cant see the title bar). And to detect if somebody made local changes (e.g. by adding an 'M' after the rev number)

Ghabry:
Considering we are full CMake and Autotools now it should be finally possible to add this without toooo much afford?

See also #2070

PR #2070, opened 2020-02-09
Add git version hash to sources and log file

fmatthew5876:
Fix: #87

This logs the git commit hash in the output log. Should help with bug reports.

For cmake this uses some cmakefiles from here: https://github.com/rpavlik/cmake-modules/blob/master/GetGitRevisionDescription.cmake

These will correctly rebuild the `git_version.cpp` each time the git hash changes.

Plaforms:
- [x] cmake
- [x] autotools
- [x] Android
- [x] 3ds
- [ ] switch
- [x] gcw0
- [x] wii
- [x] vita

Features
- [x] Detects git version correctly
- [x] Reacts to git HEAD changes and rebuilds
- [x] Builds from source tarball without `.git` - `make distcheck` in autotools

Todo:
- [ ] Remove PLAYER_ADDTL_VERSION hacks from jenkins build scripts (not part of the codebase)

### DIFF
--- a/builds/git_version.mk
+++ b/builds/git_version.mk
@@ -1,0 +1,16 @@
+GIT_ROOT = $(git rev-parse --absolute-git-dir)
+GIT_HEAD_FILE = HEAD
+GIT_REF_FILE = $(shell git symbolic-ref -q HEAD)
+ifeq ($(strip $(GIT_REF_FILE)),)
+	GIT_REV_FILE=$(GIT_ROOT)/$(GIT_HEAD_FILE)
+else
+	GIT_REV_FILE=$(GIT_ROOT)/$(GIT_REF_FILE)
+endif
+
+
+GIT_VERSION_IN = $(GIT_ROOT)/../src/git_version.cpp.in
+GIT_VERSION_OUT =$(GIT_ROOT)/../src/git_version.cpp
+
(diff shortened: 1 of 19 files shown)


builds/git_version.mk
```diff
@@ -0,0 +1,10 @@
+GIT_ROOT = $(REPO_ROOT)/.git
+GIT_HEAD = $(GIT_ROOT)/HEAD
+GIT_REV_FILE = $(shell $(REPO_ROOT)/builds/git_ref.sh $(GIT_HEAD))
+
```
carstene1ns: Can you explain why all of this magic is needed? [`git rev-parse`](https://git-scm.com/docs/git-rev-parse) will do exactly what is needed here.



fine tuned: `git rev-parse --verify -q --short=10 HEAD^{commit}`

fmatthew5876: We can and probably should use git to generate the string. However we need to figure out which file contains the hash and add it as a dependency of the rule. This ensures that `git_version.cpp` is regenerated whenever `HEAD` changes.

carstene1ns: `git rev-parse --absolute-git-dir`
`git symbolic-ref -q HEAD`
?

carstene1ns:
Closing in favor of #2774.

Issue #2073, opened 2020-03-03
Player crashes when exit


elsemieni:
#### Player platform: 
Tested on windows (at least)

#### Attach files (as a .zip archive or link them)

[coredump - Player_2020-03-03-16-39-23.zip](https://github.com/EasyRPG/Player/files/4283274/Player_2020-03-03-16-39-23.zip)

Tested with https://ci.easyrpg.org/view/Windows/job/player-win32/1083/ 

#### Describe the issue in detail and how to reproduce it:
Run player and close it. 
Seems like it's closing normally (does not hang before closing windows or anything). However, it leaves a coredump on folder, meaning it's crashing. 

Taking a look a bit at the coredump seems the problem is in https://github.com/EasyRPG/Player/blob/master/src/drawable_list.cpp#L64 , where evaluating this resulting as null (I can be wrong, maybe I got wrong sources, please check). 

A bit of screenshots of stack-trace for "link-saving" purposes. 
![image](https://user-images.githubusercontent.com/1494687/75813788-5bbbaa80-5d6f-11ea-9fea-abc469f9f1a7.png)
![image](https://user-images.githubusercontent.com/1494687/75813846-79890f80-5d6f-11ea-806b-70e30428c702.png)

Ghabry:
This ``assert(local != nullptr);`` is triggered in Debug build.

There are other debug assertion problems in the Windows build btw:

text.cpp ``const auto end = &*text.end();`` <- This dereferences "end", not liked by MSVC

player.cpp
```cpp
escape_symbol = ReaderUtil::Recode("\\", enc);
escape_char = Utils::DecodeUTF32(Player::escape_symbol).front(); <-- This line
if (escape_symbol.empty()) {
  // Bad encoding
  Output::Debug("Bad encoding: %s. Trying next.", enc.c_str());
  continue;
}
<-- Must be moved after the if (the front() triggers an assert on empty strings)
```

carstene1ns:
The `end()` dereferences are fixed in #2026.

elsemieni:
I tested Jenkins build of #2026 (https://ci.easyrpg.org/job/player-win32-pr/4023/) . Gives me an assertion error at exiting (assuming jenkins PR builds are debug ones).

`Assertion failed: local != nullptr, file C:\Jenkins\workspace\player-win32-pr\src\drawable_mgr.h, line 42`

I don't know if it's correct behaivour or not, but anyways I leave it here if it's useful or not...

[...]

Issue #2061, opened 2020-02-06
Test suite crashes on Windows


carstene1ns:
in parse test.

Ghabry:
I will update the jenkins job to get a "Build is unstable" reported when the unit tests fail. Lets see if this is possible with Batch...

Ghabry:
Appears to work :+1:

carstene1ns:
> I will update the jenkins job to get a "Build is unstable" reported when the unit tests fail. Lets see if this is possible with Batch...

If I remember correctly, I tested this too. :grinning:

Ghabry:
yes you added it for macOS. I copied the solution from there :D

Issue #2045, opened 2020-01-11
Windows Debug assertion failures


fmatthew5876:
In the messages code there are many instances of `&*text.end()` to get a pointer to the end of the string. This works fine normally, but trips a debug exception in windows debug builds.

Ghabry:
ah, the nice debug iterators. :+1: 
Considering that end()-deref is undefined behaviour I wonder why UBSan isn't catching this :thinking:

fmatthew5876:
Added fixes for these to #2026

Ghabry:
Another assert:
player.cpp
```cpp
escape_symbol = ReaderUtil::Recode("\\", enc);
escape_char = Utils::DecodeUTF32(Player::escape_symbol).front(); <-- This line
if (escape_symbol.empty()) {
  // Bad encoding
  Output::Debug("Bad encoding: %s. Trying next.", enc.c_str());
  continue;
}
<-- Must be moved after the if (the front() triggers an assert on empty strings)
```

PR #2074, opened 2020-03-04
Allow global drawables to be destroyed when there is no list + windows crash fixes

fmatthew5876:
Fix: #2073
Fix: #2061 
Fix: #2045 

@elsemieni can you confirm this fixes the crash you found? I can't reproduce it so while I suspect this is the cause I can't prove it.

### DIFF
--- a/builds/cmake/Modules/ConfigureWindows.cmake
+++ b/builds/cmake/Modules/ConfigureWindows.cmake
@@ -43,6 +43,8 @@ if(MSVC)
 
 	option(MSVC_MULTICORE "MSVC: Build using multiple cores (/MP)" ON)
 	if (MSVC_MULTICORE)
-		add_compile_options(/MP)
+		add_compile_options("/MP")
 	endif()
+
+	add_compile_options("/utf-8")
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -664,11 +664,6 @@ if(${PLAYER_AUDIO_BACKEND} MATCHES "^(SDL.*|libretro)$")
 	endif()
 endif()
 
-# MSVC: Set UTF-8 encoding for all files (important for rtp table)
-if(MSVC)
-	target_compile_options(${PROJECT_NAME} PRIVATE "/utf-8")
-endif()
-
 # Executable
 if (${PLAYER_BUILD_EXECUTABLE})
 	add_executable(${PROJECT_NAME}_exe WIN32 "src/main.cpp")
--- a/src/drawable_mgr.cpp
+++ b/src/drawable_mgr.cpp
@@ -38,6 +38,13 @@ void DrawableMgr::Register(Drawable* drawable) {
 }
 
 void DrawableMgr::Remove(Drawable* drawable) {
-	GetLocalList().Take(drawable);
+	auto* list = GetLocalListPtr();
+	// Global drawables can be singletons, which may get destroyed after all scenes due
+	// static initialization order. Non-global drawables we assume are all gone before
+	// all lists are destroyed.
+	assert(list || drawable->IsGlobal());
+	if (list) {
+		list->Take(drawable);
+	}
 }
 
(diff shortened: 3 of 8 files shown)


elsemieni:
@fmatthew5876 Used associated jenkins PR build (Win32). After splash logo it throws me an assertion error window (stdout/stderr doesn't show me any interesting about it).

```
Program: C:\msys32\home\elsemieni\nether\SCH\Player.exe
File: C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.24.28314\include\xstring
Line: 1792

Expression: cannot dereference string iterator because it is out of range (e.g. an end iterator)
```

fmatthew5876:
Ok I added another fix which hopefully covers that crash.

elsemieni:
Same assertion failure :( 
This time I attached a debugger and well... I wonder there's a bizarre stack trace over there 
image

fmatthew5876:
Alright I went ahead and fixed all the windows crashes. Tested it myself on windows and seems to work now

Ghabry:
This still fails with "error C2015: too many characters in constant".

I believe the problem is that "/utf-8" (interpret literals as utf-8) is not passed to the unit tests.

Untested (not at my laptop right now) solution:
``target_compile_options(${PROJECT_NAME} PRIVATE "/utf-8")``
but it should be ``PUBLIC`` otherwise this compile flag is not propagated to others that depend on ``${PROJECT_NAME}``.

Ghabry:
Jenkins: Test this please <3

Ghabry:
This patch fixes the Windows build (likely needs Workspace wipe):

``patch -Np1 < patch.diff``

```patch.diff
index 77051b94..1c99be62 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -664,11 +664,6 @@ if(${PLAYER_AUDIO_BACKEND} MATCHES "^(SDL.*|libretro)$")
 	endif()
 endif()
 
-# MSVC: Set UTF-8 encoding for all files (important for rtp table)
-if(MSVC)
-	target_compile_options(${PROJECT_NAME} PRIVATE "/utf-8")
-endif()
-
 # Executable
 if (${PLAYER_BUILD_EXECUTABLE})
 	add_executable(${PROJECT_NAME}_exe WIN32 "src/main.cpp")
diff --git a/builds/cmake/Modules/ConfigureWindows.cmake b/builds/cmake/Modules/ConfigureWindows.cmake
index b4f21bb2..2cebb8cb 100644
--- a/builds/cmake/Modules/ConfigureWindows.cmake
+++ b/builds/cmake/Modules/ConfigureWindows.cmake
@@ -43,6 +43,8 @@ if(MSVC)
 
 	option(MSVC_MULTICORE "MSVC: Build using multiple cores (/MP)" ON)
 	if (MSVC_MULTICORE)
-		add_compile_options(/MP)
+		add_compile_options("/MP")
 	endif()
+
+	add_compile_options("/utf-8")
 endif()
```

fmatthew5876:
Added your patch

Ghabry:
Now it builds but there are still end-derefs.
I attached a debugger: You missed two occurences of ``text.end()`` in utf.cpp

fmatthew5876:
jenkins: test this please

Ghabry:
Jenkins: Test this please

Ghabry:
Jenkins: Test this please

builds/cmake/Modules/ConfigureWindows.cmake
```diff
@@ -43,6 +43,8 @@ if(MSVC)
 
 	option(MSVC_MULTICORE "MSVC: Build using multiple cores (/MP)" ON)
 	if (MSVC_MULTICORE)
-		add_compile_options(/MP)
+		add_compile_options("/MP")
 	endif()
+
+	add_compile_options("/utf-8")
```
carstene1ns: What I dislike here, is that the comment is lost.

Ghabry: Valid concern, maybe make the comment more generic now "Interpret character literals as UTF-8"

fmatthew5876: Added comment in #2017

Issue #587, opened 2015-09-15
Support RPG Maker 2003 v1.11


Ghabry:
This is a meta-bug to track all new features in RPG2k3 v1.11 and to discuss which are not worth implementing.

Changelog: https://docs.google.com/document/d/1sXlXZIXdjxVq5gh_lPFuFLMbMRzZe4BKPpRI13ZD2Qc/edit?pli=1
- [x] Transparent Window also affects Title Screen
- [X] When Battle Mode is one (traditional) the transparency setting of on-map message boxes is still respected (before they appeared opaque) (**->0.6.2**)
- [x] Support for skipping title screen. "In TestPlay Mode the state of Show/HideTitle button in the toolbar affects the behaviour: ShowTitle: Starts New Game directly, HideTitle: Go to Load Scene directly"
- [x] The “Y” key has now the function of an additional confirm button (such as Space, Enter and “Z”)
- [x] Picture operations (Show, Move, Erase) are not blocked by messages anymore. Do we already implement this?

Improved screen display: F5 can be used to open the video options. And a lot of different zoom options. Very low priority for me.

Event Commands (needs liblcf update):
- [x] Open Load Menu
- [x] Exit Game
- [x] Toggle Fullscreen
- [x] "Toggle ATB Wait Mode" <- Is Implemented but we our battle system is always ATB Wait
- [x] "Open Video Options" <- Just shows "OpenVideoOptions: Command not supported"

Conditional branch type “Other”
- [x] Savestate is available
- [x] Playtest mode is active <- questionable because of Anti-Cheat features
- [x] ATB wait mode <- Check is implemented
- [x] Fullscreen mode is active <- Nonsense on systems that have no window manager

An unsolved problem that is out of scope is, that v1.11 is a western europe only release, because other glyphs don't work which excludes the Asian countries where RPG2k/3 is still popular. I don't know if combining the editor with EasyRPG Player resolves font issues?

Ghabry:
show_title is 0x6F in the System chunk of the LDB.
0 = Don't show, 1 = Show (Default)

```xml
<!-- Save state available? -->
<EventCommand>
	<code>12010</code>
	<indent>0</indent>
	<string></string>
	<parameters>11 0 0 0 0 0</parameters>
</EventCommand>
<!-- Play Test active? -->
<EventCommand>
	<code>12010</code>
	<indent>0</indent>
	<string></string>
	<parameters>11 1 0 0 0 0</parameters>
</EventCommand>
[...]

PR #2078, opened 2020-03-07
RPG_RT v1.11 fixes

fmatthew5876:
Fix: #587

### DIFF
--- a/src/window_message.cpp
+++ b/src/window_message.cpp
@@ -52,8 +52,15 @@ Window_Message::Window_Message(int ix, int iy, int iwidth, int iheight) :
 {
 	SetContents(Bitmap::Create(width - 16, height - 16));
 
-	if (Data::battlecommands.battle_type != RPG::BattleCommands::BattleType_traditional &&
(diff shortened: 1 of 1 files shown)


Ghabry:
Imo there is no reason to disable saving in a debug menu. When saving at the wrong location this is the fault of the user. Maybe allow saving but show a message in a WindowHelp when Save is hovered? ("Saving here can corrupt the save!" or some other wording)

Opinions?

fmatthew5876:
This is the only case I know of where it's not good to save. As far as I can tell the side effect is only visual. When you load the game the screen will no longer be erased but everything else should play out the same way. Even transition in effects will consume the same number of frames.

So if we think it's better to just allow people to save in all cases, ( I can see the argument from a debugging point of view) I can just remove the commit.

carstene1ns:
I have no strong opinion here, but I guess debug save should always be possible.

fmatthew5876:
Ok I removed the debug save thing and rebased. Now this is the just the RPG_RT 1.11 message box fix.

PR #2082, opened 2020-03-16
Switch: Enable left thumbstick, fix using B button

carstene1ns:
Issue notified over IRC.

TODO (after release): Integrate properly into input handling instead of hardcoding into
keyboard values.


carstene1ns:
(...tested the PR build, works!)

Issue #1275, opened 2017-10-07
Music Plays way too fast on 3DS


Ijustusethisforaisue:
Thank you for supporting EasyRPG Player.
Before creating an issue, please verify that your bug still occurs in the "continuous builds" of the Player.
You can get them at the bottom of our download section: https://easyrpg.org/player/downloads

---
Please fill in the following fields before submitting an issue:
#### Player platform: 3DS

Windows, Linux, Android, ...

#### Attach files (as a .zip archive or link them)
 - A savegame next to the problem (if it is not at the beginning of the game)
 - The `easyrpg_log.txt` log file (you can find it in the game directory)
[OFF Translation 2.0.zip](https://github.com/EasyRPG/Player/files/1366334/OFF.Translation.2.0.zip)

#### Describe the issue in detail and how to reproduce it:

The Music is playing way too fast in the Game. Its mostly noticable in the fight against the spectres after getting the first add-on. The Battle Theme ,,Peper Steak" is playing way too fast

Ghabry:
You used a CIA from our download page at https://easyrpg.org/player/downloads ?

Ijustusethisforaisue:
No, i used a 3dsx file from the continous builds

Ghabry:
Okay thats fine, too. Thanks for the report.

carstene1ns:
Just for reference: OFF uses MP3 music.
I see there were problems regarding Music/Sound files in the log.
Did you somehow convert the files?

Ijustusethisforaisue:
No, i unzipped the game and kopied the unzipped file on the SD card. No mp3 converting happened from my side

Ghabry:
Okay we well test this on one of our 3DS.

GeigerCounter:
Can confirm. I have this problem with OFF, using the CIA from the Downloads page. Testing with OFF. Will test with other games.

GeigerCounter:
Can't seem to reproduce in Yume Nikke or Cognitive Dissonance. Going to try the original French version of OFF to see if for some strange reason it only presents in the English translation.

GeigerCounter:
Issue does not present in the original French version of the game. However part of the install process decompresses the music files to .wav files, so that complicates things a bit. I suspect the MP3 decoding is the culprit. A potential temporary workaround for OFF specifically is to copy the decompressed waveform files from the French game into the translated game's music folder.

carstene1ns:
[...]

PR #2086, opened 2020-03-30
Fix lag when loading large SE files

Ghabry:
**a)** When playing a SE for the first time we
1. load the entire file
2. resample the entire file
3. cache it
The intention was to help slow hardware when many SE are played at once.

**b)** With the new code it is:
1. load the entire file
2. cache it

Resample/Pitch is now done in the Audio thread. So this is a trade-off where we pay the CPU time, at least the IO part is still only once.

For Witch's Heart (Limes game) the load time of SE is reduced from 90ms (with Medium resample quality 140ms) to 6ms.

Refactor: Instead of a buffer+length the SeCache returns now a real AudioDecoder. This gives us more flexibility later, e.g. short samples (TODO: What is short) could be handled as **a)** and longer samples could be handled as **b)**. This patch handles everything as **b)**.

Unfortunately the SDL2 mixer API is fire-and-forget for SE samples so we can't use the optimisation there :(. Everything else using the generic API (SDL2 [Android], Vita, libretro) works. We should really deprecate this API, get FluidSynth working and fix our embarassing Midi support on Windows.

Todo:
- [x] Fix 3DS build
- [x] 3DS Hardware test
- [x] Fix audio glitches when playing some SEs
- [x] NO REPRO. Fix latency when first sample is played

### DIFF
--- a/src/audio_3ds.cpp
+++ b/src/audio_3ds.cpp
@@ -194,25 +194,40 @@ void CtrAudio::SE_Play(std::string const& file, int volume, int pitch) {
 	}
 
 	std::unique_ptr<AudioDecoder> dec = cache->CreateSeDecoder();
+	dec->SetPitch(pitch);
 
 	int frequency;
 	AudioDecoder::Format format, out_format;
 	int channels;
 
+	std::vector<uint8_t> dec_buf;
+	std::vector<uint8_t>* out_buf = nullptr;
+	AudioSeRef se_ref;
+	bool use_raw_buf = false;
+
 	dec->GetFormat(frequency, format, channels);
-	if (!set_channel_format(ndsp_channel, format, channels, out_format)) {
-		dec->SetFormat(frequency, out_format, channels);
+	if (set_channel_format(ndsp_channel, format, channels, out_format)) {
+		// When the DSP supports the format and the audio is not pitched the raw
+		// buffer can be used directly
+		use_raw_buf = pitch == 100;
 	}
+
+	if (use_raw_buf) {
+		se_ref = cache->GetSeData();
+		out_buf = &se_ref->buffer;
+	} else {
+		dec->SetFormat(frequency, out_format, channels);
+		dec_buf = dec->DecodeAll();
+		out_buf = &dec_buf;
+	}
+
 	ndspChnSetRate(ndsp_channel, frequency);
-	dec->SetPitch(pitch);
 
 	if (se_buf[se_channel].data_pcm16 != nullptr) {
 		linearFree(se_buf[se_channel].data_pcm16);
 	}
 
-	std::vector<uint8_t> dec_buf = dec->DecodeAll();
-
-	size_t bsize = dec_buf.size();
+	size_t bsize = out_buf->size();
 	size_t aligned_bsize = 8192;
 	// Buffer must be correctly aligned to prevent audio glitches
 	for (; ; aligned_bsize *= 2) {
@@ -226,7 +241,7 @@ void CtrAudio::SE_Play(std::string const& file, int volume, int pitch) {
 	const int samplesize = AudioDecoder::GetSamplesizeForFormat(out_format);
 	se_buf[se_channel].nsamples = bsize / (samplesize * channels);
 
-	memcpy(se_buf[se_channel].data_pcm16, dec_buf.data(), dec_buf.size());
+	memcpy(se_buf[se_channel].data_pcm16, out_buf->data(), out_buf->size());
 
 	DSP_FlushDataCache(se_buf[se_channel].data_pcm16, aligned_bsize);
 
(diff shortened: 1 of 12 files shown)


Ghabry:
Can't reproduce "Fix latency when first sample is played" anymore. Maybe was some audio bug on my system.

Ghabry:
The performance enhancements won't be fully available for the 3DS because this would need a refactor (move all the SE decode handling into the audio thread and fetch it when the playback buffer is empty).

Right now it will get:
 - speexdsp low (as all platforms)
 - Because the AudioDecoder is exposed now NDSP can be configured to match the SE audio source, less resampling needed

Ghabry:
> 3DS: Add empty romfs, otherwise building fails with "Cannot open dir"

I encountered this one while building on my local system. Maybe broken since a devkitARM update?

Ghabry:
TODO: 3DS needs a hardware test, only tested in Citra. I can do one tomorrow.

carstene1ns:
We did not notice this before, because we always have the GUS patches in romfs.

carstene1ns:
Trying to fix the t3x race condition: https://github.com/EasyRPG/Player/commit/4f7ce648aab67fae17aebf2f6e23c0a6ade5afbd

Ghabry:
Doesn't help. Tested a bit further:
The problem is btw not tex3ds but ``%.t3x.o	%_t3x.h :	%.t3x`` that invokes bin2o.

Ghabry:
Made some tests on the 3DS. Works :)

carstene1ns:
Maybe this also fixes the mono issue (need to test OFF).

Ghabry:
You mean in BGM? I just checked the code and it indeed has the same bug there. This also explains the 2x speed: Wrong sample rate for Mono... Let me patch this...

Ghabry:
OFF title screen has now correct playback speed on 3DS.

Unplanned fix for #1275 :D

Ghabry:
No idea if @igotanissueforya or @VirgileVILE are still around but if you are: Please test the following build for us :)

https://ci.easyrpg.org/view/Pull%20Requests/job/player-3ds-pr/52/

fdelapena:
jenkins: test this please

VirgileVILE:
I'll have tested it with a .cia, but actually I can't.

Ghabry:
Our PR builder only generates 3dsx files. You can run them through the Homebrew Channel CIA.

VirgileVILE:
In this case I would wait for the next version to check if it's good.

carstene1ns:
@GeigerCounter @VirgileVILE @Ijustusethisforaisue Your issue should be fixed now. Please check with Player 0.6.2 (released soon™) or the 3DS nighly after 6​78.

Narwher:
(@igotanissueforya here, that was my pre dev account)

Thanks for your work on the 3DS port! Mono sound files now play at normal speed on hardware :)

However there is still some bgm crackling (sample size?) and sfx stuttering (probably just slow SD card speeds). In addition the odd repetition of sfx still occurs often, the punching sound in OFF battle sequence as an example. There's great improvement overall however, i'm very happy :)

I'd also like to request an option to turn off the bottom screen of the 3DS in order to save battery life. Would I put that in the issue section?

fmatthew5876:
Hi @Narwher thanks for testing.

For new feature requests, please do make a new github issue so we can track it.

carstene1ns:
That is nice to hear. :+1: 
How would you handle toggling the bottom screen?

Narwher:
I've forked Player repo and uploaded a rough attempt of touchscreen toggle. It will need a better keybinding but i've mapped it to the Y button (Y is already mapped to Left Shift, but what does that key do? It doesn't do anything in most games i've tried).

The feature works but there is a weird color banding glitch over the keypad overlay during gameplay. I'm still new to Github, and not sure where this should go (Should I PR/draft?)

carstene1ns:
It looks like it just does not draw. I assume it does not save much power this way.

Ghabry:
Some RPG Maker games map shift to running or jumping. Is more common than other keys used imo.

The display technology is LCD. It is backlit: When the display is on it always consumes the same amount of power. Rendering black only helps with AMOLED displays.

Just rendering black doesn't help a lot. But as a nice-to-have feature could be used (better for the lifetime because you don't have permanent static pixels on the bottom screen). Maybe we should wait until the config stuff proposed by fmatthew is in? Then we could add this as a platform specific setting somewhere.

carstene1ns:
Yeah, this is what i meant. Wanted to use `GSPLCD_PowerOffBacklight(screen)` to do it, but need to know if we can react on touch then (but should be doable). Then this logic is a lot easier ;-)

Narwher:
Yes, I didn't realize the screen was still on and rendering black until I tested with the hardware afterwards in the dark. It's my first time looking at 3DS libraries so I'm not too familiar with the functions :0

PR #2087, opened 2020-03-31
Android: Remove armeabi, add x86_64, SDL 2.0.12

### LINKED ISSUE #1911: Android: Play Store says we don't have 64bit support
For normal updates the Play Store requires now that we have 64 bit support.

We have 64bit ARM library since 2 years but it seems that some build.gradle update is required, otherwise the automatic check won't detect it -_-

Not sure what the necessary change is but we must resolve this before 0.6.2.

Ghabry:
Compiles for me (tm). Not tested if Play Store accepts this.

Fix #1911

### DIFF
--- a/builds/android/app/src/main/java/org/libsdl/app/SDLActivity.java
+++ b/builds/android/app/src/main/java/org/libsdl/app/SDLActivity.java
@@ -41,7 +41,8 @@ import android.content.pm.ApplicationInfo;
 public class SDLActivity extends Activity implements View.OnSystemUiVisibilityChangeListener {
     private static final String TAG = "SDL";
 
-    public static boolean mIsResumedCalled, mIsSurfaceReady, mHasFocus;
+    public static boolean mIsResumedCalled, mHasFocus;
+    public static final boolean mHasMultiWindow = (Build.VERSION.SDK_INT >= 24);
 
     // Cursor types
     private static final int SDL_SYSTEM_CURSOR_NONE = -1;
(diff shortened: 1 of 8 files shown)


builds/android/app/src/main/java/org/libsdl/app/SDLActivity.java
```diff
@@ -200,15 +194,18 @@ protected void onCreate(Bundle savedInstanceState) {
...
+
         // EasyRPG modification: overwrite layout
         setContentView(R.layout.player_activity);
 
         // Load shared libraries
         String errorMsgBrokenLib = "";
         try {
-            // EasyRPG modification: acquire (which loads hidapi) must be
-            // before SDL2 libs on Android <4.4
-            mHIDDeviceManager = HIDDeviceManager.acquire(this);
```
Ghabry: This is not needed anymore because "hidapi" is now loaded on startup

Ghabry:
Only SDLActivity.java has patches.
Please check the diff if I forgot to migrate some "EasyRPG:" mods.

Ghabry:
Android 9 ARM64: Works. No other devices with older Android for testing in this flat :(
And too slow internet for the emulator download.

Ghabry:
The Play Store would accept the APK.

2 Warnings: "You removed a platform" and "This APK could be smaller"

Though the list of "Not supported devices" reports "0". :D No 4.0.4+ device is armeabi.

Ghabry:
Confirmed working for armeabi-v7a 4.0.4 and Android R  x86_64

carstene1ns:
* Tested on aarch64 / android 9 (api 28)
* Test pending for armeabi-v7a / android 7 (api 24)

We loose some joystick support before API 16 (android 4.1) it seems (but who cares).

I think it is time to announce we will drop support for very old devices (before 2012/4.0.4/api 15) Since a toolchain update may break this sooner or later.

Ghabry:
Everything below API 14 is already unsupported because SDL2 dropped support for it in 2.0.8 btw ;)

carstene1ns:
Yeah I know, we already target that:
https://github.com/EasyRPG/buildscripts/blob/a4ffc2cddac3df2fe3fd3dd3a4c09297dfaa23cc/android/2_build_toolchain.sh#L61

PR #2091, opened 2020-04-01
Use custom implementation instead of iscntrl

### LINKED ISSUE #2090: Regression: Text problems with Full-Width-Glyphs
Start any Japanese game ~~that has FaceSets (I used the intro of Ib).~~

1. The title screen command box has the wrong with
2. When starting the intro the text is incorrect aligned and some glyphs are "eaten"

Game: Note that I removed many assets to make it smaller.
[IbJp.zip](https://github.com/EasyRPG/Player/files/4414414/IbJp.zip)

Left broken, right correct
a-fs8

Ghabry:
Fix #2090

Unfortunately @fmatthew5876 was unlucky with the glyph he picked for the unit test, otherwise this would have been catched by it :(

iscntrl takes a int but the docs say "Like all other functions from cctype, the behavior of std::iscntrl is undefined if the argument's value is neither representable as unsigned char nor equal to EOF" Nice API

### DIFF
--- a/src/utils.h
+++ b/src/utils.h
@@ -322,7 +322,14 @@ namespace Utils {
 	 * @return value clamped between min and max
 	 */
 	template <typename T>
-		constexpr T Clamp(T value, const T& minv, const T& maxv);
+	constexpr T Clamp(T value, const T& minv, const T& maxv);
+
+	/**
+	 * Determines if a char is a control character (0x00-0x1F and 0x7F)
+	 * @return true when ch is a control character
+	 */
+	template <typename T>
+	bool IsControlCharacter(T ch);
 
 } // namespace Utils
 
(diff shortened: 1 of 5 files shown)


src/utils.h
```diff
@@ -322,7 +322,14 @@ namespace Utils {
...
 	template <typename T>
-		constexpr T Clamp(T value, const T& minv, const T& maxv);
+	constexpr T Clamp(T value, const T& minv, const T& maxv);
+
+	/**
+	 * Determines if a char is a control character (0x00-0x1F and 0x7F)
+	 * @return true when ch is a control character
+	 */
+	template <typename T>
+	bool IsControlCharacter(T ch);
```
fmatthew5876, line 332: Let's call it `IsAsciiControlCharacter`

fmatthew5876:
Great catch!

PR #2095, opened 2020-04-02
Fix Inputs being lost when the game lags.

### LINKED ISSUE #2094: The "Show FPS" Key (F2) doesn't work properly when there is lag
Use a CPU intensive game like Frozen Triggers, then slowdown your CPU, e.g. with "sudo cpupower frequency-set -u 1G" (Reset with "cpupower frequency-set -u 5G". (5Ghz or the max val your CPU supports))

When it lags press F2, many times the FPS counter will not toggle.

Ghabry:
Input events were only processed once per frame. When the user only briefly presses a key this would queue a key-up and a key-down event. Both are processed at the same time, nullifying them.
Now they are polled in every Update.

Fix #2094

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -238,8 +238,13 @@ void Player::MainLoop() {
 
 	int num_updates = 0;
 	while (Game_Clock::NextGameTimeStep()) {
+		if (num_updates > 0) {
+			Player::UpdateInput();
+		}
+
 		Scene::old_instances.clear();
 		Scene::instance->MainFunction();
+
 		++num_updates;
 	}
 


fmatthew5876:
What if the game is running at a high frame rate and no logical frames occur during the physical frame? Say for example you're running higher than 60hz. In that case player input won't be processed at all that frame.

I was thinking a more correct fix would be to leave `UpdateInput()` where it is. Then also add it as the first statement in the `NextTimeStep()` loop, but only for the second iteration and onward.

Then it always happens once each physical frame, and also for each logical frame if multiple logical frames occur.

carstene1ns:
Seems reasonable.

Ghabry:
Oh I didn't think about the high framerate case. Valid point.

Ghabry:
@fmatthew5876 
I adopted your idea but this is still broken for > 60 FPS scenarios.

The problem is that Input::Update is responsible for determining what is (not) pressed. But we can't move this function out of Player::Update, otherwise stuff like "IsRepeated" (for cursor scrolling) breaks.

I don't see a simple solution for this :/

fmatthew5876:
But this should still work for system inputs like FPS, Screenshot, full screen, etc..

They all happen in `Player::UpdateInput()`

Or are you saying key repeats or "IsPressed()" within the game logic itself can break just by being over 60fps?

Ghabry:
It does not work for ``Player::UpdateInput()`` because the ``Input::IsPressed`` etc. update there key state when ``Input::Update`` is called.

fmatthew5876:
Can we call `Input::Update()` after the game logic loop if `num_updates == 0`? Or does that not work?

fmatthew5876:
I think it won't work.

Say I'm running at > 60fps. I press a key during a physical frame with no logical frames. The "press" event will disappear, the game logic will never see I pressed a key.

fmatthew5876:
As discussed on irc, this solution is the best for now, even though the high fps situation requires additional work.

PR #2100, opened 2020-04-04
DisplayUI cleanups - Config 1 of N

fmatthew5876:
These are a set of commits which remove dead and old code from our UI classes and cleanup various things.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -390,7 +390,7 @@ if(${PLAYER_TARGET_PLATFORM} STREQUAL "SDL2")
 		src/sdl2_ui.h)
 	target_compile_definitions(${PROJECT_NAME} PUBLIC USE_SDL=2)
 
-	find_package(SDL2 REQUIRED)
+	find_package(SDL2 2.0.5 REQUIRED)
 	target_link_libraries(${PROJECT_NAME} SDL2::SDL2main)
 elseif(${PLAYER_TARGET_PLATFORM} STREQUAL "libretro")
 	target_sources(${PROJECT_NAME} PRIVATE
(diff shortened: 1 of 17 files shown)


src/libretro_ui.cpp
```diff
@@ -67,6 +67,10 @@ enum PadInputState {
 };
 
 LibretroUi::LibretroUi(int width, int height) {
+	// Handled by libretro
+	// FIXME: Can we react to when libretro switches from full to windowed and report the status?
```
Ghabry, line 71: No. There are no APIs available to query the window layout of the frontend.

fmatthew5876: That's bad for us. If any game depends on fullscreen in there event code it breaks. Sounds like it worth a bug report / PR to libretro. I can't imagine this would be too hard to implement.

fmatthew5876:
Rebased and self-reviewed. This is ready now

src/libretro_ui.cpp
Ghabry: Too complex solution, maybe create an issue about this for a later refactor:
Because there are many platforms with no proper fullscreen support (no window manager available) maybe this should be faked in general? I mean expose ``ToggleFullscreen`` in BaseUi which updates the fullscreen flag and ``ToggleFullscreen`` calls into Ui-Implementation ``RealToggleFullscreen`` (bad name I'm uncreative :D). This ``RealToggleFullscreen`` could return if it worked and when ``false`` BaseUi fakes it by updating the vars.
Or provide some ``HasFeature()`` querying API. Could be used besides Fullscreen for mouse and keyboard detection (and tell the user "This game will not run properly on this platform"). :)

fmatthew5876: I attempt to handle these concerns in the #2108 . There I have a `ConfigParam` object which is a generic parameter. Not only can it take a value, you can also specify a valid value range, lock it to a single value, or disable it entirely.

The idea is that for things like fullscreen, vsync, etc.. Platforms can lock these to true or false and they will show up greyed out in the settings UI. Or if disabled they won't show up at all.

It's trying to come up with a way to expose platform capabilities without having a bunch of ifdefs in the settings UI, config file, and command line parsing.

I say we punt this discussion until we get to that one.

fmatthew5876:
Made sdl 2.0.5 the minimum and move checks to build system

Issue #2097, opened 2020-04-02
Input: Split "system" keys and  "game logic" keys


Ghabry:
See #2095

The system keys like F2 (Show FPS) in ``Player::UpdateInput()`` don't work properly when the FPS are > 60 because the key state depends on ``Input::Update`` which is bound to the game logic and breaks when we call it more often.

There should be a new Api ``Input::UpdateSystem`` (?) introduced that only updates the pressed state of system keys and leaves the game keys alone.

PR #2105, opened 2020-04-04
Separate system keys from game keys

fmatthew5876:
Fix: #2097

### DIFF
--- a/src/input.h
+++ b/src/input.h
@@ -53,6 +53,11 @@ namespace Input {
 	void Update();
 
 	/**
+	 * Updates Input state for only system keys not used by game logic
+	 */
+	void UpdateSystem();
+
+	/**
 	 * Resets all button states.
 	 */
 	void ResetKeys();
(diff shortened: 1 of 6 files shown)


fmatthew5876:
One thing I haven't done yet. In the log mechanism, we probably should not save system buttons. Otherwise capture and replay will break.

fmatthew5876:
Tested this, F2 works much better now @500fps.

Not entirely sure about whether the FULLSCREEN key should be a game key or a system key. Thoughts?

Ghabry:
Imo it is bad that replays can toggle the fullscreen but for proper playback you need this flag to be set correctly. So it must be this way.

Maybe the fullscreen status reported can be faked but this is more work to get correctly working - You need a real fullscreen flag and a fake fullscreen flag.

src/input.h
```diff
@@ -125,6 +130,41 @@ namespace Input {
...
+	/**
+	 * Gets if a system button is being pressed.
+	 *
+	 * @param button button ID.
+	 * @return whether the button is being pressed.
+	 */
+	bool IsSystemPressed(InputButton button);
+
+	/**
+	 * Gets if a button is starting to being pressed.
```
carstene1ns, line 142: "system"

fmatthew5876: Fixed, thanks!

Issue #2102, opened 2020-04-04
すでに私たちは地獄のまっただ中でした: Battle: dummy enemies can attack in battle


Ghabry:
Via mail bug report.
Needs investigation. Sounds like a battle system (2000) bug.

#### Name of the game: すでに私たちは地獄のまっただ中でした

http://vipkohaku2011.web.fc2.com/entry/065.html

#### Attach files (as a .zip archive or link them)
[Save08.zip](https://github.com/EasyRPG/Player/files/4431292/Save08.zip)

The last played battles are 39 and 41.

#### Describe the issue in detail and how to reproduce it:
"Please fix the problem that dummy enemies (such as friends before they are called) can attack in battle."

fmatthew5876:
Battles 39 and 41 have a yellow cat enemy 10 with actions:
* Normal Attack - 50
* Skill 57 - 100

The other 2 enemies that can show up are 8 and 9 cats, which all have rating 50 actions

fmatthew5876:
I don't understand this one. Everything seem to be working. The yellow cat always uses his 100 rating skill. Is it some other battle?

Ghabry:
More details: The problem is Battle 48. The Bird uses on first turn a skill that "Calls his friends".

Unfortunately I don't have RPG Maker set up on this PC yet and RPG_RT.exe also fails to run so I can't really tell you what is going on.

Educated guess: A "Target all enemies" skill hits invisible/escaped enemies.

fmatthew5876:
> More details: The problem is Battle 48. The Bird uses on first turn a skill that "Calls his friends".
> 
> Unfortunately I don't have RPG Maker set up on this PC yet and RPG_RT.exe also fails to run so I can't really tell you what is going on.

The bird is enemy 48, he uses skill 66 which enabled switch 14, which enables a battle event that summons the other birds.

> Educated guess: A "Target all enemies" skill hits invisible/escaped enemies.

I suspect this is it. The hidden birds don't take any actions on the first round.

PR #2107, opened 2020-04-04
Various battle fixes

fmatthew5876:
see commits

Fix: #2102
Fix: #2123

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -148,8 +148,8 @@ Game_BattleAlgorithm::AlgorithmBase::AlgorithmBase(Type ty, Game_Battler* source
 		source->SetCharged(false);
 	}
 
-	target->GetBattlers(targets);
-	current_target = targets.begin();
+	current_target = targets.end();
+	party_target = target;
 }
 
 void Game_BattleAlgorithm::AlgorithmBase::Reset() {
@@ -762,6 +762,11 @@ int Game_BattleAlgorithm::AlgorithmBase::GetSourceAnimationState() const {
 }
 
 void Game_BattleAlgorithm::AlgorithmBase::TargetFirst() {
+	if (party_target) {
+		party_target->GetBattlers(targets);
+		party_target = nullptr;
+	}
+
 	current_target = targets.begin();
 	cur_repeat = 0;
 
@@ -1168,6 +1173,7 @@ bool Game_BattleAlgorithm::Skill::IsTargetValid() const {
 		skill.scope == RPG::Skill::Scope_party) {
 		if (GetTarget()->IsDead()) {
 			// Cures death
+			// NOTE: RPG_RT 2k3 also allows this targetting if reverse_state_effect.
 			return !skill.state_effects.empty() && skill.state_effects[0];
 		}
 
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -114,6 +114,9 @@ Game_BattleAlgorithm::AlgorithmBase::AlgorithmBase(Type ty, Game_Battler* source
 	source_restriction(RPG::State::Restriction(source->GetSignificantRestriction()))
 {
 	Reset();
+	if (source) {
+		source->SetIsDefending(false);
+	}
 
 	current_target = targets.end();
 }
@@ -123,6 +126,9 @@ Game_BattleAlgorithm::AlgorithmBase::AlgorithmBase(Type ty, Game_Battler* source
 	source_restriction(RPG::State::Restriction(source->GetSignificantRestriction()))
 {
 	Reset();
+	if (source) {
+		source->SetIsDefending(false);
+	}
 
 	SetTarget(target);
 }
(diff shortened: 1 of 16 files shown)


fmatthew5876:
Tested #2102 with this patch. In master an all enemy skill hits all the hidden birds. In this PR it only hits the visible one.

src/game_enemy.cpp
```diff
@@ -169,6 +169,7 @@ int Game_Enemy::GetHue() const {
 
 void Game_Enemy::SetHidden(bool _hidden) {
 	hidden = _hidden;
+	ResetBattleStateOnDeath();
```
Ghabry, line 172: I don't think SetHidden should change anything. They are just hidden and keep everything.
Because this is unconditionally enemies have 0 HP when the battle starts.

Also you forgot the edge case (bug?) that enemies removed via "Force Flee" in 2k3 can reappear with "Enemy encounter", another reason against "SetHp(0)" ;)

fmatthew5876:
Test Cases
========

1.
--------
Start enemy hidden
Turn 0 event inflict poison
Turn 2 event unhide

Result:
Turn 1: enemy appears
Turn 2: enemy takes poison damage

2.
---------------

Set enemy action Always Defend 50%, *3+0 Self-Destruct 100%
Turn *3+1 event: Show enemy1

Turn 1: Cast fire on enemy - Normal damage
Turn 2: Use skill which inflicts poison and reduces fire element
Turn 3: Enemy self destructs
Turn 4: Enemy re-appears
Turn 5: Cast fire on enemy - High damage
Turn 5: Enemy takes poison damage


Conclusions
------------

Hide doesn't appear to change anything about enemies. They just stop participating in the battle.

* Showing an enemy does not change their states

fmatthew5876:
Test Cases Add/Rem party
-------------------------------

Battle event 1 triggered by SW1 on
```
Text "Remove"
Remove Brian
Sw1 Off
```

Battle event 2 triggered by SW2 on
```
Text "Add"
Add Brian
Sw2 Off
```

1.
------

Turn 1: Brian Defends
Turn 1: Alex uses skill which turns on Sw1
Turn 1: "Remove" - Brian removed
Turn 1: Carol uses skill which turns on Sw2
Turn 1: "Add" - Brian Added
Turn 1: Enemy uses physical skill which hits all
Turn 1: Brian takes half damage :grey_exclamation: 

2. 
--------
Turn 1: Brian Defends
Turn 1: Alex uses skill which turns on Sw1
Turn 1: "Remove" - Brian removed
Turn 1: Carol Defends
Turn 1: Enemy uses physical skill which hits all

Turn 2: Alex Defends
Turn 2: Carol uses skill which turns on Sw2
Turn 2: "Add" - Brian Added
Turn 2: Enemy uses physical skill which hits all
Turn 2: Brian takes half damage :exclamation: 

3.
-----

Turn 1: Brian uses a skill to increase fire defense on himself
Turn 1: Alex uses skill to remove brian

Turn 2: Carol uses skill to add brian
Turn 2: Enemy uses fire magic on all
Turn 2: Brian takes half damage

4. 
---
Turn 1: Alex uses a skill to increase fire defense of Brian
Turn 1: Carol uses skill to remove brian
Win the battle

Battle again
Turn 1: Alex uses a skill to add brian
Turn 1: Enemy uses fire magic on all
Turn 1: Brian takes normal damage

Conclusion
---------------

* Add/Remove from party doesn't change any battle states, including defend!
* Between battles, Battle states are cleared for all actors, even those outside of the party

fmatthew5876:
This has been cleaned up and tested.

I was able to come up with a very nice way to handle defend and charged states. 

Now every time we construct a new battle algorithm, we remove both the states in `AlgorithmBase::AlgorithmBase()`. For charged, we carry the charge flag over into the algo itself. For defend, we set the defend flag in Defend's constructor, which gets called after the base one.

This works for 2k and works very nicely for 2k3. In 2k3 it means you won't come out of defend state until you select an action, which is what RPG_RT does.

fmatthew5876:
Added the party fixes and tested all the edge cases.

Ghabry:
Thanks. Will review this soon. (I consider this important enough to get it merged before the next Android apk push, because then we have everything in).

src/game_battlealgorithm.cpp
```diff
@@ -114,6 +114,9 @@ Game_BattleAlgorithm::AlgorithmBase::AlgorithmBase(Type ty, Game_Battler* source
 	source_restriction(RPG::State::Restriction(source->GetSignificantRestriction()))
 {
 	Reset();
+	if (source) {
```
Ghabry, line 117: > Turn off defend status in AlgorithmBase constructor 

The source check is unnecessary. source is never null (there is a source-> access a few lines above btw)

fmatthew5876: Removed it

Ghabry:
There is a suspicious call to ``actor->ResetBattle()`` in ``Interpreter::CommandSimulatedAttack``.
Intention is likely to get rid of battle modifiers because of ``getDef`` and ``getSpi``

fmatthew5876:
> There is a suspicious call to `actor->ResetBattle()` in `Interpreter::CommandSimulatedAttack`.
> Intention is likely to get rid of battle modifiers because of `getDef` and `getSpi` but this event command can be called during battle :/

I saw that and wasn't sure. Since we're properly cleaning up before and after battle, it's probably safe to remove now.

PR #2108, opened 2020-04-04
Add Game_Config + CmdlineParser + video options - Config 2 of N

fmatthew5876:
The PR provides the following

- [x] A set of objects for manipulating configuration parameters, including the ability to disable them or lock them to specific values for platforms where it makes sense to do so.
- [x] A Game_Config class for storing EasyRPG configuration
- [x] Beginning support for config file parsing video options via `--config` `-c`
- [x] A new CmdlineParser class for command line arguments

### DIFF
--- a/src/config_param.h
+++ b/src/config_param.h
@@ -1,0 +1,377 @@
+/*
+ * This file is part of EasyRPG Player.
+ *
+ * EasyRPG Player is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * EasyRPG Player is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef EP_CONFIG_PARAM_H
+#define EP_CONFIG_PARAM_H
+
+#include <cstdint>
+#include <limits>
+#include <string>
+#include <vector>
+#include <algorithm>
+#include <type_traits>
+#include "flag_set.h"
+
+/** A configuration parameter with no restrictions */
+template <typename T>
+class ConfigParam {
+public:
+	using value_type = T;
+
+    explicit ConfigParam(T value = {}) : _value(std::move(value)) {}
+
+    const T& Get() const { return _value; }
+
+    bool Set(T value) {
+        if (IsValid(value)) {
+            _value = std::move(value);
+            return true;
+        }
+        return false;
+    }
+
+    bool IsValid(const T& value) const {
+		return true;
+	}
+
+	bool Enabled() const {
+		return true;
+	}
+
+	bool Locked() const {
+		return false;
+	}
+
+private:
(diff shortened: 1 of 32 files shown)


fmatthew5876:
I need to do another round of testing, but this is ready for review now.

carstene1ns:
Question: Why do we need the new command line parser? Seems like inventing the wheel when there are ready to use alternatives, e.g. [CLI11](https://github.com/CLIUtils/CLI11), [Lyra](https://github.com/bfgroup/Lyra), [cxxopts](https://github.com/jarro2783/cxxopts).

fmatthew5876:
We need a command line parser so it's easier to specify options and try to minimize the amount of work for exposing a config parameter between commandline / config file / scene settings.

Whether we use a third party one or write our own doesn't much matter to me. Command line parsing is a simple exercise. Writing one seemed like less work than evaluating all these other options and then getting them to compile on all our platforms.

This one has an unusual interface in that it parses and advances to the next in the same call. So it doesn't fit a traditional loop. This however is the most efficient way to do it and doesn't do any unnecessary extra parsing.

Whatever we choose. It needs to support unicode and case (in)sensitivity the way we use them. It also needs to be able to support the legacy RPG_RT command line options seamlessly combined with more modern ones.

src/config_param.h
```diff
@@ -0,0 +1,377 @@
+/*
```
carstene1ns, line 1: mixed whitespace.

fmatthew5876: Fixed

src/game_config.cpp
```diff
@@ -0,0 +1,171 @@
...
+		video.vsync.Set(ini.GetBoolean("video", "show-fps", false));
+	}
+	if (ini.HasValue("video", "fps-render-window")) {
+		video.vsync.Set(ini.GetBoolean("video", "fps-render-window", false));
+	}
+	if (ini.HasValue("video", "fps-limit")) {
+		video.vsync.Set(ini.GetInteger("video", "fps-limit", 0));
+	}
+	if (ini.HasValue("video", "window-zoom")) {
+		video.vsync.Set(ini.GetInteger("video", "window-zoom", 0));
```
Ghabry, line 137: This all sets the wrong field ``video.vsync`` intentional because WIP (add comment :)) or a bug?

fmatthew5876: Thats a bug for sure. Fixed!

src/config_param.h
```diff
@@ -172,189 +172,189 @@ using DoubleConfigParam = RangeConfigParam<double>;
...
+		lcf::FlagSet<E> _valid = ~lcf::FlagSet<E>();
+
+		E GetFirstValid() const {
+			for (size_t i = 0; i < _valid.size(); ++i) {
+				auto e = static_cast<E>(i);
+				if (_valid[e]) {
+					return e;
+				}
 			}
+			return E{};
```
Ghabry, line 356: the last commit added tab before public:/private:.  I think your editor is too smart with formating :)

src/game_config.cpp
```diff
@@ -0,0 +1,171 @@
...
+		of << "vsync=" << int(video.vsync.Get()) << "\n";
+	}
+	if (video.fullscreen.Enabled()) {
+		of << "fullscreen=" << int(video.fullscreen.Get()) << "\n";
+	}
+	if (video.show_fps.Enabled()) {
+		of << "show-fps=" << int(video.show_fps.Get()) << "\n";
+	}
+	if (video.fps_render_window.Enabled()) {
+		of << "show-fps=" << int(video.fps_render_window.Get()) << "\n";
```
Ghabry, line 161: must be ``fps-render-window``

Issue #2111, opened 2020-04-05
Change of System when loading the game


Br4ssman:
In the game we use 2 types of Systems:

- Blue is only for the main menu
- Brown for the rest of the game

This is so as not to break the synergy of each interface.

From the editor, we have configure the blue System as default, but once the player is in the introduction, we change to the brown System.

Here you can see for a moment how the type of System changes when loading the game:

https://youtu.be/zVx0KGCMsco?t=27

Thanks in advance.

fmatthew5876:
This happens because the regular scene transition out occurs after loading the save game and replacing the currently active system graphic.

PR #2113, opened 2020-04-05
Do an early transition out before loading save

fmatthew5876:
Prevents the load game's state from impacting the
currently visible screen.

Fix: #2111


fmatthew5876:
> Is this just me or does this break the transision? (Goes from visible to black instant)

Oops yes it does. I added a more general fix for this.

Issue #2069, opened 2020-02-08
Rain and snow effects are slightly too strong


fmatthew5876:
1. Cause any strength level of rain or snow effect.
2. Take a screen grab of both RPG_RT and Player
3. Count the particles

You'll see nearly every time Player has more particles on the screen than RPG_RT. There also seems to be a bit more randomness in RPG_RT with how many particles are on the screen each frame. 

Our algorithm for how we spawn new particles needs to be tweaked.

The difference is most obvious at "Weak" strength levels.

fmatthew5876:
Notes from Cherry on Discord about this one:

```
CherryToday at 5:04 PM
I believe it has some non-trivial thing where it uses 100 permanent particles which have a lifetime and position, and those are updated with some randomness.
I think the strength just controls how many of those particles are actually drawn

the particles are in the RPG::ScreenEffect structure in DynRPG (from PepsiOtaku's version)
however, the structure is bigger than what is declared in http://www.rewking.com/dynrpg/_screen_effect_8h_source.html

no guarantees but I think the particles start at offset 0x94
maybe I'm off by 4 or 8 bytes, you'll have to check.
it should be an array with 100 elements consisting of 3 integers each
so, 300 integers in total
so you can log them to a file every frame or whatever

struct Particle {
  int t;
  int x;
  int y;
};

Particle* particles = (Particle*)(((int)RPG::screen->screenEffect)+0x94);

// use particles[0...99]

something like that
again, I'm not 100% sure about the offset 0x94 here, so if you get weird data in the first or last particle or an access violation, adjust the offset a bit

I said the strength controls how many of those particles are actually drawn - but I think additionally it also affects the alpha used
(and the t affects alpha also, obviously.)

Our alphas for rain and sand are pretty close to perfect now, using screen captures
Rain and snow*
Sand is very strange. A lot of randomness
Thanks for all your help. With this and things before

I know the general idea is that t/x/y are initialized randomly on startup, and later on t is usually decremented if >0, and when it was 0 then only with a certain chance the particle is reinitialized (t being set back to some starting value (different per weather type), x/y being randomly set), otherwise nothing happens. x/y are modified every frame, with certain randomness as well.
the behavior is different for different weather types, obviously

you can access the internal image that is used for the weather "sprites" if needed (it is used in combination with some tint), it is an RPG::Image * 3 DWORDs before the particles array (so at +0x88 if my offset was right before)

[...]

PR #2114, opened 2020-04-06
Accurate Weather effects

fmatthew5876:
This PR is to add further improvements to weather emulation.

Fix: #2069 
Fix: #1737 

TBD:
- [x] Perfect rain emulation
- [x] Perfect snow emulation
- [x] Perfect fog emulation
- [x] Perfect sandstorm emulation

### DIFF
--- a/src/game_screen.h
+++ b/src/game_screen.h
@@ -90,7 +90,7 @@ public:
 		int16_t t = 0;
 		int16_t x = 0;
 		int16_t y = 0;
-		// Thse are only used for sandstorm particles.
+		// These are only used for sandstorm particles.
 		// RPG_RT uses double. We use float to save space.
 		int16_t alpha = 0;
 		float vx = 0.0;
--- a/src/game_screen.h
+++ b/src/game_screen.h
@@ -87,11 +87,16 @@ public:
 	int GetWeatherStrength();
 
 	struct Particle {
+		int16_t t = 0;
 		int16_t x = 0;
 		int16_t y = 0;
-		int16_t life = 0;
-		int16_t speed = 0;
-		float angle = 0.0f;
+		// Thse are only used for sandstorm particles.
+		// RPG_RT uses double. We use float to save space.
+		int16_t alpha = 0;
+		float vx = 0.0;
+		float vy = 0.0;
+		float ax = 0.0;
+		float ay = 0.0;
 	};
 
 	const std::vector<Particle>& GetParticles();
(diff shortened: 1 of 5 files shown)


fmatthew5876:
So far I've got accurate rain, snow, and fog. 

Sandstorm is almost there. I have everything figured out except how the initial acceleration values are calculated.

fmatthew5876:
Now we finally have accurate rain, snow, fog, and sandstorm weather effects!

Since this is standalone and doesn't break games, I'd like to have it in 6.2 release if we can.

src/weather.cpp
```diff
@@ -57,18 +59,22 @@ void Weather::Draw(Bitmap& dst) {
...
+static constexpr int num_sand_particles[] = { 128, 191, 255 };
 static constexpr int num_sand_colors = 4;
-static constexpr int num_sand_strength = 3;
 static constexpr auto sand_particle_rect = Rect{ 0, 0, 1, 2 };
 
 static constexpr auto sand_particle_bitmap_rect = Rect{
 	0, 0,
 	sand_particle_rect.width,
-	sand_particle_rect.height * num_sand_colors * num_sand_strength
+	sand_particle_rect.height * num_sand_colors,
```
Ghabry, line 77: is this extra comma intentional? (I'm surprised that C++ accepts trailing commas)

src/game_screen.h
```diff
@@ -87,11 +87,16 @@ class Game_Screen {
 	int GetWeatherStrength();
 
 	struct Particle {
+		int16_t t = 0;
 		int16_t x = 0;
 		int16_t y = 0;
-		int16_t life = 0;
-		int16_t speed = 0;
-		float angle = 0.0f;
+		// Thse are only used for sandstorm particles.
```
Ghabry, line 93: these

src/weather.cpp
fmatthew5876: Not intentional, but modern C++ lets do you do this now.

fmatthew5876:
> Is the movement speed of the Fog/Sand also random? In Player the left-movement is much slower than in RPG_RT. (tested multiple times)

I'll take a look at this in the debugger to verify fog values and algorithm

> 
> Imo the rain has on "lowest" the wrong colors. In RPG_RT almost all the rain is transparent but in Player you see more which are almost white.

Did you compare with true color RPG_RT? The colors are definately different without it, many instances the rain colors look more greenish on 16bit RPG_RT.

> 
> But without a direct comparison you would never notice this. Looks great!

fmatthew5876:
Looking at rain in true color RPG_RT, it looks like you're right!

In fact there is another feature I missed entirely. The colors of the rain depend on the strength level!

| Strength | Color |
| -- | -- |
| 0 | 60 * t / 12 |
| 1 | 72 * t / 12 |
| 2 | 80 * t / 12 |

fmatthew5876:
I did more analysis of rain and snow with RPG_RT true color, the pixel colors change based on strength. I also fixed the fog speeds and verified with x64dbg the fog movement algorithm matches RPG_RT.

This should be ready now

src/game_screen.h
fmatthew5876: Forgot this one. Fixed now.

PR #2116, opened 2020-04-06
Battle: Crashfix and minor bugfix

Ghabry:
Not sure if there is a better location available for the crashfix (1st commit). The animation must be part of the transition screen, can't be removed in Scene_Battle constructor.

Stacktraces fixed by crashfix:
```
  #00  pc 000000000024535c  lib/arm64/libmain.so (DrawableList::Draw(Bitmap&, int, int)+104)
  #01  pc 00000000002f88c8  lib/arm64/libmain.so (Transition::Init(Transition::Type, Scene*, int, bool)+372)
  #02  pc 00000000002d8788  lib/arm64/libmain.so (Scene::MainFunction()+452)
  #03  pc 00000000002bcda8  lib/arm64/libmain.so (Player::MainLoop()+180)
  #04  pc 00000000002bcc70  lib/arm64/libmain.so (Player::Run()+216)
  #05  pc 00000000002a8a74  lib/arm64/libmain.so (SDL_main+12)
  #06  pc 0000000000043964  lib/arm64/libSDL2.so (Java_org_libsdl_app_SDLActivity_nativeRunMain+696)
  #07  pc 0000000000031bd4  oat/arm64/base.odex (art_jni_trampoline+228)
  #08  pc 0000000000078f68  oat/arm64/base.odex (org.libsdl.app.SDLMain.run+728)
  
  
  00  pc 000000000024535c  lib/arm64/libmain.so (DrawableList::Draw(Bitmap&, int, int)+104)
  #01  pc 00000000002f88c8  lib/arm64/libmain.so (Transition::Init(Transition::Type, Scene*, int, bool)+372)
  #02  pc 00000000002d8788  lib/arm64/libmain.so (Scene::MainFunction()+452)
  #03  pc 00000000002bcda8  lib/arm64/libmain.so (Player::MainLoop()+180)
  #04  pc 00000000002bcc70  lib/arm64/libmain.so (Player::Run()+216)
  #05  pc 00000000002a8a74  lib/arm64/libmain.so (SDL_main+12)
  
  
  #00  pc 000000000022ea68  lib/arm64/libmain.so (BattleAnimation::SetFrame(int))
  #01  pc 000000000029c78c  lib/arm64/libmain.so (Game_Screen::ShowBattleAnimation(int, int, bool, int)+200)
  #02  pc 00000000002e3004  lib/arm64/libmain.so (Scene_Map::Start()+356)
  #03  pc 00000000002d8768  lib/arm64/libmain.so (Scene::MainFunction()+420)
  #04  pc 00000000002bcda8  lib/arm64/libmain.so (Player::MainLoop()+180)
  #05  pc 00000000002bcc70  lib/arm64/libmain.so (Player::Run()+216)
  #06  pc 00000000002a8a74  lib/arm64/libmain.so (SDL_main+12)
```

### DIFF
--- a/src/scene_battle.cpp
+++ b/src/scene_battle.cpp
@@ -96,6 +96,9 @@ void Scene_Battle::Start() {
 
 	Game_Battle::Init(troop_id);
 
+	// Update picture visibility
+	Main_Data::game_pictures->UpdateGraphics(true);
+
 	cycle = 0;
 	auto_battle = false;
 	enemy_action = NULL;
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -154,6 +154,7 @@ void Game_BattleAlgorithm::AlgorithmBase::Reset() {
 	revived = false;
 	reflect = -1;
 	states.clear();
+	shift_attributes.clear();
 
 	if (!IsFirstAttack()) {
 		switch_on.clear();
--- a/src/game_screen.cpp
+++ b/src/game_screen.cpp
@@ -397,9 +397,14 @@ int Game_Screen::ShowBattleAnimation(int animation_id, int target_id, bool globa
 
 void Game_Screen::UpdateBattleAnimation() {
 	if (animation) {
-		animation->Update();
-		data.battleanim_frame = animation->GetFrame();
-		if (animation->IsDone()) {
+		if (!animation->IsDone()) {
+			animation->Update();
+			data.battleanim_frame = animation->GetFrame();
+		}
+
+		if (animation->IsDone() && !Game_Battle::IsBattleRunning()) {
+			// FIXME: Lifetime is flawed but we need the animation in battle for
+			// SE and flash. Delay destruction until back on the map.
 			CancelBattleAnimation();
 		}
 	}


src/scene_map.cpp
```diff
@@ -200,6 +200,7 @@ void Scene_Map::TransitionOut(SceneType next_scene) {
 			// If screen is already erased, RPG_RT does nothing for 40 frames.
 			transition.InitErase(Transition::TransitionNone, this, 40);
 		}
+		Main_Data::game_screen->CancelBattleAnimation();
```
fmatthew5876, line 203: This breaks an edge case where when you start a battle before the animation completes, the flash and sound effects of the animation will occur during the battle.

src/scene_battle.cpp
```diff
@@ -94,6 +94,8 @@ void Scene_Battle::Start() {
 	Main_Data::game_enemyparty.reset(new Game_EnemyParty());
 	Main_Data::game_enemyparty->Setup(troop_id);
 
+	Main_Data::game_pictures->UpdateGraphics(true);
```
fmatthew5876, line 97: Should we do this after `Game_Battle::Init()`? It's in `Game_Battle::Init()` when `Game_Battle::IsBattleRunning()` becomes true. 

Now since this method takes a bool and doesn't rely on that global anymore, it should be fine. However I have a minor concern that maybe in the future someone adds something in the call chain of `UpdateGraphics` which depends on `IsBattleRunning()`

TLDR: It doesn't matter, but it might be a bit safer to move it after `Game_Battle::Init()` just to be sure.

src/scene_map.cpp
Ghabry: The cause is that the CancelBattleAnimation destroys the Animation but Drawable->Take fails to take it because it is not on the Battle stack. After returning to the map the Drawable list contains a dangling reference to this animation and crashes.

Will sharing the BattleAnimation help? Only must be set to Hidden (at the same location where I put the picture update code)

Ghabry: Imo the "Drawable->Take" when Deregistering should assert when the Take failed to catch such bugs but this breaks most Scene destructions.

fmatthew5876: Sharing and hiding should work. That will still make sound effects and screen effects occur. Ideally we would be better off to just not allocate and destroy the animation every time. Have it's lifetime match the lifetime of screen. That would also be a minor performance win.

> Imo the "Drawable->Take" when Deregistering should assert when the Take failed to catch such bugs but this breaks most Scene destructions.

The whole lifetime model of drawables is flawed. The new code makes that even more explicit. We need another refactor which better ties drawable lifetimes to scenes / lists.

Ghabry: I solved this now via delayed deletion because there is another way to crash that is not solvable through sharing: When you go from Debug scene to Battle while an animation is active (sharing doesn't work because Scene Map is not the stack top). Is this ok?

fmatthew5876: Ouch, that means there is another bug. Battle pictures and weather won't work when you trigger battles from the debug menu. 

We can punt that as it isn't critical https://github.com/EasyRPG/Player/issues/2120

This workaround is for by me for now.

PR #2118, opened 2020-04-07
Implement Ticks for: libsndfile, FmMidi, Wildmidi (kinda)

### LINKED ISSUE #1733: Return of Touhou Mother - Midi-Tick Regression
**Platform:** Android

Battles seem to hang again whenever normal attacks are used in Return of Touhou Mother. Something may have regressed the midi-tick fix in issue #744, judging by the way the game hangs whenever I try to combo with regular attacks.

### LINKED ISSUE #2106: Dの冒険Ⅶ: Hang after investigating the grave
#### Name of the game:
Dの冒険Ⅶ

https://viprpgkohaku2016.wixsite.com/kohaku2016/no-20
[Direct DL link](https://www.mediafire.com/file/e6i4cmms42wyry5/vipkh16_20_%25EF%25BC%25A4%25E3%2581%25AE%25E5%2586%2592%25E9%2599%25BA7%25E3%2580%2580%25EF%25BD%259ED%25E3%2581%25AE%25E8%25AC%258E%25EF%25BC%2581%25EF%25BC%259F%25EF%25BD%259E.zip/file)

[Save07.zip](https://github.com/EasyRPG/Player/files/4432686/Save07.zip)

Investigate the grave directly above you. It fades to a scene where it says "Chapter 8" and then it hangs with "Event 2 reached execution limit"

Good news: Not a regression. Hangs in 0.5.4 and newer

### LINKED ISSUE #2241: Return of Touhou Mother Soflocks on Battle System on Android
Hello I'm new here and I just found an issue with Return of Touhou Mother at the begging on the Battle System, well the error is it when I start to battle with the very first enemy on the game, I don't know her name, well I can make a normal attack, but the rhythm combos doesn't work, and once I attack for the second time, the game softlocks, I'm using the Android version EasyRPG maker, and I'm still not sure how to fix this issue,

Ghabry:
_libsndfile_ and our bad _wave decoder for slow platforms (FASTWAV)_ have no way to query the position. I added sample counting to get the position. This is hard to get right when "Seek" is used, so I kept this as a FIXME.

In _FmMidi_ I added the current tick count to the midi message. The value matches the harmony value with some margin of error (is a bit higher/lower, harmony somehow keeps the value constant). No idea yet how to make it match better.

_WildMidi_: This is problematic. Will probably create a PR to there library someday to export the values through there API. They only offer the samplerate (How many samples were created based on the Midi file, so this is indirectly calculated through tempo + division).
When the Midi has no tempo changes just knowing the current tempo is enough to get a quite good value but for a perfect value we need all tempo informations (thats why the midi library has to calculate it).
There is no API to get the current tempo. Assume a fixed tempo (default tempo of WildMidi) and parse the division manually from the Midi header. This makes the error small enough to fix #2106 (the tick reaches the required value before the track ends), fix #1733 and fix #2241.

The Ticks are not implemented for Midis that state that there timing is FPS based. WildMidi rejects such files and because we are not aware of any bug reports this format is very unpopular I guess.

Test for MIDI. Autorun event. Outputs the delta since the last Midi Tick call
```
@> Play BGM: 'Battle1', 100, 100, 50
@> Control Variables: [0001:Variable 0001] = MIDI Play Location (Tick)
@> Control Variables: [0003:Variable 0003] = Variable [0001]
@> Control Variables: [0001:Variable 0001] -= Variable [0002]
@> Text: \V[1]\|\^
@> Control Variables: [0002:Variable 0002] = Variable [0003]
```

### DIFF
--- a/src/decoder_fmmidi.cpp
+++ b/src/decoder_fmmidi.cpp
@@ -66,12 +66,16 @@ bool FmMidiDecoder::Open(FILE* file) {
 	}
 	seq->rewind();
 
+	ticks_per_sec = (float)seq->get_division() / tempo * 1000000;
+
 	return true;
 }
 
 bool FmMidiDecoder::Seek(size_t offset, Origin origin) {
 	if (offset == 0 && origin == Origin::Begin) {
 		mtime = 0.0f;
+		mtime_last_tempo_change = 0.0f;
+		ticks_last_tempo_change = 0;
 		seq->rewind();
 		begin = true;
 
@@ -108,7 +112,8 @@ bool FmMidiDecoder::SetPitch(int pitch) {
 }
 
 int FmMidiDecoder::GetTicks() const {
-	return 0;
+	float delta = mtime - mtime_last_tempo_change;
+	return ticks_last_tempo_change + static_cast<int>(ticks_per_sec * delta);
 }
 
 int FmMidiDecoder::FillBuffer(uint8_t* buffer, int length) {
@@ -142,8 +147,17 @@ void FmMidiDecoder::sysex_message(int, const void * data, std::size_t size) {
 	synth->sysex_message(data, size);
 }
 
-void FmMidiDecoder::meta_event(int, const void *, std::size_t) {
-	// no-op
+void FmMidiDecoder::meta_event(int event, const void * data, std::size_t size) {
+	const auto* d = reinterpret_cast<const uint8_t*>(data);
+
+	if (size == 3 && event == 0x51) {
+		tempo = (static_cast<uint_least32_t>(static_cast<unsigned char>(d[0])) << 16)
+				| (static_cast<unsigned char>(d[1]) << 8)
+				| static_cast<unsigned char>(d[2]);
+		ticks_per_sec = (float)seq->get_division() / tempo * 1000000;
+		ticks_last_tempo_change = GetTicks();
+		mtime_last_tempo_change = mtime;
+	}
 }
 
 void FmMidiDecoder::reset() {
--- a/src/decoder_wav.cpp
+++ b/src/decoder_wav.cpp
@@ -36,6 +36,7 @@ WavDecoder::~WavDecoder() {
 }
 
 bool WavDecoder::Open(FILE* file) {
+	decoded_samples = 0;
 	file_=file;
 	fseek(file_, 16, SEEK_SET);
 	fread(&chunk_size, 1, 4, file_);
@@ -105,6 +106,8 @@ bool WavDecoder::Seek(size_t offset, Origin origin) {
 	if (origin != Origin::End) {
 		offset += audiobuf_offset;
 	}
+	// FIXME: Proper sample count for seek
+	decoded_samples = 0;
 
 	bool success = fseek(file_,offset,(int)origin) == 0;
 	cur_pos = ftell(file_);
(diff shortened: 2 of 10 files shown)


src/midisequencer.h
```diff
@@ -44,6 +44,7 @@ namespace midisequencer{
         uint_least32_t message;
         int port;
         int track;
+        float ticks;
```
fmatthew5876, line 47: Does this need to be a float? Whenever I think of a concept of a number of ticks, it's always integral.

Using a float means we can accumulate rounding error and mysterious inaccuracies that are hard to track down.

Ghabry:
This kinda fixes "チンブラ戦隊ブライアン" but this uses Midi for timing in intro music so it needs perfect ticks (and is a perfect test candidate!).

src/decoder_wav.cpp
```diff
@@ -148,24 +151,32 @@ int WavDecoder::FillBuffer(uint8_t* buffer, int length) {
...
+
 	return decoded;
 }
 
-#endif
+int WavDecoder::GetTicks() const {
+	return decoded_samples / (samplerate * nchannels);
+}
+
+#endif
```
carstene1ns, line 182: missing newline

src/midisequencer.h
fdelapena: at least time (float) and time_offset (double) are, but not sure if actually required

fdelapena:
test this please

Ghabry:
oh there is progress on the PR :). I will check the float vs int comment and then this is ready for now.

Ghabry:
I reworked the midi ticks part of FmMidi. I added a function to query the divisions. Midi ticks calculation however is now from the outside. Is more accurate this way because before we had message resolution, now we have millisecond resolution.
The error margin is ~4% compared to harmony (the value is slightly lower for some reason)

I want to do the following in a later PR, is too involved for this one:
Create a GenericMidiDecoder which is responsible for initializing the "sub decoder" (FmMidi, WildMidi, Fluidsynth/Lite) and the GenericMidiDecoder will run a FmMidi sequencer that keeps track of the current position to calculate the ticks. This way we get the same ticks everywhere and this is a cheap operation. Is just parsing, not sampling.
FmMidi and Fluid could even use the same sequencer this way, this allows switching Midi audio on-the-fly.
(WildMidi does not provide suitable functions to hook in our sequencer)

Ghabry:
Here is "チンブラ戦隊ブライアン" https://easyrpg.org/play/?game=miditest

https://ux.getuploader.com/viprpg_wiki2/download/60/%E3%83%81%E3%83%B3%E3%83%96%E3%83%A9%E6%88%A6%E9%9A%8A%E3%83%96%E3%83%A9%E3%82%A4%E3%82%A2%E3%83%B3.zip

Can't compare it with RPG_RT because I can't convince Wine to play MIDI (there Wiki is terribly outdated...)

src/decoder_fmmidi.cpp
```diff
@@ -123,6 +125,7 @@ int FmMidiDecoder::FillBuffer(uint8_t* buffer, int length) {
 		seq->play(mtime, this);
 		notes = synthesize(reinterpret_cast<int_least16_t*>(buffer), samples, frequency);
 		mtime += delta;
+		ticks += ticks_per_sec * delta;
```
Ghabry, line 128: Had to change this to float because ticks_per_sec and delta are float so the error increases over time when using int.

Ghabry:
Better way: I remember the last tempo change position and use this to calculate the ticks when they are requested via GetTicks(). This avoids useless tick calculations and has much less error (result is around the same, ~3-5% error compared to RPG_RT)

Issue #2126, opened 2020-04-08
2012紅白No.016: Wrong KeyInput code path taken


Ghabry:
#### Name of the game:
2012紅白No.016 - ケツアクメ120連発　ケツアヘ傑作選
http://vipkh2012.y.ribbon.to/db/profile.cgi?_v=016&tpl=view2

#### Describe the issue in detail and how to reproduce it:
Fast forward through the intro. In the selection only Down arrow works.

The problem is that the Key input is:
``<parameters>70 0 0 0 0 0 1 0 0 1 1</parameters>``

This is detected as a 2k3 key input because the parameter is too long. When saving the event in 2k you get
``<parameters>70 0 0 0 0 0 1 0 0 1</parameters>``

see also #604

Ghabry:
Test case (We really love testcases since a while ;)):
Wait for Down press and output the pressed key (should be Down).

[Map0001.zip](https://github.com/EasyRPG/Player/files/4452566/Map0001.zip)

In short: The observation in #604 is correct except for one edge case.

I prepared the KeyInput commands:

1. 2k legacy command
``<parameters>1 1 1 0 0</parameters>``

Result 2k Value!: Reacts to all directions (2k legacy limitation)
Result 2k3 EN: Reacts to all directions

2. 2k Value! command
``<parameters>1 1 1 0 0 0 1 0 0 0</parameters>``

Result 2k Value!: Reacts to key down
Result 2k3 EN: Reacts to key down

3. 2k3 command
``<parameters>1 1 1 0 0 0 0 0 0 0 1 0 0 0</parameters>``

Result 2k Value!: Nothing happens (!)
Result 2k3 EN: Reacts to key down

Conclusion: The engines are backwards compatible using the same technique we use (even the editor handles this properly!) but 2k Value does not support the 2k3 input scheme.

carstene1ns:
Conclusion: They forked 2k3 from 2k. 2k value never got backported patches.

fmatthew5876:
2k and 2k3 definately forked on their implementations.

It's why we have this lovely code...

https://github.com/EasyRPG/Player/blob/master/src/game_interpreter.cpp#L117

PR #2127, opened 2020-04-08
Fix 2126

Ghabry:
This implements the test case observation in #2126 

But this is a interpreter change, are you brave enough to merge it? ;)


carstene1ns:
Let's get ready to rumble :8ball:

Issue #1820, opened 2019-06-21
King of Grayscale - 2k3 battles don't handle Agility accurately


NovemberJoy:
Name of the game: King of Grayscale

Download link: [https://rpgmaker.net/games/9835/downloads/9767/](Here)

Player platform:
Windows 10 Home 1709 64-bit

Attach files (as a .zip archive or link them)
 - [Save02.zip](https://github.com/EasyRPG/Player/files/3312846/Save02.zip)
 - [easyrpg_log.txt](https://github.com/EasyRPG/Player/files/3312848/easyrpg_log.txt)

Describe the issue in detail and how to reproduce it:
Load the provided save and enter the battleground to the left. Attack the enemies provided with the standard Attack command. In EasyRPG, they're guaranteed to be defeated before they get a single turn unless you miss, while in RPG_RT they'll get at least one turn before dying.

fmatthew5876:
I confirmed this is still broken in master and fixed by #2133 

However there is another unrelated issue where the actors sprites are not visible in the battle.

Issue #446, opened 2015-03-26
Picture scale differences


gadesx:
The picture representation using scale are different,
and with 50scale it show less size

![comparison](https://cloud.githubusercontent.com/assets/2506953/6854833/f16ca1f2-d3f3-11e4-8360-45688e6f40f1.png)

Test case:
http://dl.dropboxusercontent.com/u/22524283/rpg%20maker/easy/Tests%20case%2001.rar

fdelapena:
@gadesx the dropbox link is broken. Could you reupload the test case here where possible?

gadesx:
its impossible for me reupload that without internet in home, but you can make a picture similar and test with player.

Ghabry:
added [Needs Feedback] until a new example is available.

I'm pretty sure I saw a while ago some prepared image that either showed an Apple or Nothing depending on the resize algorithm of the web browser used, would be a nice test case but I can't find it anymore.

gadesx:
The old test case, maybe works fine with actual player.
[Tests case 01.zip](https://github.com/EasyRPG/Player/files/2493533/Tests.case.01.zip)

fmatthew5876:
This was also noticed in #2133 for battle gauges. Leaving a note here

rueter37:
One way to test if the battle gauges scale correctly is using the System2C graphics from the RTP. The right end of the gauge must become diagonal upon reaching 16 out of 25.

@Ghabry mentioned a possible solution for this issue in https://github.com/EasyRPG/Player/issues/2519#issuecomment-833009899.

Ghabry:
You can cheat a bit by translating the image before the scaling. Though for the issue in here with the 50% this fixes nothing.

With that code, when a picture is shown, you can use NUM+ and NUM- to adjust the offset.

When you set it to "-4" for the Yume2kki issue than the images are blitted correctly. You could try playing around with this for the gauge.

```
	static int i = 0;

	if (Input::IsTriggered(Input::PLUS)) {
		++i;
		Output::Debug("{}", i);
	} else if (Input::IsTriggered(Input::MINUS)) {
		--i;
		Output::Debug("{}", i);
	}

	Transform xform = Transform::Translation(i, i);
	xform *= Transform::Scale(zoom_x, zoom_y);
```

Ghabry:
Translation the gauge by "1, 0" before scaling makes the gauge pixel perfect but then you have obviously an offset in there, making the rest look wrong.

rueter37:
[...]

PR #2133, opened 2020-04-10
Battle 2k3 1: Fix ATB Gauge update algorithm

fmatthew5876:
I've fixed the atb gauge algorithm for 2k3 battles.

I figured out the math by exploring with x64dbg and finding it in the assembly. I verified the frame by frame numbers by dumping and comparing dynrpg output and player output.

Fix: #1820

I checked that:
- [x] Max ATB Gauge is 300000
- [x] When battler reaches max atb, the gauge doesn't increase further

- [x] Hidden monsters do not contribute to AGI sum
- [x] When a hidden monster is added back into the battle, he will contribute to AGI sum
- [x] Dead monsters *do* contribute to AGI sum
- [x] Stunned monsters *do* contribute to AGI sum
- [x] Dead actors *do* contribute to AGI sum
- [x] Stunned actors *do* contribute to AGI sum

- [x] Actors ATB gauge does not increase when they are dead
- [x] Actors ATB gauge does not increase when they are stunned
- [x] Monster ATB gauge does not increase when they are hidden
- [x] Monster ATB gauge does not increase when they are dead
- [x] Monster ATB gauge *does* increase when they are stunned

- [x] Monster ATB gauge goes to 0 when monster dies
- [x] Actor ATB gauge goes to 0 when actor dies

- [x] Double check ATB against assembly
- [x] Check what happens when actor starts battle with state restriction

### DIFF
--- a/src/game_battle.cpp
+++ b/src/game_battle.cpp
@@ -250,7 +250,7 @@ void Game_Battle::UpdateAtbGauges() {
 			if (use_2k3e_algo) {
 				const auto cur_atb = bat->GetAtbGauge();
 				const auto multiplier = std::max(1.0, static_cast<double>(275000 - cur_atb) / 55000.0);
-				increment = Utils::RoundToInt(multiplier * increment);
+				increment = Utils::RoundTo<int>(multiplier * increment);
 			}
 			bat->IncrementAtbGauge(increment);
 		}
--- a/src/game_battle.cpp
+++ b/src/game_battle.cpp
@@ -237,6 +237,8 @@ void Game_Battle::UpdateAtbGauges() {
 	Main_Data::game_enemyparty->GetBattlers(battlers);
 	Main_Data::game_party->GetBattlers(battlers);
 
+	const auto use_2k3e_algo = Player::IsRPG2k3E();
+
 	int sum_agi = 0;
 	for (auto* bat: battlers) {
 		// RPG_RT uses dead and state restricted battlers to contribute to the sum.
@@ -249,12 +251,16 @@ void Game_Battle::UpdateAtbGauges() {
 	const int max_atb = Game_Battler::GetMaxAtbGauge();
 
 	for (auto* bat: battlers) {
-		int increment = 0;
 		// RPG_RT always updates atb for non-hidden enemies, even if they can't act.
 		if (bat->Exists() && (bat->CanAct() || bat->GetType() == Game_Battler::Type_Enemy))
 		{
 			const auto agi = bat->GetAgi();
-			increment = max_atb / (sum_agi / (agi + 1));
+			auto increment = max_atb / (sum_agi / (agi + 1));
+			if (use_2k3e_algo) {
+				const auto cur_atb = bat->GetAtbGauge();
+				const auto multiplier = std::max(1.0, static_cast<double>(275000 - cur_atb) / 55000.0);
+				increment = static_cast<int>(multiplier * increment);
+			}
 			bat->IncrementAtbGauge(increment);
 		}
 	}
--- a/src/game_battle.cpp
+++ b/src/game_battle.cpp
@@ -34,6 +34,7 @@
 #include "reader_util.h"
 #include "spriteset_battle.h"
 #include "output.h"
+#include "utils.h"
 
 namespace Game_Battle {
 	const RPG::Troop* troop;
@@ -259,7 +260,7 @@ void Game_Battle::UpdateAtbGauges() {
 			if (use_2k3e_algo) {
 				const auto cur_atb = bat->GetAtbGauge();
 				const auto multiplier = std::max(1.0, static_cast<double>(275000 - cur_atb) / 55000.0);
-				increment = static_cast<int>(multiplier * increment);
+				increment = Utils::RoundToInt(multiplier * increment);
 			}
 			bat->IncrementAtbGauge(increment);
 		}
--- a/src/game_battle.cpp
+++ b/src/game_battle.cpp
@@ -155,26 +155,16 @@ bool Game_Battle::CheckWin() {
 }
 
 bool Game_Battle::CheckLose() {
-	if (!Main_Data::game_party->IsAnyActive())
-		return true;
-
-	// If there are active characters, but all of them are in a state with Restriction "Do Nothing" and 0% recovery probability, it's game over
+	// If there are active characters, but all of them are in a state with Restriction "Do Nothing" and 0% recovery probability (including death), it's game over
 	// Physical recovery doesn't matter in this case
-	int character_number = 0;
-	std::vector<Game_Battler*> actors;
 
-	Main_Data::game_party->GetActiveBattlers(actors);
-	for (auto actor : actors) {
-		for (auto id_state : actor->GetInflictedStates()) {
-			RPG::State *state = ReaderUtil::GetElement(Data::states, id_state);
-			if (state->restriction == RPG::State::Restriction_do_nothing && state->auto_release_prob == 0) {
-				++character_number;
-				break;
-			}
+	for (auto& actor : Main_Data::game_party->GetActors()) {
+		if (!actor->IsHidden() && actor->CanActOrRecoverable()) {
+			return false;
 		}
 	}
 
-	return character_number == static_cast<int>(actors.size());
+	return true;
 }
 
 Spriteset_Battle& Game_Battle::GetSpriteset() {
--- a/src/game_battle.cpp
+++ b/src/game_battle.cpp
@@ -232,25 +232,30 @@ void Game_Battle::NextTurn(Game_Battler* battler) {
 	Main_Data::game_party->IncTurns();
 }
 
-void Game_Battle::UpdateGauges() {
+void Game_Battle::UpdateAtbGauges() {
 	std::vector<Game_Battler*> battlers;
-	Main_Data::game_enemyparty->GetActiveBattlers(battlers);
-	Main_Data::game_party->GetActiveBattlers(battlers);
+	Main_Data::game_enemyparty->GetBattlers(battlers);
+	Main_Data::game_party->GetBattlers(battlers);
 
-	int max_agi = 1;
-
-	for (std::vector<Game_Battler*>::const_iterator it = battlers.begin();
-		it != battlers.end(); ++it) {
-		max_agi = std::max(max_agi, (*it)->GetAgi());
+	int sum_agi = 0;
+	for (auto* bat: battlers) {
+		// RPG_RT uses dead and state restricted battlers to contribute to the sum.
+		if (!bat->IsHidden()) {
+			sum_agi += bat->GetAgi();
+		}
 	}
+	sum_agi *= 100;
 
-	// Only affects how fast the gauges move, can be safely clamped
-	max_agi = Utils::Clamp(max_agi, 1, 1000);
+	const int max_atb = Game_Battler::GetMaxAtbGauge();
 
-	for (std::vector<Game_Battler*>::const_iterator it = battlers.begin();
-		it != battlers.end(); ++it) {
-		if (!(*it)->GetBattleAlgorithm() && (*it)->CanAct()) {
-			(*it)->UpdateGauge(1000 / max_agi);
+	for (auto* bat: battlers) {
+		int increment = 0;
+		// RPG_RT always updates atb for non-hidden enemies, even if they can't act.
+		if (bat->Exists() && (bat->CanAct() || bat->GetType() == Game_Battler::Type_Enemy))
+		{
+			const auto agi = bat->GetAgi();
+			increment = max_atb / (sum_agi / (agi + 1));
+			bat->IncrementAtbGauge(increment);
 		}
 	}
 }
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -652,12 +652,7 @@ bool Game_Battler::HasReflectState() const {
 }
 
 void Game_Battler::ResetBattle() {
-	auto atb = GetMaxAtbGauge();
-	if (!HasPreemptiveAttack()) {
-		atb /= 2;
-	}
-	SetAtbGauge(atb);
-
+	SetAtbGauge(0);
 	SetCharged(false);
 	SetIsDefending(false);
 	SetHidden(false);
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -78,16 +78,27 @@ RPG::State::Restriction Game_Battler::GetSignificantRestriction() const {
 	return State::GetSignificantRestriction(GetStates());
 }
 
-bool Game_Battler::CanActImpl(int max_arp) const {
+bool Game_Battler::CanAct() const {
 	const auto& states = GetStates();
 	for (size_t i = 0; i < states.size(); ++i) {
 		if (states[i] > 0) {
 			const auto* state = ReaderUtil::GetElement(Data::states, i + 1);
-			if (state
-					&& state->restriction == RPG::State::Restriction_do_nothing
-					&& state->auto_release_prob <= max_arp
-					)
-			{
+			assert(state);
+			if (state->restriction == RPG::State::Restriction_do_nothing) {
+				return false;
+			}
+		}
+	}
+	return true;
+}
+
+bool Game_Battler::CanActOrRecoverable() const {
+	const auto& states = GetStates();
+	for (size_t i = 0; i < states.size(); ++i) {
+		if (states[i] > 0) {
+			const auto* state = ReaderUtil::GetElement(Data::states, i + 1);
+			assert(state);
+			if (state->restriction == RPG::State::Restriction_do_nothing && state->auto_release_prob == 0) {
 				return false;
 			}
 		}
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -20,6 +20,7 @@
 #include <algorithm>
 #include <cassert>
 #include <cmath>
+#include <limits>
 #include "player.h"
 #include "game_battler.h"
 #include "game_actor.h"
@@ -77,13 +78,18 @@ RPG::State::Restriction Game_Battler::GetSignificantRestriction() const {
 	return State::GetSignificantRestriction(GetStates());
 }
 
-bool Game_Battler::CanAct() const {
-	const std::vector<int16_t> states = GetInflictedStates();
-	for (int i = 0; i < (int)states.size(); i++) {
-		// States are guaranteed to be valid
-		const RPG::State* state = ReaderUtil::GetElement(Data::states, states[i]);
-		if (state->restriction == RPG::State::Restriction_do_nothing) {
-			return false;
+bool Game_Battler::CanActImpl(int max_arp) const {
+	const auto& states = GetStates();
+	for (size_t i = 0; i < states.size(); ++i) {
+		if (states[i] > 0) {
+			const auto* state = ReaderUtil::GetElement(Data::states, i + 1);
+			if (state
+					&& state->restriction == RPG::State::Restriction_do_nothing
+					&& state->auto_release_prob <= max_arp
+					)
+			{
+				return false;
+			}
 		}
 	}
 	return true;
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -658,7 +658,8 @@ bool Game_Battler::HasReflectState() const {
 }
 
 void Game_Battler::ResetBattle() {
-	SetAtbGauge(0);
+	// Note: ATB gauge is not reset here. This is on purpose because RPG_RT will freeze
+	// the gauge and carry it between battles if !CanActOrRecoverable().
 	SetCharged(false);
 	SetIsDefending(false);
 	SetHidden(false);
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -382,7 +382,7 @@ bool Game_Battler::AddState(int state_id, bool allow_battle_states) {
 	}
 
 	if (state_id == RPG::State::kDeathID) {
-		SetGauge(0);
+		SetAtbGauge(0);
 		SetHp(0);
 		SetAtkModifier(0);
 		SetDefModifier(0);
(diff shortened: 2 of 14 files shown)


Ghabry:
Nice find.
Does this also mean that the atb gauge takes now ages to fill and you have to fast forward permanently?
Btw in 2k3e the algorithm was changed (fills faster). Do you use this one or the old one?

Ghabry:
Because this is a 2k3 battle functionality change I prefer to have it post 0.6.2 (can be submitted to the Beta channel). Because the 2k3 battle system is so fragile and breaks when you look at it and still needs a refactor (sorry, wanted to do this since months)

fmatthew5876:
> Nice find.
> Does this also mean that the atb gauge takes now ages to fill and you have to fast forward permanently?

Yes :)

Long ago that was the very thing I noticed about 2k3 and decided not to use it for my project

> Btw in 2k3e the algorithm was changed (fills faster). Do you use this one or the old one?

This is using the advocate version as that's the one I have dynrpg to help me figure out memory addresses of stuff. If I had the memory address for the atbgauge in 2k3e I could find that algo and add it.

> Because this is a 2k3 battle functionality change I prefer to have it post 0.6.2 (can be submitted to the Beta channel). Because the 2k3 battle system is so fragile and breaks when you look at it and still needs a refactor (sorry, wanted to do this since months)

I think this one is pretty safe because it just replaces one math algo with another. If it were moving stuff around in the 2k3 battle scenes I would absolutely agree with you.

Ghabry:
I bet the code is in the ultimate.dll and it just patches the ATB function.
Because the 2k3E is just a binary patched 2k3 it should be enough to look at the ATB function in a debugger while it is running. You will probably see some patched in JMP there.

Ghabry:
The ultimate.dll exports a function AtbHandlerMain and the first value it loads is 275000.

The decompiled function (via Ghidra) is:

```c
int ATBHandlerMain(int param_1,int param_2)

{
  double local_10;
  
                    /* 0x5120  1  ATBHandlerMain */
  if ((double)CONCAT44(_DAT_2dd152f4,DAT_2dd152f0) <= (double)(0x43238 - param_1) / _DAT_2dd152e8) {
    local_10 = (double)(0x43238 - param_1) / _DAT_2dd152e8;
  }
  else {
    local_10 = (double)CONCAT44(_DAT_2dd152f4,DAT_2dd152f0);
  }
  return (int)ROUND(local_10 * (double)param_2);
}
```

Looks like a very small patch? Guess you can get what it does by just executing and looking at what is at the memory adresses.

CherryDT:
It just patched the place where RM would normally do `atb += delta`

The source is:

```basic
	Function ATBHandlerMain(oldValue As Integer, delta As Integer) As Integer Export
		Return max(1, (275000 - oldValue) / 55000) * delta
	End Function
```

(I can show this since it belongs to me and I was paid only for _integrating_ it. See embedded license in the binary.)

fmatthew5876:
For posterity, in dynrpg version here are the important addresses:

| Addr | Comment |
| -- | -- |
| 0x495B69 | Sums the actor's AGI |
| 0x495BB9 | Sums the monster's AGI |
| 0x495C5E | Gets battler's AGI |
| 0x495C64 | Adds 1 to agi |
| 0x495C65 | Multiplies sum of agi by 100 |
| 0x495C6E | Divides 100 * sum / actor agi |
| 0x495C7A | Divides max_atb / previous value |
| 0x495C87 | Call which does some stuff ? and then assigns the atb value to the battler before returning |

fmatthew5876:
On the subject of rounding modes:

This code:
```
#include <cmath>

int f(double d) {
    return static_cast<int>(d);
}

int g(double d) {
    return std::round(d);
}

int h(double d) {
    return std::nearbyint(d);
}

int i(double d) {
    return std::lrint(d);
}
```

When compiled with `i686-w64-mingw32-g++ -march=i686 test.c  -c -O3 -fno-math-errno`

Produces this:
```
test.o:     file format pe-i386


Disassembly of section .text:

00000000 <__Z1fd>:
   0:   83 ec 0c                sub    $0xc,%esp
   3:   d9 7c 24 06             fnstcw 0x6(%esp)
   7:   dd 44 24 10             fldl   0x10(%esp)
   b:   0f b7 44 24 06          movzwl 0x6(%esp),%eax
  10:   66 0d 00 0c             or     $0xc00,%ax                                                                                                                                                                                              14:   66 89 44 24 04          mov    %ax,0x4(%esp)                                                                                                                                                                                           19:   d9 6c 24 04             fldcw  0x4(%esp)                                                                                                                                                                                               1d:   db 1c 24                fistpl (%esp)                                                                                                                                                                                                  20:   d9 6c 24 06             fldcw  0x6(%esp)                                                                                                                                                                                               24:   8b 04 24                mov    (%esp),%eax                                                                                                                                                                                             27:   83 c4 0c                add    $0xc,%esp                                                                                                                                                                                               2a:   c3                      ret                                                                                                                                                                                                            2b:   90                      nop                                                                                                                                                                                                            2c:   8d 74 26 00             lea    0x0(%esi,%eiz,1),%esi                                                                                                                                                                                                                                                                                                                                                                                                                              00000030 <__Z1gd>:                                                                                                                                                                                                                             30:   83 ec 2c                sub    $0x2c,%esp                                                                                                                                                                                              33:   dd 44 24 30             fldl   0x30(%esp)                                                                                                                                                                                              37:   dd 1c 24                fstpl  (%esp)                                                                                                                                                                                                  3a:   e8 00 00 00 00          call   3f <__Z1gd+0xf>                                                                                                                                                                                         3f:   d9 7c 24 1e             fnstcw 0x1e(%esp)                                                                                                                                                                                              43:   0f b7 44 24 1e          movzwl 0x1e(%esp),%eax                                                                                                                                                                                         48:   66 0d 00 0c             or     $0xc00,%ax                                                                                                                                                                                              4c:   66 89 44 24 1c          mov    %ax,0x1c(%esp)                                                                                                                                                                                          51:   d9 6c 24 1c             fldcw  0x1c(%esp)                                                                                                                                                                                              55:   db 5c 24 18             fistpl 0x18(%esp)                                                                                                                                                                                              59:   d9 6c 24 1e             fldcw  0x1e(%esp)                                                                                                                                                                                              5d:   8b 44 24 18             mov    0x18(%esp),%eax                                                                                                                                                                                         61:   83 c4 2c                add    $0x2c,%esp                                                                                                                                                                                              64:   c3                      ret                                                                                                                                                                                                            65:   8d 74 26 00             lea    0x0(%esi,%eiz,1),%esi                                                                                                                                                                                   69:   8d bc 27 00 00 00 00    lea    0x0(%edi,%eiz,1),%edi                                                                                                                                                                                                                                                                                                                                                                                                                              00000070 <__Z1hd>:                                                                                                                                                                                                                             70:   83 ec 2c                sub    $0x2c,%esp                                                                                                                                                                                              73:   dd 44 24 30             fldl   0x30(%esp)                                                                                                                                                                                              77:   dd 1c 24                fstpl  (%esp)                                                                                                                                                                                                  7a:   e8 00 00 00 00          call   7f <__Z1hd+0xf>                                                                                                                                                                                         7f:   d9 7c 24 1e             fnstcw 0x1e(%esp)                                                                                                                                                                                              83:   0f b7 44 24 1e          movzwl 0x1e(%esp),%eax                                                                                                                                                                                         88:   66 0d 00 0c             or     $0xc00,%ax                                                                                                                                                                                              8c:   66 89 44 24 1c          mov    %ax,0x1c(%esp)                                                                                                                                                                                          91:   d9 6c 24 1c             fldcw  0x1c(%esp)                                                                                                                                                                                              95:   db 5c 24 18             fistpl 0x18(%esp)                                                                                                                                                                                              99:   d9 6c 24 1e             fldcw  0x1e(%esp)                                                                                                                                                                                              9d:   8b 44 24 18             mov    0x18(%esp),%eax                                                                                                                                                                                         a1:   83 c4 2c                add    $0x2c,%esp                                                                                                                                                                                              a4:   c3                      ret                                                                                                                                                                                                            a5:   8d 74 26 00             lea    0x0(%esi,%eiz,1),%esi                                                                                                                                                                                   a9:   8d bc 27 00 00 00 00    lea    0x0(%edi,%eiz,1),%edi                                                                                                                                                                                                                                                                                    

000000b0 <__Z1id>:                                                                                                                                                                                                                             b0:   83 ec 0c                sub    $0xc,%esp                                                                                                                                                                                               b3:   dd 44 24 10             fldl   0x10(%esp)                                                                                                                                                                                              b7:   db 5c 24 04             fistpl 0x4(%esp)                                                                                                                                                                                               bb:   8b 44 24 04             mov    0x4(%esp),%eax
  bf:   83 c4 0c                add    $0xc,%esp                                                                                                                                                                                               c2:   c3                      ret                                                                                                                                                                                                            c3:   90                      nop                                                                                                                                                                                                            c4:   90                      nop                                                                                                                                                                                                            c5:   90                      nop                                                                                                                                                                                                            c6:   90                      nop                                                                                                                                                                                                            c7:   90                      nop                                                                                                                                                                                                            c8:   90                      nop                                                                                                                                                                                                            c9:   90                      nop                                                                                                                                                                                                            ca:   90                      nop                                                                                                                                                                                                            cb:   90                      nop                                                                                                                                                                                                            cc:   90                      nop                                                                                                                                                                                                            cd:   90                      nop                                                                                                                                                                                                            ce:   90                      nop                                                                                                                                                                                                            cf:   90                      nop                                      
```

The ATBHandler code looks like this:
```
2DD05460 | 55                            | push ebp                                             |
2DD05461 | 89E5                          | mov ebp,esp                                          |
2DD05463 | 83EC 0C                       | sub esp,C                                            |
2DD05466 | C745 FC 00000000              | mov dword ptr ss:[ebp-4],0                           |
2DD0546D | B8 38320400                   | mov eax,43238                                        | 275000
2DD05472 | 2B45 08                       | sub eax,dword ptr ss:[ebp+8]                         | 275000 - cur atb
2DD05475 | 50                            | push eax                                             |
2DD05476 | DB0424                        | fild st(0),dword ptr ss:[esp]                        | convert diff to double
2DD05479 | 83C4 04                       | add esp,4                                            |
2DD0547C | DC35 E862D12D                 | fdiv st(0),qword ptr ds:[<55000.0>]                  | divde by 55000.0
2DD05482 | DC1D F062D12D                 | fcomp st(0),qword ptr ds:[<1.0>]                     |
2DD05488 | DFE0                          | fnstsw ax                                            |
2DD0548A | 9E                            | sahf                                                 |
2DD0548B | 73 14                         | jae ultimate_rt_eb.2DD054A1                          |
2DD0548D | FF35 F062D12D                 | push dword ptr ds:[<1.0>]                            |
2DD05493 | FF35 F462D12D                 | push dword ptr ds:[2DD162F4]                         |
2DD05499 | 8F45 F8                       | pop dword ptr ss:[ebp-8]                             |
2DD0549C | 8F45 F4                       | pop dword ptr ss:[ebp-C]                             | Replace with 1.0 on stack
2DD0549F | EB 18                         | jmp ultimate_rt_eb.2DD054B9                          |
2DD054A1 | B8 38320400                   | mov eax,43238                                        | 275000
2DD054A6 | 2B45 08                       | sub eax,dword ptr ss:[ebp+8]                         | 275000 - cur_atb
2DD054A9 | 50                            | push eax                                             |
2DD054AA | DB0424                        | fild st(0),dword ptr ss:[esp]                        | to double
2DD054AD | 83C4 04                       | add esp,4                                            |
2DD054B0 | DC35 E862D12D                 | fdiv st(0),qword ptr ds:[<55000.0>]                  | divide by 55000.0
2DD054B6 | DD5D F4                       | fstp qword ptr ss:[ebp-C],st(0)                      | Put result on stack
2DD054B9 | DD45 F4                       | fld st(0),qword ptr ss:[ebp-C]                       | Load back into register
2DD054BC | DA4D 0C                       | fimul st(0),dword ptr ss:[ebp+C]                     | Multiply by original ATB increment
2DD054BF | DB5D FC                       | fistp dword ptr ss:[ebp-4],st(0)                     | convert to integer
2DD054C2 | 8B45 FC                       | mov eax,dword ptr ss:[ebp-4]                         |
2DD054C5 | 89EC                          | mov esp,ebp                                          |
2DD054C7 | 5D                            | pop ebp                                              |
2DD054C8 | C2 0800                       | ret 8                                                |
```

So `std::lrint()` is what we want.

CherryDT:
Ohh, ok, right, there are two different things at play. RM is written in Delphi as you know, but my DLL is written in FreeBasic.

fmatthew5876:
Heres another minor visual bug with ATB gauges 

gauge

left player, right rpg_rt

Ghabry:
looks like wrong draw order?

fmatthew5876:
Draw order is ok, the pixels are slightly off. The gauge spills over the edges.

Ghabry:
The only difference between std::round and std::rint is:

>FE_TONEAREST, then std::rint differs from std::round in that halfway cases are rounded to even rather than away from zero. 

So the only cases where you get different results are 0.5, 2.5, 4.5, ... etc.

fmatthew5876:
At least on 32 bit intel, there is a non-trivial performance difference. The `std::lrint()` is a few instructions, wheras `round()` is a function call.

It looks like on modern cpu's the difference is not as noticable.

Here it is on x86_64 `g++ test.c -c -O3 -fno-math-errno`

```
0000000000000000 <_Z1fd>:
   0:	f2 0f 2c c0          	cvttsd2si %xmm0,%eax
   4:	c3                   	retq   
   5:	90                   	nop
   6:	66 2e 0f 1f 84 00 00 	nopw   %cs:0x0(%rax,%rax,1)
   d:	00 00 00 

0000000000000010 <_Z1gd>:
  10:	48 83 ec 08          	sub    $0x8,%rsp
  14:	e8 00 00 00 00       	callq  19 <_Z1gd+0x9>
  19:	48 83 c4 08          	add    $0x8,%rsp
  1d:	f2 0f 2c c0          	cvttsd2si %xmm0,%eax
  21:	c3                   	retq   
  22:	0f 1f 40 00          	nopl   0x0(%rax)
  26:	66 2e 0f 1f 84 00 00 	nopw   %cs:0x0(%rax,%rax,1)
  2d:	00 00 00 

0000000000000030 <_Z1hd>:
  30:	48 83 ec 08          	sub    $0x8,%rsp
  34:	e8 00 00 00 00       	callq  39 <_Z1hd+0x9>
  39:	48 83 c4 08          	add    $0x8,%rsp
  3d:	f2 0f 2c c0          	cvttsd2si %xmm0,%eax
  41:	c3                   	retq   
  42:	0f 1f 40 00          	nopl   0x0(%rax)
  46:	66 2e 0f 1f 84 00 00 	nopw   %cs:0x0(%rax,%rax,1)
  4d:	00 00 00 

0000000000000050 <_Z1id>:
  50:	f2 48 0f 2d c0       	cvtsd2si %xmm0,%rax
  55:	c3                   	retq 
```

`std::lrint()` is slightly better, but not by much.

src/game_battle.cpp
```diff
@@ -232,25 +233,36 @@ void Game_Battle::NextTurn(Game_Battler* battler) {
...
+	sum_agi *= 100;
+
+	const int max_atb = Game_Battler::GetMaxAtbGauge();
+
+	for (auto* bat: battlers) {
+		// RPG_RT always updates atb for non-hidden enemies, even if they can't act.
+		if (bat->Exists() && (bat->CanAct() || bat->GetType() == Game_Battler::Type_Enemy))
+		{
+			const auto agi = bat->GetAgi();
+			auto increment = max_atb / (sum_agi / (agi + 1));
```
Ghabry, line 259: Thats one of the cases where I would discourage the use of "auto". The return type of GetAgi is hidden and the type of increment is unknown (int or float)

fmatthew5876: You can look at it 2 ways. On one hand the type is hidden. On the other hand, this expression works cleanly no matter what the type of agi is. If agi was an int, it works, if it was a double it also works.

If you made it `const int agi = bat->GetAgi()` and then the battler agi changed to double, you would get a silent cast. That's arguably a bug and a hard to find one at that.

It's trivial either way for this case.

fmatthew5876:
Only type C battles had gauge rendered improperly. I've fixed that, but now I see another issue. The way we stretch the middle part of the gauge is not the same say the way RPG_RT does it. This is most obvious with the System2C, where the right edge of the gauge is supposed to roll off diagonally. In player, you can see we stretch it wrong.

gauge2

fmatthew5876:
Those triangular gauges in the last post exposed another visual bug. I pushed a fix for it too.

fmatthew5876:
The last pieces of this are gauge initialization

| Battle Type | Monster Gauge | Actor Gauge | Actor w/w pre-emptive weapon |
| -- | -- | -- | -- |
| Normal | max / 2 | max / 2 | max |
| Normal + First Strike | 0 | max | max |
| Surprise | max | 0 | 0 |
| Back | 0 | max | max |
| Pincer | max | 0 | 0 |
| Preemptive | 0 | max | max |

This is easy to test. Make one hero and one enemy with both same AGI. Set the hero to auto battle.
Add a battle event with aborts the battle after 4-5 rounds.

fmatthew5876:
I've added ATB gauge initialization and tested it for all situations in the table above.

The stretching difference in the middle gauge has to do with our stretch blit is out of scope for this PR.

This PR is done

fdelapena:
> The stretching difference in the middle gauge has to do with our stretch blit is out of scope for this PR. I believe we should have implemented all aspects of ATB gauge now.

This one looks like #446, so this point could be managed there.

fmatthew5876:
I found a weird edge case which I'm not sure if it's a bug or not.

If you inflict a state which has a do nothing restriction and also no automatic recovery, that actors gauge will stay frozen between battles. It won't reset in the next battle.

The gauge state is not saved. So if you save and load your game, it resets back to 0. Then having the state, it stays zero each battle until the state is cured.

I can see it in the RPG_RT discompile. It literally says if the actor has a unrecoverable state with do nothing restriction, skip the whole ATB initialization.

Ghabry:
Is there a reason why std::lrint is hidden behind a Utility function?

fmatthew5876:
> Is there a reason why std::lrint is hidden behind a Utility function?

The idea was to provide an RPG_RT compatible rounding function. Otherwise the temptation is to to use `std::round()` which doesn't behave the same way.

I can remove it if we don't like this.

Ghabry:
Considering that lrint gives a better distribution for .5 roundings (up/down depending on even/odd) another option is using this function everywhere hidden behind Utils::Round

fmatthew5876:
I added `Utils::Round()`, but it looks like in our code we currently don't actually do any float to float rounding. So the method is unused.

Ghabry:
I meant renaming ``Utils::RoundToInt`` to ``Utils::Round`` and using this everywhere. Or do you consider this a bad idea?

fmatthew5876:
I'd rather be explicit that a type cast is happening, also you need to be able to choose which integer type you want. Unless we make it `template <typename T> Round()` so you have to say `Round<int>(x)`.

Ghabry:
The Gauge feels now really authenthic. Is now the same piece of crap everybody hates in 2k3. :)

3zxaic

src/game_battler.cpp
```diff
@@ -77,13 +78,18 @@ RPG::State::Restriction Game_Battler::GetSignificantRestriction() const {
...
-	for (int i = 0; i < (int)states.size(); i++) {
-		// States are guaranteed to be valid
-		const RPG::State* state = ReaderUtil::GetElement(Data::states, states[i]);
-		if (state->restriction == RPG::State::Restriction_do_nothing) {
-			return false;
+bool Game_Battler::CanActImpl(int max_arp) const {
+	const auto& states = GetStates();
+	for (size_t i = 0; i < states.size(); ++i) {
+		if (states[i] > 0) {
+			const auto* state = ReaderUtil::GetElement(Data::states, i + 1);
```
Ghabry, line 85: The comment ``// States are guaranteed to be valid`` was lost and you added a check for state validity.
(and the check only here is inconsequent. Should be everywhere or nowhere). And I prefer nowhere because the states are sanitized.

src/game_battler.cpp
```diff
@@ -77,13 +78,18 @@ RPG::State::Restriction Game_Battler::GetSignificantRestriction() const {
...
+			if (state
+					&& state->restriction == RPG::State::Restriction_do_nothing
+					&& state->auto_release_prob <= max_arp
```
Ghabry, line 88: Imo ``CanActOrRecoverable`` in confusing because it does two completely different things in one function. In one case it decides about ``Can Act right now`` and in the other case it decides about ``Can currently not act, but has a recoverable state``.

So these should be two distinct functions. Passing in 0 or max to get two completely different things makes this hard to understand (I had to think for 10 minutes about why this func works).

fmatthew5876: Is it not better to have one function for this instead of having to reproduce something like `CanAct() || !HasUnrecoverableState()` in all places this is used?

Grepping for this in the code:
```
game_battle.cpp
162:            if (!actor->IsHidden() && actor->CanActOrRecoverable()) {

game_battler.h
126:    bool CanActOrRecoverable() const;
898:inline bool Game_Battler::CanActOrRecoverable() const {

game_battler.cpp
662:    // the gauge and carry it between battles if !CanActOrRecoverable().

scene_battle_rpg2k3.cpp
55:     if (battler.IsHidden() || !battler.CanActOrRecoverable()) {
```

And it's used another time in the placement PR.

The private `CanActImpl()` method is a bit hard to read. I did it that way to avoid copy and paste, but I can split these 2 methods out if that makes it easier.

src/game_battler.cpp
fmatthew5876: I agree, will replace the comment and the if with `assert()`

src/game_battler.cpp
```diff
@@ -78,12 +79,28 @@ RPG::State::Restriction Game_Battler::GetSignificantRestriction() const {
...
+	return true;
+}
+
+bool Game_Battler::CanActOrRecoverable() const {
+	const auto& states = GetStates();
+	for (size_t i = 0; i < states.size(); ++i) {
+		if (states[i] > 0) {
+			const auto* state = ReaderUtil::GetElement(Data::states, i + 1);
+			assert(state);
+			if (state->restriction == RPG::State::Restriction_do_nothing && state->auto_release_prob == 0) {
```
Ghabry, line 101: After sleeping about this the function makes sense: The idea is to handle "Can act right now or will be able to act again without external help in the future".

CherryDT:
> 0x495C87 Call which does some stuff ? and then adds the previous value to the battler's atb

This function is `TLcfgActor * __fastcall GetPartyMember(TLcfgPartyInfo * partyInfo, int id)`

fmatthew5876:
Would like to start using `RoundTo` in other PRs without creating a mess. Can we look at this one soon?

Issue #2134, opened 2020-04-10
More accurate battle animation shake


fmatthew5876:
With dynrpg, we can perfect the screen and target shake feature from battle animations.

A quick dump shows some very strange behavior.

The values are strength, speed, x, y, and time
```
SHAKE SCREEN: 3 4 -4 0 30
SHAKE SCREEN: 3 4 -6 0 30
SHAKE SCREEN: 3 4 -6 0 30
SHAKE SCREEN: 3 4 -6 0 30
SHAKE SCREEN: 3 4 -6 0 30
SHAKE SCREEN: 3 4 -6 0 30
SHAKE SCREEN: 3 4 -6 0 30
SHAKE SCREEN: 3 4 -6 0 30
SHAKE SCREEN: 3 4 -6 0 30
SHAKE SCREEN: 3 4 -6 0 30
SHAKE SCREEN: 3 4 -6 0 29
SHAKE SCREEN: 3 4 -4 0 28
SHAKE SCREEN: 3 4 -1 0 27
SHAKE SCREEN: 3 4 2 0 26
SHAKE SCREEN: 3 4 5 0 25
SHAKE SCREEN: 3 4 7 0 24
SHAKE SCREEN: 3 4 5 0 23
SHAKE SCREEN: 3 4 2 0 22
SHAKE SCREEN: 3 4 -1 0 21
SHAKE SCREEN: 3 4 -4 0 20
SHAKE SCREEN: 3 4 -6 0 19
SHAKE SCREEN: 3 4 -6 0 18
SHAKE SCREEN: 3 4 -3 0 17
SHAKE SCREEN: 3 4 3 0 15
SHAKE SCREEN: 3 4 6 0 14
SHAKE SCREEN: 3 4 6 0 13
SHAKE SCREEN: 3 4 4 0 12
SHAKE SCREEN: 3 4 1 0 11
SHAKE SCREEN: 3 4 -2 0 10
SHAKE SCREEN: 3 4 -5 0 9
SHAKE SCREEN: 3 4 -7 0 8
SHAKE SCREEN: 3 4 -5 0 7
SHAKE SCREEN: 3 4 -2 0 6
SHAKE SCREEN: 3 4 1 0 5
SHAKE SCREEN: 3 4 4 0 4
SHAKE SCREEN: 3 4 6 0 3
SHAKE SCREEN: 3 4 6 0 2
SHAKE SCREEN: 3 4 3 0 1
[...]

PR #2136, opened 2020-04-10
More accurate battle animation shake

fmatthew5876:
RPG_RT causes the shake effect to stay locked for 10 frames.

Fix: #2134 

I still have some timing edge cases to work out.

Does anyone know exactly when in the main loop sequence the dynrpg `onFrame()` callback happens?


Ghabry:
Is there a reason for the close?

fmatthew5876:
I'm closing this as it's badly out of date. We can revisit again later.

Issue #1826, opened 2019-07-01
Captain Novolin RPG -Regression in 2k3 battles for battle animations


NovemberJoy:
Please fill in the following fields before submitting an issue:
Name of the game: Captain Novolin RPG

Download link: [Here](https://drive.google.com/open?id=1_0hS0zeV9QoAbik5LgERwGsPnFIAXB2N)

Player platform:
Windows 10 Home 1709 64-bit

Attach files (as a .zip archive or link them)
 - [easyrpg_log.txt](https://github.com/EasyRPG/Player/files/3344163/easyrpg_log.txt)

Describe the issue in detail and how to reproduce it:
Start the game and fast-forward through the introduction until the first battle starts. When it starts, all characters in the party are invisible and will not appear until their animation state changes, upon which their idle animation will begin playing properly. This persists throughout any battle the player enters.

Ghabry:
This sounds different to the animation bug in #1496.
If this is simple to fix I will do it, otherwise will be postboned to 0.5.2 as the 2k3 battle will be refactored.

PR #2137, opened 2020-04-10
Battle 2k3: Fix BattleAnim-Battlers being invisible before first action

Ghabry:
Call ResetZ before SetAnimationState is called. Otherwise BattleAnimation-Battlers will be invisible at the beginning of the battle.

Just a cosmetic thing.

~~Is there an open issue about this?~~ (Yes there is #1826). Was mentioned in the VD discord.


fmatthew5876:
Nice fix!

PR #2138, opened 2020-04-10
Wine: Disable Wow6432Node redirection for HKCU

Ghabry:
Wine incorrectly redirected HKCU, but is fixed by now.

This reverts c7d41c01


Ghabry:
Confirmed on my laptop: A fresh wine environment does not write in Wow6432, my old wine prefix does.

carstene1ns:
lol

PR #2139, opened 2020-04-10
Add safe math optimization flags

fmatthew5876:
This adds `-fno-math-errno` on gcc based platforms.

The C/C++ standard libraries have a stupid design where many of the standard math functions check for invalid input and set the global `errno`.

What could be a single instruction, turns into a branch and potential write to global memory. This is death for high performance float point. I've used `-fno-math-errno` in production critical codebases. Of all the `-ffast-math` optimizations this one is safe to use.

The reason it's not included in any of the `-ON` levels is because it's technically breaking compatibility if any of your code is actually checking errno.

Intel's compiler is the only one which enables `-fno-math-errno` by default. Probably so they can cheat and win benchmarks against people who don't know this optimization ;)

It looks like MSVC has some limited form of this with `/Oi` which is included in `/O2`

Consider this code:
```
double s(double x) {
    return std::sqrt(x);
}

long r(double x) {
    return std::lrint(x);
}
```

Compiled with `g++ -O3`
```
s(double):
        pxor    xmm2, xmm2
        movapd  xmm1, xmm0
        ucomisd xmm2, xmm0
        sqrtsd  xmm1, xmm1
        ja      .L8
        movapd  xmm0, xmm1
        ret
.L8:
        sub     rsp, 24
        movsd   QWORD PTR [rsp+8], xmm1
        call    sqrt
        movsd   xmm1, QWORD PTR [rsp+8]
        add     rsp, 24
        movapd  xmm0, xmm1
        ret
r(double):
        jmp     lrint
```

Compiled with `g++ -fno-math-errno`
```
s(double):
        sqrtsd  xmm0, xmm0
        ret
r(double):
        cvtsd2si        rax, xmm0
        ret
```

Try it yourself here: https://gcc.godbolt.org/


Ghabry:
sending this through all our optional PR builders...

Issue #2141, opened 2020-04-11
Investigate Troop::appear_randomly


fmatthew5876:
In liblcf we have a flag called `Troop::appear_randomly`.

In DynRPG, there is a `DBMonsterGroup::randomizeHidden` with a comment saying "50% chance for hidden flag"

I guess that means if you enable this flag, each monster has a 50% chance of being hidden?

There are no controls in the RPG Maker Editor for setting this.

- [x] Investigate if this flag actually does anything
- [x] Check if RPG_RT actually reads this from the database.
- [x] Check various versions of 2k and 2k3 to see which support it
- [x] Implement in Player if so
- [x] Does this effect gold / exp awards?

fmatthew5876:
Ok this is in 2k3 editor but hard to find.

You have to right click on the troop name in the troops list. Then there will be an "Appear Randomly" button you can select.

Ghabry:
See #987

fmatthew5876:
Well it may not truly be 50%

```
  FUN_00475584(param_1);
  *(int *)(param_1 + 0x24) = troop_id;
  num_not_hidden = 0;
  monsterGroup = LookupDBMonsterGroup(*(undefined **)&RPG::dbMonsterGroups,troop_id);
  num_monsters = FUN_00475714((int)monsterGroup->monsterList);
  if (-1 < num_monsters + -1) {
    mgp_id = 0;
    do {
      FUN_004755b8(param_1,mgp_id + 1);
      monGroup = LookupDBMonsterGroup(*(undefined **)&RPG::dbMonsterGroups,*(int *)(param_1 + 0x24))
      ;
      monsterGroupPosition = (MonsterGroupPosition *)FUN_0047a8dc((int)monGroup->monsterList,mgp_id)
      ;
      monster_pre = LookupMonster(param_1,mgp_id);
      FUN_004bd5f8((int *)monster_pre,monsterGroupPosition->monster_id);
      monster = LookupMonster(param_1,mgp_id);
      (monster->battler).notHidden = monsterGroupPosition->hidden ^ 1;
      if (((0 < num_not_hidden) &&
          (local_EAX_146 = LookupMonster(param_1,mgp_id),
          (local_EAX_146->battler).notHidden != false)) &&
         (pDVar1 = LookupDBMonsterGroup
                             (*(undefined **)&RPG::dbMonsterGroups,*(int *)(param_1 + 0x24)),
         pDVar1->randomizeHidden != false)) {
[...]

Issue #278, opened 2014-06-27
Automatic placement of battlers


Zegeri:
This is a RM2k3 feature that must be implemented.
Party members are placed automatically if BattleCommands.placement is False, but enemies are placed automatically if Troop.auto_alignment is True.

Open problems concerning placement are the automatic placements for battle modes in 2k3, e.g. first attack, pincer, ...

fdelapena:
Are battlers placed in x=0, y=0 in these cases? I've found some games with this behaviour.

Ghabry:
I actually started working on this feature some weeks ago. But during summer I prefer being outside instead of coding :).
Won't be ready before September.

Ghabry:
I implemented auto placement for normal attack now.
Manual placement should also work, but there are some games like OFF that have manual placement but the Y-coordinate is 0...
RPG2k3 places it correctly in mid-battle but no idea where it gets the info from, the automatic coordinate is also 0

Ghabry:
This is now semi-fixed by #851 
Open problems concerning placement are the automatic placements for battle moes in 2k3, e.g. first attack, pincer, ...

fmatthew5876:
Some notes about enemy placement in RPG_RT

Normal:
- [x] Enemies are placed as specified in editor

Initiative:
- [x] Enemies are placed as specified in editor
- [x] Enemies face left on start of battle, message, then they turn right

Surround:
- [x] Enemies are placed in a fixed column in the middle of the battle field
- [x] Odd numbered enemies (starting from 1) face left
- [x] Event numbered enemies (starting from 1) face right
- [ ] Unsure how Y positions are decided, seems to change often and not yet identifiable pattern
- [x] looks really ugly with many enemies, they all end up onto of on another.
- [x] With 1 actor, this behaves like an initiative battle 
    - [ ] Does the battle condition type change?
    - [x] Force Flee on party with "ignore if surrounded" works in this case
- [x] With 2 or more actors, it's a surround battle
- [x] If battle screen placement flag is "manual" behaves like Initiative
    - [ ] Does the battle condition type change?
- [x] If battle screen placement flag is "manual" this is disabled in battle test but not in event box

Back:
- [x] Enemies are placed as specified in editor, but flipped to the other side
- [x] Actors start facing left, then immediately turn around

Pincer:
- [x] Odd enemies on the right, even enemies on the left
- [x] Enemies automatically placed in order of id from top to bottom, and fixed Y positions
[...]

PR #2142, opened 2020-04-11
Battle 2k3 2: Implement 2k3 battler placement

fmatthew5876:
Depends on #2217 

Fix: #2141 
Fix: #278 

Battle Conditions
----------------
- [x] Convert surround to initiative when conditions not met
- [x] Convert pincer to back when conditions not met
    - [x] Handle hidden enemies
- [x] Use correct percentage for `Troop::appear_randomly`

Placement
----------
- [x] Manual enemy placement
- [x] Enemy placement for Normal / Initiative
- [x] Enemy placement for Back
- [x] Enemy placement for surround
- [x] Enemy placement for pincer
- [x] Double check edge cases for hidden enemies in pincer / surround
- [x] Verify clamping behavior on edges of screen for enemies - enemies clamp in Y only
- [x] Verify clamping behavior on edges of screen for enemies when they transform to different enemy - nothing happens
- [x] Manual actor placement
- [x] Actor placement for Normal / Initiative
- [x] Actor placement for Back
- [x] Actor placement for surround
- [x] Actor placement for pincer
- [x] Verify clamping behavior on edges of screen for actors
- [x] Verify clamping behavior on edges of screen for actors if placed on the edge and back row pushes them out of the screen

Facing Direction
------------------
- [x] Enemy initial facing direction on battle start
- [x] Actor initial facing direction on battle start
- [x] Enemies turn to face actors once when battle begins
    - [x] Test complex manual placement
- [x] Actors turn to face enemies once when battle begins
    - [x] Test complex manual placement
- [x] Actor changes direction to attack enemy
- [x] Actor changes direction to use skill on enemy
- [x] Actor does not change direction to use skill on ally
- [x] Action does not change direction on action which targets all enemies
- [x] Enemies never change direction after battle startup

Row
----
- [x] Logic which forces back actors to the front row if all front row actors incapacitated at start of battle.
- [x] Full row implementation TDB in later PR

Things to investigate
- [x] Placement depends on battle sprite sizes. Emscripten async issues?
- [ ] Check how animation battle sprites (i.e.) "large battlers" behave

### DIFF
--- a/src/window_base.cpp
+++ b/src/window_base.cpp
@@ -310,7 +310,6 @@ void Window_Base::DrawGauge(const Game_Battler& actor, int cx, int cy) const {
 	}
 
 	bool full = actor.IsAtbGaugeFull();
-	int gauge_w = 25 * actor.GetAtbGauge() / actor.GetMaxAtbGauge();
 
 	// Which gauge (0 - 2)
 	int gauge_y = 32 + 2 * 16;
@@ -320,15 +319,18 @@ void Window_Base::DrawGauge(const Game_Battler& actor, int cx, int cy) const {
 	Rect gauge_center(16, gauge_y, 16, 16);
 	Rect gauge_right(32, gauge_y, 16, 16);
 
-	// Full or not full bar
-	Rect gauge_bar(full ? 64 : 48, gauge_y, 16, 16);
-
 	Rect dst_rect(cx + 16, cy, 25, 16);
-	Rect bar_rect(cx + 16, cy, gauge_w, 16);
 
 	contents->Blit(cx + 0, cy, *system2, gauge_left, 255);
 	contents->Blit(cx + 16 + 25, cy, *system2, gauge_right, 255);
-
 	contents->StretchBlit(dst_rect, *system2, gauge_center, 255);
-	contents->StretchBlit(bar_rect, *system2, gauge_bar, 255);
+
+	const auto atb = actor.GetAtbGauge();
+	if (atb > 0) {
+		// Full or not full bar
+		int gauge_w = 25 * atb / actor.GetMaxAtbGauge();
+		Rect gauge_bar(full ? 64 : 48, gauge_y, 16, 16);
+		Rect bar_rect(cx + 16, cy, gauge_w, 16);
+		contents->StretchBlit(bar_rect, *system2, gauge_bar, 255);
+	}
 }
--- a/src/window_base.cpp
+++ b/src/window_base.cpp
@@ -326,9 +326,9 @@ void Window_Base::DrawGauge(const Game_Battler& actor, int cx, int cy) const {
 	contents->StretchBlit(dst_rect, *system2, gauge_center, 255);
 
 	const auto atb = actor.GetAtbGauge();
-	if (atb > 0) {
+	const auto gauge_w = 25 * atb / actor.GetMaxAtbGauge();
+	if (gauge_w > 0) {
 		// Full or not full bar
-		int gauge_w = 25 * atb / actor.GetMaxAtbGauge();
 		Rect gauge_bar(full ? 64 : 48, gauge_y, 16, 16);
 		Rect bar_rect(cx + 16, cy, gauge_w, 16);
 		contents->StretchBlit(bar_rect, *system2, gauge_bar, 255);
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -186,6 +186,7 @@ add_library(${PROJECT_NAME}
 	src/platform_clock.h
 	src/player.cpp
 	src/player.h
+	src/point.h
 	src/rect.cpp
 	src/rect.h
 	src/registry.h
--- a/Makefile.am
+++ b/Makefile.am
@@ -179,6 +179,7 @@ libeasyrpg_player_a_SOURCES = \
 	src/platform_clock.h \
 	src/player.cpp \
 	src/player.h \
+	src/point.h \
 	src/game_quit.cpp \
 	src/game_quit.h \
 	src/rect.cpp \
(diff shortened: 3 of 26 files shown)


fmatthew5876:
Functionality wise, I'm cutting this PR here. We now handle battler placement, facing direction, and row initialization.

The rest of row implementation will be in a follow up PR.

Once the dependencies are merged I'll rebase this again and retest.

Ghabry:
Can you give some addresses? Want to verify this asm-wise.

fmatthew5876:
DynRPG Info:

| Address | Class | Name |
| -- | -- | -- |
| 004977A4 | SceneBattle | Init() - parent function which does battle initialization sequence |
| 004B4910 | SceneBattle | InitActors() - initializes actor position and mirrored flag |
| 004BD1D0 | SceneBattle | InitActorFacing() - conditions for monster starting facing direction |
| 004BD1E8 | SceneBattle | ActorShouldMirror() - determine if actor should mirror |
| 004BE8E0 | SceneBattle | InitMonsters() - filters start conditions and initialize monster positions /mirror |
| 0047FD70 | SceneBattle | InitMonstorFacing() - conditions for monster starting facing direction |

| 004B4B6C | SceneBattle | ResetActorDirections() - resets the actors facing at end of battle init sequence - this is what causes them to turn |
| 004BED2C | SceneBattle | ResetMonsterDirections() - resets the monsters facing at end of battle init sequence |

Ghabry:
When ``DrawGauge -> gauge_w`` is 0 this results in a ``zoom_x`` of infinity. Caused by previous PR:

Ubsan reports: 
```
src/transform.h:43:4: runtime error: inf is outside the range of representable values of type 'int'
    #0 0x564b0e5fe29b in Transform::Scale(double, double) src/transform.h:43:4
    #1 0x564b0e5e37e4 in Bitmap::StretchBlit(Rect const&, Bitmap const&, Rect const&, Opacity const&) src/bitmap.cpp:651:20
    #2 0x564b0e521cac in Window_Base::DrawGauge(Game_Battler const&, int, int) const src/window_base.cpp:333:12
    #3 0x564b0ed64656 in Window_BattleStatus::RefreshGauge() src/window_battlestatus.cpp:155:5
    #4 0x564b0ed61546 in Window_BattleStatus::Refresh() src/window_battlestatus.cpp:94:2
    #5 0x564b0ebd2306 in Scene_Battle_Rpg2k3::SetupSystem2Graphics() src/scene_battle_rpg2k3.cpp:167:23
    #6 0x564b0ebd1399 in Scene_Battle_Rpg2k3::OnSystem2Ready(FileRequestResult*) src/scene_battle_rpg2k3.cpp:150:2
```

fmatthew5876:
One thing I haven't resolved is how to handle async io for web.

We have this sequence of events:
1. Create battler sprites
2. Compute battler positions based on those sprite dimensions
3. Recompute battler sprite Z values based on new battler Y values
4. Draw transition in frame containing all the sprites battlers and their Z values.
5. Perform transition in from black.

This means we have an async I/O dependency between (1) and (2). 

Also, it looks like we always have this issue with transitions. A scene start function will create the sprites (4), and then immediately TransitionIn() (4) is called in the same frame, which tries to render the transition in frame before the sprites may be loaded.

Am I overlooking something. How is it that our transition in effects on emscripten don't always look like crap?

It seems like at a minimum at least we could always wait for all graphics files when the screen is erased. A few extra frames spent spinning on a black screen won't be noticed as much as objects popping into the scene at random.

Even if we do that, I need another async hook to divide steps (1) and (2) above.

Thoughts?

Ghabry:
At least the TransitionIn is delayed when any files are pending, e.g. when loading a new map. So most assets are already there before the transisionIn happens so they look correct. Didn't you add this "wait for In" stuff? Is is broken right now?

When the position computation depends on the dimension this needs to be delayed until the async request went through then, yes.

fmatthew5876:
> At least the TransitionIn is delayed when any files are pending, e.g. when loading a new map. So most assets are already there before the transisionIn happens so they look correct. Didn't you add this "wait for In" stuff? Is is broken right now?

This is not the case. The `screen2` for transition in is snapped in `Transition::Init()` which occurs at the same frame as `Scene::Start()`. 

We need to wait for all graphics to finish loading, and then capture `screen2`. 

This can be seen very easily with your `EP_DEBUG_SIMULATE_ASYNC` define.

#2217 is trying to fix this.

fmatthew5876:
I've rebased this onto #2217 and added the async code to make placement worth correctly with async I/O.

This is ready once #2217 goes in

src/window_base.cpp
```diff
@@ -320,15 +319,18 @@ void Window_Base::DrawGauge(const Game_Battler& actor, int cx, int cy) const {
...
 	contents->Blit(cx + 0, cy, *system2, gauge_left, 255);
 	contents->Blit(cx + 16 + 25, cy, *system2, gauge_right, 255);
-
 	contents->StretchBlit(dst_rect, *system2, gauge_center, 255);
-	contents->StretchBlit(bar_rect, *system2, gauge_bar, 255);
+
+	const auto atb = actor.GetAtbGauge();
+	if (atb > 0) {
+		// Full or not full bar
+		int gauge_w = 25 * atb / actor.GetMaxAtbGauge();
```
Ghabry, line 331: This still div/0 in StretchBlit: gauge_w can become 0 here.
Then zoom_x is inf: ``src_rect.width  / dst_rect.width``

fmatthew5876: Are you sure? This logic is skipped if atb is 0 and `GetMaxAtbGauge()` is a constexpr.

Ghabry: 25 * 280 / 300000 is still 0

fmatthew5876: 🤦

PR #2146, opened 2020-04-12
SDL Audio: Fix Lock/UnlockMutex. 

Ghabry:
Our Mutex locking never worked :o

They always operated on the wrong audio device because SDL_LockAudio was used and not SDL_LockAudioDevice.

This was never noticed before because all the decoder destruction was in the Audio thread but now the BGM can be destroyed by the main thread.

But no idea how so many Android users encountered this, I had to add an artificial delay of 100ms and then it still didn't crash that often. But there are 20000 users and only 20 encountered it, so you have to be kinda unlucky :)

### DIFF
--- a/src/audio_sdl.cpp
+++ b/src/audio_sdl.cpp
@@ -80,7 +80,7 @@ SdlAudio::SdlAudio() :
 	want.userdata = this;
 
 #if SDL_MAJOR_VERSION >= 2
-	audio_dev_id = SDL_OpenAudioDevice(NULL, 0, &want, &have, SDL_AUDIO_ALLOW_FREQUENCY_CHANGE);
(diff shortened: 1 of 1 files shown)


carstene1ns:
Open question is, why do these android devices have multiple audio devices? Because it should likely work if you only have one.

Ghabry:
The legacy APIs that don't have a device ID like SDL_Lock/UnlockAudio operate on Device ID 1 but SDL_OpenAudioDevice always return a value >1.
So the mutex locking never worked anywhere. My change just increased the probability to hit the bug alot.

carstene1ns:
ah ok. So the problem was us using incompatible apis :1st_place_medal:

Ghabry:
Even after this update there are still crashes in memcpy in AudioResampler on Android -_-
But now I'm really clueless. The code is correct this way.

Also tons of new crashes in "libGLES" that never happened before. This makes also no sense at all. Why does the GPU driver crash now everywhere?

There are no known memory corruptions anymore according to ASAN...

PR #2149, opened 2020-04-13
Message regression fixes

Ghabry:
1st commit is unrelated to messages, detected by UbSan

2nd fixes a regression introduced by moving the Scene::Pop to a new location while removing Game_Temp: https://github.com/EasyRPG/Player/commit/993ef100e106c8f3d9845b92f29e8c0d3df11a77#diff-ad788c966e27c5a83ba8ac104b2b3c73R117-R121 (no idea why it was moved though)

3rd fixes \n[0]. Was broken during refactoring.
https://github.com/EasyRPG/Player/blob/0.6.1/src/window_message.cpp#L589-L605

### DIFF
--- a/src/game_message.cpp
+++ b/src/game_message.cpp
@@ -17,6 +17,8 @@
 
 // Headers
 #include "game_message.h"
+#include "game_actor.h"
+#include "game_party.h"
 #include "game_player.h"
 #include "game_battle.h"
 #include "main_data.h"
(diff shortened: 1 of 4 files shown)


src/game_message.cpp
```diff
@@ -324,10 +326,12 @@ static Game_Message::ParseParamResult ParseParamImpl(
...
 	}
 
-	// RPG_RT will replace varible substitutions that result in 0
-	// with 1 to avoid invalid actor crash.
+	// Actor 0 references the first party member
 	if (upper == 'N' && value == 0 && got_valid_number) {
-		value = 1;
+		auto* party = Main_Data::game_party.get();
+		if (party->GetBattlerCount() > 0) {
+			value = (*party)[0].GetId();
```
fmatthew5876, line 333: Nice catch! All of my tests had Alex as the first party member so I missed this one.

What does it do if the party is empty and you ask for `\n[0]`?

Ghabry: I tested this :). Gives the typical "Invalid party members specified". I implemented this by falling through with 0 which gives our typical "Invalid actor" warning.

Ghabry:
I removed the hero names because they are unused???

Down the rabbit hole. Took longer to fix then the rest...: Game_Party requires get Game_Actors to get the party members and this depends on Game_Actor. When Game_Actor was not created yet it calls Init which does a sanity check e.g. for a valid Level. When the level is wrong it creates a warning message that segfaults because there is no MessageOverlay \o/

PR #2151, opened 2020-04-14
2k3 Battle: Implement Escape Algorithm

fmatthew5876:
Depends on #2142 

This implements only the escape allowed and success fail algorithm. Implementing animations and state machine logic around escaping is out of scope.


Ghabry:
Just to have this on Github: As discussed in IRC the merge speed of PRs should be improved (multiple were stuck for months and more PRs means more conflicts), so we agreed on a new rule:

1. A single "needs improvements" review infinitly delays the PR until the problem is resolved
2. When two _accepted_ reviews are in there is now a 1 week period for further reviews.
3. When there is no response during 1 week the PR will be merged.
4. The pending reviewers can ask, without giving a reason, multiple times for a 1-week extension. With reason (vacation, very complex PR), longer extensions are possible.

Ghabry:
I backported this one now to stable - Is a useful addition because it is an algorithmic fix.
Is only one commit and only touches Escape code so was easy to backport (but won't backport Grid, is too complex and more a visual problem)

PR #2152, opened 2020-04-14
Resampler: Fix buffer overflow in mono_to_stereo_resample

Ghabry:
...  when the amount of samples read was exactly 0.

This should be last problem with the audio code...

The check didn't work because of size_t instead of int.

This is also a year old bug btw. In the old code the mono-to-stereo code was in SeCache where always the entire SE was decoded -> Buffer was never empty. :/

### DIFF
--- a/src/audio_resampler.cpp
+++ b/src/audio_resampler.cpp
@@ -385,7 +385,7 @@ int AudioResampler::FillBuffer(uint8_t* buffer, int length) {
 		int sample_size = AudioDecoder::GetSamplesizeForFormat(output_format);
 
 		// Duplicate data from the back, allows writing to the buffer directly
-		for (size_t i = amount_filled - sample_size; i > 0; i -= sample_size) {
+		for (int i = amount_filled - sample_size; i > 0; i -= sample_size) {
 			// left channel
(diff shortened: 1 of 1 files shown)


src/audio_resampler.cpp
```diff
@@ -385,7 +385,7 @@ int AudioResampler::FillBuffer(uint8_t* buffer, int length) {
 		int sample_size = AudioDecoder::GetSamplesizeForFormat(output_format);
 
 		// Duplicate data from the back, allows writing to the buffer directly
-		for (size_t i = amount_filled - sample_size; i > 0; i -= sample_size) {
+		for (int i = amount_filled - sample_size; i > 0; i -= sample_size) {
```
carstene1ns, line 388: Why not modifying above (line 380) to `<= 0` ?

Ghabry: That's a good question (and this was my first fix):
Because of the memset below:
```cpp

carstene1ns: ...but this is never reached because of the return, no?

Ghabry: Wait. Yeah this memset can be removed because the Decode function in AudioDecoder (which calls FillBuffer) handles the memset already... lemme fix this

PR #2160, opened 2020-04-15
First part of the platform restructure

carstene1ns:
This only moves files around. Everything should work as before.

TBD:
- [ ] proper cmake integration (currently not reachable, but added for later use)
- [ ] deprecation of legacy Makefiles
- [ ] drop double `3ds/3ds_` and such prefix (currently not possible since dkp Makefiles need distinct file names)

This is not intended to work on Wii, OpenDingux and Amiga currently, as these require SDL1.2 and our CMakery should be the future.
For Wii, I will write a custom UI (so it will work like 3DS, Switch and PSVita then) and the others can stay at there legacy Makefile for the time being.


carstene1ns:
I would like to have this in for 0.6.2 as this saves compilation time by not including empty files for most platforms.

CMakeLists.txt
```diff
@@ -413,22 +394,56 @@ if(${PLAYER_TARGET_PLATFORM} STREQUAL "SDL2")
...
+		src/platform/psvita/psp2_ui.h)
+	target_compile_definitions(${PROJECT_NAME} PUBLIC PSP2)
+elseif(${PLAYER_TARGET_PLATFORM} STREQUAL "switch")
+	target_sources(${PROJECT_NAME} PRIVATE
+		src/platform/switch/switch_audio.cpp
+		src/platform/switch/switch_audio.h
+		src/platform/switch/switch_clock.cpp
+		src/platform/switch/switch_clock.h
+		src/platform/switch/switch_ui.cpp
+		src/platform/switch/switch_ui.h)
```
fmatthew5876, line 434: Doesn't have to happen now, but should be call this `switch/ui.h` etc.. instead of duplicating the platform name twice?

carstene1ns: >  drop double 3ds/3ds_ and such prefix (currently not possible since dkp Makefiles need distinct file names)

src/platform/psvita/psp2_clock.h
```diff
@@ -15,8 +15,8 @@
  * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
  */
 
-#ifndef EP_PSP2_CLOCK_H
-#define EP_PSP2_CLOCK_H
+#ifndef EP_PLATFROM_PSP2_CLOCK_H
+#define EP_PLATFROM_PSP2_CLOCK_H
```
fmatthew5876, line 19: `EP_PLATFORM_PSP2_CLOCK_H`

src/platform/psvita/psp2_audio.h
```diff
@@ -14,8 +14,8 @@
  * You should have received a copy of the GNU General Public License
  * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
  */
-#ifndef EP_AUDIO_PSP2_H
-#define EP_AUDIO_PSP2_H
+#ifndef EP_PLATFROM_PSP2_AUDIO_H
+#define EP_PLATFROM_PSP2_AUDIO_H
```
fmatthew5876, line 18: `PLATFROM`

CMakeLists.txt
Ghabry: Visual Studio project files have a similiar problem. At least in VS2017 the default name for object files was without the directory. No idea if they fixed this by now though (and CMake is likely smart enough and overwrites this default)

Issue #2166, opened 2020-04-21
Android: Collection of (impossible) stacktraces


Ghabry:
Before patching them please find a way first to generate them, simply adding failure handling to silence the stacktrace is not helpful.

(with the correct debug symbols you can get the lines, until 0.6 there are only debug symbols for armeabi-v7a and x86. Newer releases also provide aarch64): https://easyrpg.org/downloads/player/dbgsym/easyrpg-player-0.6.0-android-dbgsym.tar.gz

Starting from Android 8 the C++ symbols are demangled btw.
Sorry if the stacktrace is formated incorrect, you can't copy as text from the dev console... thx google

Old issue #1666 

[ ] 1. BattleAnimation::SetFrame(int)
[ ] 2. Game_BattleAlgorithm::AlgorithmBase::HasAnimationPlayed() const
[ ] 3. Input::ResetKeys()
[ ] 4. Game_BattleAlgorithm::AlgorithmBase::HasAnimationPlayed() const
[ ] 5. libmain.so
[ ] 6. Game_BattleAlgorithm::AlgorithmBase::PlaySoundAnimation(bool, int)+463

Ghabry:
1.

```
********** Crash dump: **********
pid: 0, tid: 0 >>> org.easyrpg.player <<<
  #00  pc 000000000022c5f8  lib/arm64/libmain.so (BattleAnimation::SetFrame(int)): /src/battle_animation.cpp:318
  #01  pc 0000000000299c34  lib/arm64/libmain.so (Game_Screen::ShowBattleAnimation(int, int, bool, int)+200): /src/game_screen.cpp:375
  #02  pc 00000000002e4080  lib/arm64/libmain.so (Scene_Map::Start()+356): /src/scene_map.cpp:84
  #03  pc 00000000002d97e8  lib/arm64/libmain.so (Scene::MainFunction()+532): /src/scene.cpp:?
  #04  pc 00000000002be018  lib/arm64/libmain.so (Player::MainLoop()+208): /src/player.cpp:246
  #05  pc 00000000002bdec4  lib/arm64/libmain.so (Player::Run()+216): /src/player.cpp:226
  #06  pc 00000000002a8ccc  lib/arm64/libmain.so (SDL_main+12): /src/main.cpp:28
  #07  pc 0000000000043964  lib/arm64/libSDL2.so (Java_org_libsdl_app_SDLActivity_nativeRunMain+696): Routine ??
```

Ghabry:
[...]

PR #2167, opened 2020-04-21
Fix crash in ShowBattleAnimation() when target is invalid

fmatthew5876:
This crash can occur when loading a save game with an
invalid animation target.

Fixes crash #1 in #2166 

To test:

1. Make Ev01 show battle anim on EV02, no wait
2. Make Ev02
3. Start a new game
4. Trigger battle animation
5. save before animation completes
6. in editor, delete Ev02
7. Load your save

RPG_RT will  throw an invalid event exception. Player will crash.

This PR changes it so player will just show a warning and skip the animation

### DIFF
--- a/src/game_screen.cpp
+++ b/src/game_screen.cpp
@@ -359,17 +359,20 @@ int Game_Screen::ShowBattleAnimation(int animation_id, int target_id, bool globa
 		return 0;
 	}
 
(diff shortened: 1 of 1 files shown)


Ghabry:
Nice. Invalid animation was my initial idea but couldn't find a way to bypass the checks in the interpreter. :)

PR #2168, opened 2020-04-21
Battle 2k: Prevent crash when battle action is NULL

Ghabry:
This adds warnings to the Battle2k code in an attempt to solve 2. in #2166 with the help of our users.

My guess is that the bug is caused by an actor/enemy with multiple battle actions but I can't find a way to generate such a situation.

This crash is very rare. It affected only two users but these users 10 times so it must be a very specific edge case that is not randomly hit. Therefore I think knowing game and monster party is enough to figure this out. By showing a visual warning and not crashing maybe we can motivate the users to send us a bug report with a logfile :).

This is for inclusion in 0.6.2, but the solution is obviously not.

### DIFF
--- a/src/scene_battle.cpp
+++ b/src/scene_battle.cpp
@@ -627,6 +627,13 @@ void Scene_Battle::CreateEnemyActionSkill(Game_Enemy* enemy, const RPG::EnemyAct
 
 void Scene_Battle::ActionSelectedCallback(Game_Battler* for_battler) {
 	assert(for_battler->GetBattleAlgorithm() != nullptr);
+
+	if (for_battler->GetBattleAlgorithm() == nullptr) {
+		Output::Warning("ActionSelectedCallback: Invalid action for battler %d (%s)",
+				for_battler->GetId(), for_battler->GetName().c_str());
+		Output::Warning("Please report a bug!");
+	}
+
 	battle_actions.push_back(for_battler);
 
 	if (for_battler->GetType() == Game_Battler::Type_Ally) {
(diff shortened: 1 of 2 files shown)


fmatthew5876:
There is an `assert()` in `Scene_Battle::ActionSelectedCallback()`. I'm assuming assertions are turned off in release builds? We should add a check there too.

Given this already results in a crash, and that crash is rare, I'm more inclined to just `std::abort()` in these cases. Then we'll get a stack trace and not rely on the very few users which hit this to be so generous as to file a bug report.

Thoughts?

fmatthew5876:
Also, thinking more long term

As a refactor, we should encapsulate this battle_actions vector and the setting of states. If we can make a data structure which holds the invariants (like DrawableList) we can make these bugs impossible.

I'm worried even when we discover and fix this one, some other seemingly innocent battle change can very easily introduce it back in again.

carstene1ns:
We do not explicitely disable assertions, however, default seems off:
https://stackoverflow.com/a/9144080

fmatthew5876:
It's probably too late for 6.2, but for the next release we should consider turning them on at least for the few rounds on android apk's leading up to 6.3. Could potentially help a lot with uncovering bugs.

carstene1ns:
Yeah, we have cmake building then. So this is just a matter of passing other build type.
We could even push any beta with assertions or something.

Ghabry:
The crash is fixed (thats the return true I added).
In the typical case you will get double warnings but there are other places where ``ProcessBattleAction`` is called, so I added it there just to be safe.
I dont expect these warnings to stay forever (and they should never appear when we find and fix the bug :D) and avoided new error handling paths because error handling can introduce new bugs. Was safer to just survive until ProcessBattleAction which handles and removes the action.

@fmatthew5876 
I can enable assertions. They should never be hit in normal execution and when a user hits them this means inproper error handling from our side because all user-hit errors should report through "Output".

Ghabry:
Added the warning to ``ActionSelectedCallback``. That the crash is there is unlikely because there are no known crashes in RPG2k3 which takes the same codepath.

PR #2170, opened 2020-04-22
Move __MORPHOS_SHAREDLIBS into the Makefile for MorphOS

BSzili:
I missed this one in my previous PR, this should reduce the Amiga specific clutter in the source files to the minimum.


BSzili:
It won't hurt to keep it open for a while, but I think this is the last of it. I'll do the same with liblcf later.

Issue #2171, opened 2020-04-22
Regression: Deep 8: Pictures in opening sequence have wrong position


Ghabry:
#### Name of the game:
Deep 8
https://rmarchiv.tk/games/200
Recommend "2014-05-01 [ Demo - 1 ]". Is the oldest version but has less DynRPG & co.

#### Describe the issue in detail and how to reproduce it:
Start the game, fast forward through the landing sequence. When you can take over control over yours chars you will see some trees. You can already see that they look strange while scrolling and when you move down you can see that they are not properly aligned.

Didn't happen in 0.6.1!

Map "Vorszenario -> wald 1"
Relevant event is 239 at (19, 14). The Picture position happens once through this Event through reading the screen coordinates.
The reported coordinates are the same in both Player and RPG_RT so the problem must be somewhere else

fmatthew5876:
screenshot of the differences

![pics](https://user-images.githubusercontent.com/1198466/80114638-ca3c0e00-8551-11ea-9ad1-aaf867aaa2b3.png)

fmatthew5876:
So the bug is with pan screen and scroll with map pictures.

```
ShowPicture: X, Y scroll with map
Pan Screen
```

In RPG_RT the picture will stay where it is as the screen moves. In Player, the picture moves with the screen.

PR #2172, opened 2020-04-23
Fix rounding bug in scroll with map picture offset adjustment

fmatthew5876:
Fix: #2171


carstene1ns:
0.6.1 ftw. :grinning: 
https://github.com/EasyRPG/Player/blob/00ad8ab137261fff23559365ad50397a2d98ffcc/src/game_picture.cpp#L280

Ghabry:
Also tested with the latest Deep 8 version: Works for me

PR #2177, opened 2020-04-28
Regression fixes

### LINKED ISSUE #2176: Yume2kki pc_back Wallpaper Doesn't Load in 0.6.2
#### Player platform:
Linux

#### Attach files
https://pastebin.com/5C1LQgXU (easyrpg_log.txt)

In 0.6.1 using the pc_back customizable wallpaper in Yume2kki works:
notborked

But after updating to 0.6.2 the pc_back wallpaper doesn't load anymore.
borkednow

EDIT: Oh, forgot how to mention how to reproduce this:
Start a new game, skip the intro and head to the PC, select the Tools option, then Wallpaper, choose ON if you see a "New Wallpaper !!" notification with an ON/OFF option, and then choose the second option in the screen after that, and then the first option in the pc_back menu.

Ghabry:
That's a flaw (well, the structure made sense because there is no way to read stuff from ./, but then "creative" games appeared...) with the FileFinder data structure: The root-directory files are stored in a different location than the submembers, so there must be extra handling for finding files in the root-dir.

Fix #2176

Regression caused by 7de54b62

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -2504,6 +2504,9 @@ bool Game_Interpreter::CommandShowPicture(RPG::EventCommand const& com) { // cod
 		// transparency. >= 1.12 Editor only let you set one transparency field but it affects
 		// both chunks here.
 		params.bottom_trans = com.parameters[14];
+	} else if (Player::IsRPG2k3() && !Player::IsRPG2k3E()) {
+		// Corner case when 2k maps are used in 2k3 (pre-1.10) and don't contain this chunk
+		params.bottom_trans = params.top_trans;
 	}
 
 	if (param_size > 16) {
--- a/src/bitmap.cpp
+++ b/src/bitmap.cpp
@@ -103,9 +103,9 @@ Bitmap::Bitmap(const std::string& filename, bool transparent, uint32_t flags) {
 
 	int w = 0;
 	int h = 0;
-	void* pixels;
+	void* pixels = nullptr;
 
-	char data[4];
+	uint8_t data[4] = {};
 	size_t bytes = fread(&data, 1, 4, stream);
 	fseek(stream, 0, SEEK_SET);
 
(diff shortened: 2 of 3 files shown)


Ghabry:
Added a small fix here to address Android 10 & 12 if you don't mind.

Ghabry:
Renamed this to "Regression fixes" now.
2k3E is not affected by this btw because the bottom_trans is ignored in the Picture code.

src/game_interpreter.cpp
```diff
@@ -2504,6 +2504,9 @@ bool Game_Interpreter::CommandShowPicture(RPG::EventCommand const& com) { // cod
 		// transparency. >= 1.12 Editor only let you set one transparency field but it affects
 		// both chunks here.
 		params.bottom_trans = com.parameters[14];
+	} else if (Player::IsRPG2k3() && !Player::IsRPG2k3E()) {
+		// Corner case when 2k maps are used in 2k3 (pre-1.10) and don't contain this chunk
+		params.bottom_trans = params.top_trans;
```
fmatthew5876, line 2509: Don't think it matters realistically, but it might be cleaner to do it as:

```
params.bottom_trans = params.top_trans;
if (param_size > 14) {
    params.bottom_trans = com.parameters[14];
}
```

Then at least we're guaranteed `bottom_trans` is always initialized to a sane value.

Ghabry: The old code that got lost by your commit was ``params.bottom_trans = param_size > 14 ? com.parameters[14] : params.top_trans;``. So is exactly this one. I can add it back.

Not sure how much impact it has on "LSD compatible fixes for SavePicture::current_bot_trans". I prefer working interpreter over RPG_RT compatible LSDs.

fmatthew5876: Ahh no you're right. If we do that then this chunk will start showing up in 2k saves. Your version is correct.

carstene1ns:
I would like to pull this one in the maintenance branch.

Ghabry:
There is a bug report mail about bad images in VD2 intro: Is fixed my the ShowPicture fix

Issue #2098, opened 2020-04-03
Switch to using fmtlib for our logging


fmatthew5876:
We should change our `Output` logger to use fmtlib instead of legacy C style print()

https://github.com/fmtlib/fmt

fmtlib is coming in newer C++ standard library. Benefits for us are:

* More type safety in logging
* More flexibility in being able to log objects directly, including non null terminated strings
* In some cases fmtlib is even faster than printf

Ghabry:
I also came to the conclusion a while ago that fmtlib would be one of the useful libraries for Player.

The problem is that this will be a required dependency - Must be added to the buildscripts:
https://github.com/EasyRPG/buildscripts/

Ghabry:
fmtlib does also support positional arguments, useful for translations when we start to support this :)

PR #2181, opened 2020-04-29
Use fmtlib for Output logger

fmatthew5876:
Fix: #2098 

In order to avoid rebase hell, I will add commits to change all of our log statements only after this PR gets approved.

Looks like linux packaging for this library is not great. In Ubuntu eoan for example it only has ancient version 5.1.2 which doesn't have pkg-config and breaks on trying to print `char32_t`.

Focal has 6.1.2 which should be fine.

vcpkg is up to date with latest 6.2.0


src/output.h
```diff
@@ -67,32 +69,27 @@ namespace Output {
 	void IgnorePause(bool val);
 
 	/**
-	 * Displays a string with formatted string.
+	 * Displays an info string with formatted string.
 	 *
 	 * @param fmt formatted string to display.
```
carstene1ns, line 74: dox?

Ghabry:
So the minimal supported version should be 5.3.

Is for Debian in "Debian Stable Backports".
Fedora starting with Fedora 31.
OpenBSD Ports has 2.0 ??? o.O
Ubuntu starting from Ubuntu 20.04.

Other part are the console ports. I wonder how much frustration compiling this with newlib will give >.>

src/output.h
fmatthew5876: Added, and updated `README.md` too

Ghabry:
Library building looks good. Havn't tested Player linking yet.

You also have to add "-lfmt" to all our Makefiles. (fortunately this will solve when the CMake migration is done...)

fmatthew5876:
I tested and this compiles and links successfully with

- [x] Ubuntu eoan cmake
- [x] Ubuntu focal autotools
- [x] Windows vcpkg

Ghabry:
Build fails with

```
tests/output.cpp:13:10: error: 'Post' is not a member of 'Output'
   13 |  Output::Post("Test %s", "post");
```

fmatthew5876:
Thanks for catching that, fixed the test.

Ghabry:
Could youdo a full text search for Output::Post ? Also see a failure in the vita build.

fmatthew5876:
Hmm I did do a text search, but must have missed that one. Anyway, updated it.

builds/switch/Makefile
```diff
@@ -78,7 +78,7 @@ ASFLAGS	:=	-g $(ARCH)
 LDFLAGS	=	-specs=$(DEVKITPRO)/libnx/switch.specs -g $(ARCH) -Wl,-Map,$(notdir $*.map)
 
 LIBS	:=  -llcf -lexpat -lpixman-1 -lpng -lz \
-		-licuuc -licui18n -licudata \
+		-licuuc -licui18n -licudata -fmt \
```
Ghabry, line 81: lacks a "l"

fmatthew5876: Fixed

fmatthew5876:
Ok I've added commits to convert all log messages over. I also cleaned up and tested `Window_NumberInput`.

I did this by grepping for all files with `Output::`. Then opening them all in my editor and using search and replace in the edtior for the basic cases and manually fixing the complicated ones.

I didn't want to do something fully systematic like `sed` here because we have some `snprintf` usage and other things like removing `.c_str()` which are hard.

fmatthew5876:
Self reviewed this, fixed a few minor things. From my perspective this is ready to go.

Ghabry:
Does fmtlib support %s and %d in case we miss something?

fmatthew5876:
> Does fmtlib support %s and %d in case we miss something?

It'll just print the %s and %s literally and ignore the arguments.

I did a `ag Output:: | grep %` and it returns me nothing. Pretty sure we got them all

Ghabry:
Jenkins: Test this please <3

fmatthew5876:
jenkins: test this please

fmatthew5876:
> (Note that many new commits - At least 10 - were added since the last review)

Is that not an artifact of rebasing? I haven't changed this since my last comments.

Also, kicked off 3ds, switch, and vita builds

Ghabry:
When I looked at this the last commit was "Fix output test and vita". Then all the others appeared one day later.

fmatthew5876:
> When I looked at this the last commit was "Fix output test and vita". Then all the others appeared one day later.

Right, when you guys approved it first I hadn't done anything to actually convert our code to use fmtlib style. The other commits are changing our entire code base to use fmtlib style logging. This has to be done or all our log statements will be broken.

All that needs another review.

src/player.cpp
```diff
@@ -147,12 +147,12 @@ void Player::Init(int argc, char *argv[]) {
 	if (!addtl_ver.empty())
 		header << " " << addtl_ver;
 	header << " started";
-	Output::Debug(header.str().c_str());
+	Output::Debug(header.str());
```
Ghabry, line 150: Not your fault but because I see it: Can you change this to ``{}, header`` so we don't have a variable in the first argument?

src/player.cpp
```diff
@@ -147,12 +147,12 @@ void Player::Init(int argc, char *argv[]) {
...
 
 	unsigned int header_width = header.str().length();
 	header.str("");
 	header << std::setfill('=') << std::setw(header_width) << "=";
-	Output::Debug(header.str().c_str());
+	Output::Debug(header.str());
```
Ghabry, line 155: same here.

src/window_numberinput.cpp
```diff
@@ -49,16 +50,9 @@ Window_NumberInput::Window_NumberInput(int ix, int iy, int iwidth, int iheight)
...
-	if (show_operator) {
-		format_string += plus ? "+" : "-";
-	}
-
-	format_string += "%0*d";
-
-	snprintf(s, 9, format_string.c_str(), digits_max, number);
+	auto s = fmt::format("{0}{1:0{2}d}",
+			show_operator ? (plus ? "+" : "-") : "",
+			number, digits_max);
```
Ghabry, line 55: Marking this - Need to check how this magic works

Ghabry:
Made a build test with fmtlib 5.3 as this is the lowest version where supporting makes sense and that many distributions offer:
Compiles :+1: 

Maybe add a version check to CMake. For autotools version checking is not useful because no older version exists ;)

fmatthew5876:
Addressed review comments

fmatthew5876:
Rebased, and fixed a few in put from the rebase.

Did another set of greps to verify no stray `%` leaked in.

src/game_party.cpp
```diff
@@ -151,8 +151,7 @@ void Game_Party::LoseGold(int n) {
 
 void Game_Party::AddItem(int item_id, int amount) {
 	if (item_id < 1 || item_id > (int) Data::items.size()) {
-		Output::Debug("Can't add item to party. %04d is not a valid item ID.",
-						item_id);
+		Output::Debug("Can't add item to party. {} is not a valid item ID.", item_id);
```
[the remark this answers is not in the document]
fmatthew5876: I don't see us padding item ids or skill ids anywhere else so this looked like a bug. If there's a reason it should stay this way I change it back.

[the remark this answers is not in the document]
carstene1ns: Well, these are different as well:
https://github.com/EasyRPG/Player/pull/2181/files#diff-2aaa5ba57b150041a2e74c625a51e663R67
https://github.com/EasyRPG/Player/pull/2181/files#diff-765ecb2cb36ff21ff2bb0e8857e36aeeR58

It was just a question, so let's go on.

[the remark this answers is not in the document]
fmatthew5876: That was a call on my part to print unicode glyphs in hex instead of decimal.

Issue #2186, opened 2020-05-01
Wii: Time measurement with Platform_Clock::now is constant


Ghabry:
#### Player platform:
Wii

#### Describe the issue in detail and how to reproduce it:

The sleep_for 1s works excellent but the clock is not incrementing:

![Screenshot_20200501_174207-fs8](https://user-images.githubusercontent.com/1331889/80818301-331b2a00-8bd3-11ea-8248-a18b6f52ae08.png)

fmatthew5876:
What does the `Game_Clock` line in the log file look like?

Ghabry:
StdSteady steady=1 period=ns (1 / 1000000000)

Ghabry:
libstdc++ implementation is here: https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/src/c%2B%2B11/chrono.cc#L42

But no idea which one devkitXXX is configured to use. No idea where you configure this.

fmatthew5876:
We can try changing it to use `std::system_clock` for newlib ports.

Ghabry:
How do you do this? Sorry but I'm always lost when I look in std_clock.h. I just see template foo and hard to read chrono code.

fmatthew5876:
https://github.com/EasyRPG/Player/blob/master/src/std_clock.h#L27

Ghabry:
The broken function must be ``clock_gettime``. The value is stuck.
With ``std::time`` and ``gettimeofday`` I get proper values.

There is also a discussion here that the function is broken: https://github.com/dborth/sdl-wii/issues/50

SDL-Wii implementation: https://github.com/dborth/sdl-wii/blob/master/SDL/src/timer/wii/SDL_systimer.c

Maybe some useful constants in here?
https://github.com/devkitPro/libogc/blob/master/gc/ogc/lwp_watchdog.h

fmatthew5876:
Is it broken because of wii? Or because it's newlib?

PR #2187, opened 2020-05-01
Add Wii Platform Clock

fmatthew5876:
Fix: #2186

### DIFF
--- a/src/std_clock.h
+++ b/src/std_clock.h
@@ -19,15 +19,11 @@
 #define EP_STD_CLOCK_H
 
 #include <chrono>
-#include <type_traits>
 #include <thread>
 
 #if !defined(_3DS) && !defined(__SWITCH__) && !defined(PSP2)
 struct StdClock {
-	using clock = std::conditional_t<
-		std::chrono::high_resolution_clock::is_steady,
-		std::chrono::high_resolution_clock,
-		std::chrono::steady_clock>;
+	using clock = std::chrono::steady_clock;
 
 	using rep = clock::rep;
 	using period = clock::period;
@@ -56,12 +52,12 @@ inline void StdClock::SleepFor(std::chrono::duration<R,P> dt) {
 }
 
 constexpr const char* StdClock::Name() {
-	if (std::is_same<clock,std::chrono::high_resolution_clock>::value) {
-		return "StdHigRes";
-	} else if (std::is_same<clock,std::chrono::steady_clock>::value) {
+	if (std::is_same<clock,std::chrono::steady_clock>::value) {
 		return "StdSteady";
 	} else if (std::is_same<clock,std::chrono::system_clock>::value) {
 		return "StdSystem";
+	} else if (std::is_same<clock,std::chrono::high_resolution_clock>::value) {
+		return "StdHigRes";
 	}
 	return "Unknown";
 }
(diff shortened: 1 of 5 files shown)


Ghabry:
In Dolphin this works fine but maybe also do a hardware test. Don't have my Wii around so maybe @carstene1ns could do it?

Ghabry:
When I do ``std::chrono::time_point_cast<std::chrono::milliseconds>(Platform_Clock::now())`` I get on Linux: 9075040, 9076040 and in Dolphin: 
Some gigantic number such as "5021779348504474546" followed by "5283566805015642470"

Ghabry:
Test code for the clock.
Just put it in the main before ``Player::Init`` is called

```cpp
#include "platform_clock.h"
#include "output.h"

	Output::Debug("%s", Platform_Clock::Name());
	auto a = Platform_Clock::now();
	auto b = Platform_Clock::now();

	do {
		a = Platform_Clock::now();
		Output::Debug("Now1: %lld %lld", a - b, std::chrono::duration_cast<std::chrono::milliseconds>(a - b));
		Output::Debug("Sleep 1s");
		Platform_Clock::SleepFor(1s);
		b = Platform_Clock::now();
		Output::Debug("Now2: %lld %lld", b - a, std::chrono::duration_cast<std::chrono::milliseconds>(b - a));
		Output::Debug("Sleep 1s");
		Platform_Clock::SleepFor(1s);
	} while (true);

**Comment by Ghabry** [comment]
I'm so sorry for wasting time here. Was all my fault.

The correct way to print a chrono time is
```
std::chrono::duration_cast<std::chrono::milliseconds>(b - a).count()
```
and not
```
std::chrono::duration_cast<std::chrono::milliseconds>(b - a)
```

Somehow this worked on Linux and 3DS so I thought this is the correct way to do it :/

Even the first commit after fixing the 2 compile errors yields 1000.

**Comment by Ghabry** [comment]
Guess the reason why this printed wrong without the count() call is that the way powerpc packs 64bit different in memory (different byte order, also cpu is not 64bit) :(

Wii: ``39f8cc40000940d``
Linux: ``3b9c9635``

File: src/std_clock.h [hunk 1]
```diff
@@ -19,15 +19,11 @@
...
-#include <type_traits>
 #include <thread>
 
 #if !defined(_3DS) && !defined(__SWITCH__) && !defined(PSP2)
 struct StdClock {
-	using clock = std::conditional_t<
-		std::chrono::high_resolution_clock::is_steady,
-		std::chrono::high_resolution_clock,
-		std::chrono::steady_clock>;
+	using clock = std::chrono::steady_clock;
```
carstene1ns, line 26: Looks like the logic about high res clock can be removed in `Name()` function as well?

Ghabry:
btw Android actually reports highres clock but I assume this is an artifact of the evaluation order. When the check for HighRes clock in Name() would be the last this would report a different clock.

```
Clock: StdHigRes steady=1 period=ns (1 / 1000000000)
```

carstene1ns:
Makes sense, so it is both HighRes and Steady.

carstene1ns:
I can confirm that this works on hardware.
@fmatthew5876: Question is, do you still want to revert some part of it (as @Ghabry said it worked already in a former state).

Once merged, I will pull this into maintenance branch.

Ghabry:
This is already the first version.
He reverted it for me once I found the problem in my tests. :)

carstene1ns:
Okay, then ready to merge after having heard about the `StdClock::Name()`.

fmatthew5876:
Re-ordered the name check. I kept the ifs in there just in case later we decide we need to use some of the other standard clocks. They are compile time ifs so performance is not affected.

fmatthew5876:
jenkins: test this please

PR #2188, opened 2020-05-02
Amiga build fix

BSzili:
I managed to switch to SDL2 on MorphOS with some outside help. The same will happen for AmigaOS4 once I'm able to test it myself. There's still no SDL2 for AROS.

### DIFF
--- a/src/sdl2_ui.cpp
+++ b/src/sdl2_ui.cpp
@@ -255,8 +255,8 @@ bool Sdl2Ui::RefreshDisplayMode() {
 	int display_height = current_display_mode.height;
 
 #ifdef SUPPORT_ZOOM
-	display_width *= current_display_mode.zoom;
-	display_height *= current_display_mode.zoom;
+		display_width *= current_display_mode.zoom;
+		display_height *= current_display_mode.zoom;
 #endif
 
 	if (!sdl_window) {
@@ -289,9 +289,11 @@ bool Sdl2Ui::RefreshDisplayMode() {
 
 		uint32_t rendered_flag = 0;
 
+#ifndef __MORPHOS__
 		if (Player::vsync) {
 			rendered_flag |= SDL_RENDERER_PRESENTVSYNC;
 		}
+#endif
 
 		sdl_renderer = SDL_CreateRenderer(sdl_window, -1, rendered_flag);
 		if (!sdl_renderer) {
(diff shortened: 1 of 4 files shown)


Ghabry:
Needs Rebase. Wrong master used?

SDL2 support is great. We want to drop SDL someday when nobody uses it anymore.

Btw are there some CMake toolchain files available for these platforms? At least with SDL2 support it should be quite easy now to get this working.

src/sdl2_ui.cpp
```diff
@@ -255,8 +255,8 @@ bool Sdl2Ui::RefreshDisplayMode() {
 	int display_height = current_display_mode.height;
 
 #ifdef SUPPORT_ZOOM
-	display_width *= current_display_mode.zoom;
-	display_height *= current_display_mode.zoom;
+		display_width *= current_display_mode.zoom;
```
carstene1ns, line 258: rebase leftover.

src/sdl2_ui.cpp
```diff
@@ -411,23 +413,23 @@ void Sdl2Ui::ToggleZoom() {
...
-		current_display_mode.zoom = 1;
-	}
+		// get current window size, calculate next bigger zoom factor
+		int w, h;
+		SDL_GetWindowSize(sdl_window, &w, &h);
+		last_display_mode.zoom = std::min(w / SCREEN_TARGET_WIDTH, h / SCREEN_TARGET_HEIGHT);
+		current_display_mode.zoom = last_display_mode.zoom + 1;
+
+		// get maximum usable window size
+		int display_index = SDL_GetWindowDisplayIndex(sdl_window);
```
carstene1ns, line 423: same here.

BSzili:
CMake is now part of the official MorphOS SDK, there's also a third party port to AmigaOS4. AROS again falls by the wayside, but I cross-compile everything. When you decide to drop SDL1.2, I'll try to come up with something. Unless you plan to use multiple windows it should be possible to make an SDL2 shim for SDL1.2 on AROS.

carstene1ns:
We will likely have some SDL1.2 support for the next years. However, only fixes, no new features.

fmatthew5876:
Unrelated to the PR, @BSzili what gcc compiler versions are supported for these platforms? Is there work being done on upgrading?

I'd like to get us on C++17 as soon as all our ports are on a viable gcc version.

BSzili:
MorphOS has GCC 9, and AmigaOS4 has GCC 8, those should be good. For AROS ABIv0 the last compiler is GCC 6, which only has limited C++17 support :\

carstene1ns:
Also pulling this into maintenance branch.

PR #2189, opened 2020-05-02
Filesystem [1/N]

Ghabry:
This is a partial patchset from the VFS changes rebased to upstream.
It only changes how File input/output works across Player in a consistent way by now. You don't get any new features by this but it makes future VFS work muuuuuch easier when these stream changes are kept being updated upstream.

For reviewing I recommend to look at the change as a whole. The changes by ChrisBreiti are a bit old :)

- All uses of ``FILE*`` replaced with C++ streams
- All C++ streams use now ``Filesystem::InputStream/Filesystem::OutputStream`` which is by now a shallow wrapper around ``std::filebuf``. 

Basicly the VFS code will do the following later:

1. You open a file like ``/home/user/game.zip/RPG_RT.ldb``
2. This returns a ``Filesystem::InputStream`` that inherits from ``istream``. So most of C++ fstream API will work as expected.
3. Use normal istream read funcs on it.

Is a draft. Need to test if I broke any Read/Write code. All these raw buffer APIs are hard to get right when a shared pointer is mixed in ^^'. Type confusion.

### DIFF
--- a/src/audio_decoder.cpp
+++ b/src/audio_decoder.cpp
@@ -104,7 +104,7 @@ public:
 		error_message = std::string("WMA audio files are not supported. Reinstall the\n") +
 			"game and don't convert them when asked by Windows!\n";
 	}
-	bool Open(Filesystem::InputStream stream) override { return false; }
+	bool Open(Filesystem::InputStream) override { return false; }
 	bool IsFinished() const override { return true; }
 	void GetFormat(int&, Format&, int&) const override {}
 	bool Seek(std::streamoff, std::ios_base::seekdir) override { return false; }
--- a/src/audio_decoder.cpp
+++ b/src/audio_decoder.cpp
@@ -114,8 +114,10 @@ const char wma_magic[] = { (char)0x30, (char)0x26, (char)0xB2, (char)0x75 };
 
 std::unique_ptr<AudioDecoder> AudioDecoder::Create(std::shared_ptr<FileFinder::istream> stream, const std::string& filename, bool resample) {
 	char magic[4] = { 0 };
-	stream->read(magic, sizeof(magic));
-	stream->seekg(std::ios::ios_base::beg);
+	if (stream->read(magic, sizeof(magic)).gcount() == 0) {
+		return nullptr;
+	}
+	stream->seekg(0, std::ios::ios_base::beg);
 
 #if !(defined(HAVE_WILDMIDI) || defined(HAVE_XMP))
 	/* WildMidi and XMP are the only audio decoders that need the filename passed
@@ -164,10 +166,12 @@ std::unique_ptr<AudioDecoder> AudioDecoder::Create(std::shared_ptr<FileFinder::i
 	// Try to use internal OGG decoder
 	if (!strncmp(magic, "OggS", 4)) { // OGG
 #ifdef HAVE_OPUS
-		fseek(file, 28, SEEK_SET);
-		if (fread(magic, 4, 1, file) != 1)
+		stream->seekg(28, std::ios::ios_base::beg);
+		if (stream->read(magic, sizeof(magic)).gcount() == 0) {
 			return nullptr;
-		fseek(file, 0, SEEK_SET);
+		}
+		stream->seekg(0, std::ios::ios_base::beg);
+
 		if (!strncmp(magic, "Opus", 4)) {
 			if (resample) {
 				return std::unique_ptr<AudioDecoder>(new AudioResampler(std::unique_ptr<AudioDecoder>(new OpusDecoder())));
@@ -179,7 +183,9 @@ std::unique_ptr<AudioDecoder> AudioDecoder::Create(std::shared_ptr<FileFinder::i
 
 #if defined(HAVE_TREMOR) || defined(HAVE_OGGVORBIS)
 		stream->seekg(29, std::ios::ios_base::beg);
-		stream->read(magic, sizeof(magic));
+		if (stream->read(magic, sizeof(magic)).gcount() == 0) {
+			return nullptr;
+		}
 		stream->seekg(0, std::ios::ios_base::beg);
 
 		if (!strncmp(magic, "vorb", 4)) {
--- a/src/audio_decoder.cpp
+++ b/src/audio_decoder.cpp
@@ -19,7 +19,6 @@
 #include <cassert>
 #include <cstring>
 #include "audio_decoder.h"
-#include "filefinder.h"
 #include "output.h"
 #include "system.h"
 #include "utils.h"
@@ -105,7 +104,7 @@ public:
 		error_message = std::string("WMA audio files are not supported. Reinstall the\n") +
 			"game and don't convert them when asked by Windows!\n";
 	}
-	bool Open(FILE*) override { return false; }
+	bool Open(std::shared_ptr<FileFinder::istream> stream) override { return false; }
 	bool IsFinished() const override { return true; }
 	void GetFormat(int&, Format&, int&) const override {}
 private:
@@ -113,11 +112,10 @@ private:
 };
 const char wma_magic[] = { (char)0x30, (char)0x26, (char)0xB2, (char)0x75 };
 
-std::unique_ptr<AudioDecoder> AudioDecoder::Create(FILE* file, const std::string& filename, bool resample) {
+std::unique_ptr<AudioDecoder> AudioDecoder::Create(std::shared_ptr<FileFinder::istream> stream, const std::string& filename, bool resample) {
 	char magic[4] = { 0 };
-	if (fread(magic, 4, 1, file) != 1)
-		return nullptr;
-	fseek(file, 0, SEEK_SET);
+	stream->read(magic, sizeof(magic));
+	stream->seekg(std::ios::ios_base::beg);
 
 #if !(defined(HAVE_WILDMIDI) || defined(HAVE_XMP))
 	/* WildMidi and XMP are the only audio decoders that need the filename passed
@@ -268,8 +266,9 @@ std::unique_ptr<AudioDecoder> AudioDecoder::Create(FILE* file, const std::string
 		}
 
 		// Parsing MP3s seems to be the only reliable way to detect them
-		if (Mpg123Decoder::IsMp3(file)) {
-			fseek(file, 0, SEEK_SET);
+		if (Mpg123Decoder::IsMp3(stream)) {
+			stream->clear();
+			stream->seekg(0, std::ios::ios_base::beg);
 			if (resample) {
 				mp3dec = new AudioResampler(std::unique_ptr<AudioDecoder>(new Mpg123Decoder()));
 			} else {
@@ -287,7 +286,8 @@ std::unique_ptr<AudioDecoder> AudioDecoder::Create(FILE* file, const std::string
 	}
 #endif
 
-	fseek(file, 0, SEEK_SET);
+	stream->clear();
+	stream->seekg(0, std::ios::ios_base::beg);
 	return nullptr;
 }
 
--- a/src/audio_decoder.cpp
+++ b/src/audio_decoder.cpp
@@ -107,6 +107,7 @@ public:
 	bool Open(Filesystem::InputStream stream) override { return false; }
 	bool IsFinished() const override { return true; }
 	void GetFormat(int&, Format&, int&) const override {}
+	bool Seek(std::streamoff, std::ios_base::seekdir) override { return false; }
 private:
 	int FillBuffer(uint8_t*, int) override { return -1; };
 };
@@ -337,7 +338,7 @@ int AudioDecoder::GetVolume() const {
 }
 
 void AudioDecoder::Rewind() {
-	if (!Seek(0, Origin::Begin)) {
+	if (!Seek(0, std::ios_base::beg)) {
 		// The libs guarantee that Rewind works
 		assert(false && "Rewind");
 	}
@@ -379,11 +380,7 @@ bool AudioDecoder::SetPitch(int) {
 	return false;
 }
 
-bool AudioDecoder::Seek(size_t, Origin) {
-	return false;
-}
-
-size_t AudioDecoder::Tell() const {
+std::streampos AudioDecoder::Tell() const {
 	return -1;
 }
 
--- a/src/audio_decoder.cpp
+++ b/src/audio_decoder.cpp
@@ -115,11 +115,10 @@ const char wma_magic[] = { (char)0x30, (char)0x26, (char)0xB2, (char)0x75 };
 
 std::unique_ptr<AudioDecoder> AudioDecoder::Create(Filesystem_Stream::InputStream& stream, const std::string& filename, bool resample) {
 	char magic[4] = { 0 };
-	stream.ReadIntoObj(magic);
 	if (!stream.ReadIntoObj(magic)) {
 		return nullptr;
 	}
-	stream.seekg(0, std::ios::ios_base::beg);
+	stream.seekg(0, std::ios::beg);
 
 #if !(defined(HAVE_WILDMIDI) || defined(HAVE_XMP))
 	/* WildMidi and XMP are the only audio decoders that need the filename passed
@@ -275,7 +274,7 @@ std::unique_ptr<AudioDecoder> AudioDecoder::Create(Filesystem_Stream::InputStrea
 		// Parsing MP3s seems to be the only reliable way to detect them
 		if (Mpg123Decoder::IsMp3(stream)) {
 			stream.clear();
-			stream.seekg(0, std::ios::ios_base::beg);
+			stream.seekg(0, std::ios_base::beg);
 			if (resample) {
 				mp3dec = new AudioResampler(std::unique_ptr<AudioDecoder>(new Mpg123Decoder()));
 			} else {
--- a/src/audio_decoder.cpp
+++ b/src/audio_decoder.cpp
@@ -178,10 +178,9 @@ std::unique_ptr<AudioDecoder> AudioDecoder::Create(std::shared_ptr<FileFinder::i
 #endif
 
 #if defined(HAVE_TREMOR) || defined(HAVE_OGGVORBIS)
-		fseek(file, 29, SEEK_SET);
-		if (fread(magic, 4, 1, file) != 1)
-			return nullptr;
-		fseek(file, 0, SEEK_SET);
+		stream->seekg(29, std::ios::ios_base::beg);
+		stream->read(magic, sizeof(magic));
+		stream->seekg(0, std::ios::ios_base::beg);
 
 		if (!strncmp(magic, "vorb", 4)) {
 			if (resample) {
--- a/src/audio_decoder.cpp
+++ b/src/audio_decoder.cpp
@@ -136,7 +136,7 @@ std::unique_ptr<AudioDecoder> AudioDecoder::Create(Filesystem::InputStream strea
 #ifdef HAVE_WILDMIDI
 		static bool wildmidi_works = true;
 		if (wildmidi_works) {
-			auto mididec = std::unique_ptr<AudioDecoder>(new WildMidiDecoder(filename));
+			auto mididec = std::unique_ptr<AudioDecoder>(new WildMidiDecoder());
 			if (mididec->WasInited()) {
 				if (resample) {
 					mididec = std::unique_ptr<AudioResampler>(new AudioResampler(std::move(mididec)));
--- a/src/audio_decoder.cpp
+++ b/src/audio_decoder.cpp
@@ -104,7 +104,7 @@ public:
 		error_message = std::string("WMA audio files are not supported. Reinstall the\n") +
 			"game and don't convert them when asked by Windows!\n";
 	}
-	bool Open(std::shared_ptr<FileFinder::istream> stream) override { return false; }
+	bool Open(Filesystem::InputStream stream) override { return false; }
 	bool IsFinished() const override { return true; }
 	void GetFormat(int&, Format&, int&) const override {}
 private:
@@ -112,7 +112,7 @@ private:
 };
 const char wma_magic[] = { (char)0x30, (char)0x26, (char)0xB2, (char)0x75 };
 
-std::unique_ptr<AudioDecoder> AudioDecoder::Create(std::shared_ptr<FileFinder::istream> stream, const std::string& filename, bool resample) {
+std::unique_ptr<AudioDecoder> AudioDecoder::Create(Filesystem::InputStream stream, const std::string& filename, bool resample) {
 	char magic[4] = { 0 };
 	if (stream->read(magic, sizeof(magic)).gcount() == 0) {
 		return nullptr;
--- a/src/audio_decoder.cpp
+++ b/src/audio_decoder.cpp
@@ -104,7 +104,7 @@ public:
 		error_message = std::string("WMA audio files are not supported. Reinstall the\n") +
 			"game and don't convert them when asked by Windows!\n";
 	}
-	bool Open(Filesystem::InputStream) override { return false; }
+	bool Open(Filesystem_Stream::InputStream) override { return false; }
 	bool IsFinished() const override { return true; }
 	void GetFormat(int&, Format&, int&) const override {}
 	bool Seek(std::streamoff, std::ios_base::seekdir) override { return false; }
@@ -113,12 +113,13 @@ private:
 };
 const char wma_magic[] = { (char)0x30, (char)0x26, (char)0xB2, (char)0x75 };
 
-std::unique_ptr<AudioDecoder> AudioDecoder::Create(Filesystem::InputStream stream, const std::string& filename, bool resample) {
+std::unique_ptr<AudioDecoder> AudioDecoder::Create(Filesystem_Stream::InputStream& stream, const std::string& filename, bool resample) {
 	char magic[4] = { 0 };
-	if (stream->read(magic, sizeof(magic)).gcount() == 0) {
+	stream.ReadIntoObj(magic);
+	if (!stream.ReadIntoObj(magic)) {
 		return nullptr;
 	}
-	stream->seekg(0, std::ios::ios_base::beg);
+	stream.seekg(0, std::ios::ios_base::beg);
 
 #if !(defined(HAVE_WILDMIDI) || defined(HAVE_XMP))
 	/* WildMidi and XMP are the only audio decoders that need the filename passed
@@ -167,11 +168,11 @@ std::unique_ptr<AudioDecoder> AudioDecoder::Create(Filesystem::InputStream strea
 	// Try to use internal OGG decoder
 	if (!strncmp(magic, "OggS", 4)) { // OGG
 #ifdef HAVE_OPUS
-		stream->seekg(28, std::ios::ios_base::beg);
-		if (stream->read(magic, sizeof(magic)).gcount() == 0) {
+		stream.seekg(28, std::ios::ios_base::beg);
+		if (stream.read(magic, sizeof(magic)).gcount() == 0) {
 			return nullptr;
 		}
-		stream->seekg(0, std::ios::ios_base::beg);
+		stream.seekg(0, std::ios::ios_base::beg);
 
 		if (!strncmp(magic, "Opus", 4)) {
 			if (resample) {
@@ -183,11 +184,11 @@ std::unique_ptr<AudioDecoder> AudioDecoder::Create(Filesystem::InputStream strea
 #endif
 
 #if defined(HAVE_TREMOR) || defined(HAVE_OGGVORBIS)
-		stream->seekg(29, std::ios::ios_base::beg);
-		if (stream->read(magic, sizeof(magic)).gcount() == 0) {
+		stream.seekg(29, std::ios::ios_base::beg);
+		if (stream.read(magic, sizeof(magic)).gcount() == 0) {
 			return nullptr;
 		}
-		stream->seekg(0, std::ios::ios_base::beg);
+		stream.seekg(0, std::ios::ios_base::beg);
 
 		if (!strncmp(magic, "vorb", 4)) {
 			if (resample) {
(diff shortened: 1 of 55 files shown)


Ghabry:
Jenkins: Test this please

Ghabry:
There are still some crashes to resolve e.g. when project/save path are invalid.

But besides this: rdy for review

src/bitmap.cpp
```diff
@@ -106,21 +106,21 @@ Bitmap::Bitmap(const std::string& filename, bool transparent, uint32_t flags) {
...
-		img_okay = ImageBMP::ReadBMP(stream, transparent, w, h, pixels);
+		img_okay = ImageBMP::ReadBMP(*stream, transparent, w, h, pixels);
 	else if (bytes >= 4 && strncmp((char*)(data + 1), "PNG", 3) == 0)
-		img_okay = ImagePNG::ReadPNG(stream, (void*)nullptr, transparent, w, h, pixels);
+		img_okay = ImagePNG::ReadPNG(*stream, transparent, w, h, pixels);
 	else
 		Output::Warning("Unsupported image file %s (Magic: %02X)", filename.c_str(), *reinterpret_cast<uint32_t*>(data));
 
-	fclose(stream);
+	stream.reset(); //Deletes the stream and closes the file
```
fmatthew5876: Not strictly needed, unless you have a reason to want to close the file early and release the resources.

src/filefinder.cpp
```diff
@@ -668,6 +668,32 @@ std::shared_ptr<std::fstream> FileFinder::openUTF8(const std::string& name,
...
+		name.c_str(),
+#endif
+		m));
+	return (*ret) ? ret : std::shared_ptr<std::ifstream>();
+}
+
+std::shared_ptr<std::ostream> FileFinder::openUTF8Output(const std::string& name,
+	std::ios_base::openmode m)
+{
+	std::shared_ptr<std::ofstream> ret(new std::ofstream(
```
fmatthew5876: `std::make_shared`

src/filefinder.cpp
```diff
@@ -668,6 +668,32 @@ std::shared_ptr<std::fstream> FileFinder::openUTF8(const std::string& name,
 	return (*ret)? ret : std::shared_ptr<std::fstream>();
 }
 
+std::shared_ptr<std::istream> FileFinder::openUTF8Input(const std::string& name,
+	std::ios_base::openmode m)
+{
+	std::shared_ptr<std::ifstream> ret(new std::ifstream(
```
fmatthew5876: `std::make_shared`

src/image_png.cpp
```diff
@@ -31,6 +31,13 @@ static void read_data(png_structp png_ptr, png_bytep data, png_size_t length) {
 	*bufp += length;
 }
 
+static void read_data_istream(png_structp png_ptr, png_bytep data, png_size_t length) {
+	std::istream* bufp = (std::istream*)png_get_io_ptr(png_ptr);
```
fmatthew5876: Use `auto` on the left and `reintrerpret_cast` on the right.

src/image_png.cpp
```diff
@@ -31,6 +31,13 @@ static void read_data(png_structp png_ptr, png_bytep data, png_size_t length) {
...
+	if (bufp != NULL&&*bufp) {
```
fmatthew5876, line 36: `nullptr` and whitespace

src/output.cpp
```diff
@@ -249,8 +249,8 @@ bool Output::TakeScreenshot() {
 }
 
 bool Output::TakeScreenshot(std::string const& file) {
-	std::shared_ptr<std::fstream> ret =
-		FileFinder::openUTF8(file, std::ios_base::binary | std::ios_base::out | std::ios_base::trunc);
+	std::shared_ptr<std::ostream> ret =
```
fmatthew5876: I would use `auto` here

src/player.cpp
```diff
@@ -1034,21 +1043,16 @@ std::string Player::GetEncoding() {
...
-
-		std::vector<std::string> encodings;
-		std::ifstream is(ldb, std::ios::binary);
-		// Stream required due to a liblcf api change:
-		// When a string is passed the encoding of the string is detected
-		if (is) {
-			encodings = ReaderUtil::DetectEncodings(is);
-		}
+		auto ldb_stream = FileFinder::openUTF8Input(ldb, std::ios::ios_base::in | std::ios::ios_base::binary);
+		std::vector<std::string> encodings = ReaderUtil::DetectEncodings(*ldb_stream);
```
fmatthew5876, line 1055: `auto`

src/audio_decoder.cpp
```diff
@@ -105,19 +104,18 @@ class WMAUnsupportedFormatDecoder : public AudioDecoder {
...
+	bool Open(std::shared_ptr<FileFinder::istream> stream) override { return false; }
 	bool IsFinished() const override { return true; }
 	void GetFormat(int&, Format&, int&) const override {}
 private:
 	int FillBuffer(uint8_t*, int) override { return -1; };
 };
 const char wma_magic[] = { (char)0x30, (char)0x26, (char)0xB2, (char)0x75 };
 
-std::unique_ptr<AudioDecoder> AudioDecoder::Create(FILE* file, const std::string& filename, bool resample) {
+std::unique_ptr<AudioDecoder> AudioDecoder::Create(std::shared_ptr<FileFinder::istream> stream, const std::string& filename, bool resample) {
```
fmatthew5876: Is `AudioDecoder` taking ownership of the stream? If so I would just pass `istream&` and forget the `shared_ptr`.

fmatthew5876:
So one general comment instead of a sea of nits.

I don't think we need to use `shared_ptr` for this, in fact we don't event need `unique_ptr`.

The `istream` interface as of C++11 is movable.

https://en.cppreference.com/w/cpp/io/basic_iostream/basic_iostream

Also istream/ostream itself already encapsulates the encapsulates the real I/O inside of filebuf. So using pointers is adding another level of indirection we may not need.

So what I would suggest you explore is if we can just directly return `FileFinder::istream` by value, utilizing it's move constructor. Then it behaves pretty much like a `unique_ptr` and we'll get slightly better performance through fewer indirections.

If it turns out we can't do this, I would try `unique_ptr` over shared ptr. I'm not sure we really need or even want shared semantics for file handles.

src/filefinder.cpp
```diff
@@ -646,26 +646,37 @@ void FileFinder::Quit() {
...
-#ifdef _WIN32
-	return _wfopen(Utils::ToWideString(name_utf8).c_str(),
-				   Utils::ToWideString(mode).c_str());
+Filesystem::InputStream FileFinder::OpenInputStream(const std::string& name,
+	std::ios_base::openmode m)
+{
+	std::streamsize size = FileFinder::GetFileSize(name);
+	auto *buf = new std::filebuf();
+
+	Filesystem::InputStream ret(new Filesystem::InputStreamRaw(buf->open(
```
fmatthew5876, line 655: `std::make_shared`

src/filesystem.h
```diff
@@ -0,0 +1,88 @@
...
+constexpr std::ios_base::seekdir Filesystem::CSeekdirToCppSeekdir(int origin) {
+	switch (origin) {
+		case SEEK_SET:
+			return std::ios_base::beg;
+		case SEEK_CUR:
+			return std::ios_base::cur;
+		case SEEK_END:
+			return std::ios_base::end;
+		default:
+			assert(false);
```
fmatthew5876, line 69: This should fail to compile in debug builds. Sadly, you can't put `assert()` in constexpr functions.

src/image_xyz.cpp
```diff
@@ -70,13 +71,11 @@ bool ImageXYZ::ReadXYZ(const uint8_t* data, unsigned len, bool transparent,
...
-bool ImageXYZ::ReadXYZ(FILE* stream, bool transparent,
+bool ImageXYZ::ReadXYZ(Filesystem::InputStream& stream, bool transparent,
 					   int& width, int& height, void*& pixels) {
-	fseek(stream, 0, SEEK_END);
-	long size = ftell(stream);
-	fseek(stream, 0, SEEK_SET);
+	long size = stream->get_size();
 	std::vector<uint8_t> buffer(size);
-	long size_read = fread((void*) &buffer.front(), 1, size, stream);
+	long size_read = stream->read(reinterpret_cast<char*>(&buffer.front()), size).gcount();
```
fmatthew5876, line 78: `buffer.data()`

src/input.cpp
```diff
@@ -57,10 +57,10 @@ bool Input::IsWaitingInput() { return wait_input; }
...
 
 void Input::Init(
-	ButtonMappingArray buttons,
-	DirectionMappingArray directions,
-	const std::string& replay_from_path,
-	const std::string& record_to_path
+		ButtonMappingArray buttons,
+		DirectionMappingArray directions,
+		const std::string& replay_from_path,
+		const std::string& record_to_path
```
fmatthew5876, line 63: whitespace change intentional?

src/output.cpp
```diff
@@ -206,30 +206,28 @@ static void HandleErrorOutput(const std::string& err) {
...
 	char* buf = new char[log_size];
 
-	std::ifstream in;
-	in.open(FileFinder::MakePath(Main_Data::GetSavePath(), OUTPUT_FILENAME).c_str());
-	if (!in.bad()) {
-		in.seekg(0, std::ios_base::end);
-		if (in.tellg() > log_size) {
-			in.seekg(-log_size, std::ios_base::end);
+	auto in = FileFinder::OpenInputStream(FileFinder::MakePath(Main_Data::GetSavePath(), OUTPUT_FILENAME).c_str(),std::ios::ios_base::in);
+	if (in&&!in->bad()) {
```
fmatthew5876, line 218: whitespace

src/output.cpp
```diff
@@ -206,30 +206,28 @@ static void HandleErrorOutput(const std::string& err) {
...
-			out.write(buf, read);
-			out.close();
+			in->getline(buf, 1024 * 100);
+			in->read(buf, 1024 * 100);
+			size_t read = in->gcount();
+			in.reset();
+
+			auto out = FileFinder::OpenOutputStream(FileFinder::MakePath(Main_Data::GetSavePath(), OUTPUT_FILENAME).c_str(), std::ios::ios_base::out);
+			out->write(buf, read);
+			out.reset();
```
fmatthew5876, line 230: Don't need the calls to `reset`

src/platform/3ds/3ds_audio.cpp
```diff
@@ -67,15 +67,15 @@ void CtrAudio::BGM_Play(std::string const& file, int volume, int pitch, int fade
...
+	if (!filestream) {
 		Output::Warning("Couldn't play BGM %s: File not readable", file.c_str());
 		return;
 	}
 
 	LockMutex();
-	bgm_decoder = AudioDecoder::Create(filehandle, file);
-	if (bgm_decoder && bgm_decoder->Open(filehandle)) {
+	bgm_decoder = AudioDecoder::Create(filestream, file);
+	if (bgm_decoder && bgm_decoder->Open(filestream)) {
```
fmatthew5876, line 78: For the `bgm_decoder` why do we have to pass the stream to the create method and then open it again?

src/audio_decoder.cpp
```diff
@@ -105,19 +104,21 @@ class WMAUnsupportedFormatDecoder : public AudioDecoder {
...
 private:
 	int FillBuffer(uint8_t*, int) override { return -1; };
 };
 const char wma_magic[] = { (char)0x30, (char)0x26, (char)0xB2, (char)0x75 };
 
-std::unique_ptr<AudioDecoder> AudioDecoder::Create(FILE* file, const std::string& filename, bool resample) {
+std::unique_ptr<AudioDecoder> AudioDecoder::Create(Filesystem::InputStream stream, const std::string& filename, bool resample) {
 	char magic[4] = { 0 };
-	if (fread(magic, 4, 1, file) != 1)
+	if (stream->read(magic, sizeof(magic)).gcount() == 0) {
```
fmatthew5876, line 118: should it be `!= sizeof(magic)` what if I tried to read a 1, 2 or 3 byte file?

src/filesystem.h
```diff
@@ -0,0 +1,88 @@
...
+	class vfs_istream : public std::istream {
+	public:
+		explicit vfs_istream(std::streambuf* sb, std::streamsize size);
+		~vfs_istream() override;
+
+		std::streamsize get_size() const;
+
+	private:
+		std::streamsize size;
+	};
```
fmatthew5876, line 47: As we saw in https://github.com/EasyRPG/liblcf/pull/343  `seek` and `tell` result in a system call every single time they are used. For I/O which does a lot of telling, if you track the file offset yourself you get a massive speedup. For I/O which does a lot of small seeking, a call to `read()` into an empty buffer can be faster than a small seek.

If we wanted to fix this at the stream level, it would have to be done by implementing our own custom replacement for `filebuf`. Something to consider for future PR's. Maybe it could speed up the audio system?

Once all this work is done, an optimized `filebuf` could be worth exploring.

src/filesystem.h
```diff
@@ -0,0 +1,88 @@
...
+
+	using InputStreamRaw = vfs_istream;
+	using OutputStreamRaw = vfs_ostream;
+
+	using InputStream = std::shared_ptr<InputStreamRaw>;
+	using OutputStream = std::shared_ptr<OutputStreamRaw>;
+
+	class vfs_istream : public std::istream {
+	public:
+		explicit vfs_istream(std::streambuf* sb, std::streamsize size);
```
fmatthew5876, line 40: Implement or delete move operations. `istream` is movable, so I believe your class will inherit it's move constructor, and then if a stream is moved into a new one, you'll get double delete on your rdbuf.

src/filefinder.cpp
```diff
@@ -646,26 +646,37 @@ void FileFinder::Quit() {
...
 }
 
-FILE* FileFinder::fopenUTF8(const std::string& name_utf8, char const* mode) {
-#ifdef _WIN32
-	return _wfopen(Utils::ToWideString(name_utf8).c_str(),
-				   Utils::ToWideString(mode).c_str());
+Filesystem::InputStream FileFinder::OpenInputStream(const std::string& name,
+	std::ios_base::openmode m)
+{
+	std::streamsize size = FileFinder::GetFileSize(name);
```
fmatthew5876, line 652: You should be able to get the file size from the `filebuf`, this can save you from having to open the file twice.

src/decoder_wildmidi.cpp
```diff
@@ -272,7 +297,12 @@ WildMidiDecoder::WildMidiDecoder(const std::string file_name) {
 	}
 	Output::Debug("WildMidi: Using %s as configuration file...", config_file.c_str());
 
+#if LIBWILDMIDI_VERSION >= 1027 // at least 0.4.3
```
fmatthew5876, line 300: Do we need to support old wildmidi still?

src/audio_sdl_mixer.cpp
```diff
@@ -44,6 +44,58 @@
...
+		delete stream_ref;
+
+		context->hidden.unknown.data1 = nullptr;
+		return 0;
+	}
+
+	SDL_RWops* create_StreamRWOps(Filesystem::InputStream stream){
+		SDL_RWops * ret = SDL_AllocRW();
+		// create a new shared pointer to avoid deletion of the content when the scope of this function ends
+		ret->hidden.unknown.data1 = new RefHolder { std::move(stream) };
```
fmatthew5876, line 89: If this was `unique_ptr`, you could drop an indirection by releasing the `unique_ptr` here and deleting the stream itself in `vio_close`. No need for `RefHolder`

src/bitmap.cpp
```diff
@@ -116,11 +116,11 @@ Bitmap::Bitmap(const std::string& filename, bool transparent, uint32_t flags) {
 	else if (bytes > 2 && strncmp((char*)data, "BM", 2) == 0)
 		img_okay = ImageBMP::ReadBMP(stream, transparent, w, h, pixels);
 	else if (bytes >= 4 && strncmp((char*)(data + 1), "PNG", 3) == 0)
-		img_okay = ImagePNG::ReadPNG(stream, (void*)nullptr, transparent, w, h, pixels);
+		img_okay = ImagePNG::ReadPNG(stream, transparent, w, h, pixels);
 	else
 		Output::Warning("Unsupported image file %s (Magic: %02X)", filename.c_str(), *reinterpret_cast<uint32_t*>(data));
 
-	fclose(stream);
+	stream.reset();
```
fmatthew5876, line 123: No need for `reset` unless you want to close the file early on purpose.

src/decoder_fmmidi.cpp
```diff
@@ -50,15 +49,11 @@ int read_func(void* instance) {
 	return fmmidi->file_buffer[fmmidi->file_buffer_pos++];
 }
 
-bool FmMidiDecoder::Open(FILE* file) {
-	this->file = file;
-
+bool FmMidiDecoder::Open(Filesystem::InputStream stream) {
```
fmatthew5876, line 52: If the decoder doesn't take ownership, it should pass by reference to avoid a refcount.

`unique_ptr` makes this easier as you'll get compiler errors unless you either pass it by reference for `std::move` it.

src/decoder_opus.cpp
```diff
@@ -24,26 +24,33 @@
...
-	return ftell(f);
+	return 0;
 }
 
-static int custom_close(void* stream) {
-	FILE* f = reinterpret_cast<FILE*>(stream);
-	return fclose(f);
+static opus_int64 vio_tell_func(void* stream) {
+	auto* f = reinterpret_cast<Filesystem::InputStreamRaw*>(stream);
+	return f->tellg();
```
fmatthew5876, line 44: I'm pretty sure this is `int64_t` everywhere, but maybe we want  a `static_cast<opus_int64>` just in case?

src/decoder_wav.cpp
```diff
@@ -30,24 +30,21 @@ WavDecoder::WavDecoder()
...
+	stream->read(reinterpret_cast<char*>(&nchannels), sizeof(nchannels));
 	Utils::SwapByteOrder(nchannels);
-	fread(&samplerate, 1, 4, file_);
+	stream->read(reinterpret_cast<char*>(&samplerate), sizeof(samplerate));
 	Utils::SwapByteOrder(samplerate);
-	fseek(file_, 6, SEEK_CUR);
+	stream->seekg(6, std::ios::ios_base::cur);
 	uint16_t bitspersample;
-	fread(&bitspersample, 1, 2, file_);
+	stream->read(reinterpret_cast<char*>(&bitspersample), sizeof(bitspersample));
```
fmatthew5876, line 47: This is a good case where doing a single read into a buffer and copying the pieces out would be faster than all these individual reads and seeks.

fmatthew5876:
So one general comment about the overall approach. I really like what you're doing here and I see the goal is to encapsulate I/O so we can load resources from anything. This is great.

We should carefully consider whether we really wanted shared ownership semantics here (i.e. `shared_ptr`).

For resources like files, you really want to know when they are opened and when they get closed. When you have a `shared_ptr`, this becomes difficult because anything could be holding a reference. It's possible to write a bug where you keep the file around too long. We should consider using `unique_ptr` instead. Not only will this make the resource lifetime more explicit, it'll also give us better performance as we don't pay for atomic ref counts anymore.

In fact we don't event need `unique_ptr`.

The `istream` interface as of C++11 is movable.

https://en.cppreference.com/w/cpp/io/basic_iostream/basic_iostream
https://en.cppreference.com/w/cpp/io/basic_istream/operator%3D

Also istream/ostream itself already encapsulates the real I/O inside of filebuf. So using pointers is adding another level of indirection we don't actually need.

Essentially, the `istream` itself can act as a `unique_ptr` for the `filebuf` it contains. By doing this you remove one level of indirection and we don't have to deal with pointers and null checks in our code.

Consider getting rid of the pointers and `RawInputStream` and just using `InputStream` directly by implementing the move operations and `std::move`ing the stream object itself when you want to pass ownership to another entity.

Ghabry:
Thanks for the long and detailed review appreciate it.
Because this is a huge and invasive change it is important to get the overall API correct in the first attempt to avoid more gigantic refactoring.

The usage of ``shared_ptr`` was recommended in a stackoverflow answer but your explanation makes sense. Shared ptr is not really needed here as there is always exclusive ownership on the stream. Will try to get everything (std)moved.

src/decoder_wildmidi.cpp
Ghabry: This feature will be only useful when you want to load instrument files from a non-filesystem path (e.g. zip or APK) and it was actually submitted by CBreiti for us.
So there is no reason to reject older wildmidi Versions here as they will still work perfectly fine.

src/filesystem.h
Ghabry: I think this will be useful to implement. Other streams like compressed archive streams that must be baked by memory for performance will have a fast tell so "Real" Filestreams should have the same improvements
Also even zip reading has the fstream as underlying stream so everything will profit from it.

src/input.cpp
Ghabry: Rebase error :/

src/filesystem.h
Ghabry: C++14 allows non-constexpr functions in constexpr when they are in an untaken codepath when a constant is passed in.

(Is described somewhere here according to other websites but no idea where: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3652.html )

fmatthew5876: As we discussed on irc, this is now working correctly in C++14.

src/filesystem.h
fmatthew5876: For the file only implementation, one might be able to get away with simply creating a new streambuf which wraps `filebuf` and intercepts it's seek and tell calls. This would need to be serparately unit tested and benchmarked carefully. I would definatly save any attempt for later PR.

For the zip file version, the details are notfully clear to me. But the overall design you'd want separate stream objects for each use like the file case. Each of those streams can reference a shared backend with 1 file handle. The buffer could be an LRU or ARC cache shared by all streams, or each stream could have it's own buffer. For this use case having a `shared_ptr` hidden inside the stream class could make a lot of sense. The zipfile resources go away exactly when all streams are closed.

Or would it just be easier to open same zipfile multiple times as read-only? One could cache and index to the file somewhere so new streams can jump directly to the part of the file they need to read without scanning.

fmatthew5876:
Another consideration.

I think at the end of this, we'll want the stream implementation in liblcf. Then we can build tools around zipping and unzip games etc..

If that part gets merged to player first, are we going to have to deal with license discussions to port it to liblcf?

Ghabry:
This sounds again as liblcf as the "catch all" library for everything that could be useful.

At least an efficient fstream would make sense in liblcf because this directly influences the parsing speed. (the Read(std::string) version of the API could use it, then all tools profit from it).

For zip: This adds dependencies to liblcf and zipping/unzipping games is out of scope for this PR. I will only do on-the-fly decompress in memory. Is not clear to me why they should use our unpack tool instead of unar?
The only packing use-case I see is somthing like "Make gamedisk" but this is an editor feature.

---
For everything that is related to "core" functionality like Video/Audio/Input/IO you can assume that I agree on MIT-licensing.

fmatthew5876:
I think best way is to just proceed putting it in player and we can move pieces later as needed.

Ghabry:
Syntax question:
I do ``auto is = FileFinder::OpenInputStream (return InputStream&&)``. How do I move this now up the call hierarchy to make it "owned" by a unique_ptr?

Lets say I do "Audio::Open(is)". What kind of argument must Open accept? ``&`` or ``&&``. How do I assign it to a unique_ptr? I tried different combination with and without std::move and they all give compiler errors.

fmatthew5876:
> Syntax question:
> I do `auto is = FileFinder::OpenInputStream (return InputStream&&)`. How do I move this now up the call hierarchy to make it "owned" by a unique_ptr?

Not sure what you mean here. That syntax is not valid C++.

> 
> Lets say I do "Audio::Open(is)". What kind of argument must Open accept? `&` or `&&`. How do I assign it to a unique_ptr? I tried different combination with and without std::move and they all give compiler errors.

If `Open()` is going to take ownership of the stream, have it take the `unique_ptr<InputStream>` by value. At the callsite, you move it in. If it's going to just use the stream, take it as either `InputStream*` or `InputStream&`.

src/audio_decoder.cpp
Ghabry: added now such a ReadIntoObj, just not using it alot

src/platform/3ds/3ds_audio.cpp
Ghabry: strange API design: Create will return an instance of a suitable audio decoder and Open() will do the actual processing. Should be improved but not here (and not before this PR, I don't want to rebase this :P)

Ghabry:
still WIP, just addressed the shared_ptr part. Also crashes on startup in LOG_FILE for some reason.

I added this shared_ptr -> Movable as a new commit because this is terrible to rebase. If you don't want this I have to squash everything upto "Add Filesystem::Input/OutputStream and use it everywhere."

src/input_source.cpp
```diff
@@ -88,7 +88,7 @@ bool Input::Source::InitRecording(const std::string& record_to_path) {
 	if (!record_to_path.empty()) {
 		auto path = record_to_path.c_str();
 
-		record_log = FileFinder::OpenOutputStream(path, std::ios::out | std::ios::trunc);
+		record_log = std::make_unique<Filesystem_Stream::OutputStream>(FileFinder::OpenOutputStream(path, std::ios::out | std::ios::trunc));
```
fmatthew5876, line 91: This doesn't make sense, you create an `OutputStream` temporary, and then move it into a unique_ptr.

Now that your `OutputStream` itself is movable we don't need the `unique_ptr` at all.

Just do 

`auto record_log = FileFinder::OpenOutputStream(path, std::ios::out | std::ios::trunc);`

If someone wants to use the stream:
````
void use(OutputStream& s);
use(record_log)
```

If someone want so take ownership
```
take(OutputStream s)
take(std::move(s));
```

src/filesystem_stream.h
```diff
@@ -0,0 +1,170 @@
...
+		~InputStream() override;
+		InputStream(const InputStream&) = delete;
+		InputStream& operator=(const InputStream&) = delete;
+		InputStream(InputStream&& is) : std::istream(std::move(is)) {
+			set_rdbuf(is.rdbuf());
+			is.set_rdbuf(nullptr);
+		}
+		InputStream& operator=(InputStream&& is) {
+			std::istream::operator=(std::move(is));
+			set_rdbuf(is.rdbuf());
```
fmatthew5876, line 41: Need to test against self assignment here. i.e. `x = std::move(x)`

src/filesystem_stream.h
```diff
@@ -0,0 +1,170 @@
...
+	class InputStream : public std::istream {
+	public:
+		InputStream() = default;
+		explicit InputStream(std::streambuf* sb, std::streamsize size);
+		~InputStream() override;
+		InputStream(const InputStream&) = delete;
+		InputStream& operator=(const InputStream&) = delete;
+		InputStream(InputStream&& is) : std::istream(std::move(is)) {
+			set_rdbuf(is.rdbuf());
+			is.set_rdbuf(nullptr);
```
fmatthew5876, line 37: You forgot to move `streamsize`. Since we're setting rdbuf to nullptr in `is`, we should also make `is.streamsize = 0`

src/filesystem_stream.h
```diff
@@ -0,0 +1,170 @@
...
+#define EP_FILESYSTEM_STREAM_H
+
+// Headers
+#include <cassert>
+#include <istream>
+#include <ostream>
+#include "utils.h"
+
+namespace Filesystem_Stream {
+	class InputStream : public std::istream {
```
fmatthew5876, line 28: Since these stream objects are going to be used direct and not behind a pointer, should we make them `final` so people don't try to inherit from them and get slicing behavior?

Also if this is final. the `size` member can be `private` and not `protected`

Ghabry: The ``protected`` was a logic error. I thought the Filesystem-Implementations will inherit from it but they will all only implement there own buffer objects, so this can be made ``final+private``

src/filefinder.cpp
Ghabry: I'm a bit indifferent here. Which of the ways if faster should be also explored via benchmarking. 
At least for compressed streams seek-to-end and back to begin is bad because it forces a full decompression that could be avoided in some cases.

Also the file size isn't really used a lot except for some buffer preallocation optimisations. Imo we can remove it completely later and just read until eof (but needs more testing if this breaks some libs)

Ghabry:
Imo this is ready for reviewing & careful playtesting (any IO errors) on different platforms.

I will still improve some stuff but this is 99% ready.

fmatthew5876:
I tested some games with this including HH3. LGTM!

This one is a potential source of rebase hell. So I would prefer it goes in sooner rather than later.

Ghabry:
I try to remove the GetSize() call. In all except two places it can be replaced with ``Utils::ReadStream``.

Problematic is libsndfile and SDL RW Ops. They both want a filesize function. But I guess both will be happy without a size information because not all streams are seekable or have a known size.

fmatthew5876:
Optimization which can be done later:

- [ ] `Utils::ReadStream` should take the vector by reference and fill it in. This will allow clients to reuse a single buffer and not allocate every time. You can provide 2 overloads, an optimal one which the user passes the buffer, and a convenience once which allocates and returns it.
- [ ] For large files, growing linearly is not efficient. Try starting with 8k and doubling the size each time.

Ghabry:
The XMP IsModule test does not go through the Filesystem stuff.
I submitted a patch to them in January 2019 which is not merged yet :(
https://github.com/cmatsuoka/libxmp/pull/134

Ghabry:
XMP is finally being developed again and my patch was integrated :+1:

Issue #2183, opened 2020-04-30
Charset: Wrong facing direction 


zell180:
#### Name of the game:
Saint Seiya -Il Mito-

#### Player platform:
Android

#### Attach files (as a .zip archive or link them)
![95170714_254785425646180_8719222479228239872_n](https://user-images.githubusercontent.com/6542956/80708183-8619b200-8aeb-11ea-9aa2-cb446c3759ec.jpg)

#### Describe the issue in detail and how to reproduce it:
In this screen charset Marin (the girl) should watch Seiya (the boy) in down direction but during this scene the charset rotates automatically itself on right.
To test this part of game 
Start Game -> Select "Medio" -> Skip Intro -> Select "La Rinascita" and you will be there after the first intro scene :D

Ghabry:
Good news: This is NO regression ;)

Ghabry:
Here a testcase to reproduce this.
There are some messages, after a while the chara will look to the right. Then looks right while moving down. In RPG_RT faces downwards while moving down.

[issue-2183.zip](https://github.com/EasyRPG/Player/files/4560250/issue-2183.zip)

https://easyrpg.org/play/master/?game=issue-2183

fmatthew5876:
This one is rather simple. RPG_RT resets the "direction fixed" flag on page change and player doesn't.

Much simpler test case:

EV01 page 1 with custom move route
```
Direction Fix On
Move right
Sw1 on
```

page 2, activiated by SW01
```
Move up
```

The event should move right while facing down, then move up while facing up

fmatthew5876:
Move Route Page Change Test cases

For each of these EV01 has 2 pages, page 1 and page 2 activiatedby SW01

1. custom routes
--------------------

Page 1 Trigger custom move route freq 8 `Move Right, SW1 on: repeat`
Page 1 Trigger custom move route freq 8 `Move Up: repeat`

Result: Right, Up, 

2. custom routes
--------------------------

Page 1 Trigger custom move route freq 7 `Move Right, SW1 on: repeat`
Page 1 Trigger custom move route freq 7 `Move Up: repeat`

Result: Right Up

3. event routes
-----------------

Page 1 *
```
Set MoveRoute: This Event, Repeat, Right, SW1 on
WaitForAllMovement
```

Page 2 *
```
SetMoveRoute: This Event, Repeat, Up
```

Results:

| Page 1 | Page 2 | Result |
| -- | -- | -- |
[...]

PR #2193, opened 2020-05-03
Reset facing lock flag on event page change

fmatthew5876:
This fixes the direction issue reported by the user in #2183 It does not fix the other timing issues I found with movement in that issue.


Ghabry:
Hm. This fixes the problem in my test case but not in the game Il Mito :(.
So my test case was bad...
Will try to find a way to give you the relevant part of this unpublished game.

Ghabry:
@zell180

If you want to test:
https://ci.easyrpg.org/job/player-linux-pr/4136/ or https://ci.easyrpg.org/job/player-win32-pr/4299/

PR #2201, opened 2020-05-07
Fix syntax error in libretro port

guoyunhe:
Array syntax error. `,` should be placed after each element.

### DIFF
--- a/src/platform/psp/psp_input_buttons.cpp
+++ b/src/platform/psp/psp_input_buttons.cpp
@@ -23,19 +23,19 @@
 
 Input::ButtonMappingArray Input::GetDefaultButtonMappings() {
 	return {
-		,{DECISION, Keys::JOY_1} // Circle
-		,{DECISION, Keys::JOY_2} // Cross
-		,{CANCEL, Keys::JOY_3} // Square
+		{DECISION, Keys::JOY_1}, // Circle
+		{DECISION, Keys::JOY_2}, // Cross
+		{CANCEL, Keys::JOY_3}, // Square
 
-		,{TOGGLE_FPS, Keys::JOY_0} // Triangle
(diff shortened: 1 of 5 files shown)


Ghabry:
Oh. We forgot to test libretro before merging the PR :/. Sorry.

Could you just remove the leading comma on line 30 and keep it as is. Thanks!

There was a reason for the leading commas:

>>
> Ghabry: Is there a reason for using leading commas?
>
> fmatthew5876: Yes, with leading commas you don't need to worry about the last element being different. This makes it easier handling all these #ifdef

guoyunhe:
For modern C++ compilers, the last comma will always be ignored.

You can check the code of RetroArch for reference https://github.com/libretro/RetroArch/blob/master/retroarch.c#L411

fmatthew5876:
He's right this was was my mistake we don't need to do this anymore.

We should have all our files consistent.

@guoyunhe Would you mind also applying this change to `input_buttons_desktop.cpp` and all the `platform/*/*_input+buttons.cpp`?

Thanks for the patch!

Ghabry:
Well I don't want to outsource the work of cleaning up our code base to external contributors.
I can push a commit to the PR to fix this.
Thx for the contribution and for moving to the upstream repo for libretto opensuse :)

Ghabry:
Jenkins: test this please

src/platform/psp/psp_input_buttons.cpp
```diff
@@ -23,19 +23,19 @@
 
 Input::ButtonMappingArray Input::GetDefaultButtonMappings() {
 	return {
-		,{DECISION, Keys::JOY_1} // Circle
```
Ghabry, line 26: ^-- same syntax error as in libretro input

Issue #2207, opened 2020-05-11
Yume2kki: Shogi Board debug save difference


purupurupuruin:
#### Name of the game:
Yume2kki (Version 0.112 without hotfix patches)

#### Player platform:
Linux

#### Attach files (as a .zip archive or link them)
[shogiboarddebugsavebug.zip](https://github.com/EasyRPG/Player/files/4606975/shogiboarddebugsavebug.zip) 
[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/4606977/easyrpg_log.txt)

#### Describe the issue in detail and how to reproduce it:
Since I was told any incompatibility with debug saves is a bug I decided to report this.
In Shogi Board, loading a debug save in Player or entering the area normally the map looks like this:
![shogiunbug](https://user-images.githubusercontent.com/64451846/81513727-53d43580-92df-11ea-9db4-085c7011144f.png)

However, loading a debug save made in Shogi Board in RPG_RT the area looks like this:
![shogibug](https://user-images.githubusercontent.com/64451846/81513726-533b9f00-92df-11ea-97a4-19aaf715c04a.png)
The flying shadows are all glitched looking.
To reproduce:
I attached 2 saves
Save 1 is immediately before Shogi Board, Save 2 is inside Shogi Board
Load the second save in both Player and RPG_RT, wander around the area to see the difference.
Then load the first save in RPG_RT and interact with the pit, Shogi Board looks normal, like when loading the second save in Player.

Ghabry:
This is fixed by #2208 

When loading the attached Player save in #2208 the shadows look glitched the same way as in RPG_RT.
When saving again with #2208 Player and loading in RPG_RT the shadows look correct :+1:

Old savegames are "corrupted" and not fixed but switching the map solves it.

Issue #2197, opened 2020-05-04
Barrels do not roll


zell180:
#### Name of the game:
Saint Seiya -Il Mito-

#### Player platform:
Android

#### Attach files (as a .zip archive or link them)
[barrels_not_roll.zip](https://github.com/EasyRPG/Player/files/4575441/barrels_not_roll.zip)

#### Describe the issue in detail and how to reproduce it:
Load the save and talk to the boy in front of you, select one of the choices (first play 1 level of game, 4 play 4 levels) and a minigame starts.
If you be hitten by the barrel you lose a life and the game restart but in EasyRPG the barrels no longer roll

Ghabry:
It seems that when two event pages have **Exactly the same custom move commands** then the move command **continues** at the same position. Just with the new frequency settings applied. WTF.

fmatthew5876:
Confirmed in the page reset function at `0x4AAF50` in dynrpg RPG_RT. Happens near the end.

```
if (old_page == nullptr || old_move_route != new_move_route) {
   original_move_route_index = 0;
}
```

Ghabry:
Are you sure this is the line for this? This looks like the normal index reset on page change code.
For determining whether the move route is the same I expect some loop or memcmp on the chunk.

fmatthew5876:
`0x4AAF50` is the start of the page change function. `0x4AB200` calls a function which compares the arrays.

In this case it's not a true `memcpy`, it looks like it only checks the command codes. So if you had for example if you had

`Move Right, Sw1 on` `Move Right, Sw2 on`, they would still compare equal and the index would not be reset. This should be verified with a test.

Ghabry:
Just tested it: Your observation is correct. Sometimes reverse engenering is useful to find edge-cases in edge-cases -_-

Ghabry:
This seems to be also the case when you set the Move Route with "Change Move Route" and then switch to a new page with a Custom route that matches the "Event cmd move route".

Though I observe something strange:

I have RRRRDDDDLLLLUUUU for both. Then I switch the page (movement becomes slow) and somehow when the initial route was custom the U movement is suddenly fast again.

Will share the map you will see what I mean:

Next to you (tile on the right) is an event that switches the page:

[Map0003.zip](https://github.com/EasyRPG/Player/files/4587373/Map0003.zip)

Ghabry:
Ah nevermind. I had "Sw1" turned on in my tests which was the page trigger. Still a good test maybe ^^

[...]

Issue #2196, opened 2020-05-04
Charset disappear after talk


zell180:
#### Name of the game:
Saint Seiya -Il Mito-

#### Player platform:
Android

#### Attach files (as a .zip archive or link them)
[Save01.zip](https://github.com/EasyRPG/Player/files/4575272/Save01.zip)

#### Describe the issue in detail and how to reproduce it:
After talk with NPC she disappear. Select "No, ho troppo sonno adesso" to short talk

Ghabry:
This is slowly becoming too much work for me:

1. Can I share the game you provided to me with other devs?
2. If not: Please create a minimal testcase to reproduce this

zell180:
Solved. Error in our code.

zell180:
I've shared the game. Issue persist, if you use save not happens.

fmatthew5876:
I can reproduce this with the second save zell180 uploaded.

This one has something to do with something in the save game. If you edit and save the map, the bug no longer happens. If you then comment out the `save_count` code, the bug occurs again.

It happens immediately after Event 349 / page 3 finishes executing.

zell180:
I think this information can help you: the bug occurs even if you get to that part of the game without loading the save

fmatthew5876:
This happens right when the event finishes executing. I added a message at the end of the event to verify.

1. Turns on switch for page 4
2. Refresh runs, setting up new page
3. Message appears
4. Message closes
5. sprite dissapears

If you edit the game map, you have to disable the `save_count` logic in `Game_Map`, otherwise the bug will not happen.

fmatthew5876:
Also checked with ubsan and address sanitizer. Nothing on this bug

fmatthew5876:
Ok I've figured this out.

The save game has `direction=4` (UpRight) and `sprite_direction=1` (right)

When you talk to the event, her `sprite_direction` changes to face you. Once the event finishes, Player will do `sprite_direction = direction`.

After that the sprite rendering code computes a bounding rect with `sprite_direction == 4` which ends up out of bounds. So you get a `Sprite::src_rect` with `height == 0` and this is what causes the event to dissapear.

This also explains why the bug doesn't occur when `save_count` logic triggers, because all the events get reset. Her direction must have changed to 4 from a previous event and saved into the save game.

Resetting the events `sprite_direction` after the event finishes is correct, but we do it wrong. First, in RPG_RT `sprite_direction` or `facing` as they call it can never be a diagonal direction.

[...]

Issue #1859, opened 2019-07-31
Interpreter PR Todo Items


fmatthew5876:
While we slog through the interpreter PR's here is the list of todo items which are being postponed to avoid rebase issues

- [x] Remove `Game_Map::RefreshType`
- [x] Cleanup `Game_Interpreter::GetFrame()`
- [x] Streamline `Game_Map::RequestMap()` logic in one place. Currently in `Game_Player`, `Scene_Map`, and `Scene_Debug`. Should be after #1784 is merged.
- [x] Remove Async globals and store the async reason and parameters in the `AsyncContext`
- [x] Should we verify save game command chunks against event commands in #1884? How does RPG_RT do it?
- [x] Are level up / skill learned messages word wrapped? Check EXP, Level, and Class Change.
- [x] Can self destruct / escape be undone with show enemy command? - YES
- [ ] Verify what happens if we change event page after event erased. What happens to LSD chunks?

Ghabry:
TBD: Fix performance issues caused by #1740 . Need to confirm my test was correct when back home.

fmatthew5876:
Confirmed the following in RPG_RT:

- [x] Enemy escapes, event show enemy - reappears
- [x] Enemy self-destructs, event show enemy - reappears
- [x] Enemy dies, event show enemy - stays dead

fmatthew5876:
Remove the MIDI bullet as thats already tracked by #1733

PR #2208, opened 2020-05-11
Refactor Game_Character classes + add move route test suite

fmatthew5876:
This is a large refactor for the `Game_Character` class heirarchy.

Depends on: https://github.com/EasyRPG/liblcf/pull/372

Goals
------

- [x] Remove use of `Main_Data:;game_data` from characters and map #1954
- [x] Restructure `Game_Character` to match RPG_RT design more closely. This will fix bugs we don't even know about yet, make it easier to fix all of the current bug set, and make it very easy to fix bugs discovered later.
- [x] Fix remaining incompatibilities with RPG_RT w.r.t characters and move routes
- [x] Simplify code and remove global variables

Issues Fixed
------------

- [x] Fix: #1645 
- [x] Fix: #2197 
- [x] Fix: #1859
- [x] Fix: #1852 
- [x] Fix: #2183
- [x] Fix: #2196 
- [x] Fix: #2207
- [x] Fix: #2224 
- [x] Fix: #2255 
- [x] Fix: #2286
- [x] Fix: #2316

Unit Test coverage:
- [ ] Fix: #1567 
- [ ] Fix: #1626 
- [ ] Fix: #1675 
- [ ] Fix: #1691 
- [ ] Fix: #1699 
- [x] Initialization
- [x] Movement / speed / frequency / stop count
- [x] Character Animation Types / frames
- [x] Vehicle animation frames
- [x] Move Routes
- [ ] Vehicle boarding / unboarding
- [ ] Vehicle ascent / descent
- [ ] Player movement when boarded syncs to vehicle
- [ ] Event Triggering
- [x] Flash
- [ ] Bush Depth
- [x] MoveTo()
- [ ] Player input and movement
- [ ] Player input and menu / encounter calling + situations which clear the flags
- [x] Panning
- [x] Save Counts
- [ ] Through flag properly set and reset in places
- [ ] MakeWay movement - maybe punt to later PR
- [ ] MakeWay jumping - maybe punt to later PR
- [ ] MakeWay land/disembark - maybe punt to later PR
- [ ] Event page refresh and page conditions

Additional Incompatibilities discovered and fixed
------------------

- [x] CommandHeroTranspareny resets the through flag
- [x] Verify stop animation flag is ignored for spin
- [x] through flag correctly used
- [x] Several additional cases when `menu_calling` and `encounter_calling` are cleared
- [x] airship will trigger collision events but not touch events. And only on `Game_Event`'s update sequence.
- [x] GetOnOffVehicle replaces diagonal directions with `sprite_direction`
- [x] Fix incorrect timing of spin with speed 2
- [ ] Post-boarding sets `phasing = hidden || slipThrough` ??

Testing Plan
-------------

- [x] Add a unit test suite
- [ ] Test a list of unique edge cases for this PR
- [ ] Retest the issue list
- [ ] A code reviewer should independently review this PR against RPG_RT in an RE tool. 
- [ ] Play a few games like a normal user would

Issue Retest List
----------

- [x] #1871 - retested a few cases
- [x] #1878 - tested Yume2kki music room
- [x] #1783  - tested a few multi-teleport cases
- [x] #1732
- [x] #1704 
- [x] #1707 - only move route chunk related
- [x] #1699 
- [x] #1696 
- [x] #1693 
- [x] #1691 
- [x] #1664 
- [x] #1659 
- [x] #1638 
- [x] #1604 
- [x] #1588 
- [x] #1936 
- [x] #1937 
- [x] #1938 
- [x] #579 
- [x] #937
- [x] #1267 
- [x] #1293 
- [x] #1346 
- [x] #1429 
- [x] #1585
- [x] #1587 
- [x] #1593 
- [x] #1654 
- [x] #1671

Lcf Chunk Renames to follow
---------
- [ ] move_route.repeated -> done
- [ ] sprite_direction -> facing
- [ ] through vs route_through -> something?
- [ ] sprite_transparent -> sprite_hidden

Dynrpg Info
-------------

Vtable and Structure layouts:
https://github.com/fmatthew5876/DynRPG/blob/master/Character.h

| Address | Class | Symbol |
| -- | -- | -- |
| 004C3D60 | Character | VTABLE |	
| 004A89A0 | Hero | VTABLE |
| 004B3BF4 | Vehicle | VTABLE |
| 004AAC8C | Event | VTABLE |
| 004A35D0 | SceneMap | Update() |
| 004AAF50 | SceneMap | RefreshAllPages() |
| 004B1938 | Interpreter | OnFinishEvent() |
| 004AEBD4 | Interpreter | CommandTeleport() |
| 004AEF30 | Interpreter | CommandSetEventLocation() |
| 004AFBF0 | Interpreter | CommandSetMoveRoute() |
| 004AF09C | Interpreter | CommandSwapEventLocation() |
| 004AD618 | Interpreter | CommandChangeParty() |
| 004AF610 | Interpreter | CommandPanScreen() |

Notes to self
-----------
- [x] Vehicles pending move on load / map change
- [x] Visible flag only used by hero, we use it in other places
- [x] CommandHeroTranspareny resets the through flag
- [x] Embarking a vehicle resets the through - test this?
- [x] Teleport / vehicle location / event location commands - review direction and facing
- [x] move route on inactive event
- [x] Check command halt all movement
- [x] Collision code sets menu calling
- [x] Get on/off vehicle not blocked by boarding or unboarding?
- [x] Check change party


fmatthew5876:
| Address | Class | Name |
| -- | -- | -- |
| 004C1150 | Battler | ShouldMirrorBattleAnimations() - not used in this PR, but shows how battle aninimation mirroring depends on battler mirrored flag |
| ??? | SceneBattle | Logic which makes the character turn only if they perform an action on a single enemy - I need to find this again |
| 004977A4| SceneBattle | InitBattle() - initializes the battle all calls all the below routines |
| 004B4910 | SceneBattle | InitBattleActors() - does actor placement |
| 004BE8E0 | SceneBattle | InitBattleMonsters() - filters encounter condition and does enemy placement |
| 004B4B6C | SceneBattle | ResetPartyDirections() - Resets mirrored flag for all party at end of battle init sequence |
| 004BED2C | SceneBattle | ResetMonsterDirections() - Resets mirrored flag for all monster at end of battle init sequence |

This is rebased but the rebase was a little nasty. I need to retest again.

Ghabry:
btw you can use ``REQUIRE`` with ``==, >, <, ...`` instead of ``REQUIRE_EQ/_GT``. It will still show the evaluated values when the test fails.

fmatthew5876:
Interesting jump edge case:

```
Move UpRight
Jump
Move Forward
Land
Move Forward
```

So when a jump lands the `UpRight` direction gets replaced by Up

fdelapena:
The PR fixes event collisions from airship, at least when calling battle, which is currently broken in 0.6.2.1 and master:

vehibug

Reported by @MudarraP

fmatthew5876:
This is finally ready to be reviewed.

I still need to work through a lot of test cases but this is now in it's final state. I don't expect major changes from here, only bug fixes.

I didn't get to do all the full unit test coverage that I wanted to, but a significant portion is covered, including all the fundamental per frame behaviors of basic move routes.


This PR is essentially a rewrite of the entire `Game_Character` class hierarchy based on RE tools. The code is structured like RPG_RT so long term we will approach 100% correctness here very quickly.

It goes without saying that despite anyones best efforts, there will be regressions. Therefore I'd like us to take a hard think about how much is realistic via code review, how much testing each of us can realistically achieve on our own, and when it makes sense to just let it go to android testers so they can help us find bugs.

What I don't want to have happen is this sets for months. That is only going to make the situation worse as the code and my memory of it bitrots.

Know that once this goes in, I will be available and 100% committed to fixing regressions as they are found.

Ghabry:
The compromise I can offer is that

- This is pushed to the Android BETA channel (currently up to 2500 testers)
- We ask for tests on Twitter etc.
- any PRs that conflicts with this will be delayed.

The review window will be around 4 weeks. This is so complex that it needs at least a weekend or two for proper reviewing & playtesting.

Ghabry:
Regression:
Get Yume2kki 0.113a from https://yume2kki.fandom.com/wiki/Downloads

Load this save: 
[Save06.zip](https://github.com/EasyRPG/Player/files/5142931/Save06.zip)

You see miniteleporters (they activate on touch)
After walking on them the animation stops but in current master and RPG_RT they continue (looks like some animation type continuous page change bug)

fmatthew5876:
> Regression:
> Get Yume2kki 0.113a from https://yume2kki.fandom.com/wiki/Downloads

Nice catch! Fixed

Ghabry:
Patch for the emscripten test runner:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index c4f9c88c..6b3747a7 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -891,6 +891,10 @@ if(PLAYER_ENABLE_TESTS)
 	endif()
 	add_dependencies(check_player test_runner_player)
 	add_dependencies(check check_player)
+
+	if(CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
+		set_property(TARGET test_runner_player APPEND_STRING PROPERTY LINK_FLAGS "-s ALLOW_MEMORY_GROWTH")
+	endif()
 endif()
 
 # Instrumentation framework
```

Ghabry:
I still have problems with synchronizing the inputs but I can see that for the first 65 frames (this is move right, it desyncs on move up) that RPG_RT and your PR are exactly the same!

For comparison: Master desyncs on Frame 6.

So even this very basic test already shows improvements.

Ghabry:
Regression:

[Map0003.zip](https://github.com/EasyRPG/Player/files/5146900/Map0003.zip)

On this map (uses 2k3 RTP) EV0006 faces right in your PR and down in 0.6.2.1 and RPG_RT.

fmatthew5876:
> Regression:
> 
> [Map0003.zip](https://github.com/EasyRPG/Player/files/5146900/Map0003.zip)
> 
> On this map (uses 2k3 RTP) EV0006 faces right in your PR and down in 0.6.2.1 and RPG_RT. I think the sprite direction is always off-by-one. In your PR it is one higher than usual. Tested multiple directions.

Another simple one, rebased to master and fixed.

Ghabry:
More accuracy stuff: The initialisation appears to be off-by-one:
I recorded now based on a hook after ``++frameCounter`` in RPG_RT and after ``IncFrame`` in Player.

[Map0002.zip](https://github.com/EasyRPG/Player/files/5147390/Map0002.zip)

A "Move right" is injected in Frame 50, which is processed in Frame 51. As you can see the stop count of the hero is correct after the move.

[frames.zip](https://github.com/EasyRPG/Player/files/5147405/frames.zip)

fmatthew5876:
> I recorded now based on a hook after ++frameCounter in RPG_RT and after IncFrame in Player.

This is what is causing the difference. RPG_RT increments the frame counter *before* updating the scene, player does it at the end.


Detailed RPG_RT behavior:

* Increments the frame counter at the beginning of each frame
* SceneTitle: ++frame_counter
* When new game is selected:
* Calls a function which resets all game objects, including setting the frame counter to 0
* Sets up a teleport using the start position
* Calls the same `FinalizeTeleport` function used by event teleport, which runs pre-update (frame_counter == 0 when this happens)
* SceneMap: ++frame_count (==1)
* Runs the first map update


So essentially what happens here is the first time you hook RPG_RT, you only capture pre-update. When you hook player, you capture both pre-update and the first update.

So the fix here is to move the player frame increment before the first call to update.

Ghabry:
this works almost flawless now. When I apply my Player patch below I get a perfectly synched input session (until a msg box opens because of subcommand and active flag).

the RPG_RT structure is

```
++frameCounter;
UpdateInput(); <- My hook (Input and Save) is here
switch (current_scene)->Update
```

In Player we will need something like this:

```
	if (update_scene) {
		IncFrame();
	}

	Input::Update(); <- Record/Inject input
	Audio().Update();

	if (update_scene) {
		/* Saved here */

		Scene::instance->Update();
	}
```

It would make more sense when I save after the update because then you get the state after the frame finished but the current solution only needs one hook so I prefer it (and in the end its the same, only the frame counter in the LSD differs by one).

fmatthew5876:
Added `message_active`, so now you should be able to match on that one.

Ghabry:
Together with my subcommand-hook I get now a perfect match.
Recording of anything on the map that is not depending on RNG should be possible now :)

Ghabry:
Status update: 

I need a bit more time but I add a hard deadline here now:
The merge is in the last full September week (21-27).
Based on the last commits it is unlikely that I find a bug that needs a significant change so feel free to create new PRs that use this one as a base.

Ghabry:
Could you rebase this?

I currently get:

```cpp
FAILED: CMakeFiles/check_player 
cd /data/tmp/easyrpg && /data/tmp/easyrpg/test_runner
easyrpg/easyrpg-player/tests/doctest.h:1114:5: runtime error: load of value 4294967295, which is not a valid value for type 'const lcf::rpg::EventPage::Trigger'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior easyrpg/easyrpg-player/tests/doctest.h:1114:5 in 
=================================================================
==28382==ERROR: AddressSanitizer: global-buffer-overflow on address 0x55a7b8ddc044 at pc 0x55a7b652d06b bp 0x7ffd578bb430 sp 0x7ffd578babe0
READ of size 1 at 0x55a7b8ddc044 thread T0
    #0 0x55a7b652d06a in __interceptor_strlen.part.0 (/data/tmp/easyrpg/test_runner+0x197906a)
    #1 0x7f5b912cd14e in std::char_traits<char>::length(char const*) /build/gcc/src/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/char_traits.h:357:25
    #2 0x7f5b912cd14e in std::basic_ostream<char, std::char_traits<char> >& std::operator<<<std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*) /build/gcc/src/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/ostream:612:44
    #3 0x55a7b6e9818a in Game_Map::SetupCommon() easyrpg/easyrpg-player/src/game_map.cpp:304:6
    #4 0x55a7b6e960a5 in Game_Map::Setup(std::unique_ptr<lcf::rpg::Map, std::default_delete<lcf::rpg::Map> >) easyrpg/easyrpg-player/src/game_map.cpp:136:2
    #5 0x55a7b681dceb in MockGame::MockGame(MockMap) easyrpg/easyrpg-player/tests/mock_game.cpp:46:2
    #6 0x55a7b6799b3d in _DOCTEST_ANON_FUNC_48() easyrpg/easyrpg-player/tests/game_character_anim.cpp:260:17
    #7 0x55a7b6adc429 in doctest::Context::run() easyrpg/easyrpg-player/tests/doctest.h:5856:21
    #8 0x55a7b6ae27d9 in main easyrpg/easyrpg-player/tests/doctest.h:5940:71
    #9 0x7f5b90fc0151 in __libc_start_main (/usr/lib/libc.so.6+0x28151)
    #10 0x55a7b64e58ad in _start (/data/tmp/easyrpg/test_runner+0x19318ad)

0x55a7b8ddc044 is located 60 bytes to the left of global variable 'lcf::DBString::npos' defined in 'easyrpg/easyrpg-player/lib/liblcf/src/dbarray.cpp:16:41' (0x55a7b8ddc080) of size 4
0x55a7b8ddc044 is located 0 bytes to the right of global variable 'lcf::DBArrayAlloc::_empty_buf' defined in 'easyrpg/easyrpg-player/lib/liblcf/src/dbarray.cpp:15:49' (0x55a7b8ddc040) of size 4
SUMMARY: AddressSanitizer: global-buffer-overflow (/data/tmp/easyrpg/test_runner+0x197906a) in __interceptor_strlen.part.0
Shadow bytes around the buggy address:
```

Ghabry:
@fmatthew5876  On some platforms the unit tests are crashing, can you check this?

src/game_character.h
```diff
@@ -940,7 +940,7 @@ inline bool Game_Character::IsFacingLocked() const {
 }
 
 inline void Game_Character::SetFacingLocked(bool locked) {
-	data()->lock_facing = locked;
+	data()->lock_facing = locked || IsDirectionFixedAnimationType(GetAnimationType());
```
Ghabry, line 971: Minor: I find it confusing that SetFacingLocked can silently fail now. From such setters I would expect that they "work"

src/game_map.cpp
```diff
@@ -149,8 +149,11 @@ void Game_Map::ResetPendingMove() {
 	}
 }
 
-void Game_Map::Setup() {
+void Game_Map::Setup(std::unique_ptr<RPG::Map> map_in) {
```
Ghabry: uh, this is nice, this will allow implementing a ``--map`` (without ID) feature

src/game_event.cpp
```diff
@@ -55,6 +55,20 @@ void Game_Event::SetSaveData(lcf::rpg::SaveMapEvent save)
 	data()->ID = event->ID;
 	SetMapId(map_id);
 
+	// Sanity checks on bad save data
+	auto checkMoveRoute = [&](const auto& mr, auto& idx, const auto& name) {
+		const auto n = static_cast<int>(mr.move_commands.size());
+		if (idx < 0 || idx > n) {
+			idx = n;
+			Output::Warning("Event {}: Save Data invalid {}={}. Fixing ...", data()->ID, name, idx);
```
Ghabry, line 63: The warning should mention that it is a MoveRoute

src/game_character.cpp
```diff
@@ -216,6 +216,9 @@ void Game_Character::UpdateMoveRoute(int32_t& current_index, const lcf::rpg::Mov
 	if (is_overwrite && !IsMoveRouteOverwritten()) {
 		return;
 	}
+	const auto num_commands = static_cast<int>(current_route.move_commands.size());
+	assert(current_index >= 0);
+	assert(current_index <= num_commands);
```
Ghabry, line 220: Can this trigger when a hacked LDB is used? If yes this should be a warning and not an assert

Ghabry (commented):
-

src/game_event.cpp
fmatthew5876: The lambda gets passed in `move_route_index` or `original_move_route_index` string name

src/game_character.cpp
fmatthew5876: No, but it can trigger on a corrupted save game. I found this from an unrelated bug, that if the move route index is invalid it causes an infinite loop. The piece of code you previously commented on sanitizes the save data to prevent this from ever happening, and so the assert here is used to make sure that invariant is held.

I will add a comment here explaining this

src/game_character.h
fmatthew5876: This is how RPG_RT does it. To make the interface more sane we would need a careful refactor to ensure the correct behavior is not disturbed.

I say punt for now.

Ghabry (approved):
Let's get the counter close to 200

Issue #2204, opened 2020-05-09
Yume2kki: Cloudy World's Crying Eyes Cave doesn't update properly


purupurupuruin:
#### Name of the game:
Yume2kki (Version 0.112 without hotfix patches)

#### Player platform:
Linux

#### Attach files (as a .zip archive or link them)
[Video demonstrating bug](https://github.com/EasyRPG/Player/files/4602993/cryingeyesvideocomparison.zip)
[Debug save right before cave](https://github.com/EasyRPG/Player/files/4602939/CloudyWorldCryingEyesDebugSave.zip)
[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/4602938/easyrpg_log.txt)

#### Describe the issue in detail and how to reproduce it:
EasyRPG Player:
![cryingeyeseasyrpg](https://user-images.githubusercontent.com/64451846/81468367-fed7d880-9193-11ea-96fd-3db8e0bd3a7a.png)
RPG_RT:
![cryingeyesrpgrt](https://user-images.githubusercontent.com/64451846/81468369-013a3280-9194-11ea-9cb5-df493c6ddaf8.png)

In Cloudy World there's a cave with a pair of crying eyes that show up if you interact with the cliff.
The eyes change depending on some of the effects you can use (Eyeball Bomb, Teru Teru Bozu, and Marginal), but the appearance of the eyes don't always update in EasyRPG player.
Since I'll have a hard time describing all the interactions, I made 2 screenshots showing one possible interaction, and a video showcasing all the various glitchy interactions with the eyes I found.

To reproduce:
Load the provided save and enter the staircase.
Or head to the Crying Eyes Cave in Cloudy World the normal way:
https://yume2kki.fandom.com/wiki/Cloudy_World
Then mess around with the Marginal, Teru Teru Bozu, and Eyeball Bomb effects, as demonstrated in the video.

(Sorry for the the video being laggy.)
(Also sorry for all the bug reports I made so quickly.)

Ghabry:
Nice thing about Yume2kki is that the relevant events are usually at the top-left corner of the map and there is documentation!

This uses Tile substitution to go from Close -> Half Open -> Open.
The eyes are event tiles. Maybe some interaction problem with subsitution and event tiles?

fmatthew5876:
All differences are caused by one issue. A bug in `Sprite` when the bitmap changes while the event is off screen don't cause the effects bitmap to change.

All issues in the video fixed by #2209

PR #2209, opened 2020-05-13
Sprite: Don't reset bitmap_changed state when off screen

fmatthew5876:
Fixes a bug where graphic state changes would be skipped in some
cases when the sprite is off screen.

Fix: #2204


fmatthew5876:
jenkins: test this please

PR #2210, opened 2020-05-13
Scene_Debug: refactor + call map and battle events

fmatthew5876:
- [x] Support calling map event pages
- [x] Support calling battle event pages
- [x] Refactor Scene_Debug to make adding new features easy

### DIFF
--- a/src/game_battle.cpp
+++ b/src/game_battle.cpp
@@ -37,7 +37,7 @@
 #include "utils.h"
 
 namespace Game_Battle {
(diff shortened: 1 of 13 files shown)


fmatthew5876:
https://github.com/rueter37/Player/commits/fixes-0.6.2 has some of the things fixed here.

This PR is too heavy for 6.2.1 but I can pull out those fixes to a separate PR if we want them?

Issue #2198, opened 2020-05-04
Different animation behaviour


zell180:
#### Name of the game:
Saint Seiya -Il Mito-

#### Player platform:
Android

#### Attach files (as a .zip archive or link them)
[rotation_comparison.zip](https://github.com/EasyRPG/Player/files/4577257/rotation_comparison.zip)

#### Describe the issue in detail and how to reproduce it:
Comet animation seems not has the same behaviour in EasyRPG.
To test this part of game
Start Game -> Select "Medio" -> Skip Intro -> Select "La Rinascita" and you will be there after the first intro scene :D

fmatthew5876:
This is us incorrectly emulating an edge case. When a rotation is canceled on a picture, it'll continue rotating until it completes a full revolution.

Somehow we do the math and/or timing wrong and it shows up here.

PR #2211, opened 2020-05-14
Fix accuracy bugs in picture rotation algorithm

fmatthew5876:
Fix: #2198

DynRPG info:

| Address | Class | Name |
| -- | -- | -- |
| 004C1C20 | Picture | Update() |

### DIFF
--- a/src/game_pictures.cpp
+++ b/src/game_pictures.cpp
@@ -44,6 +44,22 @@ static bool IsEmpty(const RPG::SavePicture& data) {
 	return IsEmpty(data, data.frames);
 }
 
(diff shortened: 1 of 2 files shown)


fmatthew5876:
@zell180 please confirm the bug looks fixed to you.

zell180:
For us is fixed

fmatthew5876:
There was a divide by 0 bug hiding in the previous implementation.

Also we can use the copy fast path on frames when `time_left <= 1`. I checked in vtune and this `SyncCurrentToFinish()` optimization is worth it.

PR #2212, opened 2020-05-14
Add support for async message ops - Fix broken inn sequence

fmatthew5876:
Fixes inn when costs money being skipped.

### DIFF
--- a/src/scene_map.cpp
+++ b/src/scene_map.cpp
@@ -426,14 +426,13 @@ void Scene_Map::OnAsyncSuspend(F&& f, AsyncOp aop, bool is_preupdate) {
 
 	if (aop.GetType() == AsyncOp::eCallInn) {
 		activate_inn = true;
+		inn_started = false;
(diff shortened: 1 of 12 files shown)


src/scene_map.cpp
```diff
@@ -480,12 +479,29 @@ void Scene_Map::FinishInn() {
...
+
+	if (!inn_started) {
+		Transition::instance().InitErase(Transition::TransitionFadeOut, Scene::instance.get());
+		inn_started = true;
+
+		AsyncNext([=]() { StartInn(); });
+		return;
+	}
+
+	if (Audio().BGM_IsPlaying() && Audio().BGM_PlayedOnce()) {
```
Ghabry, line 501: Not tested but this is not the inverse of the old condition. They inverse would be ``Audio().BGM_IsPlaying() && !Audio().BGM_PlayedOnce()``

fmatthew5876: Fixed, thanks for catching that.

PR #2215, opened 2020-05-14
Fix Physical attack to hit algorithm + class EXP + crit flash

fmatthew5876:
DynRPG Info:

0x49B4E1 is where the math is done


fmatthew5876:
Class one is easy.

Start a new game, change your class, save, load.

fmatthew5876:
Critical hits @ dynrpg 0x49B64F

PR #2217, opened 2020-05-15
Transitions - wait for graphics loading if screen erased - dependency for battle2k3 PRs

fmatthew5876:
This PR improves support for asynchronous functions during scene initialization. It's required for 2k3 battle and also fixes some issues with maps.

### DIFF
--- a/src/transition.cpp
+++ b/src/transition.cpp
@@ -389,12 +389,15 @@ void Transition::Update() {
 		return;
 	}
 
-
 	// FIXME: Break this dependency on DisplayUI
 	if (transition_type != TransitionNone && !screen2) {
 		// Wait for all graphics to load before drawing screens.
 		if (FromErase() && AsyncHandler::IsGraphicFilePending()) {
 			return;
+		}
+
+		if (scene) {
+			scene->UpdateGraphics();
 		}
 
 		if (!screen1) {
--- a/src/transition.cpp
+++ b/src/transition.cpp
@@ -23,6 +23,7 @@
 #include <cmath>
 
 #include "transition.h"
+#include "async_handler.h"
 #include "bitmap.h"
 #include "game_player.h"
 #include "graphics.h"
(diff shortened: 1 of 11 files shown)


fmatthew5876:
I tested this with `EP_DEBUG_SIMULATE_ASYNC`. 

One good test is to create an event which does
```
Hide Screen
Creates 300 pictures all over the screen
ShowScreen
```

With master, transition in effects look broken. In this PR, the screen stays black until all are loaded and then the transition in looks correct.

This implementation is not complete. It only supports map and battle through the new `Scene::UpdateGraphics()` callback. Once #2218 this will always work properly for all scenes.

Frozen triggers seems like a poor example. Even with this PR, things pop into the scene. Not sure what they are doing but maybe they don't use transitions?

fmatthew5876:
I added some more logic to async defer the call to `Scene::TransitionIn()` if any graphics are loading. 

This is going to be needed in for `Scene_Battle_Rpg2k3`  #2142 to properly sequence LoadBattleGraphics -> Place battlers -> TransitionIn.

It's also fixing bugs we have now of a similar nature with `Scene_Map` where if you chain multiple teleports during scene startup and different maps and graphics need to be loaded.

fmatthew5876:
Tested with this scenario:

Map1 EV01 parallel `Teleport Map2`
Map2 EV01 parallel `Teleport Map3`
Map3

With `EP_DEBUG_SIMULATE_ASYNC`

In master, the player and maps 1, 2, and 3 pop into the scene. In this PR, the screen stays black, and then fades into map 3.

I also verified with debug prints that `Scene::TransitionIn()` does not get called until the end of the chain of async events.

src/transition.cpp
```diff
@@ -391,23 +385,53 @@ void Transition::Draw(Bitmap& dst) {
 }
 
 void Transition::Update() {
-	if (IsActive()) {
-		if (flash_iterations > 0) {
+	if (!IsActive()) {
+		return;
+	}
+
+	// FIXME: Break this dependency on DisplayUI
```
Ghabry, line 392: this FIXME was not supposed to be moved. No DisplayUI here

fmatthew5876: But it is used a few lines down to create the bitmaps.

Ghabry:
This works beautiful for yume nikki. During the intro and e.g. for "loading please wait" all the ugly pop-ins are gone.

Issue #2218, opened 2020-05-15
Remove `UpdateGraphics()` sequence.


fmatthew5876:
Right now we have:
1. Update() - does game logic updates - locked to 60fps
2. UpdateGraphics() - updates the state of drawables based on game logic - locked to 60 fps
3. Draw() - draws drawables - runs at configured fps / vsync

I would argue that (2) is completely unnecessary. In Draw() we can grab the state from the game objects and then draw it directly.

This will simplifiy the main loop, particularly with transitions and drawing to other bitmaps. Every call to draw will be guaranteed correct. It will also simplify the game logic.

PR #2220, opened 2020-05-16
Pictures and Screen: optimize and remove UpdateGraphics() method + fix debug battle pics / weather

### LINKED ISSUE #2120: Battle pictures and weather don't work when starting a battle from the Debug Menu
On the map, either create a picture which is visible in battle or set a weather effect. Now start a battle from debug menu.

The picture and weather will not be visible.

This happens because the `Scene_Debug` at the top of the stack prevents the shared scenes from being transferred.

When fixing this, we should also consider a more elegant way to handle `Game_Screen::InitGraphics()`. Right now it's called in both the map scene and the battle scene for battle test only. Is there a way we can more robustly initialize the shared drawables?

### LINKED ISSUE #2262: [WIP custom battle system] Wrong picture loading at loading save + event behaivour
#### Name of the game:
Technical demo of a custom menu and battle system from a unreleased game (La Leyenda Heroen - 2003, by @Soeufans ).
Download link (with author permission): https://drive.google.com/file/d/16RNJ1Bna3sEH-bERumFv4KerPI0adQ6w/view?usp=sharing

Note: It has lots of patches file and unrelated map files, but do not fear; the game does not uses it. 

#### Player platform:
Tested with lastest master Player (30-jun-2020).

#### Description
I wanted to make a more resumed bug report with test cases, but I think it will better idea upload the whole thing here, beacuse there's several interesting things to deal here. I will enumerate one each one.

#### Wrong picture loading

Steps:
-Start game.
-Press cancel to enter custom menu.
-Select Guardar (Save). Save in any slot.
-Select Cargar (Load), load the saved file. 

Expected behaivour (RPG_RT.exe):
image

Current behaivour (Player master):
image

Note: It's not related about saving. RPG_RT can load both saves from itself and Player correctly, and Player loads saves from itself and RPG_RT incorrectly. 

#### Event behaivour
This is currently a mixture of several ones, I don't know if it's related to the quantity of events, or programming style of the author. 

Steps:
-Start game.
-Talk to the first pillar below.

Expected behaivour (RPG_RT, wrong one?)
<del>MIDI music doesn't play</del> -> EDIT: Fixed, not a bug.

Current behaivour (Player master)
<del>-MIDI music plays.</del> -> EDIT: Fixed, not a bug.
-Lots, and I mean LOTS of warnings in easyrpg.log, like an unoptimized engine or something like that... of these:

````
Debug: Image size out of bounds: System/Azul7 (160x80 < 160x640 < 160x80) [2679x] EDIT -> Non standard system graphic dimensions, explained below. 

Debug: Event 0 exceeded execution limit
````

The first one can be explained that the game uses a non-standard System resource, but it got that debug warning lots of times, I feel something is wrong. 
And the second one, Event 0 maybe came from a common event, but I can't relate who? (Or some sort of a strange/faulty event) ? 

I will try meanwhile to isolate these cases and bring more info If I have more about this case, but I think it will be a bit interesting at least. 

Grettings!

fmatthew5876:
- [x] Fixes picture and save for #2218 
- [x] Optimizes picture drawing
- [x] Optimizes low frame rates at expense of high frame rates
- [x] Simplifies code
- [x] Fix #2120
- [x] Fix #2262

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -255,6 +255,8 @@ add_library(${PROJECT_NAME}
 	src/sprite_battler.h
 	src/sprite_character.cpp
 	src/sprite_character.h
+	src/sprite_picture.h
+	src/sprite_picture.cpp
 	src/sprite.cpp
 	src/sprite.h
 	src/spriteset_battle.cpp
--- a/Makefile.am
+++ b/Makefile.am
@@ -257,6 +257,8 @@ libeasyrpg_player_a_SOURCES = \
 	src/sprite_battler.h \
 	src/sprite_character.cpp \
 	src/sprite_character.h \
+	src/sprite_picture.cpp \
+	src/sprite_picture.h \
 	src/sprite_timer.cpp \
 	src/sprite_timer.h \
 	src/spriteset_battle.cpp \
--- a/src/game_battle.cpp
+++ b/src/game_battle.cpp
@@ -136,7 +136,6 @@ void Game_Battle::UpdateAnimation() {
 
 void Game_Battle::UpdateGraphics() {
 	spriteset->Update();
-	Main_Data::game_screen->UpdateGraphics();
 
 	if (need_refresh) {
 		need_refresh = false;
--- a/src/game_battle.cpp
+++ b/src/game_battle.cpp
@@ -137,7 +137,6 @@ void Game_Battle::UpdateAnimation() {
 void Game_Battle::UpdateGraphics() {
 	spriteset->Update();
 	Main_Data::game_screen->UpdateGraphics();
-	Main_Data::game_pictures->UpdateGraphics(true);
 
 	if (need_refresh) {
 		need_refresh = false;
--- a/src/game_pictures.cpp
+++ b/src/game_pictures.cpp
@@ -203,7 +203,6 @@ bool Game_Pictures::Picture::Show(const ShowParams& params) {
 	data.flags.affected_by_tint = (params.flags & 16) == 16;
 	data.flags.affected_by_flash = (params.flags & 32) == 32;
 	data.flags.affected_by_shake = (params.flags & 64) == 64;
-	last_spritesheet_frame = -1;
 
 	const auto num_frames = NumSpriteSheetFrames();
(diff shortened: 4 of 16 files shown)


fmatthew5876:
VTune Performance measurements
=======================

All tests run with --no-vsync. Results from vtune on ubuntu 20.04 x64.

Value is % of cpu-time spent in function and the things it calls.

All of these tests are the start of Frozen Triggers from 5s to 20s.

master
-------

| --fps-limit | Game_Picture::UpdateGraphics() |
| -- | -- |
| 10 | 2.3% |
| 20 | 1.8% |
| 60 | 1.4% |
| 120 | 1.0% |
| 0 | 0.1 % |

The total percentage of time spent in `Game_Picture::UpdateGraphics()` increases  as the frame rate goes down. This is because at low frame rate scenarios, the game loop and therefore `UpdateGraphics()` is called multiple times per frame. Conversely, at high frame rate scenarios the game loop is called very infrequently and so total time spent in `UpdateGraphics` is much lower.

This PR
--------


| --fps-limit | Sprite_Picture::Draw() | Nominal FPS |
| -- | -- |
| 10 | 0.2% |
| 20 | 0.4% |
| 60 | 0.7% |
| 120 | 1.2% |
| 0 (unlimited) | 1.5% |

Now we replace `Game_Picture::UpdateGraphics()` with `Sprite_Picture::Draw()` and now the relationship reverses. In low frame rate scenarios we only do the state transfer once per frame, whereas in high frame rate scenarios we do it every frame. So we're effectively optimizing low frame rate scenarios at the expense of a high frame rate scenarios, which is the correct trade off to make.

We can see at 60 fps, the optimizations in this PR have reduced the amount of time spent in this state transfer routine significantly. 

Finally, and most importantly this PR simplifies our code a lot. As now our already too complex main loop doesn't have to handle updating the picture graphics state before drawing.

fmatthew5876:
Retested several picture scenarios, including battle pictures.

This PR is ready.

Ghabry:
There was no further feedback since one month, so I'm merging this now.

PR #2223, opened 2020-05-20
Version 0.6.2.1

carstene1ns:


src/version.h
```diff
@@ -29,13 +29,19 @@
...
+#else
 #define PLAYER_VERSION TO_STRING(PLAYER_MAJOR) "." TO_STRING(PLAYER_MINOR) "." TO_STRING(PLAYER_PATCH)
+#endif
 
 /**
  * Version written to the easyrpg_data.version field in savegames.
  * Increment the last digit (0) only when having a good reason.
  */
-#define PLAYER_SAVEGAME_VERSION (PLAYER_MAJOR * 1000 + PLAYER_MINOR * 100 + PLAYER_PATCH * 10 + 0)
+#define PLAYER_SAVEGAME_VERSION (PLAYER_MAJOR * 1000 + PLAYER_MINOR * 100 + PLAYER_PATCH * 10 + 1)
```
Ghabry, line 45: Could be changed to ``PLAYER_TWEAK`` and ``Increment the last digit (0) only when having a good reason.`` because we have a good reason now

Ghabry: actually... it is irrelevant to if we do this here. This targets the stable branch so if we want to keep this change to ``version.h`` it will need another PR to master anyway.

carstene1ns: Yeah, I want to touch that again when we have no more custom makefiles/git versioning.

Issue #2226, opened 2020-05-24
Picture spritesheet chunks are not saved properly when there is no sprite sheet


fmatthew5876:
Start a 2k3e game, create a picture at (160, 120) with transparency 50%

Save your game with player and load it with RPG_RT.

The picture will appear in RPG_RT at (0,0) fully opaque.

The reason is because we don't save the correct values in spritesheet_cols, spritesheet_rows, and spritesheet_frame. My save example had (0,0,-2) where it should be the defaults of (1,1,0).

Issue #2225, opened 2020-05-24
Problem about "batch" in variables


Soeufans:
Im creating my Game Legend of Heroen, to Rm2k3 and porting it with easyRPG

I found some bug, about the batches, if i put a number from a variable into a batch to IDV 0001 to 0001 it does not work properly.

Here is a example with an engine with showing numbers, with RM2k3 there is no problem, but with EasyRPG palyer does not work  as it should

[Caso raro.zip](https://github.com/EasyRPG/Player/files/4674011/Caso.raro.zip)

Thank you for supporting EasyRPG Player.
Before creating an issue, please verify that your bug still occurs in the "continuous builds" of the Player.
You can get them at the bottom of our download section: https://easyrpg.org/player/downloads

---
Please fill in the following fields before submitting an issue:
#### Attach files (as a .zip archive or link them)
 - A savegame next to the problem (if it is not at the beginning of the game)
 - The `easyrpg_log.txt` log file (you can find it in the game directory)

elsemieni:
More information of the bug:

**Test case:**
@> Control Variables: [0001] = -1 
@> Control Variables: [0002] = 1234567 
@> Control Variables: [0001..0001] = Variable [0002]

**Expected output:**
v[0001] = 1234567

**Current output:**
v[0001] = 0 (seems it writes zero)

Tested with:
-Player master 1169 https://ci.easyrpg.org/view/Player/job/player-win32/1169/ 
-RPG_RT 1.12a

fmatthew5876:
This is a regression from the previous range variable fix.

Fixed in #2227

PR #2227, opened 2020-05-24
Picture save game fixes + variable fix

fmatthew5876:
Fix: #2226 
Fix: #2225

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -957,15 +957,11 @@ bool Game_Interpreter::CommandControlVariables(lcf::rpg::EventCommand const& com
 			break;
 		case 1:
 			// Var A ops B
-			if (com.parameters[0] != 1) {
-				value = Main_Data::game_variables->Get(com.parameters[5]);
-			}
+			value = Main_Data::game_variables->Get(com.parameters[5]);
 			break;
 		case 2:
 			// Number of var A ops B
-			if (com.parameters[0] != 1) {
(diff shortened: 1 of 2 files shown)


fmatthew5876:
jenkins: test this please

fmatthew5876:
jenkins: test this please

src/game_interpreter.cpp
```diff
@@ -957,15 +957,11 @@ bool Game_Interpreter::CommandControlVariables(RPG::EventCommand const& com) { /
 			break;
 		case 1:
 			// Var A ops B
-			if (com.parameters[0] != 1) {
-				value = Main_Data::game_variables->Get(com.parameters[5]);
-			}
+			value = Main_Data::game_variables->Get(com.parameters[5]);
```
carstene1ns, line 960: Always interesting to see things getting back to their former state.

fmatthew5876: This was a silly optimization gone wrong.

Ghabry: This is why we need a test suite. Would have catched this. Maybe, is an edge case. 🤔

fmatthew5876:
jenkins test this please

src/game_interpreter.cpp
fmatthew5876: Even if the test suite didn't catch it, once the bug is reported we make a new test and it never happens again no matter how many ways we try to refactor or optimize the underlying code.

PR #2230, opened 2020-05-26
Emscripten: Disable SDL2 async handling.

Ghabry:
This allows the usage of ASYNCIFY_IGNORE_INDIRECT and is much faster.


Ghabry:
closing this, will add a commit to #2253

PR #2239, opened 2020-06-06
Allow debug flag from gamebrowser

tyrone-sudeium:
Adds ability to start a game with debug flag enabled by pressing SHIFT (or its mappings) to launch it instead of ENTER. Also maps the Vita START button to F9 if running with debug flag instead of ESCAPE, which should allow you to open the debug menu with the Vita's START button if running in this mode.

This is useful for playing certain RPGMaker games that are buggy or are otherwise incomplete, since tweaking variables and switches would otherwise be the only method to get past certain events.

This is my first PR so I'm very keen to hear any feedback, especially if I've broken any style guides or other rules. I'm really keen to help out more with this project!

### DIFF
--- a/src/scene_gamebrowser.cpp
+++ b/src/scene_gamebrowser.cpp
@@ -39,6 +39,7 @@ Scene_GameBrowser::Scene_GameBrowser() {
 }
 
 void Scene_GameBrowser::Start() {
+	initial_debug_flag = Player::debug_flag;
 	Game_System::SetSystemGraphic(CACHE_DEFAULT_BITMAP, lcf::rpg::System::Stretch_stretch, lcf::rpg::System::Font_gothic);
 	CreateWindows();
 	Game_Clock::ResetFrame(Game_Clock::now());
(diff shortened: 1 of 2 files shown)


Ghabry:
The Shift for booting with debug options sounds good to me but do you need this button mapping override? You can access the debug menu from the in game menu already.

(And - undocumented - when the game blocks the menu you can still call it with 5x Cancel while holding shift)

fdelapena:
> You can access the debug menu from the in game menu already.

Is menu accessible in testplay mode when the game sets the disable menu flag? If not, those games still require some alternate debug access.

carstene1ns:
Reading the first paragraph and ignoring the second:
like-a-boss

fdelapena:
> Reading the first paragraph and ignoring the second:
> like-a-boss

Oh, I'ive misread the "when the game blocks the menu". Well, this is interesting to know.

tyrone-sudeium:
Also didn't know about that menu trick, hence the mapping! I'm happy to remove it now that I know this exists, if we'd prefer to leave the button mapping untouched.

Ghabry:
btw there is one case that isn't handled: Imo the state of the original debug_flag should be preserved.

Right now pressing Shift will always enable the debug flag for all game runs.

The flag could be saved for both (debug and non-debug) starts and restored in ``Scene_GameBrowser::Continue``

tyrone-sudeium:
Shows my inexperience! I wasn't even aware you could get back to the gamebrowser after picking a game, but it seems obvious in hindsight. I've made the change.

fmatthew5876:
Thanks for the contribution!

Ghabry:
Jenkins: test this please

PR #2243, opened 2020-06-11
Mac: create app bundle, use OpenGL SDL renderer

tyrone-sudeium:
# Summary
Fundamentally the problem I'm trying to solve here is that on macOS right now, starting the Player will always cause your graphics card to switch to its discrete GPU (if it has one). This is a serious problem for battery life on devices, and is completely unnecessary considering the Player doesn't actually use any 3D rendering features (I assume!).

# The Fix...
To solve it, we can make use of macOS's `NSSupportsAutomaticGraphicsSwitching` key in the `Info.plist`.

# Graphics Switching
The Player is currently distributed as a raw Mach-O executable. `Info.plist` keys only work when the app is distributed as a macOS App Bundle. I've written a quick-and-dirty script to take the raw Mach-O executable and all the associated metadata (the `Info.plist`, and the `Player.icns`) and package it up to an `.app` bundle.

Even then, `NSSupportsAutomaticGraphicsSwitching` is only supported for apps that use OpenGL. Apps that use Metal are expected to enumerate all the GPUs on the device and manually select the appropriate ones. Unfortunately, SDL's Metal renderer (the default) always selects the "default" device, which always selects the discrete GPU.

# Regarding App Bundles
I'm not attempting to completely solve the distribution problem here, since my knowledge of autoconf and cmake is elementary at best, and it would also require changes to the Jenkins jobs, of which I'm unfamiliar. This PR is just the first baby steps toward being able to distribute app bundles to end users to make the user experience feel more seamless to the macOS platform. The primary goal of this PR is to fix the graphics switching issue. We can work toward app bundles more seriously later.

# Future Work?

None of this is related to this PR specifically, but I have some thoughts about how to be better citizens on the macOS platform. We should consider code signing the resulting `.app` bundle, and also amending the release pipelines so that when creating a new release, the `.app` bundle is sent off to Apple for [notarization](https://developer.apple.com/documentation/xcode/notarizing_macos_software_before_distribution).

I do this kind of stuff all the time as my full-time job so I'm happy to contribute towards this effort if we agree we want to tackle it!

Another issue that's come up is it's natural on macOS to download an `.app` bundle, drag it to the `/Applications` folder, and double click it from there. However, that'll never work because there'll never be an RPG Maker game in `Player.app/Contents/MacOS`, the working dir when launched this way...

I can work around it for now by doing something like:

    open -a "Player.app" --args --project-path "/path/to/game"

But we can't really expect end-users to open their terminals to do this, right?

On macOS perhaps we should differ the `project-path` strategy like this:
  1. First check for a command line arg, otherwise,
  2. Check for an `RPG_RT.ldb` in `path_to_app.app/Contents/Resources`, otherwise,
  3. Check for projects in `$HOME/Library/Application Support/EasyRPG Player/Project`, otherwise,
  4. Check in current working dir

Number 2 means game developers could ship games on macOS as a `.app` and all the RPG Maker files would be self-contained (though they'd be on their own with regards to code signing).

Number 3 means that end users can simply drag-and-drop the `Player.app` to `/Applications`, as with any other Mac app, and then put all their RPG Maker games in `$HOME/Library/Application Support/EasyRPG Player/Project` and have them be auto discovered by the gamebrowser. This also sidesteps some annoying Catalina permission popups. However, users would also need to know that they have to do this...


src/sdl2_ui.cpp
```diff
@@ -268,6 +268,14 @@ bool Sdl2Ui::RefreshDisplayMode() {
 		SDL_GL_SetAttribute(SDL_GL_BLUE_SIZE, 5);
 		#endif
 
+		#if defined(__APPLE__) && defined(TARGET_OS_MAC)
```
tyrone-sudeium, line 271: I'm not sure if we ever build for iOS, but I've done this instead of `defined(__APPLE__) && defined(__MACH__)` (which I found elsewhere in the project), because that will return true for both macOS and iOS. We want this behaviour _specifically_ on Macs. On iOS and other Apple platforms we definitely want to use the SDL default renderer instead.

carstene1ns: `TARGET_OS_MAC` can not be used for this purpose. It is defined as `0`.
http://sealiesoftware.com/blog/archive/2010/8/16/TargetConditionalsh.html

tyrone-sudeium: `TARGET_OS_OSX` it is then

tyrone-sudeium:
> icon change?

The old icon was corrupt in some way, Mac Preview failed to open it, and the `.app` binary didn’t render an icon, just using the default placeholder instead. I generated a new one using the existing `.ico` assets, with variants at 256x256, 32x32, and 16x16.

src/sdl2_ui.cpp
tyrone-sudeium: Switched to `TARGET_OS_OSX` and removed the `defined` since it will be defined-but-zero on non-Mac Apple platforms

fdelapena:
Thank you for path information, this will be useful also for global and per-game settings later.

Regarding code signing (and the most recent notarization step), I had a developer ID with a generated certificate somewhere, but not sure if they get revoked on not renewing the yearly fee, as the certificate duration is for 5 years.

Ghabry:
Did you optimize the PNG in the ICNS? (with "advpng -z4" or a similiar tool)

tyrone-sudeium:
@Ghabry nope. `iconutil` already does a decent job of "crushing" the input pngs when it converts to `.icns`. Out of curiosity, I installed advancecomp on my Mac and ran all the input images through `advpng -z4`, then re-ran `iconutil`.

`Player.icns` before: 700,891 bytes
`Player.icns` after: 700,891 bytes

Ghabry:
Jenkins: test this please

Ghabry:
For future development:

- CMake can create bundles, look in the CMake docs (add_executable MACOS_BUNDLE)
- It would be nice if we could get rid of "easyrpg-player" completely and just use "Player.app"
- Finder sucks and doesn't set the working directory properly (is the home directory m( ), we have a workaround for this in main_data.cpp but when using "Player.app" it fails to find games in the current directory. Imo this App should support three use cases (checked in this order):

1. "Is there a game inside the bundle file?" -> Run this game (good save directory???)
2. "Is there a game in the same directory" -> Run/list game(s)
3. "$HOME/Library/Application Support/EasyRPG Player/Games"

PR #2245, opened 2020-06-13
FmMidi: naive harmony-like loop support

tyrone-sudeium:
Pretty naive implementation but follows the same MIDI message that harmony/rm2k3 uses for looping, namely the `b06f` event, which is a Control Change event on channel 1 with value `111`. You can quickly look for that by seeing if the MIDI message is `0x6fb0`, which is what that magical number check is.

Unfortunately fmmidi looks like it's the only decoder that lets me inspect the MIDI messages as they're passing through like this, which is what made this such a simple patch.

### DIFF
--- a/src/decoder_fmmidi.cpp
+++ b/src/decoder_fmmidi.cpp
@@ -73,10 +73,15 @@ bool FmMidiDecoder::Open(FILE* file) {
 
 bool FmMidiDecoder::Seek(size_t offset, Origin origin) {
 	if (offset == 0 && origin == Origin::Begin) {
-		mtime = 0.0f;
-		mtime_last_tempo_change = 0.0f;
-		ticks_last_tempo_change = 0;
-		seq->rewind();
+		mtime = seq->rewind_to_loop();
+		mtime_last_tempo_change = mtime_ltc_loopstart;
+		ticks_last_tempo_change = ticks_ltc_loopstart;
+		if (mtime > 0.0f) {
+			// Bit of a hack, prevent stuck notes
+			// TODO: verify with a MIDI event stream inspector whether RPG_RT does this?
+			synth->all_note_off();
+			tempo = tempo_loopstart;
+		}
 		begin = true;
 
 		return true;
@@ -140,6 +145,13 @@ int FmMidiDecoder::synthesize(int_least16_t * output, std::size_t samples, float
 }
 
 void FmMidiDecoder::midi_message(int, uint_least32_t message) {
+	// If message is a B06F loopstart control change...
+	if ((message & 0xFFFF) == 0x6FB0) {
+		// Save the state of these fields to restore them when looping
+		tempo_loopstart = tempo;
+		mtime_ltc_loopstart = mtime_last_tempo_change;
+		ticks_ltc_loopstart = ticks_last_tempo_change;
+	}
 	synth->midi_event(message);
 }
 
--- a/src/decoder_fmmidi.cpp
+++ b/src/decoder_fmmidi.cpp
@@ -20,11 +20,14 @@
 #ifdef WANT_FMMIDI
 
 // Headers
+#include <algorithm>
 #include <cstdio>
 #include <cassert>
 #include "audio_decoder.h"
 #include "output.h"
 #include "decoder_fmmidi.h"
+
+constexpr int FmMidiDecoder::midi_default_tempo;
 
 FmMidiDecoder::FmMidiDecoder() {
 	note_factory.reset(new midisynth::fm_note_factory());
@@ -66,22 +69,33 @@ bool FmMidiDecoder::Open(FILE* file) {
 	}
 	seq->rewind();
 
-	ticks_per_sec = (float)seq->get_division() / tempo * 1000000;
+	tempo.emplace_back(this, midi_default_tempo);
 
 	return true;
 }
 
 bool FmMidiDecoder::Seek(size_t offset, Origin origin) {
+	assert(!tempo.empty());
+
 	if (offset == 0 && origin == Origin::Begin) {
 		mtime = seq->rewind_to_loop();
-		mtime_last_tempo_change = mtime_ltc_loopstart;
-		ticks_last_tempo_change = ticks_ltc_loopstart;
+
 		if (mtime > 0.0f) {
+			// Throw away all tempo data after the loop point
+			auto rit = std::find_if(tempo.rbegin(), tempo.rend(), [&](auto& t) { return t.mtime <= mtime; });
+			auto it = rit.base();
+			if (it != tempo.end()) {
+				tempo.erase(it, tempo.end());
+			}
+
 			// Bit of a hack, prevent stuck notes
 			// TODO: verify with a MIDI event stream inspector whether RPG_RT does this?
 			synth->all_note_off();
-			tempo = tempo_loopstart;
+		} else {
+			tempo.clear();
+			tempo.emplace_back(this, midi_default_tempo);
 		}
+
 		begin = true;
 
 		return true;
@@ -117,8 +131,9 @@ bool FmMidiDecoder::SetPitch(int pitch) {
 }
 
 int FmMidiDecoder::GetTicks() const {
-	float delta = mtime - mtime_last_tempo_change;
-	return ticks_last_tempo_change + static_cast<int>(ticks_per_sec * delta);
+	assert(!tempo.empty());
+
+	return tempo.back().GetTicks(mtime);
 }
 
 int FmMidiDecoder::FillBuffer(uint8_t* buffer, int length) {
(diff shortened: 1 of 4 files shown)


Ghabry:
Looks fine to me. Will also work for anything else that uses the sequencer.
Maybe add a small comment that this equals event 111 on channel 1.

Do you have any midi files for testing?

fdelapena:
jenkins: ok to test

Ghabry:
Seeking in fmmidi works by replaying all messages from the beginning until the time is reached. This makes sense because you need e.g. tempo information.

That fmmidi fires everything at once is problematic because of my external tempo tracking for the ticks. To keep this working a seek must be programmed as a slow play-loop or all the callbacks must be patched to also send the time.

tyrone-sudeium:
@Ghabry I've made some fundamental changes to how this works now. I'd be interested to see what the behaviour is with regards to tempo changes that you mentioned. I used the MIDI loop test provided by @fdelapena but I added an instrument change in the middle of the looping section, so I could see whether RPG_RT reverted the instrument back to harp after the loop. It doesn't. This made me suspect RPG_RT doesn't play back all the instructions between the start and the loop point, so I've replicated that behaviour. Unfortunately that necessitates modifying midisequencer, I hope that's ok. The only alternative is to roll our own sequencer, which wouldn't be a terrible idea, but the fmmidi one is already pretty decent, IMO.

fdelapena:
Jenkins buildbot: test this please

Ghabry:
okay one test I would still like to see then in harmony is "tempo change, [...], loop-point, [...], tempo change 2, [...], loop".

Based on your tests it should continue at the speed of "tempo change 2" when the loop is hit.
Low priority imo but this would be hard to replicate in fmmidi because the time is precalculated while parsing :(. (but not impossible because you could apply a multiplier from the outside)

tyrone-sudeium:
> "tempo change, [...], loop-point, [...], tempo change 2, [...], loop".

I made this with 120BPM -> Loop Start -> 200 BPM -> Loop.

Harmony/RPG_RT plays at 120BPM until it hits 200BPM the first time, then retains 200BPM permanently for all subsequent loops.

With fmmidi in this PR, it reverts to 120BPM at the loop start, and only switches back to 200BPM after crossing the 200BPM event. So, we're not consistent with Harmony/RPG_RT. 😭 

There's an argument to be made that what FmMidi is doing is more "correct", but I suppose we want to follow RPG_RT, warts and all, right? I can think of a few ways to fix it but it makes my brain hurt......

Ghabry:
I made a quick analysis through 28404 MIDI files of mostly German games and only 1034 files use a loop.

So this is less than <4%. Don't think it is worth to support this edge case. The feature is in general seldom used and when it was used I guess they were aware of the limitations :)

Ghabry:
After removing duplicates and only considering MIDIs with more than one tempo command only 398 files are left. 
So I don't see a need in supporting this.

fmatthew5876:
It doesn't have to be done in this PR, but fuzzing or unit testing this midi control stuff would make a lot of sense. Especially if we can do it in a way that covers all the relevant combinations of audio libraries.

How hard would it be to generate some synthetic midi example streams in code, play all the samples instantaneously and piped to /dev/null, and test the edge cases with unit tests? Particularly a unit test on the tempo issue with a comment saying RPG_RT does x and but we do y.

tyrone-sudeium:
As part of rebasing this I had to also make sure the looping didn't break the ticks support. @Ghabry can you please sanity check me here? I'm not completely sure I understand how the ticks work so I made some guesses

src/decoder_fmmidi.cpp
```diff
@@ -66,21 +69,33 @@ bool FmMidiDecoder::Open(FILE* file) {
...
+		while (it > tempo.begin()) {
+			--it;
+			if (it->mtime <= mtime) {
+				++it;
+				break;
+			}
+		}
+		if (it != tempo.end()) {
+			tempo.erase(it, tempo.end());
+		}
```
Ghabry, line 93: @fmatthew5876 
Is there a more idiomatic way to do this?
I search from the end as I know that all elements to delete are at the end but "erase" with reverse iterators gives a compiler error :(

Ghabry:
@tyrone-sudeium 
I rewrote my tick code. Is now stack-based. Everytime a tempo event is encountered it is pushed on a stack. When Seeking all tempo data after the seek position is discarded (yeah, not Harmony compatible, but fmmidi isn't either ;)).

Test case (csvmidi bell.csv > bell.mid):
Also contains an edge case test for "multiple tempo changes at same time as 111".

```csv
0, 0, Header, 1, 2, 120
1, 0, Start_track
1, 0, Sequencer_specific, 10, 0, 0, 91, 35, 2, 0, 3, 0, 51, 0
1, 0, MIDI_port, 0
1, 0, Tempo, 500000
1, 0, Time_signature, 4, 2, 24, 8
1, 0, End_track
2, 0, Start_track
2, 0, MIDI_port, 0
2, 0, Program_c, 0, 46
2, 360, Note_on_c, 0, 76, 60
2, 479, Note_on_c, 0, 76, 0
2, 480, Note_on_c, 0, 74, 60
2, 599, Note_on_c, 0, 74, 0
2, 600, Note_on_c, 0, 72, 60
2, 719, Note_on_c, 0, 72, 0
2, 720, Note_on_c, 0, 71, 60
2, 839, Note_on_c, 0, 71, 0
2, 840, Note_on_c, 0, 69, 60
2, 959, Note_on_c, 0, 69, 0
2, 960, Note_on_c, 0, 67, 60
2, 960, Tempo, 100000
2, 960, Control_c, 0, 111, 0
2, 960, Tempo, 200000
2, 1079, Note_on_c, 0, 67, 0
2, 1080, Note_on_c, 0, 74, 60
2, 1199, Note_on_c, 0, 74, 0
2, 1200, Note_on_c, 0, 67, 60
2, 1319, Note_on_c, 0, 67, 0
2, 1320, Note_on_c, 0, 74, 60
2, 1439, Note_on_c, 0, 74, 0
2, 1440, Note_on_c, 0, 62, 60
2, 1441, Tempo, 300000
2, 1559, Note_on_c, 0, 62, 0
2, 1560, Note_on_c, 0, 69, 60
2, 1580, Tempo, 400000
2, 1679, Note_on_c, 0, 69, 0
2, 1680, Note_on_c, 0, 71, 60
2, 1799, Note_on_c, 0, 71, 0
2, 1800, Note_on_c, 0, 69, 60
2, 1919, Note_on_c, 0, 69, 0
2, 1919, End_track
0, 0, End_of_file
```

src/decoder_fmmidi.cpp
```diff
@@ -66,22 +68,39 @@ bool FmMidiDecoder::Open(FILE* file) {
...
+			auto it = tempo.end();
+			while (it > tempo.begin()) {
+				--it;
+				if (it->mtime <= mtime) {
+					++it;
+					break;
+				}
+			}
+			if (it != tempo.end()) {
+				tempo.erase(it, tempo.end());
```
fmatthew5876, line 93: I believe this whole loop can be:
```
auto riter = std::find_if(tempo.rbegin(), tempo.rend(), [&](auto& t) { return t .mtime <= seq->rewind_to_loop; });
auto iter = std::next(riter).base();
if (iter != tempo.end()) {
   tempo.erase(iter);
}
```

I've never tried the `base()` trick with reverse iterator before.  So test before you try this.

src/decoder_fmmidi.cpp
```diff
@@ -66,22 +69,34 @@ bool FmMidiDecoder::Open(FILE* file) {
...
 	if (offset == 0 && origin == Origin::Begin) {
 		mtime = seq->rewind_to_loop();
-		mtime_last_tempo_change = mtime_ltc_loopstart;
-		ticks_last_tempo_change = ticks_ltc_loopstart;
+
 		if (mtime > 0.0f) {
+			// Throw away all tempo data after the loop point
+			auto rit = std::find_if(tempo.rbegin(), tempo.rend(), [&](auto& t) { return t.mtime <= mtime; });
+			--rit;
+			auto it = std::next(rit).base();
```
fmatthew5876, line 87: Why `--rit` and then `std::next(rit)` (which is just ++rit)? Maybe we just need to call `base()` directly? Again haven't tried this, so not sure what combination is needed to avoid off by 1 errors.

Ghabry: thats just my lack of iter API knowledge. I searched for a std::cur() to call base on xD. This works.

Ghabry:
Tested with:
- The example above
- Multiple tempo before loop
- No tempo at all

They all work. The last hits the sentinel default tempo on the stack at "mtime = 0".

Loop at 0.0 hits the ".clear()" codepath.

fmatthew5876:
> Tested with:
> 
> * The example above
> * Multiple tempo before loop
> * No tempo at all
> 
> They all work. The last hits the sentinel default tempo on the stack at "mtime = 0".
> 
> Loop at 0.0 hits the ".clear()" codepath.

What about tempo after the loop? Can it be erasing 1 too many?

Ghabry:
- The example above but all tempo removed except 300000 and 400000: Only the sentinel value survives -> correct

Issue #2246, opened 2020-06-14
Some constexpr code fails to compile on GCC < 7.2


Ghabry:
Some of our code appears to fail to compile with older GCC as noticed by someone on Twitter:

See 
- https://twitter.com/_imai/status/1271352248115683329
- http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#1684
- https://pastebin.com/29RUdk99
- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66297

Maybe this worked in gcc5 but broke in 5.1 - 7.1?

fmatthew5876:
This is a bug anyway, those functions should be static. I fixed it in my huge PR, but I can backport this small fix to unblock people.

fmatthew5876:
Also would it make sense / be a lot of work to setup a new linux / gcc5 build? Basically a jenkins build that only runs on merge master which tests our advertised minimum supported compiler version.

PR #2247, opened 2020-06-14
Fix constexpr compilation in Game_Character

fmatthew5876:
Fix: #2246


Ghabry:
Have to setup the right gcc compiler first, maybe gcc5 really works and broken since 5.1

PR #2249, opened 2020-06-16
Mac: cmake .app support, use NSBundle for finding app dir

tyrone-sudeium:
First pass at making cmake support building the `.app` bundles. Also I fixed the `FIXME` in `main_data.cpp` where it was using SDL to get the directory of the executable / app bundle. Now it uses the `NSBundle` API from Apple, so it can always return either the folder the raw mach-o executable is in if it's not an app bundle, or the folder the app bundle is in if it is.

This fixes the bug with app bundles so now you can simply drag `Player.app` into an RPG Maker project directory and double click it to start, like you could on any other platform.


CMakeLists.txt
```diff
@@ -692,30 +706,41 @@ endif()
...
+		set_target_properties(${EXE_NAME} PROPERTIES OUTPUT_NAME "Player")
 	endif()
 
-	target_link_libraries(${PROJECT_NAME}_exe ${PROJECT_NAME})
+	target_link_libraries(${EXE_NAME} ${PROJECT_NAME})
 
-	# installation
-	include(GNUInstallDirs)
-	install(TARGETS ${PROJECT_NAME}_exe RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
+	# installation - don't use install on macs
```
tyrone-sudeium, line 736: Did this because I'm getting a weird error I don't know how to fix, and I'm also not entirely sure where cmake would put a `.app` if it were to install it...

src/sdl2_ui.cpp
```diff
@@ -773,7 +773,7 @@ void Sdl2Ui::ProcessFingerEvent(SDL_Event& evnt) {
 }
 
 void Sdl2Ui::SetAppIcon() {
-#if defined(__WINRT__) || defined(__MORPHOS__)
+#if defined(__WINRT__) || defined(__MORPHOS__) || TARGET_OS_OSX
```
tyrone-sudeium, line 776: Added this because on macOS, running the code in this function causes the app icon in the Dock to become really blurry.

Ghabry:
Finally somebody who is brave enough to write ObjectiveC code.

src/system.h
```diff
@@ -127,3 +127,7 @@
 #endif
 
 #endif
+
+#if defined(__APPLE__) && defined(__MACH__)
+#include <TargetConditionals.h>
```
tyrone-sudeium, line 132: This adds the `TARGET_OS_OSX` and friends. On Macs there's this standard thing called "precompiled headers", which is essentially a header file that every compilation unit silently includes, and it's in here that this TargetConditionals is usually included. We don't have one of those, so this is as close as I could get.

Not strictly necessary, but I thought it was a bit cleaner than littering the code with these 3 lines every time we need to check if we're a Mac or not.

resources/osx/Info.plist
```diff
@@ -4,7 +4,7 @@
 <dict>
 
     <key>CFBundleExecutable</key>
-    <string>EasyRPG Player</string>
+    <string>Player</string>
```
Ghabry, line 7: Is this string shown anywhere when the application is installed?

The full name is "EasyRPG Player" - Also "Player" is ambigious.

tyrone-sudeium: I'll see if I can get cmake and friends to emit "EasyRPG Player"

Ghabry: CMake supports a plist template file "plist.in". Maybe is useful: https://stackoverflow.com/questions/42313808/customising-macos-bundles-info-plist-file-with-cmake

https://cmake.org/cmake/help/v3.18/prop_tgt/MACOSX_BUNDLE_INFO_PLIST.html

Ghabry:
Jenkins: Test this please

src/sdl2_ui.cpp
carstene1ns: Sure, because it gets replaced by a 32x32 icon. No idea of the exact strategy macos uses, but I assume it uses the icon of the app bundle otherwise.
However, this does not apply to Homebrew for example.

tyrone-sudeium: Good point about homebrew! I could make it specifically check if it's running in an app bundle and then early exit?

Ghabry:
Jenkins: test this please

Ghabry:
Your code does not support out-of-tree builds (build in folder "build" and using "cmake .." or "configure ..")

must prefix MACOSX_BUNDLE_INFO_PLIST and bundle icon with ${CMAKE_CURRENT_SOURCE_DIR}.

Btw Cmake has "install(BUNDLE" for installing bundles but I can't find any sane default for this -_-

configure.ac
```diff
@@ -93,6 +93,25 @@ AS_IF([test "x$BASHCOMPLETION_DIR" = "x"],[
...
+
+# Detect the target system
+case "${host_os}" in
+    darwin*)
+        EP_BUILD_MAC=YES
+        ;;
+esac
+
+AS_IF([test "$EP_BUILD_MAC" = "YES"],[
+	AC_PROG_OBJCXX
```
carstene1ns, line 109: This needs to be run unconditionally.

src/sdl2_ui.cpp
carstene1ns: Seems like a good idea. This way we still have an icon when not run as app bundle.

---
General problem is the SDL way of setting this is limited, so I already added a windows specific way.
With high resolutions / HighDPI desktops this needs to be addressed sooner or later.
We already have high resolution variants of our icon, problem is we do not ship them with the raw executable, because it would increase the size. So a 32x32 icon was added to not make the executable much bigger.
On Linux-like platforms we could use the freedesktop.org icon specification to get the svg (or scaled pngs), for macos (with homebrew) we could put the `.icns` somewhere, for Windows use the included `.ico`.
However, this is unrelated to this Pull Request.

Ghabry:
macOS fails now with "ninja: error: unknown target 'install'". Which is expected as there is no "raw" executable anymore to install anymore in CMake. (Jenkins update needed)

When somebody complains later there could be a ``PLAYER_ENABLE_BUNDLE=OFF`` setting which will build a "raw" executable like autotools does.

tyrone-sudeium:
If someone wants a raw mach-o executable just copy it out of the bundle, it's `EasyRPG Player.app/Contents/MacOS/EasyRPG Player`

src/sdl2_ui.cpp
```diff
@@ -796,6 +801,11 @@ void Sdl2Ui::SetAppIcon() {
 	SetClassLongPtr(window, GCLP_HICON, (LONG_PTR) icon);
 	SetClassLongPtr(window, GCLP_HICONSM, (LONG_PTR) icon_small);
 #else
```
carstene1ns, line 803: I would like a comment here that macos automatically uses the bundle icon

CMakeLists.txt
```diff
@@ -733,14 +733,15 @@ if (${PLAYER_BUILD_EXECUTABLE})
 
 	target_link_libraries(${EXE_NAME} ${PROJECT_NAME})
 
-	# installation - don't use install on macs
-	if(NOT APPLE)
-		include(GNUInstallDirs)
+	include(GNUInstallDirs)
+	if(APPLE)
+		install(TARGETS ${EXE_NAME} RUNTIME DESTINATION BUNDLE DESTINATION "${CMAKE_BINARY_DIR}/Package")
```
Ghabry, line 738: Is the use of CMAKE_BINARY_DIR intentional here? That will install in the build-tree.

tyrone-sudeium: Yes. Where else to put it? `/usr/local` is not inappropriate, and Jenkins wants to call install...

Issue #17, opened 2012-05-27
Use FluidSynth instead of internal timidity from SDL_Mixer


ghost:
# FluidSynth

SDL_Mixer supports FluidSynth since version 1.2.12.
FluidSynth has better MIDI support than the forked version of Timidity used in SDL_Mixer.

We need it to support some MIDI effects made in RPG Maker, not supported by the basic Timidity:
- Seeking support (there is a command in RPG Maker to memorize MIDI position).
- Possibility to configure loop on event 111 (used by RTP songs).
- Change speed support.
## SoundFont support

Furthermore, FluidSynth supports SoundFonts, internal timidity only supports GUS Patches (.pat) and freepats package is incomplete and heavy. We can use a lighter, more complete, SoundFont. There is a enough good free SoundFont on major GNU/Linux distributions: TimGM6mb.sf2, from the musescore-soundfont-gm package, used by the great MuseScore music notation editor. Soundfont is 6 MB only, several times smaller than the incomplete freepats.
## Performance tuning

Fluidsynth command line parameters for less CPU and RAM usage:
- Disable chorus and reverb for less CPU usage.
- Increase buffer size for less CPU usage.
- Reduce buffer count for less RAM usage.

Miscellaneous player tweaks:
- Disable real time priority, prevents warning.
- Disable MIDI in, not needed for playing.
- Disable shell, not needed for playing.

Command line example:

```
fluidsynth \
    --audio-bufcount 3 \
    --audio-bufsize 8192 \
    --audio-driver alsa \
    --chorus no \
    --reverb no \
    -o audio.realtime-prio=0 \
    --no-shell \
    --no-midi-in \
    /usr/share/sounds/sf2/TimGM6mb.sf2 \
    Music/Battle1.mid
```

Results:
- FluidSynth application RAM usage: About 18 MB.
- CPU usage: 15-25% on a 800 MHz CPU.

PR #2250, opened 2020-06-17
Add Fluidsynth/FluidLite

### LINKED ISSUE #834: Embric of Wulfhammer's Castle: Sound effect repeating incorrectly [MIDI loop point]
This game has a scene in which the Duchess goes to sleep. It is accompanied by a short sound effect, which in RPG_RT plays only once. In EasyRPG, however, the sound effect incorrectly constantly repeats itself.
- Game download: https://aeresland.net/EmbricDELUXE_2.2.2.7z
- Savegame demonstrating the problem: https://www.heavylobster.com/files/Save02.lsd

To reproduce, load the save game and talk to Alice (immediately in front of the Duchess in the savegame)

Edit: added [MIDI loop test](https://github.com/EasyRPG/Player/files/3006572/midi-loop-test.zip) case and [online test](https://easyrpg.org/play/master/?game=issue-834) in Emscripten (with FMMIDI).

-----

See #2199

### LINKED ISSUE #1322: FluidSynth/FluidLite MIDI playback support (SF2/SF3 files)
Despite SDL_mixer already supports FluidSynth, some ports may take advantage of FluidLite because it is a lightweight version of FluidSynth not requiring GLib. However, it still requires some additional work to provide standard MIDI file (SMF) parsing (which might be reused and mapped somehow from FMMIDI loader and/or try to import some minimal stuff from the larger FluidSynth).

The advantage of SF2/SF3 is a single file, replaceable by users (at least a couple of users asked for this feature), and SF3 uses Vorbis for samples. I've [compressed](https://github.com/EasyRPG/Player/files/1590768/test.zip) the 30MB (28MB in zip) GeneralUser.sf2 into a 6.5MB SF3 (3.7MB in zip). This allows a larger, more complete, higher quality sound bank (better drums, no missing instruments) for ports such Android, saving a significant amount of package space without sacrificing sound quality.

FluidSynth/FluidLite license is LGPLv2.1, which is GPLv3 compatible.

There is also [FluidSynth Lite](https://github.com/dotfloat/fluidsynth-lite), which strips glib too, provides .mid support, but not SF3 support currently.

### LINKED ISSUE #2029: Tempo changes the Tone of MIDI songs
Play a MIDI modifying the Tempo.
In RPG_RT, Tempo will change the speed of a song (if it is a MIDI file, if it is MP3 will change speed and Tone), but in Player change the speed and tone, like MP3 files.

Ghabry:
Fix #1322 
Fix #834
Fix #2029 

- [ ] TODO: Update Makefiles - Do some console tests 

Depends on FS/1 because this would cause too much rebase pain otherwise.

Features:
- Refactored MIDI Api. Is now managed by a Create function ``MidiDecoder`` (similiar how ``AudioDecoder`` works) which handles instantiation of an appropriate subclas. A ``GenericMidiDecoder`` implements most of the API already and manages a FmMidi Sequencer and an underlying MIDI library.
- When the MIDI library supports external sequencers only ``FillBuffer`` and Midi messages must be implemented. Ticks, loops etc. are managed by ``GenericMidiDecoder``.
- The Api also handles libraries that lack external sequencer support (``WildMidi``). ``GenericMidiDecoder`` is aware of it and will call appropriate functions and do _sample counting_. For ``WildMidi`` this means that ``Seek`` to implement Loops will work!
- FluidSynth/Lite support (Synth has priority in detection)
- The Fluid synth instance is recycled for BGMs to avoid long delays when loading the soundfont
- VIO for soundfonts through Filesystem API

TBD: Good default filenames/search locations for Soundfonts


fmatthew5876:
> TBD: Good default filenames/search locations for Soundfonts

We may want to punt on this question until we tackle engine configuration more generally in the config PRs sequence. The midi engine and soundfont are definitely a big part of that.

I'd be fine with some env var or command line parameter hack for now. Maybe also some hard coded default to search like `RPG_RT.sf[23]` in the game directory for game devs. I wouldn't want this important PR to get bogged down in that somewhat orthogonal discussion.

Ghabry:
With FS1 merged this is now open for review.

Right now the soundfont is hardcoded to "easyrpg.soundfont", will keep it this way for now.

src/audio_midi.h
```diff
@@ -0,0 +1,192 @@
...
+#define EP_AUDIO_MIDI_H
+
+// Headers
+#include "audio_decoder.h"
+#include "midisequencer.h"
+
+#if defined(GEKKO) || defined(_3DS)
+#  define EP_MIDI_FREQ 22050
+#else
+#  define EP_MIDI_FREQ 44100
```
fdelapena, line 28: I'd suggest 22050 also for other platforms. After comparing with scc1t2.sf2, they sound similar to the original GM.DLS DirectMusic playback, so it seems Windows plays at that rate from RPG_RT.exe and editor. DLS samples are 22 KHz according to Awave Studio for middle note playback, so there wouldn't be a major gain with higher playback rate.

src/decoder_fluidsynth.cpp
```diff
@@ -0,0 +1,241 @@
...
+	if (once)
+		return init;
+	once = true;
+
+#ifdef HAVE_FLUIDLITE
+	fluid_set_default_fileapi(&fluidlite_vio);
+#endif
+
+	if (!global_settings) {
+		global_settings.reset(new_fluid_settings());
```
fdelapena, line 153: Are chorus and reverberation disabled by default? According to the FluidSynth 2.0 API, they are enabled by default, not just in the command line player. They are a huge CPU hog, even more than high sampling rates.

Tuning buffer size and count also affects cpu/latency significantly, and FluidSynth seems to use a small buffer oriented for pro audio low latency, so CPU usage can be reduced there, too. I didn't check how Fluid buffer is being filled here or if needs to be tweaked to match the expected Player audio buffer size. FluidSynth API has some buffer related settings related to period-size and periods. Maybe these are for its own playback.

On Linux, setting realtime-prio to 0 disables Fluid attempting higher priority, which prevents some permission warnings, likely not requiring guards as should be ignored on other platforms.

See #17 for some early CLI tests.

src/decoder_fluidsynth.cpp
```diff
@@ -0,0 +1,213 @@
...
+	if (fluid_synth_sfload(synth, "easyrpg.soundfont", 1) == FLUID_FAILED) {
+		error_message = "Could not load soundfont.";
+		return;
+	}
+
+	fluid_synth_set_interp_method(synth, -1, 7);
+
+	double sample_rate = 0;
+	fluid_settings_getnum(settings, "synth.sample-rate", &sample_rate);
+	assert(sample_rate != 0);
```
fmatthew5876: If this is coming from a runtime configuration, assert is not valid here. We have to handle the error by exiting. assert is only for bugs in code.

fmatthew5876:
Other than the minor nit, looks ok to me.

src/decoder_fluidsynth.cpp
Ghabry: It seems that line is already gone

src/decoder_fluidsynth.cpp
Ghabry: Here are the default settings:
```
  fluid_settings_register_str(settings, "synth.verbose", "no", 0, NULL, NULL);
  fluid_settings_register_str(settings, "synth.dump", "no", 0, NULL, NULL);
  fluid_settings_register_str(settings, "synth.reverb.active", "yes", 0, NULL, NULL);
  fluid_settings_register_str(settings, "synth.chorus.active", "yes", 0, NULL, NULL);
  fluid_settings_register_str(settings, "synth.ladspa.active", "no", 0, NULL, NULL);
  fluid_settings_register_str(settings, "midi.portname", "", 0, NULL, NULL);
  fluid_settings_register_str(settings, "synth.drums-channel.active", "yes", 0, NULL, NULL);

  fluid_settings_register_int(settings, "synth.polyphony",
                             256, 16, 4096, 0, NULL, NULL);
  fluid_settings_register_int(settings, "synth.midi-channels",
                             16, 16, 256, 0, NULL, NULL);
  fluid_settings_register_num(settings, "synth.gain",
                             0.2f, 0.0f, 10.0f,
                             0, NULL, NULL);
  fluid_settings_register_int(settings, "synth.audio-channels",
                             1, 1, 256, 0, NULL, NULL);
  fluid_settings_register_int(settings, "synth.audio-groups",
                             1, 1, 256, 0, NULL, NULL);
  fluid_settings_register_int(settings, "synth.effects-channels",
                             2, 2, 2, 0, NULL, NULL);
  fluid_settings_register_num(settings, "synth.sample-rate",
                             44100.0f, 22050.0f, 96000.0f,
                             0, NULL, NULL);
  fluid_settings_register_int(settings, "synth.min-note-length", 10, 0, 65535, 0, NULL, NULL);
```

About 2nd and 3rd paragraph: The real-time stuff is not applicable because fluidlite does not start any daemon. The buffer/latency is the default buffer size. This means for fast music it will sound out-of-sync because the MIDI resolution is ~4ms. This will be resolved after tyrone-sudeiums PR. Then the MIDI can play in real-time and the system ALSA Midi player can be used.

Ghabry: Shall I disable reverb and chorus everywhere or only on slower platforms? Is there a hearable difference?

fmatthew5876: Maybe only on slower platforms? I don't want you destroyed my beautiful game music when I play on my PC :)

Also, once all this is done can one of you guys catalogue all the audio configuration knobs in an github issue? These things should all eventually be modifiable via command line / config / scene settings.

fmatthew5876: Or rather, are these features enabled in RPG_RT? We could follow the standard as our default.

fdelapena: Default playback with DirectMusic on RPG_RT seems to use a barely hearable chorus and reverb, I feel them mostly undetectable by most users. Default chorus and reverb thresholds on FluidSynth are too high and sounds pretty different compared to the original RPG_RT. It sounds a lot closer when chorus and reverb are off, actually. I'd suggest hear this by your self with scc1t2.sf2 at 22050 with disabled chorus and reverb to compare.

Ghabry: well there could be a config option to turn them on again later, so I make it default OFF for now. Configuration can be decided when the config code is more advanced. I will prepare a list for you with all useful settings per individual audio decoder.

When the system-wide MIDI support is ready this is also a non-issue for Win/Lin/macOS because the OS configuration is used (but don't ask me how to correctly configure MIDI out on Linux xD)

src/audio_midi.h
```diff
@@ -0,0 +1,188 @@
...
+ */
+
+#ifndef EP_AUDIO_MIDI_H
+#define EP_AUDIO_MIDI_H
+
+// Headers
+#include "audio_decoder.h"
+#include "midisequencer.h"
+
+#define EP_MIDI_FREQ 22050
```
elsemieni, line 25: Watch out. I know 22050 is less process usage but this preset (at least in my windows x86) is making melodies sounds a bit "uncoordinated" in difference with 44100. 

Here's a little example with Battle3.mid (sorry, github doesn't like audios =(  )
With 44100:
https://drive.google.com/file/d/1ODdgq7aU0tljQwopjRmnyjY-p2kokUv_/view?usp=sharing
With 22050:
https://drive.google.com/file/d/1ti2VP5uFELNgIE7TgTlJJiJuPeV-VO1P/view?usp=sharing

fmatthew5876: Would like to know how much less process usage really is this? Especially on desktop platforms?

I'm all for saving CPU cycles but 22khz is sacrificing audio quality in a big way. I would prefer to default to "reasonable" audio quality that "sounds like RPG_RT or better" and only cut back if we find we really need to on limited hardware.

Also on modern systems, using low frequencies like this for music is very uncommon and so downstream software / OS audio stacks may up-sample it poorly.

For all these conjectures about processing costs of audio features, we really need some benchmarks and performance numbers. Some of them could be costing much less or more more CPU than we think.

That being said, I also really want to see some initial version of this merged to master so I can being playing with it while testing other PRs. So I'm fine collecting the outstanding questions into a github issue, getting this into a base working state, and merging it.

elsemieni:
Also a little observation with RTP2003 MIDIs. 
Most (if not all) RTP2003 MIDI's uses these commands in the beginning of each track:

image

In channel 10 (drums) this causes troubles and FluidLite fallbacks to a non-drum instrument. If you manually remove these events from MIDI, melodies play normally. 

I don't know if it's related with FluidLite directly (I don't have other software reference who uses FluidLite to test), or with the interface that "connects" FluidLite and player (at least I tested the PR without soundfont and works fine).

Ghabry:
I will only:

- [ ] Recheck pitch
- [ ] undo the 22050 change
- [ ] Apply fdelas suggestions to fix missing Notes

Then this is ready.

The stuck instruments when doing BGM Stop -> BGM play require some work I would postbone to another PR. They only happen in Fluid, not in Wildmidi or FmMidi.

Ghabry:
I'm done here. I was able to fix the incorrectly pitched Midi. This was a design issue in the Resampler (the pitch stuff was unused for years) which explains why FmMidi always had the wrong pitch. Midis play now correctly for FmMidi and FluidLite/Synth.

Can't be fixed in WildMidi because it is not Midi Message based. (at least I was able to implement Midi ticks through tricks here)

I postbone the following two bugs:

- [ ] Investigate wrong CC command
- [ ] Stuck instruments when BGM_Stop -> BGM_Play

fmatthew5876:
Sounds good to me, lets get those items into a github issue so we don't forget them

PR #2252, opened 2020-06-18
N3DS Touchscreen toggle 

Narwher:
The numpad on the 3DS touchscreen may not be used for some games. This change makes it so that it is hidden by default upon startup (the LCD is turned off as well) and is activated by touch input.

There is currently no way to turn it back off. Almost all button bindings are used by the interpreter, so it would likely have to be a something involving the touchscreen again. The '0' button on the numpad occupies the width of two normal keys - perhaps dividing it in half and adding a button for LCD toggle would work?

Side note: I also experimented with changing the sample rate for 3DS audio but have since changed it back. As I have not cherry-picked it all of these commits are in the PR. Please only examine commits from 18 June onwards.

Thanks for your consideration!


Ghabry:
About a way to configure this / turn it off again:
By default of sounds fine.

We currently also work on a settings UI. It could also contain an option for toggling the LCD later.

Narwher:
With [8c6a6c3](https://github.com/EasyRPG/Player/pull/2252/commits/8c6a6c3ee2fb39e720bbf09db811472db31f432e) I made a rough draft for a possible extra button on the numpad overlay, to turn off the display. There's no function tied to it yet though.

keyboard

src/platform/3ds/3ds_ui.cpp
```diff
@@ -220,6 +227,13 @@ void CtrUi::ProcessEvents() {
 		touch_y = pos.py;
 
 		keys[keys_tbl[col + (row * 4)]] = true;
+
+		if (!touchscreen) {
+			touchscreen = true;
+			gspLcdInit();
```
carstene1ns, line 233: Please with error checking:
```suggestion
			if (R_SUCCEEDED(gspLcdInit())) {
				GSPLCD_PowerOnBacklight(GSPLCD_SCREEN_BOTTOM);
				gspLcdExit();
			}
```

src/platform/3ds/3ds_ui.cpp
```diff
@@ -78,13 +80,19 @@ CtrUi::CtrUi(int width, int height)
 		osSetSpeedupEnable(true);
 	}
 
+	// Turn off bottom screen initially
```
carstene1ns, line 83: why?

src/platform/3ds/3ds_ui.cpp
```diff
@@ -277,49 +291,46 @@ void CtrUi::UpdateDisplay() {
...
-		u8 pos_x = col * button_width;
-		u8 pos_y = row * button_height;
-
-		// "0" is handled specially
-		u8 draw_width = button_width;
-		if (col < 2 && row == 3) {
-			draw_width *= 2;
-			if (col == 1)
-				pos_x = 0;
+		C2D_SceneBegin(bottom_screen);
```
carstene1ns, line 294: The indentation is completely off here.

fdelapena:
Jenkins CI: ok to test

troybonneau:
This is marked for 7.0, so I was wondering whether or not this needed a bit more work before it gets in. I think I could probably finish up the graphics/code if no one is available.

Ghabry:
#2617 replaced this one

PR #2253, opened 2020-06-19
Add Emscripten support to CMake

### LINKED ISSUE #2228: Webplayer hangs on tab switching
It seems to hang on tab switching since 0.6.2. In 0.6.1 does not seem to happen. Tested on Firefox and Chromium. It could be related with interpreter or music pause on losing focus like desktop does after switching to the new loop mechanism not required with the new llvm backend.

Ghabry:
Fix #2228

How to build:
```
source EMSCRIPTEN_PATH/emsdk_env.fish
emcmake cmake . -GNinja -DPLAYER_JS_BUILD_SHELL=ON|OFF
```

```
cmake --build . --target check
[2/2] /bin/node /home/gabriel/code/easyrpg/easyrpg-player/test_runner.js
[doctest] doctest version is "2.3.5"
[doctest] run with "--help" for options
============================================
[doctest] test cases:     98 |     98 passed |      0 failed |     10 skipped
[doctest] assertions:   3391 |   3391 passed |      0 failed |
[doctest] Status: SUCCESS!
```

The ``LANGUAGES C`` is required for the ".js" files and I can't figure out a different way to enable C.

This allows changing the path with ``-DPLAYER_JS_GAME_URL`` and it will only recompile this one file! (at least when Ninja is used, Makefiles do a full recompile)
```set_property(SOURCE src/async_handler.cpp APPEND PROPERTY COMPILE_DEFINITIONS "EM_GAME_URL=\"${PLAYER_JS_GAME_URL}\"")``

File system access for unit tests would be possible in a Node environment but this is a huge change I don't want to do.

The Toolchain file must be patched right now because there is still this open issue:
https://github.com/emscripten-core/emscripten/pull/10184
I explain there why ``-DCMAKE_FIND_ROOT_PATH`` can't be used :/

Any suggestion for ``README.md`` edits?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.7)
 
-project(EasyRPG_Player VERSION 0.6.2 LANGUAGES C CXX)
+project(EasyRPG_Player VERSION 0.6.2 LANGUAGES CXX)
 
 # Extra CMake Module files
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/builds/cmake/Modules")
@@ -778,18 +778,16 @@ if (${PLAYER_BUILD_EXECUTABLE})
 	target_link_libraries(${EXE_NAME} ${PROJECT_NAME})
 
 	if(CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
-		# These must be set before calling em_link_*
-		# See emscripten repo: tests/cmake/target_js/CMakeLists.txt#L94
-		set_property(TARGET ${PROJECT_NAME}_exe PROPERTY LINK_FLAGS "-s ALLOW_MEMORY_GROWTH=1 -s ASYNCIFY=1")
+		set_property(TARGET ${PROJECT_NAME}_exe PROPERTY LINK_FLAGS "-s ALLOW_MEMORY_GROWTH=1 -s ASYNCIFY=1  --pre-js ${PLAYER_JS_PREJS} --post-js ${PLAYER_JS_POSTJS}")
+		set_source_files_properties("src/main.cpp" PROPERTIES OBJECT_DEPENDS "${PLAYER_JS_PREJS};${PLAYER_JS_POSTJS}")
 
 		if(PLAYER_JS_BUILD_SHELL)
 			set_target_properties(${PROJECT_NAME}_exe PROPERTIES SUFFIX ".html")
-			set_property(TARGET ${PROJECT_NAME}_exe PROPERTY LINK_FLAGS "--shell-file ${PLAYER_JS_SHELL}")
+			set_property(TARGET ${PROJECT_NAME}_exe APPEND_STRING PROPERTY LINK_FLAGS " --shell-file ${PLAYER_JS_SHELL}")
 		endif()
 
 		target_link_libraries(${PROJECT_NAME}_exe "idbfs.js")
-		em_link_pre_js(${PROJECT_NAME}_exe "${PLAYER_JS_PREJS}")
-		em_link_post_js(${PROJECT_NAME}_exe "${PLAYER_JS_POSTJS}")
+
 	endif()
 
 	# installation
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.7)
 
-project(EasyRPG_Player VERSION 0.6.2 LANGUAGES CXX)
+project(EasyRPG_Player VERSION 0.6.2 LANGUAGES C CXX)
 
 # Extra CMake Module files
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/builds/cmake/Modules")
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

**Comment by Ghabry** [comment]
Toolchain File Update was merged. So only have to wait for next emscripten release

File: CMakeLists.txt [hunk 1]
```diff
@@ -453,6 +453,15 @@ if(APPLE)
 		src/platform/macos/utils.h)
 endif()
 
+if(CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
+	option(PLAYER_JS_BUILD_SHELL "Build the Player executable as a shell file (.html) instead of a standalone javascript file (.js)" OFF)
+	set(PLAYER_JS_PREJS "${CMAKE_CURRENT_SOURCE_DIR}/resources/emscripten/emscripten-pre.js" CACHE STRING "File to use for --pre-js")
+	set(PLAYER_JS_POSTJS "${CMAKE_CURRENT_SOURCE_DIR}/resources/emscripten/emscripten-post.js" CACHE STRING "File to use for --post-js")
+	set(PLAYER_JS_SHELL "${CMAKE_CURRENT_SOURCE_DIR}/resources/emscripten/emscripten-shell.html" CACHE STRING "Shell file to use (--shell-file). Requires PLAYER_JS_BUILD_SHELL=ON")
+	set(PLAYER_JS_GAME_URL "games/" CACHE STRING "Game URL/directory where the web player searches for games")
+	set_property(SOURCE src/async_handler.cpp APPEND PROPERTY COMPILE_DEFINITIONS "EM_GAME_URL=\"${PLAYER_JS_GAME_URL}\"")
```
carstene1ns, line 462: This is a nice feature. For autotools this involves either a convenience lib or manual makefile rules, which is both ugly.
I would suggest we remove the `EM_GAME_URL` completely from there, since the additional steps are not implemented as well.

CMakeLists.txt
```diff
@@ -453,6 +453,15 @@ if(APPLE)
 		src/platform/macos/utils.h)
 endif()
 
+if(CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
```
carstene1ns, line 456: Any reason why all of this is configurable?
If someone forks Player they will simply replace the files and not mess with cmake options.

Ghabry: There is not really a reason. Options where it is useful is ``PLAYER_JS_BUILD_SHELL`` (To decide if you want .js or .html as output) and ``PLAYER_JS_GAME_URL``.
The others are "for style". But I can get rid of them.

carstene1ns: Yeah, that is what I figured. They could stay, but I really do not see a use for them.

carstene1ns:
Reason why we need `C` is the "dummy library" they use:
https://github.com/emscripten-core/emscripten/blob/master/cmake/Modules/Platform/Emscripten.cmake#L325-L335

I guess they could check which language the project uses (and use c++ then, but out-of-scope here) or we could pass the flags manually, because the additional logic is only for rebuilding.

carstene1ns:
After thinking about this again, I came up with this:
```patch
index 2df8b7b1..2934e8eb 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.7)
 
-project(EasyRPG_Player VERSION 0.6.2 LANGUAGES C CXX)
+project(EasyRPG_Player VERSION 0.6.2 LANGUAGES CXX)
 
 # Extra CMake Module files
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/builds/cmake/Modules")
@@ -743,18 +743,16 @@ if (${PLAYER_BUILD_EXECUTABLE})
 	target_link_libraries(${EXE_NAME} ${PROJECT_NAME})
 
 	if(CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
-		# These must be set before calling em_link_*
-		# See emscripten repo: tests/cmake/target_js/CMakeLists.txt#L94
-		set_property(TARGET ${PROJECT_NAME}_exe PROPERTY LINK_FLAGS "-s ALLOW_MEMORY_GROWTH=1 -s ASYNCIFY=1")
+		set_property(TARGET ${PROJECT_NAME}_exe PROPERTY LINK_FLAGS "-s ALLOW_MEMORY_GROWTH=1 -s ASYNCIFY=1  --pre-js ${PLAYER_JS_PREJS} --post-js ${PLAYER_JS_POSTJS}")
+		set_source_files_properties("src/main.cpp" PROPERTIES OBJECT_DEPENDS "${PLAYER_JS_PREJS};${PLAYER_JS_POSTJS}")
 
 		if(PLAYER_JS_BUILD_SHELL)
 			set_target_properties(${PROJECT_NAME}_exe PROPERTIES SUFFIX ".html")
-			set_property(TARGET ${PROJECT_NAME}_exe PROPERTY LINK_FLAGS "--shell-file ${PLAYER_JS_SHELL}")
+			set_property(TARGET ${PROJECT_NAME}_exe APPEND_STRING PROPERTY LINK_FLAGS " --shell-file ${PLAYER_JS_SHELL}")
 		endif()
 
 		target_link_libraries(${PROJECT_NAME}_exe "idbfs.js")
-		em_link_pre_js(${PROJECT_NAME}_exe "${PLAYER_JS_PREJS}")
-		em_link_post_js(${PROJECT_NAME}_exe "${PLAYER_JS_POSTJS}")
+
 	endif()
 
 	# installation
```

Basically passing the flags directly without convenience library. The "rebuild-if-pre/post-js-changed" behaviour is retained using a dependency on main object..

`set_property` overwrites already set flags, adding `APPEND_STRING` fixes this.

fdelapena:
> Toolchain File Update was merged. So only have to wait for next emscripten release

Emscripten 2.0.0 was tagged 9 days ago 🙌. [Drops Python 2. Only the new wasm backend is supported now](https://github.com/emscripten-core/emscripten/blob/master/ChangeLog.md#200-08102020).

CMakeLists.txt
Ghabry: did this now

src/player.cpp
```diff
@@ -259,6 +259,9 @@ void Player::MainLoop() {
 
 	auto frame_limit = DisplayUi->GetFrameLimit();
 	if (frame_limit == Game_Clock::duration()) {
+#ifdef EMSCRIPTEN
+		emscripten_sleep(0);
```
Ghabry, line 263: this was the reason why my other ASYNC PR never worked: I forgot to add a sleep to this return here. (this sleep 0 is "yield to browser" otherwise nothing will render)

Ghabry:
```
[30/32] cd /...mscripten/easyrpg-player/lib/liblcf/test_runner.js
[doctest] doctest version is "2.3.5"
[doctest] run with "--help" for options
===============================================================================
[doctest] test cases:     10 |     10 passed |      0 failed |      0 skipped
[doctest] assertions:     60 |     60 passed |      0 failed |
[doctest] Status: SUCCESS!
[32/32] cd ..ldscripts/emscripten/easyrpg-player/test_runner.js
[doctest] doctest version is "2.3.5"
[doctest] run with "--help" for options
Debug: Test debg
Warning: Test test
Info: Test info
===============================================================================
[doctest] test cases:    107 |    107 passed |      0 failed |     10 skipped
[doctest] assertions:   3498 |   3498 passed |      0 failed |
[doctest] Status: SUCCESS!
```

Ghabry:
This is now finished. Only needs a Jenkins update now to use CMake.

Ghabry:
Jenkins: Test this please

Ghabry:
Soundfont test case: https://easyrpg.org/play/pr2253/?game=soundfont

Ghabry:
The fix linking commit also fixes linking on macOS now. (#2303)

PR #2254, opened 2020-06-24
Properly check Teleport Targets on add/remove

carstene1ns:


carstene1ns:
https://easyrpg.org/play/master/?start-map-id=2&start-position=14%207
To reproduce: Talk to the npc on your left, then talk to the npc above you twice and go to menu -> tech skill -> alex -> teleport.

Ghabry:
Yeah it should have been lower bound... There is even a commit that replaced lower found with find_if:
https://github.com/EasyRPG/Player/commit/9d94abe2f2bf2f4acab3db7bc700af325d05883f >.>

Anyway. This is a regression since 0.6.2. Should be backported.

PR #2258, opened 2020-07-01
Android : add a favorite feature

BlisterB:
Hello :)

I've added a favorite feature.
A favorite game will be listed first.

I've also done some code clean.


builds/android/app/src/main/java/org/easyrpg/player/game_browser/GameScanner.java
```diff
@@ -94,8 +94,20 @@ private void scanGames(){
             first_directory = false;
         }
 
-        // Sort the games list
-        Collections.sort(gameList);
+        // Sort the games list : alphabetically ordered, favorite in first
```
carstene1ns, line 97: This is ugly. Should be possible to rewrite `compareTo` to do this.

BlisterB: This is indeed done in GameInformation.

carstene1ns: I mean you do not need 2 lists and sort them each on their own.

Ghabry: btw the compare logic could be made more local (where sort is called) by using an inner class or a lambda expresion: https://www.baeldung.com/java-8-sort-lambda

BlisterB: Because compareTo have been rewrite in GameInformation, could you tell me why it's "ugly" and how to improve it?

Ghabry: What he means is that you can just use one list and do a sort with multiple conditions. Is also much more efficient.

carstene1ns: Since GameInformation knows if a game is favourited, you can add that in the comparison.

I am a bit rusty in java, but you get the idea:

```java
	public int compareTo(GameInformation gameInformation) {
		if (this.isFavorite() && ! gameInformation.isFavorite())
			return 1;
		if (! this.isFavorite() && gameInformation.isFavorite())
			return -1;
		return this.title.compareTo(gameInformation.title);
	}
```

BlisterB: I have done 2 separate sort because if I do one only sort, it won't put the favorite game in first.

BlisterB: Also, please note that 2 lists could be usefull for the future. Example if we want to have a "favorite" separate from other game.

BlisterB: Oh, our messages have been intertwined, please ignore the order.

builds/android/app/src/main/java/org/easyrpg/player/game_browser/GameScanner.java
```diff
@@ -94,8 +94,20 @@ private void scanGames(){
...
+        List<GameInformation> favoriteGameList = new ArrayList<GameInformation>();
+        List<GameInformation> noFavoriteGameList = new ArrayList<GameInformation>();
+        for(GameInformation game : gameList) {
+            if(game.isFavorite()) {
+                favoriteGameList.add(game);
+            } else {
+                noFavoriteGameList.add(game);
+            }
+        }
+        Collections.sort(favoriteGameList);
```
Ghabry, line 107: I'm also a bit rusty. Havn't used Java since Java 6. Here is the solution in Java 8 using the new "comparing" API that accepts Functions (omg, you can pass _functions_ as arguments, _finally_)
```java
gameList.sort(
  Comparator.comparing(GameInformation::isFavorite, Comparator.reverseOrder()).
  thenComparing(GameInformation::getTitle));
```

BlisterB:
I've simplified the part sorting the game as suggested by @carstene1ns .
I've also reduce the system call induced by `isFavorite()`

@Ghabry I've haven't coded for several years, I feel that the current implement is still "java compliant" and pretty easy to read. Do we really need to use streams in order to be compliant with new best practices?

BlisterB:
@fdelapena @Ghabry @carstene1ns 
I'm changing the design of the app (not a big deal, but it's definitively an improvement)). I have not finished but I already know that the pull request will be in conflict with this one. Do you want me to wait the merge of this pull request?

Ghabry:
@BlisterB 
The easiest solution in this case is to simply base your new branch on this branch.
Then when this PR is merged you simply rebase on upstream/master to get rid of the now merged commits.

BlisterB:
Okay nice, that's what I did :).

PR #2267, opened 2020-07-13
macOS: Always link against Foundation framework 

Ghabry:
(fix non-SDL2 build such as libretro core)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,6 @@ __MACOSX
 
 # Android
 *.iml
+
+# libretro
+easyrpg_libretro.*
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -478,6 +478,9 @@ if(APPLE)
 	target_sources(${PROJECT_NAME} PRIVATE
 		src/platform/macos/utils.mm
 		src/platform/macos/utils.h)
+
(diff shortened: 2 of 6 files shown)


Ghabry:
Markign this as draft because this fails on Windows with:
"LINK : warning LNK4098: defaultlib 'MSVCRT' conflicts with use of other libs; use /NODEFAULTLIB:library". No idea why, I enforce "/MT" :(

Ghabry:
I must have missed this while reading the CMake changelogs:
Since 3.15 they support a new option ``MSVC_RUNTIME_LIBRARY`` which finally solves this ``MT/MD`` chaos when using Visual Studio.
I set it now automatically based on the VCPKG triplet configuration (which means that only VCPKG works from now on when using MSVC - But we never supported anything else this is even documented).
CMake <3.15 is still supported but I issue noisy warnings.

Ghabry:
I will tag this as "backport to stable". It is veeeery unlikely that there will be a new stable release (I can just cherry-pick these commits in the libretro fork) but just in case.

fmatthew5876 (approved):

Ghabry:
Don't merge yet. I found another problem when compiling libretro-core I currently solve. ... fPIC problems >.>

Ghabry:
now is done

PR #2270, opened 2020-07-15
Allow building with libfmt older than 5.3

carstene1ns:
This is to allow building on debian stable, e.g. Raspberry Pi.


carstene1ns:
test this please

Issue #1920, opened 2019-10-06
Games with altered file extensions


Ghabry:
A few games, e.g. Laxius Power use modified file extensions.

Player will fail to detect such games because RPG_RT.ldb is missing.
No idea how to support this.

File extension changes:
Rpg_rt.ldb to Rpg_rt.set
Rpg_rt.lmt to Rpg_rt.lxb
and all the MapXXXX.lmu to MapXXXX.ont

sorlok:
So I'd like to try fixing this. There's basically 3 things that need to be done:

1) Detect that a folder *might* have renamed game files.
2) Figure out which files were renamed.
3) Fix any internal string renames (things like "LcfMapUnit" are sometimes mangled on purpose).

Note that for all 3 steps, I'd like to avoid reading bytes from RPG_RT.exe directly if possible.

(1) is easy: count the number of "RPG_RT.*" files (should be 3) and "MapXXXX.*" files (should be >0).

For (2), we can assume from (1) what the Map extension is. Of the 3 RPG_RT.* files, 1 is ".ini", and the *larger* of the remaining 2 is almost certainly the database. We could also try parsing the guessed "LMT" file and seeing if it passes, but this seems like overkill. Finally, I'd like to add an option to read the custom extensions from metadata (since we have the RPG_RT.* checksums).

For (3), I think we'll need to just disable the "LcfXXX" checks inside the liblcf parser whenever a custom file extension is in use. Alternatively, if we used metadata for (2), we can also retrieve the "LcfXXX" replacements from metadata, which is a little more reliable.

This approach should support custom extensions with minimal fuss. Let me know if this seems like a good approach and I'll start on a prototype.

Ghabry:
Thanks for working on this :)

This issue can be split in two categories:

Main issue:
 - [ ] Support loading of "folders" with altered file extensions

Secondary issues:
 - [ ] Support in game browser (Game with renamed files listed)
 - [ ] Support in Android Game Browser. See https://github.com/EasyRPG/Player/blob/master/builds/android/app/src/main/java/org/easyrpg/player/game_browser/GameBrowserHelper.java#L45

I would consider this resolved (and open a new issue) when the *Main issue* is fixed but you can decide if you want to fix all of them.

-----------------------------------

> (1) is easy: count the number of "RPG_RT." files (should be 3) and "MapXXXX." files (should be >0).

Better check for 2 as some games have no .ini file.

[...]

PR #2271, opened 2020-07-23
Fix for games with altered file extensions

sorlok:
Ready to merge (from my point of view). 

This is my implementation for Issue #1920; please see the initial discussion there.

Regarding @Ghabry's comments:
1) Check only 2 files: Done, but see note
2) Assume larger==LDB: Done
3) LcfXXX checks: Done, but see note

Some notes on this fix:
1) What's the easiest way to cheaply check if a directory *might* be a game project? Right now I check for RPG_RT.ini, but as you said that's not always present. I expect RPG_RT.exe might also be stripped. Is there any good file to check, or should I just disable this minor performance hack?  ---Side note: it might be worth keeping if *all games with hacked extensions* have .ini files.
2) The Meta file now has "LmtHeaderString=XXX", which allows us to pass the "XXX" string in to liblcf (i.e., to tell it what string to expect within the file header). As you said, errors are disabled if this string doesn't match  --but it still shows a warning. Do we want to allow passing in the expected header string to liblcf when parsing to silence the warning, or is this not worth the trouble?
3) We cannot apply the Meta file until *after* scanning the directory. Main reason: the Meta instance is stored in player.cpp. Sub reason: we don't know which file to calculate the CRC32 of until we scan (since they were renamed). This is not a huge problem, since the scanning is pretty fast.
4) I scan the names of all mapXXXX.ext files to ensure the extension is consistent across files. This might be overkill; we could (a) just stop on the first file or (b) check for the starting map filename (pulled from LMT), assuming we've parsed LMT at this point (which I think we have).
5) This fix assumes all "rpg_rt.XXX" files are renamed by extension ONLY --i.e., if you rename it to "abc.lmt" then this fix won't find it. Do any hacks do this, and do we care?
6) Just reiterating that the meta file can be used to overcome most limitations as discussed (e.g., slow lmu scanning and small LDB files).

### DIFF
--- a/builds/android/app/src/main/java/org/easyrpg/player/game_browser/GameBrowserHelper.java
+++ b/builds/android/app/src/main/java/org/easyrpg/player/game_browser/GameBrowserHelper.java
@@ -30,7 +30,7 @@ import java.util.LinkedList;
 
 public class GameBrowserHelper {
     //Files' names
-    private final static String DATABASE_NAME = "RPG_RT.ldb", TREEMAP_NAME = "RPG_RT.lmt", INI_FILE = "RPG_RT.ini";
+    private final static String DATABASE_NAME = "RPG_RT.ldb", TREEMAP_NAME = "RPG_RT.lmt", INI_FILE = "RPG_RT.ini", EXE_FILE = "RPG_RT.exe";
 
     private final static String TAG_FIRST_LAUNCH = "FIRST_LAUNCH";
     private static int GRANTED_PERMISSION = 0;
@@ -50,6 +50,9 @@ public class GameBrowserHelper {
         boolean databaseFound = false;
         boolean treemapFound = false;
 
+        // Create a lookup by extension as we go, in case we are dealing with non-standard extensions.
+        int rpgrtCount = 0;
+
         for (File entry : dir.listFiles()) {
             if (entry.isFile() && entry.canRead()) {
                 if (!databaseFound && entry.getName().equalsIgnoreCase(DATABASE_NAME)) {
@@ -58,10 +61,24 @@ public class GameBrowserHelper {
                     treemapFound = true;
                 }
 
+                // Count non-standard files.
+                // NOTE: Do not put this in the 'else' statement, since only 1 extension may be non-standard and we want to count both.
+                if (entry.getName().toLowerCase().startsWith("rpg_rt.")) {
+                    if (!(entry.getName().equalsIgnoreCase(INI_FILE) || entry.getName().equalsIgnoreCase(EXE_FILE))) {
+                        rpgrtCount += 1;
+                    }
+                }
+
                 if (databaseFound && treemapFound) {
                     return true;
                 }
             }
+        }
+
+        // We might be dealing with a non-standard extension.
+        // Show it, and let the C++ code sort out which file is which.
+        if (rpgrtCount == 2) {
+            return true;
         }
 
         return false;
--- a/builds/android/app/src/main/java/org/easyrpg/player/game_browser/GameScanner.java
+++ b/builds/android/app/src/main/java/org/easyrpg/player/game_browser/GameScanner.java
@@ -19,11 +19,6 @@ public class GameScanner {
     // (Such as caching games' thumbnail, avoiding some syscall ...)
     private static volatile GameScanner instance = null;
 
-    //Files' names
-    private final static String DATABASE_NAME = "RPG_RT.ldb",
-            TREEMAP_NAME = "RPG_RT.lmt",
-            INI_FILE = "RPG_RT.ini";
-
     private List<GameInformation> gameList;
     private List<String> errorList;
     private Activity context;
@@ -106,42 +101,11 @@ public class GameScanner {
         Log.i("Browser", gameList.size() + " games found : " + gameList);
     }
 
-    /**
-     * Tests if a folder is a RPG2k Game.
-     * (contains DATABASE_NAME and TREEMAP_NAME)
-     * @param dir Directory to test
-     * @return true if RPG2k game
-     */
-    private static boolean isRpg2kGame(File dir) {
-        if (!dir.isDirectory() || !dir.canRead()) {
-            return false;
-        }
-
-        boolean databaseFound = false;
-        boolean treemapFound = false;
-
-        for (File entry : dir.listFiles()) {
-            if (entry.isFile() && entry.canRead()) {
-                if (!databaseFound && entry.getName().equalsIgnoreCase(DATABASE_NAME)) {
-                    databaseFound = true;
-                } else if (!treemapFound && entry.getName().equalsIgnoreCase(TREEMAP_NAME)) {
-                    treemapFound = true;
-                }
-
-                if (databaseFound && treemapFound) {
-                    return true;
-                }
-            }
-        }
-
-        return false;
-    }
-
     private void scanFolder(File[] list, int depth) {
         if (list != null) {
             for (File file : list) {
                 if (!file.getName().startsWith(".")) {
-                    if (isRpg2kGame(file)) {
+                    if (GameBrowserHelper.isRpg2kGame(file)) {
                         gameList.add(new GameInformation(file.getName(), file.getAbsolutePath()));
                     } else if (file.isDirectory() && file.canRead() && depth > 0) {
                         // Not a RPG2k Game but a directory -> recurse
--- a/src/filefinder.cpp
+++ b/src/filefinder.cpp
@@ -41,6 +41,7 @@
 #include "options.h"
 #include "utils.h"
 #include "filefinder.h"
+#include "fileext_guesser.h"
 #include "output.h"
 #include "player.h"
 #include "registry.h"
@@ -726,7 +727,7 @@ std::string FileFinder::FindDefault(const DirectoryTree& tree, const std::string
 }
 
 bool FileFinder::IsValidProject(DirectoryTree const & dir) {
-	return IsRPG2kProject(dir) || IsEasyRpgProject(dir);
+	return IsRPG2kProject(dir) || IsEasyRpgProject(dir) || IsRPG2kProjectWithRenames(dir);
 }
 
 std::string FileFinder::FindDefault(FileFinder::DirectoryTree const &tree, const std::string &dir, const std::string &name,	const char **exts) {
@@ -747,6 +748,10 @@ bool FileFinder::IsEasyRpgProject(DirectoryTree const& dir){
 		lmt_it = dir.files.find(Utils::LowerCase(TREEMAP_NAME_EASYRPG));
 
 	return(ldb_it != dir.files.end() && lmt_it != dir.files.end());
+}
+
+bool FileFinder::IsRPG2kProjectWithRenames(DirectoryTree const& dir) {
+	return !FileExtGuesser::GetRPG2kProjectWithRenames(dir).Empty();
 }
 
 bool FileFinder::HasSavegame() {
--- a/src/game_map.cpp
+++ b/src/game_map.cpp
@@ -271,17 +271,14 @@ void Game_Map::SetupCommon(int _id, bool is_load_savegame) {
 	location.map_id = _id;
 
 	// Try loading EasyRPG map files first, then fallback to normal RPG Maker
-	std::stringstream ss;
-	ss << "Map" << std::setfill('0') << std::setw(4) << location.map_id << ".emu";
-
-	std::string map_file = FileFinder::FindDefault(ss.str());
+	std::string map_name = Game_Map::ConstructMapName(location.map_id, true);
+	std::string map_file = FileFinder::FindDefault(map_name);
 	if (map_file.empty()) {
-		ss.str("");
-		ss << "Map" << std::setfill('0') << std::setw(4) << location.map_id << ".lmu";
-		map_file = FileFinder::FindDefault(ss.str());
+		map_name = Game_Map::ConstructMapName(location.map_id, false);
+		map_file = FileFinder::FindDefault(map_name);
 
 		if (map_file.empty()) {
-			Output::Error("Loading of Map {} failed.\nThe map was not found.", ss.str());
+			Output::Error("Loading of Map {} failed.\nThe map was not found.", map_name);
 		}
 
 		auto map_stream = FileFinder::OpenInputStream(map_file);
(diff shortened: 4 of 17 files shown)


Ghabry:
One quick comment: std::regex is broken on some platforms and can't be used :(

fdelapena:
Jenkins bot: ok to test

fdelapena:
Jenkins: test this please

fdelapena:
jenkins: test this please

Ghabry:
Thanks for your contribution. Really cool to have support for this. Reviewing can currently take longer because it is summer and vacation time, so don't worry if there is less feedback . We are more active during the cold months ;).

Ghabry:
(I was mentioning std::regex before but this should work properly since gcc5.0 so can be ignored)

- What does "Snh" mean?

> What's the easiest way to cheaply check if a directory might be a game project? Right now I check for RPG_RT.ini, but as you said that's not always present. I expect RPG_RT.exe might also be stripped.

When a game tree is parsed the file list is already in memory so the expensive operation (Read directory and alloc all the strings) is already done. I don't think doing some string testing will be really noticable. Also your code will only execute for the "not a game" case which shouldn't happen too often so don't optimize too much.

Sometimes games rename both RPG_RT and the ini. When the Exe is renamed it also changes the ini file it searches for (Game.exe will search for Game.ini)

> [LCF Header] As you said, errors are disabled if this string doesn't match --but it still shows a warning. Do we want to allow passing in the expected header string to liblcf when parsing to silence the warning, or is this not worth the trouble?

The "warning" is only printed on terminal so I don't consider this important

> I scan the names of all mapXXXX.ext files to ensure the extension is consistent across files.

I agree on you this is overkill. Just using the first mapfile you can find is good enough, we can still adopt if any game ever breaks this heuristic... (because there is some Map0001.bak there)

> [RPG_RT prefixed changed] Do any hacks do this, and do we care?

Not sure if any game does this. Is also a bit hard to analyse. When you provide me with some script I can use to scan hundreds of games easily I can tell you quickly ;)

sorlok:
> What does "Snh" mean?
Just my initials; I usually prefix branches with this to minimize the chance of conflict. Let me know if you prefer an alternative naming scheme (I can file a new PR).

> I don't think doing some string testing will be really noticeable.
Great, I removed the INI check.

> The "warning" is only printed on terminal so I don't consider this important
Great, I removed this from the sample .ini file

> Just using the first mapfile you can find is good enough
Done

> When you provide me with some script I can use to scan hundreds of games easily I can tell you quickly
That's not a bad idea. I'll write a quick Python script that can check this stuff.

sorlok:
Rebased and squashed. This branch is ready for merge (from my point of view --let me know if anything needs to change).

sorlok:
Hey all,

Just got back from vacation; wanted to check if there's anything more needed with this branch. (Also if you are all still on vacation that's fine too.) Thanks!

Ghabry:
I'm back from vacation and currently prepare running your script (extracting all the games takes a while \*g\*)

Ghabry:
Some games are molebox encrypted or have corrupted zip paths (\ instead of /) but for ones that work (around 1900): Only 2 are affected by this PR:

- Beyond Black Horizon (https://rmarchiv.de/games/443) - This has changed file extensions
- Verloren (https://rmarchiv.de/games/1333) - Uses default extensions but the maps are in a folder "MM" with names 0001.lmu etc. _out of scope_

sorlok:
> I'm back from vacation and currently prepare running your script (extracting all the games takes a while *g*)
> 
> Minor patch (recurses in a subdirectory when the subdir only has one folder)
> 
> ```
> 91,96c91
> <   dirlist = list(os.listdir(full_dir))
> <   dirs = list(filter(lambda x: os.path.isdir(os.path.join(full_dir, x)), dirlist))
> <   if len(list(dirs)) == 1:
> <     return scan_project(full_dir, dirs[0])
> < 
> <   for filepath in dirlist:
> ---
> >   for filepath in os.listdir(full_dir):
> ```

Updated, thanks

Ghabry:
This needs now a minor rebase

src/filefinder.cpp
```diff
@@ -739,6 +741,132 @@ bool FileFinder::IsEasyRpgProject(DirectoryTree const& dir){
...
+
+	// We need to figure out the names of the map tree and the DB (maps come later).
+	std::vector<RPG2KNonStandardFilenameGuesser::ExtAndSize> candidates;
+	for (const std::pair<std::string, std::string>& item : dir.files) {
+		// Check for any RPG_RT files
+		std::smatch m;
+		if(std::regex_search(item.first, m, rgx)) {
+			std::string ext = m[1].str();
+			if (ext != "exe" && ext != "ini") {
+				candidates.push_back({item.second, ext, GetFileSize(item.second)});
```
Ghabry, line 760: this doesn't work when launched from the Gamebrowser (wrong relative path).

Use:
```
candidates.push_back({item.second, ext, GetFileSize(FileFinder::FindDefault(dir, item.second))});
```

Ghabry:
In my first quick test it worked. Good job :)

src/filefinder.cpp
sorlok: Fixed in 9e361a9bf1052, thanks

sorlok:
Rebased; I'm running a few quick tests (and waiting on builds).

sorlok:
Tested again on Linux and Android; confirmed it's working.

src/filefinder.cpp
```diff
@@ -749,6 +751,132 @@ bool FileFinder::IsEasyRpgProject(DirectoryTree const& dir){
 	return(ldb_it != dir.files.end() && lmt_it != dir.files.end());
 }
 
+bool FileFinder::IsRPG2kProjectWithRenames(DirectoryTree const& dir) {
+	return !GetRPG2kProjectWithRenames(dir).Empty();
+}
+
+FileFinder::RPG2KNonStandardFilenameGuesser FileFinder::GetRPG2kProjectWithRenames(DirectoryTree const& dir) {
```
Ghabry, line 758: Can you move the entire guess/remap logic in a new file? My problem here is that I have a huge PR touching FileFinder in the works and such a huge change is terrible to handle for me :(.

No idea. Make a name up: "filefinder_guesser.cpp/h", "file_guesser.cpp/h", "project_guesser.cpp/.h".

Move ``GetRPG2kProjectWithRenames, GuessAndAddLmuExtension`` and ``RPG2KNonStandardFilenameGuesser, RPG2KFileExtRemap`` to this

``IsRPG2kProjectWithRenames`` can stay as it fits with the rest.

builds/android/app/src/main/java/org/easyrpg/player/game_browser/GameScanner.java
```diff
@@ -128,12 +132,26 @@ private static boolean isRpg2kGame(File dir) {
...
                 }
             }
         }
 
+        // We might be dealing with a non-standard extension.
+        // Show it, and let the C++ code sort out which file is which.
+        if (rpgRTcount == 2) {
+            return true;
+        }
+
```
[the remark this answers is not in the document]
sorlok: Done.

builds/android/app/src/main/java/org/easyrpg/player/game_browser/GameBrowserHelper.java
```diff
@@ -50,6 +50,9 @@ public static boolean isRpg2kGame(File dir) {
         boolean databaseFound = false;
         boolean treemapFound = false;
 
+        // Create a lookup by extension as we go, in case we are dealing with non-standard extensions.
+        int rpgRTcount = 0;
```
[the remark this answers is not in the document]
sorlok: Done.

src/game_map.cpp
```diff
@@ -271,17 +271,14 @@ void Game_Map::SetupCommon(int _id, bool is_load_savegame) {
 	location.map_id = _id;
 
 	// Try loading EasyRPG map files first, then fallback to normal RPG Maker
-	std::stringstream ss;
-	ss << "Map" << std::setfill('0') << std::setw(4) << location.map_id << ".emu";
-
-	std::string map_file = FileFinder::FindDefault(ss.str());
+	std::string mapName = Game_Map::ConstructMapName(location.map_id, true);
```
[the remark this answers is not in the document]
sorlok: Done.

src/game_map.cpp
```diff
@@ -271,17 +271,14 @@ void Game_Map::SetupCommon(int _id, bool is_load_savegame) {
...
 	if (map_file.empty()) {
-		ss.str("");
-		ss << "Map" << std::setfill('0') << std::setw(4) << location.map_id << ".lmu";
-		map_file = FileFinder::FindDefault(ss.str());
+		mapName = Game_Map::ConstructMapName(location.map_id, false);
+		map_file = FileFinder::FindDefault(mapName);
 
 		if (map_file.empty()) {
-			Output::Error("Loading of Map {} failed.\nThe map was not found.", ss.str());
+			Output::Error("Loading of Map {} failed.\nThe map was not found.", mapName.c_str());
```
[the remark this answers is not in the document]
sorlok: Done, thanks for the info (I don't know much about this output api).

src/game_map.cpp
```diff
@@ -1703,11 +1701,18 @@ int Game_Map::GetTargetPanY() {
 	return location.pan_finish_y;
 }
 
-FileRequestAsync* Game_Map::RequestMap(int map_id) {
+std::string Game_Map::ConstructMapName(int map_id, bool isEasyRpg) {
```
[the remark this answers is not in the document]
sorlok: Done.

src/options.h
```diff
@@ -48,13 +48,26 @@
...
+/** Suffixes for their EasyRPG equivalents */
+#define SUFFIX_EDB "edb"
+#define SUFFIX_EMT "emt"
+#define SUFFIX_EMU "emu"
+
 /** lcf::Database filename. */
-#define DATABASE_NAME "RPG_RT.ldb"
-#define DATABASE_NAME_EASYRPG "EASY_RT.edb"
+#define DATABASE_NAME RPG_RT_PREFIX "." SUFFIX_LDB
+#define DATABASE_NAME_EASYRPG RPG_RT_PREFIX "." SUFFIX_EDB
```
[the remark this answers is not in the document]
sorlok: Fixed. Sorry about that, it was not intentional.

src/filefinder.cpp
sorlok: Done.

sorlok:
Changes made and rebased. Tested and works on Linux. I'm still building on Android.

sorlok:
Android is working great. That's everything on my side.

src/filext_guesser.h
```diff
@@ -0,0 +1,97 @@
...
+	 * Scans a directory tree and tries to identify the LMT/LDB files, but with non-standard extensions.
+	 *
+	 * @param dir The directory tree of the project in question
+	 * @return An object that contains the candidates (check with .Empty())
+	 */
+	RPG2KNonStandardFilenameGuesser GetRPG2kProjectWithRenames(FileFinder::DirectoryTree const& dir);
+
+}
+
+#endif
```
Ghabry, line 97: new line at end of file missing

src/filext_guesser.h
```diff
@@ -0,0 +1,97 @@
...
+
+class Meta;
+namespace FileFinder {
+	class DirectoryTree;
+}
+
+/**
+ * FilextGuesser contains helper methods for guessing the extensions used on non-standard RPG Projects.
+ */
+namespace FilextGuesser {
```
Ghabry, line 32: I think there is a typo here. Should be FileExtGuesser and the file fileext_guesser.

Ghabry (commented):
Except for my remaining nits this looks now fine for me.

Do you know how to "squash" in a rebase? If yes please squash the two "Code review" commits into one.

I marked it as "Has PR deps" because I want to merge this after #2293
(that PR changes some string functions and want to attempt a regex -> starts/ends_with change as I expect this to be faster here)

src/filext_guesser.h
sorlok: Fixed.

src/filext_guesser.h
sorlok: Hmm, I was thinking "filext", but I now think that might be trademarked by the associated ".com" address. I'll change this after work.

sorlok:
Renamed filext to fileext (and similar).
Switched from regex to string_view.
Rebased, squashed.

Code seems like it's still working (tested loading two games on Linux).

src/filefinder.cpp
```diff
@@ -26,6 +26,7 @@
 #include <string>
 #include <vector>
 #include <sstream>
+#include <regex>
```
Ghabry, line 29: unnecessary include (thanks for changing this :+1: )

Ghabry (commented):
Want to quickly test this but it looks now fine for me! Thanks! Very useful feature :)

The commit message can be improved. Instead of "Squashed commits" can you use something like "Support games with altered file extensions" (you can keep the verbose message part as is)

src/filefinder.cpp
sorlok: Fixed (and you're welcome!)

sorlok: Also fixed the commit message.

Ghabry (changes requested):
I planned to merge it but I just found an important bug (sorry).

When starting a renamed game the output is "Debug: Encoding not detected". This means that for non-English games the language detection is broken (and they won't run). The encoding detection still only uses RPG_RT.ldb.

I think you must reorder the code a bit. The following must happen before "GetEncoding" is called:

1. Load Meta file
2. Guess extensions
3. GetEncoding
4. [...] stuff
5. LoadDatabase

sorlok:
Ah, interesting. Glad you caught this bug before merging! I'm a bit busy this weekend, but I'll check it out next week.

Ghabry:
About the build failures: That was a very bad timing for updating the PR. There is currently a breaking change from liblcf pending/rebuilding.
You will have to rebase again in ~1h ^^'.

sorlok:
> About the build failures: That was a very bad timing for updating the PR. There is currently a breaking change from liblcf pending/rebuilding.
> You will have to rebase again in ~1h ^^'.

No worries; I'm still working on the fix.

sorlok:
Ok, the fix is in (tested on Laxius Power and a standard game).
1) I left it as its own commit for now, since I think that makes it easier to review. Let me know if/when you want it squashed.
2) Are there any EasyRPG format games I can test this on? I want to make sure I didn't break anything with the new format.

Ghabry (approved):
This works perfect now. Tested it. Thanks for this great contribution!

Ghabry:
@sorlok 
FYI
Made an announcement: https://twitter.com/EasyRPG/status/1301108541185548288

sorlok:
Cool, thanks!

PR #2272, opened 2020-08-03
Fix "man" target not working properly with out-of-source builds

suve:
When performing an out-of-source build, that is instead of:
```
cmake ./
```
doing
```
mkdir build/
cd build/
cmake ../
```

The man page would fail to build and install. This changeset fixes those issues.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -752,13 +752,14 @@ set(MAN_NAME easyrpg-player.6)
 find_program(A2X_EXECUTABLE NAMES a2x a2x.py)
 if(NOT A2X_EXECUTABLE STREQUAL "A2X_EXECUTABLE-NOTFOUND")
 	add_custom_command(OUTPUT resources/${MAN_NAME}
+		COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/resources
 		COMMAND ${A2X_EXECUTABLE} -a player_version="${PROJECT_VERSION}" -f manpage -D ${CMAKE_CURRENT_BINARY_DIR}/resources ${CMAKE_CURRENT_SOURCE_DIR}/resources/${MAN_NAME}.adoc
 		DEPENDS resources/${MAN_NAME}.adoc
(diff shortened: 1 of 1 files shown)


fdelapena:
Thanks!
Jenkins: ok to test

CMakeLists.txt
```diff
@@ -752,13 +752,14 @@ set(MAN_NAME easyrpg-player.6)
 find_program(A2X_EXECUTABLE NAMES a2x a2x.py)
 if(NOT A2X_EXECUTABLE STREQUAL "A2X_EXECUTABLE-NOTFOUND")
 	add_custom_command(OUTPUT resources/${MAN_NAME}
+		COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/resources
```
carstene1ns, line 755: Somehow the directory is created here even without this, but is more correct this way.

CMakeLists.txt
```diff
@@ -752,13 +752,14 @@ set(MAN_NAME easyrpg-player.6)
...
 		COMMAND ${A2X_EXECUTABLE} -a player_version="${PROJECT_VERSION}" -f manpage -D ${CMAKE_CURRENT_BINARY_DIR}/resources ${CMAKE_CURRENT_SOURCE_DIR}/resources/${MAN_NAME}.adoc
 		DEPENDS resources/${MAN_NAME}.adoc
 		COMMENT "(Re-)building manpage ${MAN_NAME}"
 		VERBATIM)
 	if(UNIX)
 		add_custom_target(man ALL DEPENDS resources/${MAN_NAME})
-		install(FILES resources/${MAN_NAME} DESTINATION ${CMAKE_INSTALL_MANDIR}/man6)
+		install(FILES ${CMAKE_CURRENT_BINARY_DIR}/resources/${MAN_NAME} DESTINATION ${CMAKE_INSTALL_MANDIR}/man6)
```
carstene1ns, line 762: Thanks, this was not notified before, because we only tested release archives thoroughly and the manual page is already included there.

carstene1ns:
jenkins: test this please.

PR #2274, opened 2020-08-09
Add if()s around circular find_package() calls

suve:
While packaging EasyRPG Player for the Fedora Linux distribution ([link](https://src.fedoraproject.org/rpms/easyrpg-player)), I stumbled across this problem with the circular dependency between Feeetype and Harfbuzz:
1) CMake would find Freetype
2) It would then try to find HarfBuzz, as wanted by FreeType
3) It would find HarfBuzz
4) It would then try to find FreeType, as wanted by HarfBuzz
5) This second search for FreeType would fail.

This patch adds some `if(NOT TARGET)` guards around the circular `find_package()` calls, solving the problem described above.

### DIFF
--- a/builds/cmake/Modules/FindHarfbuzz.cmake
+++ b/builds/cmake/Modules/FindHarfbuzz.cmake
@@ -55,7 +55,11 @@ if(HARFBUZZ_FOUND)
 
 	if(NOT TARGET Harfbuzz::Harfbuzz)
 		add_library(Harfbuzz::Harfbuzz UNKNOWN IMPORTED)
-		find_library(Freetype REQUIRED)
+
+		if(NOT TARGET Freetype::Freetype)
+			find_library(Freetype REQUIRED)
+		endif()
+
 		set_target_properties(Harfbuzz::Harfbuzz PROPERTIES
 			INTERFACE_INCLUDE_DIRECTORIES "${HARFBUZZ_INCLUDE_DIRS}"
 			INTERFACE_LINK_LIBRARIES Freetype::Freetype
--- a/builds/cmake/Modules/FindHarfbuzz.cmake
+++ b/builds/cmake/Modules/FindHarfbuzz.cmake
@@ -57,7 +57,7 @@ if(HARFBUZZ_FOUND)
 		add_library(Harfbuzz::Harfbuzz UNKNOWN IMPORTED)
 
 		if(NOT TARGET Freetype::Freetype)
-			find_library(Freetype REQUIRED)
+			find_package(Freetype REQUIRED)
 		endif()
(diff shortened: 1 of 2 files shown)


carstene1ns:
I just had this issue today when testing your other pull request. :+1:
The upstream cmake find modules are a bit different (they do not handle the circular dependency), so might be worth to avoid that at all.

carstene1ns:
jenkins: ok to test.

adriaandegroot:
Clever. I just solved this in packaging EasyRPG-Player on FreeBSD with CMake 3.18 in a marginally-different way,

```
--- builds/cmake/Modules/FindHarfbuzz.cmake.orig        2020-08-13 08:40:25 UTC
+++ builds/cmake/Modules/FindHarfbuzz.cmake
@@ -55,7 +55,13 @@ if(HARFBUZZ_FOUND)
 
        if(NOT TARGET Harfbuzz::Harfbuzz)
                add_library(Harfbuzz::Harfbuzz UNKNOWN IMPORTED)
-               find_library(Freetype REQUIRED)
+               # Handle circular dependency: FindFreetype.cmake asks
+               # for a QUIET lookup.
+               if (Harfbuzz_FIND_QUIETLY)
+                       find_library(Freetype QUIET)
+               else()
+                       find_library(Freetype REQUIRED)
+               endif()
```

but it comes down to the same thing.

Ghabry:
Interesting. Never saw this error before, but also havn't updated my system since a month.

Slow internet + Arch = Almost no updates :(

Ghabry:
The real bug is here that ``find_library(Freetype REQUIRED)`` is used instead of ``find_package(Freetype REQUIRED)``.
CMake will do a simple name-lookup instead of using the find module, though no idea why this broke now.

But I agree that wrapping them in ``if NOT TARGET`` checks is good style :+1:

builds/cmake/Modules/FindHarfbuzz.cmake
```diff
@@ -55,7 +55,11 @@ if(HARFBUZZ_FOUND)
 
 	if(NOT TARGET Harfbuzz::Harfbuzz)
 		add_library(Harfbuzz::Harfbuzz UNKNOWN IMPORTED)
-		find_library(Freetype REQUIRED)
+
+		if(NOT TARGET Freetype::Freetype)
+			find_library(Freetype REQUIRED)
```
Ghabry, line 60: could you change this to ``find_package`` pls?

suve: Sure. Done in 199c70c52fb1993a47f04382d43f753282cedace.

PR #2282, opened 2020-08-17
Upstream DynRpg Parser

Ghabry:
This PR has the goal to upstream the DynRPG Parser.
This one was written as part of commisioned work in 2017/18 for the Deep8 game.

The main motivation for the submission is that this code bitrots and always updating to latest Player code is a huge task and this is imo useful code.
Right now I won't submit any plugins.
The only plugin worth upstreaming right now is the DynTextPlugin, all others are of terrible code quality or custom modifications for Deep8.

Obviously there is no plugin system in the Player, a "plugin" is just a class inheriting from DynRpgPlugin and registred in ``create_all_plugins``.

The code is not very invasive, I only added the following "Hook points":
 - Comment Event (parses and executes DynRPG commands)
 - Load
 - Save
 - Update (Spriteset graphics update)
 - Reset

Most code in dynrpg.cpp is the parser, which works by parsing the line char-by-char and keeps track of the current state through an enum. Nothing special, typical way to implement a parser. The parser is tested as it is used in Deep8 since 2 years but this is code you better don't ever touch again :).

A sample plugin (dynrpg_easyrpg) is provided. When DynRPG is enabled (dynloader.dll present) it will export the following functions:

 - \@call: This is a forwarder for RPGSS to get better warning output. It just takes the first argument as a function name and forwards the other arguments to the function
- \@easyrpg_add: 1st argument is the output variable, further arguments are the values that are added, example: ``@easyrpg_add 5, 10, V2, 30 <- V[5] = 10 + V[2] + 30``.
- \@easyrpg_output: Outputs to the console. 1st arg: Debug, Info, Warning, Error, 2nd vararg string, 3+ varargs, example: ``@easyrpg_output Info, "Hello $1! Var5 is $2", World, V5`` (this is a unique feature, DynRpg itself has no vararg support)

Basic DynRpg function syntax:
- The function name is always prefixed with a \@
- An argument is either a Token (a string without ""), a string (in "") or a Variable (Token prefixed with V)

How the Parser in the Player interprets different things: When parsed for simplicity everything is stored as a string and converted to int on demand.
- Hello - This is a Token, stored as "Hello"
- Hello World - This is a Token, stored as "HelloWorld" (DynRpg compatible - Whitespace is removed)
- "Hello World" - stored as "Hello World"
- V1 - Substituted with the variable, then the result stored as string. Is a valid number.
- VV1 - Variable Indirection, as above. Is a valid number.
- 10 - This is a Token, stored as "Hello". Is a valid number.

The Sample Plugin writes savedata. The save format of DynRPG is very simple:
- Header: DYNSAVE1
- Size of the Plugin Identifier (4 byte)
- Plugin Identifier
- Size of the Plugin data
- Plugin data
- [Repeat until EOF]

The sample plugin writes "EasyRpgPlugin" as identifier and the Player version as 4 byte as an indicator that EasyRPG wrote this file and not the official DynRpg.

Loading works by looking for a plugin that has the same identifier as the one in the save file.

### DIFF
--- a/src/dynrpg_easyrpg.h
+++ b/src/dynrpg_easyrpg.h
@@ -1,0 +1,38 @@
+/*
+ * This file is part of EasyRPG Player.
+ *
+ * EasyRPG Player is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * EasyRPG Player is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef EP_DYNRPG_EASYRPG_H
+#define EP_DYNRPG_EASYRPG_H
+
+#include "dynrpg.h"
+
+namespace DynRpg {
+	/**
+	 * A DynRPG plugin that provides EasyRPG specific built-in functions.
+	 * Mostly for testing.
+	 */
+	class EasyRpgPlugin : public DynRpgPlugin {
+	public:
+		EasyRpgPlugin() : DynRpgPlugin("EasyRpgPlugin") {}
+
+		void RegisterFunctions() override;
+		void Load(std::vector<uint8_t>& buffer) override;
+		std::vector<uint8_t> Save() override;
+	};
+}
+
+#endif
--- a/src/dynrpg_easyrpg.h
+++ b/src/dynrpg_easyrpg.h
@@ -30,7 +30,7 @@ namespace DynRpg {
 		EasyRpgPlugin() : DynRpgPlugin("EasyRpgPlugin") {}
 
 		void RegisterFunctions() override;
-		void Load(std::vector<uint8_t>& buffer) override;
+		void Load(const std::vector<uint8_t>& buffer) override;
 		std::vector<uint8_t> Save() override;
 	};
 }
--- a/src/player.h
+++ b/src/player.h
@@ -144,8 +144,9 @@ namespace Player {
 	 * Loads savegame data.
 	 *
 	 * @param save_file Savegame file to load
+	 * @param save_id ID of the savegame to load
 	 */
-	void LoadSavegame(const std::string& save_file);
+	void LoadSavegame(const std::string& save_file, int save_id = 0);
 
 	/**
 	 * Starts a new game
--- a/tests/dynrpg.cpp
+++ b/tests/dynrpg.cpp
@@ -114,8 +114,10 @@ TEST_CASE("Variable names with junk") {
 }
 
 TEST_CASE("Arg parse") {
+	// 2.5x not tested here because the result differs between different C++ libraries
+	// See: https://bugs.llvm.org/show_bug.cgi?id=17782
 	LogGuard lg;
-	std::vector<std::string> args = {"A", "1x", "2.5y"};
+	std::vector<std::string> args = {"A", "1y", "2.5 x"};
 	bool okay = false;
 
 	std::string s, s2;
@@ -134,15 +136,28 @@ TEST_CASE("Arg parse") {
 	CHECK(i2 == 0);
 	CHECK(f == 0.0);
 
+	std::tie(f, i,s) = DynRpg::ParseArgs<float, int, std::string>("func", args, &okay);
+	CHECK(!okay);
+	CHECK(f == 0.0f);
+	CHECK(i == 0);
+	CHECK(s == "");
+
 	std::tie(s, s2, i) = DynRpg::ParseArgs<std::string, std::string, int>("func", args, &okay);
 	CHECK(okay);
 	CHECK(s == "A");
-	CHECK(s2 == "1x");
+	CHECK(s2 == "1y");
 	CHECK(i == 2);
 
 	std::tie(std::ignore, std::ignore, s) = DynRpg::ParseArgs<std::string, std::string, std::string>("func", args, &okay);
 	CHECK(okay);
-	CHECK(s == "2.5y");
+	CHECK(s == "2.5 x");
+
+	// DynRPG reports string here but for convenience empty strings are 0
+	args = {"", ""};
+	std::tie(f, i) = DynRpg::ParseArgs<float, int>("func", args, &okay);
+	CHECK(okay);
+	CHECK(f == 0.0f);
+	CHECK(i == 0);
 }
 
 TEST_CASE("easyrpg dynrpg invoke") {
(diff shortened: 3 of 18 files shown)


src/player.h
```diff
@@ -43,6 +43,15 @@ namespace Player {
 		EngineEnglish = 8
 	};
 
+	/** Bitmask for activated patches */
+	enum PatchType {
+		PatchNone = 0,
+		/** DynRPG */
+		PatchDynRpg = 1,
+		/** ManiacPatch */
+		PatchManiacPatch = 2
```
fmatthew5876, line 52: Why not just `PatchManiac`

Ghabry: no particular reason, Patch is the enum prefix and ManiacPatch the patch name ;). But I changed it.

src/dynrpg_easyrpg.h
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+#ifndef EP_DYNRPG_EASYRPG_H
+#define EP_DYNRPG_EASYRPG_H
+
+#include "dynrpg.h"
+
+namespace DynRpg {
+	/**
+	 * A DynRPG plugin that provides EasyRPG specific built-in functions.
```
fdelapena, line 25: It would be convenient a reminder somewhere on how license affects plugins if game devs want to add their own to Player.

Ghabry: Because a game is not a derived work I don't see the problem here. Is like adding new event commands.

fmatthew5876 (approved):
As you mentioned this has been working for years and is "don't touch" code. I am approving without too thorough of a review.

One thing I would suggest, which doesn't have to be done now, is getting a good suite of unit tests around this parser to make sure future code changes and refactors don't break it.

Ghabry:
Yeah Unit tests is a good idea.

I also consider replacing the macros with templates as a little challenge ^^'

Ghabry:
Replaced the vector with Span (nice API) and the macro parsing with a std::tuple-integer-sequence-.....stuff.
Think I will add some tests :)

Ghabry:
@CherryDT

Quick dynrpg question:

```
@ABc "bla",N2,123
```
and this plugin:

```cpp
// This handler is called when an event comment was encountered
bool onComment(const char *text, const RPG::ParsedCommentData *parsedData,
	RPG::EventScriptLine *nextScriptLine, RPG::EventScriptData *scriptData,
	int	eventId, int pageId, int lineId, int *nextLineId)
{
	printf("Cmd: %s %s\n", parsedData->command, text);
	printf("Args: %d\n", parsedData->parametersCount);
	for (int i = 0; i < parsedData->parametersCount; ++i) {
		const auto& p = parsedData->parameters[i];
		if (p.type == RPG::PARAM_NUMBER)
			printf("Num: %d %f\n", i, p.number);
		else
			printf("%s: %d %s\n", p.type == RPG::PARAM_TOKEN ? "Tok" : "Str",  i, p.text);
	}
	return true; // If we weren't able to handle the command, we return true!
}
```

Doesn't matter what I pass for "N" the string is always empty. Shouldn't I receive the actor name?

CherryDT:
Yes you should receive name of actor 2... (as type string)

Maybe this works only if the name was changed (not the DB default) but that would be a bug then. The N feature didn't seem to receive much use... Which DynRPG version are you testing against?

Ghabry:
@CherryDT
I get an empty "type string" string.

I tested with PepsiOtaku 0.32 and with your release.

In both I'm not receiving anything.

Also rewrote the database string with the Change Name event command.

Ghabry:
hmm, testing the parser actually shows many "interesting" edge cases that I don't want to support before they are relevant in existing code.

E.g. "V1.5" is interpreted as "Read value from var 2" because 1.5 is rounded to two?! Wat.

Ghabry:
@fmatthew5876 

I found a strange behaviour in Variables (and Switches).
https://github.com/EasyRPG/Player/blob/master/src/game_variables.h#L122
Here and a few lines below it accesses ``lcf::rpg::Variables``. Is this intentional? That's the only place where this global data is accessed. Considering that the class should be decoupled from lcf::rpg::variables this is strange?

When I do a unit test with the following code:
```
Main_Data::game_variables = std::make_unique<Game_Variables>(-99999, 99999);
std::vector<int32_t> vars = {100, 4, 2, 1};
Main_Data::game_variables->SetData(vars);
CHECK(DynRpg::ParseCommand("@FunC V2,VV3,VVVV3", args) == "func");
```
All arguments of DynRPG will be 0 because the check ``!Main_Data::game_variables->IsValid(number)`` fails.

Can I fix this or is the code in variables/switches correct?

Ghabry:
hm wait. This is just not intuitive. IsValid checks if it is in the database range (for the Debug dialog).
So not a bug, I just have to remove the check from the dynrpg code (no idea why it is even there, there is no var limit o.O)

Ghabry:
This is ready now when I can get macOS to accept my unit test...

Found actually two bad bugs in the parser:

 - Tokens (everything that is not a string literal) must be lowercased
 - "@bla ," was not parsed as "one empty string arg"

Ghabry:
ohh, that error is interesting and shows that unit tests are useful:

To check if it is a float I do:
```
std::istringstream iss(str);
iss >> f;
parse_okay = !iss.fail();
```
libstc++ and MSVC do not set the fail-bit for "1.5x" but libc++ (clang) does. I relied on the first behaviour.

Ghabry:
Well fixing this is tricky so I attempted to make the problem less likely to appear instead:
The int parsing is not done via float parsing anymore (also faster this way).

It will now only happen when:
- The parameter is a float AND
- The float contains trailing text, e.g. "1.5x" (but "1.5 x" is fine).
So only "wrong" input to float parameters will trigger this behaviour.

Ghabry:
rebased. Imo this PR is done.

fdelapena (approved):

tests/dynrpg.cpp
```diff
@@ -0,0 +1,204 @@
...
+	CHECK(args.size() == 3);
+	CHECK(args[0] == "4");
+	CHECK(args[1] == "4");
+	CHECK(args[2] == "100");
+
+	Output::SetLogLevel(LogLevel::Error);
+}
+
+TEST_CASE("Variable names with junk") {
+	Main_Data::game_variables = std::make_unique<Game_Variables>(-99999, 99999);
```
fmatthew5876, line 83: Lets allocate `game_variables` in `MockActor` and use it here. That way the variables get destroyed when this test exits.

Since all unit tests run under one binary, we have to be super careful about resetting all global state.

Once I've refactored singletons for Player, I'll also refactor for tests. For now we're stuck with this messy half situation.

tests/dynrpg.cpp
```diff
@@ -0,0 +1,204 @@
...
+		lvl = Output::GetLogLevel();
+		Output::SetLogLevel(LogLevel::Error);
+	}
+
+	~LogGuard() {
+		Output::SetLogLevel(lvl);
+	}
+
+	LogLevel lvl;
+};
```
fmatthew5876, line 21: `MockActor` does this as well, might as well use it and not have to invent a new guard.

Ghabry: Addressed your two comments. I still like my pun on LockGuard :P

fmatthew5876 (approved):

PR #2283, opened 2020-08-18
Warn when DynRPG or ManiacPatch are detected

Ghabry:
first commit extracted from #2282.

This one is worth backporting.

### DIFF
--- a/src/game_pictures.cpp
+++ b/src/game_pictures.cpp
@@ -124,6 +124,10 @@ int Game_Pictures::GetDefaultNumberOfPictures() {
 		return 1000;
 	}
 	else if (Player::IsMajorUpdatedVersion()) {
+		if (Player::IsPatchDynRpg() && Player::IsRPG2k3()) {
+			// DynRPG only exists for RPG Maker 2003 1.08
+			return 2000;
+		}
 		return 50;
 	}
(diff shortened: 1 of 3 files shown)


src/player.h
```diff
@@ -223,6 +232,16 @@ namespace Player {
 	 */
 	bool IsCP1251();
 
+	/**
+	 * @return True when the DynRPG patch is active
+	 */
+	bool IsDynRpg();
```
fmatthew5876, line 238: I know i'm being pedantic, but should these be `IsPatchDynRpg` `IsPatchManiac` etc.. Basically all of the are called `IsPatchSomething` so we know they all are detecting patches, no matter how many we support.

src/game_pictures.cpp
```diff
@@ -124,6 +124,10 @@ int Game_Pictures::GetDefaultNumberOfPictures() {
 		return 1000;
 	}
 	else if (Player::IsMajorUpdatedVersion()) {
+		if (Player::IsPatchDynRpg() && Player::IsRPG2k3()) {
+			// DynRPG only exists for RPG Maker 2003 1.08
+			return 2000;
```
Ghabry, line 129: If you don't mind: I added a commit that allocates 2000 pic slots for savegame compat.

PR #2285, opened 2020-08-19
RFC: Add db:: accessor for LDB and LMT data

fmatthew5876:
Depends on: https://github.com/EasyRPG/liblcf/pull/384

Fix: #2281 

This is a new interface to access the lcf database and treemap from all Player code.

Essentially we will change from this:
```
auto& actors = lcf::Data::actors;
auto* actor = lcf::ReaderUtil::GetElement(actors, 1);
```

To this
```
auto& actors = db::Actors();

// Assumed id is always valid, asserts in debug build
auto& actor = db::Get(actors, 1);

// Requires you to check the pointer for validity
auto* actor = db::GetPtr(actors, 1);

// Like before, but also logs a warning if id is invalid
auto* actor = db::GetPtr(actors, 1, "MyFunction");
```

The reasons for changing this interface are:
* The database singleton is now part of player code and not lcf
* Enables us refactor player *before* removing `lcf::Data` and all singletons from liblcf.
* More concise name `db::`
* Potentially faster compile speeds. Including `db.h` doesn't bring in any lcf/rpg headers, only forward declarations.

I'm posting this for review on the interface. Once everyone is on board with the design, I'll add commits to this porting Player code to use the new interface. Need to avoid rebase hell on this one.

Issue #2048, opened 2020-01-13
Yume2kki: Message Box Show Choice looks different


Ghabry:
#### Name of the game:
Yume2kki 

Lets stay on 109f for now, doesn't matter.

#### Attach files
https://github.com/EasyRPG/Player/files/3490326/yume2kkiaudio.zip

* At the Computer select "Music"
* Press ESC to open the menu.
* Select ..... [1/2] (thats the last option)
* Select the 3rd option (サウンドルーム...)

Screenshot comparison: (unreadable one is RPG_RT xD)
![Screenshot_20200113_115952-fs8](https://user-images.githubusercontent.com/1331889/72250610-44c4cb80-35fc-11ea-990b-2e68bf1412a7.png)

Ghabry:
@fmatthew5876 
Marking this for 0.6.2 for now as this one here looks suspicious, maybe is a real message box bug.

fmatthew5876:
Here is the code where this triggers, dumped from my `hh3tool`

```
0827:     code=ConditionalBranch(12010) indent=4 str="" params=[0 4383 1 0 0 1]
0828:      code=ShowMessage(10110) indent=5 str="\>\C[10]\C[0]ގI" params=[]   
0829:      code=ShowMessage_2(20110) indent=5 str="\>" params=[]                 
0830:      code=ShowChoice(10140) indent=5 str="ގ" params=[5]                         
0831:      code=ShowChoiceOption(20140) indent=5 str="\>ގ\N[6]֖߂" params=[0]           
0832:       code=Comment(12410) indent=6 str="----------------------------------------
----------------------------------" params=[]                                 
0833:       code=ControlSwitches(10210) indent=6 str="" params=[0 3942 3942 1]        
0834:       code=Comment(12410) indent=6 str="223B̉o܂B-Nulsdodage" params=[]           
0835:       code=Comment(12410) indent=6 str="----------------" params=[]      
0836:       code=END(10) indent=6 str="" params=[]                 
0837:      code=ShowChoiceOption(20140) indent=5 str="\>SRֈړ" params=[1]              
0838:       code=Comment(12410) indent=6 str="----------------------------------------
----------------------------------" params=[]                                         
0839:       code=ControlSwitches(10210) indent=6 str="" params=[0 3942 3942 1]
0840:       code=Comment(12410) indent=6 str="223B̉o܂B-Nulsdodage" params=[]           
0841:       code=Comment(12410) indent=6 str="----------------" params=[]             
[...]

PR #2287, opened 2020-08-20
Initial Work on Using Translations Via LcfTrans

sorlok:
~~DO NOT MERGE: This is for discussion only.~~

~~This is a draft PR;~~ see Issue #797 for the initial discussion. Further code discussion will go here.

### DIFF
--- a/src/scene_title.h
+++ b/src/scene_title.h
@@ -121,15 +121,16 @@ private:
 
 	/**
 	 * Moves a window (typically the New/Continue/Quit menu) to the middle or bottom-center of the screen.
-	 * @param centerVertical If true, the menu will be centered vertically. Otherwise, it will be at the bottom of the screen.
+	 * @param window The window to resposition.
+	 * @param center_vertical If true, the menu will be centered vertically. Otherwise, it will be at the bottom of the screen.
 	 */
-	void RepositionWindow(Window_Command& window, bool centerVertical);
+	void RepositionWindow(Window_Command& window, bool center_vertical);
 
 	/**
 	 * Picks a new language based and switches to it.
-	 * @param langStr If the empty string, switches the game to 'No Translation'. Otherwise, switch to that translation by name.
+	 * @param lang_str If the empty string, switches the game to 'No Translation'. Otherwise, switch to that translation by name.
 	 */
-	void ChangeLanguage(const std::string& langStr);
+	void ChangeLanguage(const std::string& lang_str);
 
 	void HideTranslationWindow();
 
@@ -165,11 +166,11 @@ private:
 	 * Stored in a struct for easy resetting, as Scene_Title can be reused.
 	 */
 	struct CommandIndices {
-		int new_game       =  0;
-		int continue_game  =  1;
-		int import         = -1;
-		int translate      = -1;
-		int exit           =  2;
+		int new_game =  0;
+		int continue_game =  1;
+		int import = -1;
+		int translate = -1;
+		int exit =  2;
 	};
 	CommandIndices indices;
 
--- a/src/scene_title.h
+++ b/src/scene_title.h
@@ -52,6 +52,16 @@ public:
 	void CreateCommandWindow();
 
 	/**
+	 * Creates the Window displaying available translations.
+	 */
+	void CreateTranslationWindow();
+
+	/**
+	 * Creates the Help window and hides it
+	 */
+	void CreateHelpWindow();
+
+	/**
 	 * Plays the title music.
 	 */
 	void PlayTitleMusic();
@@ -90,6 +100,12 @@ public:
 	void CommandImport();
 
 	/**
+	 * Option Translation.
+	 * Shows the Translation menu, for picking between multiple languages or localizations
+	 */
+	void CommandTranslation();
+
+	/**
 	 * Option Shutdown.
 	 * Does a player shutdown.
 	 */
@@ -103,17 +119,59 @@ public:
 private:
 	void OnTitleSpriteReady(FileRequestResult* result);
 
+	/**
+	 * Moves a window (typically the New/Continue/Quit menu) to the middle or bottom-center of the screen.
+	 * @param centerVertical If true, the menu will be centered vertically. Otherwise, it will be at the bottom of the screen.
+	 */
+	void RepositionWindow(Window_Command& window, bool centerVertical);
+
+	/**
+	 * Picks a new language based and switches to it.
+	 * @param langStr If the empty string, switches the game to 'No Translation'. Otherwise, switch to that translation by name.
+	 */
+	void ChangeLanguage(const std::string& langStr);
+
+	void HideTranslationWindow();
+
 	/** Displays the options of the title scene. */
 	std::unique_ptr<Window_Command> command_window;
+
+	/** Displays all available translations (languages). */
+	std::unique_ptr<Window_Command> translate_window;
+
+	/** Displays help text for a given language **/
+	std::unique_ptr<Window_Help> help_window;
+
+	/** Contains directory names for each language; entry 0 is resverd for the default (no) translation */
+	std::vector<std::string> lang_dirs;
+
+	/** Contains help strings for each language; entry 0 is resverd for the default (no) translation */
+	std::vector<std::string> lang_helps;
 
 	/** Background graphic. */
 	std::unique_ptr<Sprite> title;
 
-	/** Offsets for each selection, in case "Import" is enabled. */
-	int new_game_index =  0;
-	int continue_index =  1;
-	int exit_index     =  2;
-	int import_index   = -1;
+	/**
+	 * Current active window
+	 *   0 = command
+	 *   1 = translate
+	 */
+	int active_window = 0;
+
+	/** 
+	 * Offsets for each selection, in case "Import" or "Translate" is enabled.
+	 *   Listed in the order they may appear; exit_index will always be last,
+	 *   and import appears before translate, if it exists.
+	 * Stored in a struct for easy resetting, as Scene_Title can be reused.
+	 */
+	struct CommandIndices {
+		int new_game       =  0;
+		int continue_game  =  1;
+		int import         = -1;
+		int translate      = -1;
+		int exit           =  2;
+	};
+	CommandIndices indices;
 
 	/** Contains the state of continue button. */
 	bool continue_enabled = false;
--- a/src/translation.cpp
+++ b/src/translation.cpp
@@ -458,12 +458,9 @@ namespace {
 
 
 
-std::vector<std::vector<std::string>> Translation::TranslateMessageStream(const Dictionary& dict, const std::stringstream& msg1, const std::stringstream* msg2, char trimChar) {
+std::vector<std::vector<std::string>> Translation::TranslateMessageStream(const Dictionary& dict, const std::stringstream& msg, char trimChar) {
 	// Prepare source string
-	std::string msgStr = msg1.str();
-	if (msg2!=nullptr) {
-		msgStr += msg2->str();
-	}
+	std::string msgStr = msg.str();
 	if (msgStr.size()>0 && msgStr.back() == trimChar) {
 		msgStr.pop_back();
 	}
@@ -507,60 +504,32 @@ void Translation::RewriteEventCommandMessage(const Dictionary& dict, std::vector
 	// have to track the current index directly in this function.
 	CommandIterator commands(commandsOrig);
 	while (!commands.Done()) {
-		// Logic: build up both the Message stream and the Choice stream, since we'll need to 
-		//        deal with both eventually. Then, if they can be merged do that.
-		if (commands.CurrIsShowMessage() || commands.CurrIsShowChoice()) {
-			// First build up the lines of Message texts
+		// We only need to deal with either Message or Choice commands
+		if (commands.CurrIsShowMessage()) {
+			// Build up the lines of Message texts
 			std::stringstream msg_str;
 			std::vector<size_t> msg_indexes;
 			commands.BuildMessageString(msg_str, msg_indexes);
 
-			// Next, build up the lines of Choice elements
-			std::stringstream choice_str;
-			std::vector<size_t> choice_indexes; // Number of entries == number of choices
-			commands.BuildChoiceString(choice_str, choice_indexes);
-
-			// Will they fit on screen if we combine them?
-			bool combine = false; //msg_indexes.size()>0 && msg_indexes.size()+choice_indexes.size() <= 4;
-
 			// Go through messages first, including possible choices
 			if (msg_indexes.size()>0) {
 				// Get our lines, possibly including "combined"
-				std::vector<std::vector<std::string>> msgs = TranslateMessageStream(dict, msg_str, (combine?&choice_str:nullptr), '\n');
+				std::vector<std::vector<std::string>> msgs = TranslateMessageStream(dict, msg_str, '\n');
 				if (msgs.size()>0) {
 					// The complex replacement logic is based on the last message box, then all remaining things are simply left back in.
 					std::vector<std::string>& lines = msgs.back();
 
-					// There is a special case here: if we are asked to remove a message box, we should cancel the "combine" action and do nothing further
+					// There is a special case here: if we are asked to remove a message box, we should do nothing further
 					// This command is *only* respected as the first line of a message box.
 					if (lines[0]==TRCUST_REMOVEMSG) {
-						// Update the index of all choice items first
-						for (size_t& index : choice_indexes) {
-							index -= msg_indexes.size();
-						}
-
-						// Now clear all message boxes in reverse order.
+						// Clear all message boxes in reverse order.
 						while (!msg_indexes.empty()) {
 							commands.RemoveByIndex(msg_indexes.back());
 							msg_indexes.pop_back();
 						}
-
-						// Finally, reset the "combine" flag so that we process the message box command correctly.
-						combine = false;
 					} else {
-						// We only need the last X Choices from the translation, since we can't change the Choice count.
-						size_t maxLines = 4;
-						if (combine) {
-							// Go backwards through our lines/choices
-							while (!choice_indexes.empty() && !lines.empty()) {
-								commands.ReWriteString(choice_indexes.back(), lines.back());
-								choice_indexes.pop_back();
-								lines.pop_back();
-								maxLines -= 1;
-							}
-						}
-
 						// Trim lines down to allowed remaining (with choices).
+						const size_t maxLines = 4;
 						while (lines.size() > maxLines) {
 							lines.pop_back();
 						}
@@ -589,10 +558,17 @@ void Translation::RewriteEventCommandMessage(const Dictionary& dict, std::vector
 				}
 			}
 
-			// Go through choices second.
-			if (!combine && choice_indexes.size()>0) {
+			// Note that commands.Advance() has already happened within the above code.
+		} else if (commands.CurrIsShowChoice()) {
+			// Build up the lines of Choice elements
+			std::stringstream choice_str;
+			std::vector<size_t> choice_indexes; // Number of entries == number of choices
+			commands.BuildChoiceString(choice_str, choice_indexes);
+
+			// Go through choices. 
+			if (choice_indexes.size()>0) {
 				// Translate, break back into lines.
-				std::vector<std::vector<std::string>> msgs = TranslateMessageStream(dict, choice_str, nullptr, '\n');
+				std::vector<std::vector<std::string>> msgs = TranslateMessageStream(dict, choice_str, '\n');
 				if (msgs.size()>0) {
 					// Logic here is also based on the last message box.
 					std::vector<std::string>& lines = msgs.back();
--- a/src/translation.cpp
+++ b/src/translation.cpp
@@ -1,0 +1,784 @@
+/*
+ * This file is part of EasyRPG Player.
+ *
+ * EasyRPG Player is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * EasyRPG Player is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "translation.h"
+
+// Headers
+#include <fstream>
+#include <iomanip>
+#include <lcf/data.h>
+#include <lcf/rpg/terms.h>
+#include <lcf/rpg/map.h>
+#include "lcf/rpg/mapinfo.h"
+
+#include "cache.h"
+#include "main_data.h"
+#include "game_actors.h"
+#include "game_map.h"
+#include "player.h"
+#include "output.h"
+#include "utils.h"
+
+
+// Name of the translate directory
+#define TRDIR_NAME "languages"
+
+// Name of expected files
+#define TRFILE_RPG_RT_LDB    "rpg_rt.ldb.po"
+#define TRFILE_RPG_RT_COMMON "rpg_rt.ldb.common.po"
+#define TRFILE_RPG_RT_BATTLE "rpg_rt.ldb.battle.po"
+#define TRFILE_RPG_RT_LMT    "rpg_rt.lmt.po"
+#define TRFILE_META_INI      "META.INI"
+
+// Message box commands to remove a message box or add one in place.
+#define TRCUST_REMOVEMSG        "__EASY_RPG_CMD:REMOVE_MSGBOX__"
+#define TRCUST_ADDMSG           "__EASY_RPG_CMD:ADD_MSGBOX__"
+
+
+std::string Tr::TranslationDir() {
+	return Player::translation.RootDir();
+}
+
+std::string Tr::CurrTranslationId() {
+	return Player::translation.GetCurrLanguageId();
+}
+
+void Translation::Reset()
+{
+	ClearTranslationLookups();
+
+	languages.clear();
+	currLanguage = "";
+	translationRootDir = "";
+}
+
+void Translation::InitTranslations()
+{
+	// Reset
+	Reset();
+
+	// Determine if the "languages" directory exists, and convert its case.
+	auto tree = FileFinder::GetDirectoryTree();
+	auto langIt = tree->directories.find(TRDIR_NAME);
+	if (langIt != tree->directories.end()) {
+		// Save the root directory for later.
+		translationRootDir = langIt->second;
+
+		// Now list all directories within the translate dir
+		auto translation_path = FileFinder::MakePath(FileFinder::GetDirectoryTree()->directory_path, translationRootDir);
+		auto translation_tree = FileFinder::CreateDirectoryTree(translation_path, FileFinder::RECURSIVE);
+		if (translation_tree == nullptr) {  return; }
+
+		// Now iterate over every subdirectory.
+		for (const auto& trName : translation_tree->directories) {
+			Language item;
+			item.langDir = trName.second;
+			item.langName = trName.second;
+
+			// If there's a manifest file, read the language name and help text from that.
+			std::string metaName = FileFinder::FindDefault(*translation_tree, trName.second, TRFILE_META_INI);
+			if (!metaName.empty()) {
+				lcf::INIReader ini(metaName);
+				item.langName = ini.GetString("Language", "Name", item.langName);
+				item.langDesc = ini.GetString("Language", "Description", "");
+			}
+
+			languages.push_back(item);
+		}
+	}
+}
+
+std::string Translation::GetCurrLanguageId() const 
+{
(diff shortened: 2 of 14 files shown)


fmatthew5876 (dismissed):
Doing all this text conversion on the fly is going to slow down games that don't use this feature. It also adds a lot of complexity over the code and you're bound to miss a few cases.

Can't we do it on startup? LIke on game initialization, retranslate the whole database. On load map, retranslate all strings in the map data. On load save game, retranslate all strings in the save game etc..

Maybe we can add a nice autogenerated hook to liblcf that lets you iterate over all string members of all lcf struct structs.

sorlok:
Hmm, I was thinking the performance hit wouldn't be that big, but then again for things like Nintendo 3DS and Android it might start to matter. Translating the DB on the fly whenever the translation is selected makes sense, although we'd need to reload the original LCF if the translation was changed at runtime (so: slower to switch translations, but that's actually fine).

Do you know if the LCF data structure can be modified at runtime? I.e., variables are not "const", and nothing is cached that could interfere with changing the data at runtime?

If a hook is added to iterate over all strings, it would make sense to have it include the "contexts" as well --and LcfTrans should probably use it so that both code paths are aligned. Ideally there would be a similar function to iterate + change all strings.

fmatthew5876:
> Hmm, I was thinking the performance hit wouldn't be that big, but then again for things like Nintendo 3DS and Android it might start to matter. 

It's not just performance, there are going to be endless whack-a-mole style bugs where we forget to translate a string or 2 here and there. 

The key here is that for something like this we have make it systematic so we solve the problem once and only once. Sprinkling `Tr()` everywhere is just not going to work.


> Translating the DB on the fly whenever the translation is selected makes sense, although we'd need to reload the original LCF if the translation was changed at runtime (so: slower to switch translations, but that's actually fine).

QuickSave method:

If you want to support hot reload of translations, then as discussed in irc, one way to go is if you change the translation we can do an automatic quicksave/quickload and reload all the game structures. Then the translation logic is localized to one place and always works. I think it's perfectly acceptable to have to wait for an automatic reload if you change the translation.


> Do you know if the LCF data structure can be modified at runtime? I.e., variables are not "const", and nothing is cached that could interfere with changing the data at runtime?

Right now they can, but later they will be behind a const facade #2285 . Still this could be modified to allow translation or if we do the quicksave/quickload approach it doesn't matter.

> If a hook is added to iterate over all strings, it would make sense to have it include the "contexts" as well --and LcfTrans should probably use it so that both code paths are aligned. Ideally there would be a similar function to iterate + change all strings.

The hook should be a very general `template <typename F> ForEachString(const F& f)` in liblcf. 

Like this one:
https://github.com/EasyRPG/Player/blob/master/src/utils.h#L367

Liblcf just provides a mechanism to call a lambda function on all string members, then it's up to player to supply the translate function into the lambda.

Ghabry:
@fmatthew5876 
You thought of something like this?

https://github.com/EasyRPG/liblcf/compare/master...Ghabry:foreachstring?expand=1

usage

```
lcf::Data::actors[0].ForEachString([](const std::string& val, const std::string& ctx) {
	return val + "TRANS";
});
```

(not PR'ing yet, do this after StringView and DBString)

fmatthew5876:
> @fmatthew5876
> You thought of something like this?
> 
> https://github.com/EasyRPG/liblcf/compare/master...Ghabry:foreachstring?expand=1
> 
> usage
> 
> ```
> lcf::Data::actors[0].ForEachString([](const std::string& val, const std::string& ctx) {
> 	return Translate(val, "actor." + ctx);
> });
> ```
> 
> (not PR'ing yet, do this after StringView and DBString)

yeah exactly, something like that. (use const F&, not F&&)

fdelapena:
[Jenkins]: ok to test

sorlok:
I removed all the old runtime changes (except message boxes), so most of the translation now happens in ForEachString(). Still a lot of things to fix, but I made a brief demo of Yume 2kki:
https://drive.google.com/file/d/18gn6QgDXShxdExXfWrT1kZ6wl-JJKjdH/view?usp=sharing

Unzip it so that "translations" appears in the game's Project directory, then you can select "en" from the menu. I copied the existing translation strings/pictures from wolfen's translation  --this is a tech demo, so please don't redistribute beyond this thread.

The cool thing about doing translations this way is that wolfen's translation was made for 0.112b, but you can run my link on 0.113e (or later) since it's just doing string substitution. I can definitely see the value of a system like this for projects that receive updates.

By the way, I think I will switch my demo game to something else. Yume 2kki is really interesting, but it doesn't have a lot of "standard" RPG Maker stuff (like combat effects, states, or some menu items), so it doesn't really stress the software. Suggestions are welcome (or maybe I'll just do Don Miguel's demo).

Ghabry:
Yeah I made the "-u" update switch mostly for Yume2kki. The updates are so fast that normal translators are always behind because it is so cumbersome to update this. The 0.112 to 0.113 is actually quite interesting for testing the merge algorithm:
One game dev of Yume2kki left and they asked to remove everything they contributed from the game. Really nice to test the update+stale feature :+1: 

-------------

Btw a tinygettext replacement: Here is the Po parser from lcftrans (used by the update feature so you can consider it tested and working now ;)). (I renamed the class to Dictionary because you already use Translation). Only needs: The starts_with converted to StringView startswith and maybe helpers to find entries. Otherwise is just "call fromPo" and then magic happens.

```cpp
class Entry {
public:
	std::string original; // msgid
	std::string translation; // msgstr
	std::string context; // msgctxt
};

class Dictionary
{
public:
	const std::vector<Entry>& getEntries() const;

	static Dictionary fromPO(std::istream& in);

private:
	std::vector<Entry> entries;
};

const std::vector<Entry>& Dictionary::getEntries() const {
	return entries;
}

Dictionary Dictionary::fromPO(std::istream& in) {
	// Super simple parser.
	// Only parses msgstr, msgid and msgctx

	Dictionary t;

	std::string line;
	bool found_header = false;
	bool parse_item = false;

	Entry e;

	auto extract_string = [&line](int offset) {
		std::stringstream out;
		bool slash = false;
		bool first_quote = false;

		for (char c : line.substr(offset)) {
			if (c == ' ' && !first_quote) {
				continue;
			} else if (c == '"' && !first_quote) {
				first_quote = true;
				continue;
			}

			if (!slash && c == '\\') {
				slash = true;
			} else if (slash) {
				slash = false;
				switch (c) {
					case '\\':
						out << c;
						break;
					case 'n':
						out << '\n';
						break;
					case '"':
						out << '"';
						break;
					default:
						std::cerr << "Parse error " << line << " (" << c << ")\n";
						break;
				}
			} else {
				// no-slash
				if (c == '"') {
					// done
					return out.str();
				}
				out << c;
			}
		}

		std::cerr << "Parse error: Unterminated line" << line << "\n";
		return out.str();
	};

	auto read_msgstr = [&]() {
		// Parse multiply lines until empty line or comment
		e.translation = extract_string(6);

		while (std::getline(in, line, '\n')) {
			if (line.empty() || starts_with(line, "#")) {
				break;
			}
			e.translation += extract_string(0);
		}

		parse_item = false;
		t.addEntry(e);
	};

	auto read_msgid = [&]() {
		// Parse multiply lines until empty line or msgstr is encountered
		e.original = extract_string(5);

		while (std::getline(in, line, '\n')) {
			if (line.empty() || starts_with(line, "msgstr")) {
				read_msgstr();
				return;
			}
			e.original += extract_string(0);
		}
	};

	while (std::getline(in, line, '\n')) {
		if (!found_header) {
			if (starts_with(line, "msgstr")) {
				found_header = true;
			}
			continue;
		}

		if (!parse_item) {
			if (starts_with(line, "msgctxt")) {
				e.context = extract_string(7);

				parse_item = true;
			} else if (starts_with(line, "msgid")) {
				parse_item = true;
				read_msgid();
			}
		} else {
			if (starts_with(line, "msgid")) {
				read_msgid();
			} else if (starts_with(line, "msgstr")) {
				read_msgstr();
			}
		}
	}

	return t;
}
```

sorlok:
Awesome, thanks! I'll integrate the PO reader.

Ghabry:
Jenkins: Test this please

Ghabry:
You can't use dynamic cast. It requires rtti which results in binary bloat (and I forgot to disable rtti+exceptions in the web player. Huge waste of space).
You can use ``Game_Battle::IsBattleRunning``.
Though there is currently no good way to tell if an event is a common event. I'm thinking about a solution (maybe custom lcf chunks that contain the event type)

Ghabry:
From an esthethic POV I like that the Language selection is part of the title scene and just spawns a new command box :).
Also nice fade effect after language selection (guess this is just the Title Scene::Pop/Push ^^')

Works nice in Yume2kki!

Ghabry:
I recorded a video here: Nice job!

[yume2kki-translation-0.6.3.zip](https://github.com/EasyRPG/Player/files/5218726/yume2kki-translation-0.6.3.zip)

@sorlok 
One problem I noticed is that it won't work when the language directory is not full lower case (English doesnt work, english does) but I guess there will be some Ini file later to solve this (and use this string instead).

I general imo the folder should be "Languages". In other user interfaces you usually see the word "Languages", not "Translations". (same for the folder -> "languages")

E.g. for each language a "Meta.ini" file:

```csv
[Language]
Name=English
Code=en
Description=Shiny translation by Team TheEnglishSpeakazzzz <- Could be shown as a one-line help window at the top
```

sorlok:
> You can't use dynamic cast. It requires rtti which results in binary bloat (and I forgot to disable rtti+exceptions in the web player. Huge waste of space).
> You can use `Game_Battle::IsBattleRunning`.
> Though there is currently no good way to tell if an event is a common event. I'm thinking about a solution (maybe custom lcf chunks that contain the event type)

Removed the dynamic cast. Right now I just scan *both* maps and common events (since there's no good way to tell). That will work fine for now.

sorlok:
Languages should load correctly now regardless of case. Also, the directory was renamed to "Languages", and the Meta.ini file works as described. There is also Help text shown.

Ghabry:
You can't use exceptions. They are disabled on the same performs as rtti ;).
Also this is bad style. You never throw an exception and you do a "catch all". What do you want to catch there? Fix the code bug instead of hiding it.

fdelapena:
> From an esthethic POV I like that the Language selection is part of the title scene and just spawns a new command box :).
> Also nice fade effect after language selection (guess this is just the Title Scene::Pop/Push ^^')
> 
> Works nice in Yume2kki!

Just as a note, some games ( https://github.com/EasyRPG/Player/pull/2035#issuecomment-624349707 ) will require additional tweaks to integrate additional options.

A possible approach to work around this without modifying the original game data:

From the built-in system in Player, get the blinking arrow sprites used for scrollable window selectables and add them when there are more than three options (e.g. Translations, Options), then when the selectable goes further the quit, display the built-in system window over the original with the 3 original selectors (with some built-in names) + the additional ones.

elsemieni:
I was testing this a bit and I figured when you apply a translation, the file mapping does not cover any audio resources (only graphical ones). Until now I don't know if this is a bug (well, expectable as this is a draft) or a design choice. If it's the last option, I suggest file mapping must apply for all resources, not just audio ones (practical use case: Audio dubs). 

Anyways, awesome work @sorlok ! <3

sorlok:
Thanks @elsemieni, that's a good point. I'll add overloads to the audio loader as well.

sorlok:
@fdelapena That is a good point, but I think I'm going to call that out of scope for the initial implementation. I also think we'll need some discussion on the appropriate solution in those cases (ideally we wouldn't have too much code for dealing with edge cases).

sorlok:
So inspired by another thread, I reworked the Message translation code to actually translate the lcf::rpg::Commands themselves (when switching to that map), instead of translating the Message string when the box is shown. This is a little more complicated, but it solves a key issue: we can now easily distinguish between Maps, Common Events, and Battle. 

Choice boxes also work now. In addition, adding or deleting Message Boxes should be pretty simple going forward.

One big question: in Game_Interpreter::GetChoices(), I see something like this:
```
for (int index_temp = index + 1; index_temp < static_cast<int>(list.size()); ++index_temp) {
  if (static_cast<Cmd>(com.code) == Cmd::ShowChoiceOption && com.parameters.size() > 0 && com.parameters[0] < max_num_choices) {
    // Choice found
    s_choices.push_back(ToString(list[index_temp].string));
  }
}
```

Can I assume that each Cmd::ShowChoiceOptions contains exactly one line of text in string? If that's true, then why is `com.parameters[0]` checked in the loop? Shouldn't that always be 1? (This is a fundamental question on the RPG data structures --it looks like GetChoices() is overly complicated, but perhaps it's implementing some important, non-obvious RPG Maker behavior?)

Ghabry:
``com.parameters[0]`` is the number of the choice. It must be validated because of games with corrupted choice handlers.
See https://github.com/EasyRPG/Player/issues/2048

sorlok:
Interesting, thanks! I had a hunch it was going to end up being something like this. I'll modify my code to accommodate.

Ghabry:
Imo this could be simplified alot by changing the way how ``PendingMessage`` works.
Imo pending message is a good hook point for this.

The current problem with pending message is that it directly substitutes \V etc. When this substitution would be delayed until the entire message is pushed then this can be used. (Or add a new string to pending message that contains the unprocessed message)

sorlok:
I was going to suggest those changes to PendingMessage, but I think it might be better for the Translation code to just iterate over the data structure directly, making changes on map load.

sorlok:
So I added some special text seperators:
```
__EASY_RPG_CMD:REMOVE_MSGBOX__
__EASY_RPG_CMD:ADD_MSGBOX__
```
These can be used to add or remove message boxes, and it works quite well.

I've run into a problem overlaying SE and Music resources. Unlike images, these use the async_loader, which first checks if the resource is cached and then sets up a load request otherwise. However, in the Translated case, I don't know how to tell if I should request the Lanuages/English/SE/* file or the SE/* version. 

What is the right way to do this? Should I try to check (in AsyncHandler::RequestFile) if Languages/English/SE/* exists, and then request the SE/* version otherwise? Or does the async_loader end up using some other function that is easier to hook? (I get this impression because the SE loader still respects the RTP.) 

Thanks.

Ghabry:
Ignore the async handler for now.
This will be hard to do right currently because "Exists" doesn't work without fetching the file.
I will try to solve this later.

sorlok:
Ok, thanks. I'll polish up what I have and then ask for a merge.

sorlok:
This is ready for merging, as far as I'm concerned. Let me know what changes need to be made --there's a lot in here, but I tried to make it as simple as possible.

Ghabry:
Great! We have lots of open PRs and this one is big so I would say we prioritize this one here in the merge order lower. But will review and test it :)
Really great work!

sorlok:
Example export of RPG Advocate's Aurora's Tear translation is now "done":
https://drive.google.com/file/d/1qUSQwWjoPLr7dKYBpvqDdqyMhQqR5eXW/view?usp=sharing

...however, note that this is still not a 1.0 translation, since the way Choice text is stored will change in LcfTrans (to be independent of the Message text). Once that happens, I'll update this.

fdelapena (approved):

sorlok:
Aurora's Tear translation is now at 1.0:
https://drive.google.com/file/d/15uqWzc125y0J8XldIv7e60XhKSF6Qapi/view?usp=sharing

Keep in mind that RPG Advocate's translation was not 100% to begin with --this simply copies over everything they worked on.

I need to rebase this branch and do a minor code cleanup, then it will be ready to merge.

sorlok:
Hi all, I am currently on a bit of a vacation, so updates on this will be slow. I will try to do a final pass sometime in the next few weeks.

sorlok:
Ok, I managed to get some time to look at this. I've rebased it and recompiled, and it all works as expected. Note that @fmatthew5876's outstanding request was dealt with some time ago (we modify the LDB/Maps at load instead of when strings are requested). So this Pull Request should be good to go!

sorlok:
@fmatthew5876 , @Ghabry (just bumping this up in case you missed it).

Ghabry:
sorry, I'm kinda overloaded with work since a month, so I can't really focus on Player right now... So had no time to check this yet again :/

sorlok:
No worries, I'll check again in a few weeks. Best of luck!

src/scene_title.h
```diff
@@ -103,17 +119,59 @@ class Scene_Title : public Scene {
 private:
 	void OnTitleSpriteReady(FileRequestResult* result);
 
+	/**
+	 * Moves a window (typically the New/Continue/Quit menu) to the middle or bottom-center of the screen.
+	 * @param centerVertical If true, the menu will be centered vertically. Otherwise, it will be at the bottom of the screen.
```
Ghabry, line 124: lacks ``@window`` documentation.
Also in general your naming convention again: ``centerVertical -> center_vertical``

src/scene_title.h
```diff
@@ -103,17 +119,59 @@ class Scene_Title : public Scene {
...
+	 */
+	void RepositionWindow(Window_Command& window, bool centerVertical);
+
+	/**
+	 * Picks a new language based and switches to it.
+	 * @param langStr If the empty string, switches the game to 'No Translation'. Otherwise, switch to that translation by name.
+	 */
+	void ChangeLanguage(const std::string& langStr);
```
Ghabry, line 132: lang_str

src/scene_title.h
```diff
@@ -103,17 +119,59 @@ class Scene_Title : public Scene {
...
+	 *   Listed in the order they may appear; exit_index will always be last,
+	 *   and import appears before translate, if it exists.
+	 * Stored in a struct for easy resetting, as Scene_Title can be reused.
+	 */
+	struct CommandIndices {
+		int new_game       =  0;
+		int continue_game  =  1;
+		int import         = -1;
+		int translate      = -1;
+		int exit           =  2;
```
Ghabry, line 172: don't align the "=", just move them to the left next to the var name.

src/translation.cpp
```diff
@@ -0,0 +1,760 @@
...
+// Name of expected files
+#define TRFILE_RPG_RT_LDB    "rpg_rt.ldb.po"
+#define TRFILE_RPG_RT_COMMON "rpg_rt.ldb.common.po"
+#define TRFILE_RPG_RT_BATTLE "rpg_rt.ldb.battle.po"
+#define TRFILE_RPG_RT_LMT    "rpg_rt.lmt.po"
+#define TRFILE_META_INI      "META.INI"
+
+// Message box commands to remove a message box or add one in place.
+#define TRCUST_REMOVEMSG        "__EASY_RPG_CMD:REMOVE_MSGBOX__"
+#define TRCUST_ADDMSG           "__EASY_RPG_CMD:ADD_MSGBOX__"
```
Ghabry, line 49: Applies to remove and add: Is this still used? I see them checked in ifs but never assigned?

src/translation.cpp
```diff
@@ -0,0 +1,760 @@
...
+
+
+std::string Tr::TranslationDir() {
```
Ghabry, line 52: Is a getter, call it ``GetTranslationDir``

src/translation.cpp
```diff
@@ -0,0 +1,760 @@
...
+	return Player::translation.RootDir();
+}
+
+std::string Tr::CurrTranslationId() {
```
Ghabry, line 56: Is a getter, call it ``GetCurrentTranslationDir``.
In general for all funcs: can you change from ``Curr`` to ``Current``?

src/translation.cpp
```diff
@@ -0,0 +1,760 @@
...
+	/**
+	 * Helper class for iterating over and rewriting event commands.
+	 * Starts at index 0.
+	 */
+	class CommandIterator {
+	public:
+		CommandIterator(std::vector<lcf::rpg::EventCommand>& commands) : commands(commands) {}
+
+		/// Returns true if the index is past the end of the command list
+		bool Done() const {
```
Ghabry, line 277: Wrong doc comments, should be ```/** */```

src/translation.h
```diff
@@ -0,0 +1,283 @@
...
+}
+
+
+/**
+ * Properties of a language
+ */
+struct Language {
+	std::string langDir;  // Language directory (e.g., "en", "English_Localization")
+	std::string langName; // Display name for this language (e.g., "English")
+	std::string langDesc; // Helper text to show when the menu is highlighted
```
Ghabry, line 137: should be ``lang_dir, lang_name, lang_desc``

src/translation.h
```diff
@@ -0,0 +1,283 @@
...
+	 */
+	void ParsePoFile(const std::string& path, Dictionary& out);
+
+	/**
+	 * Parse all .po files for the given language.
+	 * 
+	 * @param langId The ID of the language to parse, or "" for Default (no parsing is done)
+	 * @return True if the language directory was found; false otherwise
+	 */
+	bool ParseLanguageFiles(const std::string& langId);
```
Ghabry, line 221: ``lang_id``

src/translation.h
```diff
@@ -0,0 +1,283 @@
...
+	std::unordered_map<std::string, std::unique_ptr<Dictionary>> maps;  // map<id>.po, indexed by map name
+
+	// Our list of available Languages (translations, localizations), determined by scanning the files on disk.
+	std::vector<Language> languages;
+
+	// The "languages" directory, but with appropriate capitalization.
+	std::string translationRootDir;
+
+	// The translation we are currently showing (e.g., "English_1")
+	std::string currLanguage;
```
Ghabry, line 280: translation_root_dir, current_language

src/translation.cpp
```diff
@@ -0,0 +1,760 @@
...
+
+	return res;
+}
+
+
+void Translation::RewriteEventCommandMessage(const Dictionary& dict, std::vector<lcf::rpg::EventCommand>& commandsOrig) {
+	// A note on this function: it is (I feel) necessarily complicated, given what it's actually doing.
+	// I've tried to abstract most of this complexity away behind an "iterator" interface, so that we do not
+	// have to track the current index directly in this function.
+	CommandIterator commands(commandsOrig);
```
Ghabry, line 505: Marking this as "Reviewing later", can you explain me first how this TRCUST_REMOVEMSG stuff works? Considering that this is never pushed anywhere I'm confused.

Ghabry (changes requested):
Well for me this all works excellent. I want to look at ``RewriteEventCommandMessage`` again later. This function is really complecated, maybe can be simplified ^^

Please make the requested changes in a _new commit_

src/translation.cpp
sorlok: These strings are supposed to be added by the translator. Here's how it works; let's say I have a message box:

```
msgid ""
"Hero: Hello there!"
```

...and I want that message box to be removed from the translation entirely. In that case, I replace it with:

```
msgstr ""
"__EASY_RPG_CMD:REMOVE_MSGBOX__"
```

Let's say, however, that in the target language it makes more sense for the other party to shout at the hero first. So I need a new message box to appear BEFORE the current one. I'll replace it with:

```
msgstr ""
"Townsperson: Mr. Hero, hello!"
"__EASY_RPG_CMD:ADD_MSGBOX__"
"Hero: Hello there!"
```

This will ADD a message box to the event list, so now we'll have two message boxes (rather than a single two-line message box).

They are magic strings, unfortunately, but I think they're pretty unlikely to appear in source games. (And I can't think of a more robust way to do this that interacts cleanly with gettext).

src/translation.cpp
sorlok: See the earlier comment about the translator adding the strings into the translated text. 

Given that, the basic idea is to take a CommandIterator on the command stream, and then .insert() and .remove() as needed to get the requested translation. It's fairly generic around MessageBoxes, and a little bit stricter around Choices. I think the comments do a reasonable job of explaining what's going on, but let me know if there are some areas that are not clear and I'll tidy up the docs.

I did a pass to simplify the function (a month ago); it's possible it can be simplified a little further, or we could encapsulate some of these activities (e.g., "trim down to 4 lines") into their own tiny functions, if that helps.

src/translation.cpp
Ghabry: ah now I get it so that's some kind of in-band signaling. Will see if I can find a way to improve this.

For "insert new page" we already have the the form feed "\f" (needed for some battle messages) but I guess this will need changes to the po parser.

For "delete message" one could use gettext flags. They are marked via ``#,``, something like ``#, easyrpg-delete`` (see https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html). Will also need parser changes (to ensure lcftrans keeps them)

src/scene_title.h
sorlok: Fixed missing documentation.
Fixed all the camel case names I could find.

src/scene_title.h
sorlok: Fixed

src/scene_title.h
sorlok: Fixed

src/translation.cpp
sorlok: Fixed

src/translation.cpp
sorlok: Fixed "Get".
Fixed all usages of "Curr".

src/translation.cpp
sorlok: Fixed all usages of ///

src/translation.h
sorlok: Fixed

src/translation.h
sorlok: Fixed

src/translation.h
sorlok: Fixed

sorlok:
Thanks for the feedback; all items were fixed (and I added two major comments describing how add/remove work).

I also rebased, and loading maps now crashes on Linux (see below), but since it does this for new games with translations turned off, I think it is probably not related to my minor code cleanup:

```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff503e8b1 in __GI_abort () at abort.c:79
#2  0x00007ffff5693957 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff5699ae6 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff5699b21 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00005555556d022c in nonstd::span_lite::detail::report_contract_violation(char const*) ()
#6  0x00005555556d05da in nonstd::span_lite::span<unsigned char const, 18446744073709551615ul>::operator[](unsigned long) const ()
#7  0x00005555556cea61 in TilemapLayer::CreateTileCache(std::vector<short, std::allocator<short> > const&) ()
#8  0x00005555556cfeaf in TilemapLayer::SetPassable(std::vector<unsigned char, std::allocator<unsigned char> >) ()
#9  0x00005555556a71e1 in Tilemap::SetPassableDown(std::vector<unsigned char, std::allocator<unsigned char> >) ()
#10 0x00005555556a6d71 in Spriteset_Map::OnTilemapSpriteReady(FileRequestResult*) ()

```

src/translation.cpp
sorlok: Yep, that's right. 

I think I *might* prefer to keep the current strings, since they are very self-descriptive for translators, and I don't think that "\f" or gettext flags will clean up our own processing code. But this is certainly up for debate.

Ghabry: at least \f will not easily work because our code filters away control characters... Have to sleep about this ^^'

Ghabry:
try updating to the latest liblcf version. Here it works.

sorlok:
> try updating to the latest liblcf version. Here it works.

Yep, that fixed it, thanks. (I had to do a clean rebuild of both sources.)

sorlok:
I've been thinking about `__EASY_RPG_CMD:ADD_MSGBOX__` and `__EASY_RPG_CMD:REMOVE_MSGBOX__` these past few days, and I think I prefer to keep them as strings. Reasoning:
 * Translators may not be coders; if they see "\f", it will be hard for them to find a good search result on Google.
 * They're somewhat self-documenting: if we add other "EASY_RPG_CMD's" in the future, it will be simple to search for them.
   * I'm not planning this, but adding EASY_RPG_CMD's for inserting **all** types of Event Commands would allow translators greater freedom when dealing with custom menus/combat and RTL or top-to-bottom languages.
 * I don't like having two different mechanisms for add (\f) and remove (gettext flags)
 * I like keeping our .po parser simple (no need for gettext flags)
 * I like keeping tools simple (a .po editor might treat \f differently because it's an escape character)

The final decision is of course up to EasyRPG devs.

Ghabry:
Maybe make the commands less clunky by using something like ``<easyrpg:new_page>`` and ``<easyrpg:delete_page>``. Then it looks a bit like XML :D.
But in general I'm by now fine with this solution.

sorlok:
That sounds good; I'll change the commands to XML-like later today.

sorlok:
Changes made; we now use `<easyrpg:new_page>`and `<easyrpg:delete_page>`.

The sample translation of Aurora's Tear has also been updated to match:
https://drive.google.com/file/d/1NjsOs96j-nhkyooE7YQL3t9qzR67iKJs/view?usp=sharing

As far as I'm concerned, this feature is ready.

Ghabry (approved):
Except for Scene_Title this is minimally invasive as all rewrites are on Database/Map load.

``<easyrpg:delete_page>`` and ``<easyrpg:new_page>`` also work as intended and help alot when the target language has a much shorter (delete unneeded messages) or a much longer (add more pages to handle e.g. a "Actor name" on Line 1) translation.

PR #2288, opened 2020-08-20
Provide access to raw key states and mouse position in Input class.

Ghabry:
This change exposes the raw keys of the DisplayUi to InputSource and Input can query now the Released/Pressed/Triggered state of raw keys. They are not remappable as the goal is here to read the keyboard directly.

This is required by Inelukis Key Patch (Requires Triggered and Released) and ManiacPatch (Required Pressed), I'm just exposing the API already.

Mouse is required by both. The ManiacPatch also has a function to set the Mouse offset but this sounds more like a workaround for Direct3D issues, I ignore this use-case for now.

Performance: The change adds two copies for input keys and each frame a loop to update the raw state. So this is minimal slower than before.

Compatibility: Such games will only run on systems with keyboards. Not solvable right now - A warning could be displayed. Except for Android our other systems are kinda niche and the Plugins are also niche. So "niche + niche = don't care" ;)


src/input.cpp
```diff
@@ -147,6 +151,14 @@ void Input::Update() {
...
 		else if (dirpress[Direction::DOWNLEFT] > 0)	dir8 = Direction::DOWNLEFT;
 	}
+
+	// Determine pressed & released keys from raw keystate
+	auto& raw_pressed_now = source->GetPressedKeys();
+	for (unsigned i = 0; i < Input::Keys::KEYS_COUNT; ++i) {
+		raw_triggered[i] = raw_pressed_now[i] && !raw_pressed[i];
+		raw_released[i] = !raw_pressed_now[i] && raw_pressed[i];
+	}
+	raw_pressed = source->GetPressedKeys();
```
fmatthew5876, line 161: `raw_pressed = raw_pressed_now`

src/input.cpp
```diff
@@ -147,6 +151,14 @@ void Input::Update() {
 		else if (dirpress[Direction::DOWNRIGHT] > 0)	dir8 = Direction::DOWNRIGHT;
 		else if (dirpress[Direction::DOWNLEFT] > 0)	dir8 = Direction::DOWNLEFT;
 	}
+
+	// Determine pressed & released keys from raw keystate
+	auto& raw_pressed_now = source->GetPressedKeys();
```
fmatthew5876, line 156: `const auto`

src/input_source.h
```diff
@@ -68,13 +73,22 @@ namespace Input {
 
 		bool InitRecording(const std::string& record_to_path);
 
+		void GetMousePosition(int& x, int& y) const {
```
fmatthew5876, line 76: We have  `Point` struct now which contains x and y, if you don't want to use out params here.

fmatthew5876:
Nice, this will also be useful in `Scene_Settings` to be able to do input remapping.

src/input_source.h
Ghabry: Valid "Point". I changed this mouse_x/y to Point everywhere.

Ghabry:
adressed the comments.
Great that there is a use-case outside of patch support :)

src/input.h
```diff
@@ -197,6 +199,41 @@ namespace Input {
...
+	 */
+	bool IsRawKeyReleased(Input::Keys::InputKey key);
+
+	/**
+	 * Obtains the position of the mouse cursor relative to the screen
+	 *
+	 * @param x Cursor x position
+	 * @param y Cursor y position
+	 */
+	Point GetMousePosition();
```
fdelapena, line 235: Doxygen params mismatch.

Issue #2284, opened 2020-08-19
Key Input Bug in Heaven and Hell Episode 3


fmatthew5876:
Start a new game, go up and enter the building. In the large hallway, go to the right and talk to the white door.

While the message box is open, press the [Escape] key. Then finish the dialogue.

RPG_RT: Message window closes, then nothing happens
Player: Message window closes, then custom menu system immediately called.

This is an interpreter bug, it's allowing parallel events to register key presses for KeyInputProc while messages are active.

Related Issue to retest upon fixing this: #1691

PR #2290, opened 2020-08-21
Block key input processing while message is active

fmatthew5876:
Fix: #2284

Confirmed behavior in RE tools.

I would consider this for backporting to stable.


Ghabry (approved):
That's a good test for the regression test.

PR #2295, opened 2020-08-22
Spell reflect fix

### LINKED ISSUE #2251: Spell reflect issue
#### Name of the game:
Vampires Dawn II

#### Player platform:
Linux

When you cast a spell which hits the entire enemy party (e.g. "Apokalypse" or "Starke Flut") and it gets reflected because one or more of the enemies have the "Spiegel" state (state which reflects spells) the spell hits only the caster. In RPG_RT the spell is returned to the entire party of the caster.

rueter37:
This PR fixes #2251. Moreover a second commit has been added to check if an item has been used because skills invoked by items bypass spell reflection (compare with the Item "Heiliger Sand" in "Vampires Dawn II". Using that item removes all positive states on the enemies (including "Spiegel") and it doesn't get reflected in RPG_RT).

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -757,7 +757,6 @@ int Game_BattleAlgorithm::AlgorithmBase::GetSourceAnimationState() const {
 void Game_BattleAlgorithm::AlgorithmBase::TargetFirst() {
 	if (party_target) {
 		if (reflect_set) {
-			PlayAnimation();
 			targets.clear();
 			source->GetParty().GetActiveBattlers(targets);
 		} else {
@@ -766,7 +765,6 @@ void Game_BattleAlgorithm::AlgorithmBase::TargetFirst() {
 		party_target = nullptr;
 	} else {
 		if (reflect_set) {
-			PlayAnimation();
 			targets.clear();
 			targets.push_back(source);
 		}
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -147,6 +147,7 @@ Game_BattleAlgorithm::AlgorithmBase::AlgorithmBase(Type ty, Game_Battler* source
 
 	current_target = targets.end();
 	party_target = target;
+	party_target_set = true;
 }
 
 void Game_BattleAlgorithm::AlgorithmBase::Reset() {
@@ -236,7 +237,13 @@ void Game_BattleAlgorithm::AlgorithmBase::PlayAnimation(bool on_source) {
 	}
 
 	if (on_source) {
-		Game_Battle::ShowBattleAnimation(GetAnimation()->ID, { GetSource() });
+		if (!party_target_set) {
+			Game_Battle::ShowBattleAnimation(GetAnimation()->ID, { GetSource() });
+		} else {
+			std::vector<Game_Battler*> anim_targets;
+			source->GetParty().GetActiveBattlers(anim_targets);
+			Game_Battle::ShowBattleAnimation(GetAnimation()->ID, anim_targets);
+		}
 		has_animation_played = true;
 		return;
 	}
@@ -607,7 +614,13 @@ Game_Battler* Game_BattleAlgorithm::AlgorithmBase::GetSource() const {
 
 Game_Battler* Game_BattleAlgorithm::AlgorithmBase::GetTarget() const {
 	if (IsReflected()) {
-		return source;
+		if (!party_target_set) {
+			return source;
+		} else {
+			std::vector<Game_Battler*> newtargets;
+			source->GetParty().GetActiveBattlers(newtargets);
+			return *(newtargets.begin());
+		}
 	}
 
 	if (current_target == targets.end()) {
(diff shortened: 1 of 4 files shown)


rueter37:
I have added a new commit which tries to address the issues you mentioned. But I'm not sure if the changes meet your requirements, moreover it is possible that I did something horribly wrong. The tests with the new commit were successful with one exception: the battle animation is only played once - at the new targets. This is supposed to be the second animation, the first is unfortunately missing.

src/game_battlealgorithm.cpp
```diff
@@ -1176,6 +1156,32 @@ void Game_BattleAlgorithm::Skill::Init() {
 			Output::Warning("Algorithm Skill: Invalid skill animation ID {}", skill.animation_id);
 		}
 	}
+
+	// Skill reflect checks
+	// Skills invoked by items ignore reflect
+	if (!item) {
```
fmatthew5876, line 1162: Ahh my bad. There is an edge case I forgot.

Suppose there are 4 enemies and one of them has reflect, and you decide to cast a spell on the enemy party. Now suppose the enemy with reflect is killed before the spell starts, in that case we should not reflect the spell.

TL;DR Unfortunately I think this means this logic needs to move to `TargetFirst()`. You can make TargetFirst call a virtual function which skill overloads.

src/game_battlealgorithm.cpp
```diff
@@ -237,13 +236,7 @@ void Game_BattleAlgorithm::AlgorithmBase::PlayAnimation(bool on_source) {
 	}
 
 	if (on_source) {
```
fmatthew5876: This `on_source` stuff was part of the old reflect logic, lets get rid of it entirely. Now that you're actually changing the target of the algo back to the source in `TargetFirst()` this hack is no longer needed.

src/game_battlealgorithm.cpp
```diff
@@ -773,8 +756,20 @@ int Game_BattleAlgorithm::AlgorithmBase::GetSourceAnimationState() const {
 
 void Game_BattleAlgorithm::AlgorithmBase::TargetFirst() {
 	if (party_target) {
-		party_target->GetBattlers(targets);
+		if (reflect_set) {
+			PlayAnimation();
```
fmatthew5876, line 760: We should not be playing animations here. This function only changes the targets and returns the first one.

fmatthew5876:
> I have added a new commit which tries to address the issues you mentioned. But I'm not sure if the changes meet your requirements, moreover it is possible that I did something horribly wrong. The tests with the new commit were successful with one exception: the battle animation is only played once - at the new targets. This is supposed to be the second animation, the first is unfortunately missing.

Ohh.. I see what you mean. I also missed that part. RPG_RT will animate on the original targets, and then animate again on the new targets.

I still think it's better that the algorithm targets are changed, so that all the effects naturally apply to the battlers who are actually affected by the spell and we don't have this `IsReflected()` noise everywhere in the code.

What we need then is to flip the animation logic.

The current scene battle rpg2k3 does "animate on original targets using normal logic" "animate on new targets using special logic".

We should reverse this. The "animating on original targets" should be the special case which happens first, Then the normal logic flows.

So probably in `TargetFirst()` you need to save the original targets, and then change `PlayAnimation(bool on_source)` to `PlayAnimation(on_original_targets)` and then have that one done *before* the normal animation logic in the battle scene.

I hope that made sense?

rueter37:
Oops, that PlayAnimation() was a dumb oversight. I'll remove it in the next commit.

fdelapena:
Jenkins CI: ok to test

rueter37:
I have added a commit which re-add the IsReflected() method as the virtual method you mentioned which contains the reflect check again but this method gets now used in TargetFirst() only so that the IsReflected() noise is eliminated. And the second commit fixes the battle animation issue.

fmatthew5876:
This looks great!

The last thing I would do is tidy up the code a little bit.

`TargetFirst()` should first do the old logic to resolve the targets, call is reflected, and then adjust the targets like this:

````
if (party_target) {
    party_target->GetBattlers(targets);
}

If (IsReflected()) {
  original_targets = std::move(targets);
  if (party_target) {
    source->GetParty().GetBattlers(targets);
  } else {
    targets = { source };
  }
}
party_target = nullptr;

// Rest of the code
```

And then you can greatly simplify `IsReflected()` to just loop over the targets. Since this is only called once, I don't think we need that `reflect` variable either?

```
if (!item) {
for (auto* target: targets) {
   if(GetSource()->GetType() != target->GetType() && GetTarget()->HasReflectState()) {
    return true;
   }
}
}
return false;
}
```

Also once this is done, can you rebase and smash all your commits into 1? That will help us review a single diff.

fmatthew5876:
Thanks for the contribution! This looks like it works. Will test it on my side to double check and if so then I think it's good to go.

Ghabry:
Thanks for your contribution!

If you want please consider adding yourself to the AUTHORS file.

Also we plan to relicense the Player later so I have to ask you: _Do you agree on relicensing the Player to a more liberal license_? (GPLv3 is not suitable for selling games because of NDA clauses in Steamworks, xbox, playstation and Switch)

rueter37:
I fully agree on the relicensing.

fmatthew5876:
I tested this and it works!

Ghabry:
Jenkins: test this please

Ghabry (approved):

Ghabry:
Thanks! This one and your other PR were non-invasive enough for a "stable backport".

Our stable branch contains hand-picked fixes / small enhancements for a faster release cycle while there is major progress on the master branch. (the next stable release will be in one or two weeks).

Issue #2297, opened 2020-08-23
Preparations for RPG_RT - Player unit testing


Ghabry:
With PRs such as #2208 where normal code reviews fail we finally need an automatic regression tester.

@fmatthew5876 I could need your help here. I ask you to provide two things. It would help me when you can contribute them (best would be as a seperate PR, so I can use it in both master and your character PR)

My idea is the following:

1. Add a "unit test" mode to Player which saves the game each frame
2. Replay the input in RPG_RT through injected DLL while saving the game each frame (for synchronization the Game_System.frame_counter is used)
3. Compare the save files
4. .....
5. PROFIT

----------

**TODO**

- [ ] Invent a better input recording format [1] @Ghabry
- [ ] Add a new SavegameManager class that provides a Save(int save_id) function. @fmatthew5876 can you help with this? A Load function isn't needed yet but I need a globally accesible Save. [2]
- [ ] Write a unit test mode that saves the game each frame (easy to add when Save() exists)
- [ ] Write a DynRpg plugin that can replay input and save each frame @Ghabry
- [ ] Write an automatic difftool (python based?), could parse lcf2xml output and skip chunks that usually don't match (see [2])
- [ ] Record some tests

**Known issues**

- [ ] DynRPG only writes up to Save99.lsd, needs some trickery (file renaming) to record longer
- [ ] The recording desyncs when there is a teleport. Is off by 5 frames?!

[1] My suggestion: https://github.com/EasyRPG/Player/commit/6531fd8cc4c08d1b53af0a33c7e343dea5e2da54 The important lines are the "F" lines, the format is ``game_system_frame,keys....``. A frame can occur multiple times because Game_System frames are reset in some cases, but this is no problem.

```
H EasyRPG Player Recording
V 2
D 2020-08-23-23-27-37
F 45,DECISION
F 0,DECISION
F 5,RIGHT
F 6,RIGHT
[...] <- A line is only written when there was an input in this frame
F 28,RIGHT,UP
F 29,RIGHT,UP
F 30,RIGHT,UP
```

[2]

_Notes about savegame incompatibility_

- ``title.timestamp`` - this varies (obviously)
- ``system.scene`` is 0 on the map
- ``easyrpg.version`` - only we write this

_Strange observations_

- ``system.save_count`` when using dynrpg save this is always 0
- ``system.save_slot`` when using dynrpg save this is always 1

_Incompatible save storage_

[...]

PR #2298, opened 2020-08-24
Save game compatibility fixes for regtesting

fmatthew5876:
For #2297

### DIFF
--- a/src/scene.cpp
+++ b/src/scene.cpp
@@ -73,6 +73,45 @@ enum PushPopOperation {
 
 int Scene::push_pop_operation = 0;
 
+lcf::rpg::SaveSystem::Scene Scene::rpgRtSceneFromSceneType(SceneType t) {
+	switch (t) {
+		case Null:
+		case GameBrowser:
+		case SceneMax:
+		case Logo:
+			break;
+		case Title:
+			return lcf::rpg::SaveSystem::Scene_title;
+		case Map:
+			return lcf::rpg::SaveSystem::Scene_map;
+		case Menu:
+		case Item:
+		case Skill:
+		case Equip:
+		case ActorTarget:
+		case Status:
+		case Teleport:
+		case Order:
+		case End:
+			return lcf::rpg::SaveSystem::Scene_menu;
+		case File:
+		case Save:
+		case Load:
+			return lcf::rpg::SaveSystem::Scene_file;
+		case Battle:
+			return lcf::rpg::SaveSystem::Scene_battle;
+		case Shop:
+			return lcf::rpg::SaveSystem::Scene_shop;
+		case Name:
+			return lcf::rpg::SaveSystem::Scene_name;
+		case Gameover:
+			return lcf::rpg::SaveSystem::Scene_game_over;
+		case Debug:
+			return lcf::rpg::SaveSystem::Scene_debug;
+	}
+	return lcf::rpg::SaveSystem::Scene(-1);
+}
+
 Scene::Scene() {
 	type = Scene::Null;
 }
--- a/src/scene.h
+++ b/src/scene.h
@@ -24,6 +24,7 @@
 #include "drawable_list.h"
 #include <vector>
 #include <functional>
+#include <lcf/rpg/savesystem.h>
 
 /**
  * Scene virtual class.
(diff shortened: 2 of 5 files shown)


fmatthew5876:
For the states thing I tested:

- [x] actor in starting party - array saved
- [x] actor added to party - array saved
- [x] actor never in party - array clear
- [x] actor failed to add because party full - array clear
- [x] state added to actor not in party - array saved
- [x] state which actor doesn't have, removed from actor not in party - array saved

Ghabry:
I observed now something strange: When I save manually I get this 0 0 0 0 0 0 0 0 thing but when I use ``game_system->save`` from DynRPG the states array is empty. So there must be some preprocessing here. Have to check the assembly.

Ghabry:
Figured it out: This is some obscure RPG_RT stuff: The "<status>" is empty for all units UNTIL you move or open a scene.

This is not worth emulating, will try to patch this. Thanks for the other fixes though!

fmatthew5876:
> The resize happens the first time a condition is read. In RPG_RT this happens when moving because of a condition damage step (poison) or when opening the Menu scene. I can inject a function call that "touches" the states or _do you want to emulate this_?!

I would really prefer not to, this is hacky and bad for performance. Also it means a simple getter function can no longer be const. This whole change is hacky, would it make sense just to touch all the states for all actors in DynRPG when starting up?

Ghabry:
Great. Now the only incompatibilities are: 

- ``title.timestamp``
- ``system.scene`` <- can patch this in dynrpg
- ``easyrpg.version``

Any good API idea for setting timestamp and version to 0?

fmatthew5876:
> * `system.scene` <- can patch this in dynrpg

Why don't we fix this Player, it kind of makes sense, especially for the maniac patch save feature.

When saving, we can look at the scene type at the top of the scene stack and map it back to the RPG_RT scene type.

> * `title.timestamp`
> * `easyrpg.version`

Timestamp needs a patch to liblcf to allow overriding the time in `LSD_Reader::PrepareSave`. Version could be added as an argument to the save function as it's already passed in.

Let's take a step back however. While we can patch these, what happens later when we start support RPG_RT patches with custom save chunks? Or easyrpg custom features?

Ghabry:
> Why don't we fix this Player, it kind of makes sense, especially for the maniac patch save feature.

Oh right I forgot the Maniac Patch. Yes, better fix this.

> Timestamp needs a patch to liblcf

Actually why don't we just skip the LSD PrepareSave call when doing the regression test recording.
I have to manually patch the save_count for DynRPG anyway so just skipping these 3 lines is actually better.

```
++save.system.save_count;
save.title.timestamp = LSD_Reader::GenerateTimestamp();
save.easyrpg_data.version = version;
```

> Let's take a step back however. While we can patch these, what happens later when we start support RPG_RT patches with custom save chunks? Or easyrpg custom features?

I'm not sure what you mean with custom RPG_RT save chunks. Nothing like this exists right now. Do you think about future patches?

For EasyRPG custom features... Well, one could record a proper input when the feature is invented and then use this recording for regression testing?

Ghabry:
If you want to get rid of the RPG_RT status compat commit: I added probing. Couldn't find this in the DynRPG header:

```
for (int i = 0; i < RPG::actors.count(); ++i) {
	RPG::Actor* actor = RPG::actors[i + 1];
	asm volatile("call *%%esi"
		:
		: "S" (0x4BFF8C), "a" (actor), "d" (1)
		: "ecx", "cc", "memory");
}
```

I have no idea why it must be "S", when I change this I get a runtime error o.O  But it works!

fmatthew5876:
added the requested items,

Removed the status stuff, but left it as a reverted commit unless we change our minds later.

Ghabry:
made a quick test. Should be fine this way.

Keep it as is for now. Will continue working on this at the weekend - is a complex problem.

Ghabry (approved):
I do all my regtests with these commits picked so they work. Pls just remove the state commit+revert.

The "Touching" of the states through a plugin works fine. And stuff like "subhook junk data" shows that patching RPG_RT data can't be fully avoided for our use case.

fmatthew5876:
Removed the states commits

PR #2302, opened 2020-08-25
MIDI: native midiout system

tyrone-sudeium:
## Todo

- [x] Readd checks whether native API was initialized
- [x] Ordering of Midi: FluidSynth, WildMidi, Native, FmMidi
- [x] Fix WildMidi
- [x] Resampler does not forward certain api calls 👎 

## What do we get?
* RPG_RT-like looping, fading, and tempo support with the native synthesizer.
* Also attempted to implement the "skip silence at start" behaviour from RPG_RT. I've tested it to work OK with RTP2000, and some MIDI tracks from some games I have on hand, such as Legion Saga III and my own private games.
* Lets us ditch SDL_mixer 2 and all of its problems. We mainly used it for its ability to utilise the OS MIDI synthesizer, this PR lets us do that without it.

### Duplicated code
There's a lot of very similar code between `GenericMidiOut` and `GenericMidiDecoder`. Both these features were developed in parallel, and I was rapidly changing the structure of my work at the same time that work on `GenericMidiDecoder` was changing. They do pretty different things but it should still be possible to factor out a common base class, or to have some kind of "MIDI Sequencer Controller" class that both of these utilise. It's also possible that most of the code in these could live directly in the MIDI sequencer itself.

### Off by default
Currently, this PR shouldn't change behaviour of a production build, which will still use SDL_mixer 2 where it can, thus disabling the midi out. If you want to try out this stuff, it automatically becomes active when you disable SDL_mixer 2, so just set `PLAYER_WITH_SDL2_MIXER` to `False` and it should become active, if you're on a supported OS. Speaking of supported OS...

### Mac and Windows only
I've written midiout device implementations for Windows and macOS. It should theoretically be possible for iOS, Linux, and libretro, but I haven't written these myself.

### Ticks
There's also quite a lot of duplicated code to support ticks. Honestly this is just my C++ skills letting me down. I also don't completely understand how RPG Maker's GetTicks works and don't have any games that use it. See the `MidiTempoData` struct and related nastiness in `audio_midiout.cpp`, which is suspiciously similar to stuff in `decoder_midigeneric.[h,cpp]`.


src/audio_sdl.cpp
```diff
@@ -99,9 +125,27 @@ SdlAudio::SdlAudio() :
 #else
 	SDL_PauseAudio(0);
 #endif
+
+	// Start midiout polling thread
+	// TODO: don't make a thread if we're on a platform that won't use MidiOut
+	midiout_mutex = SDL_CreateMutex();
+	midiout_thread = SDL_CreateThread(MidioutThreadMain, "MidioutThread", this);
```
fmatthew5876, line 132: Why not `std::thread` and `std::mutex`? Then we avoid sdl version issues and use standard library facilities.

tyrone-sudeium: I do love me some `std::thread` and `std::mutex` but I've been told it's less portable, some platforms don't have good implementations of it. My original thread implementation used it, but doing it this way keeps the threading inside SDL, though at the expense of a bit of ifdef hell

src/audio_generic.cpp
```diff
@@ -34,6 +34,10 @@ std::vector<float> GenericAudio::mixer_buffer;
 
 GenericAudio::GenericAudio() {
 	for (auto& BGM_Channel : BGM_Channels) {
+		if (BGM_Channel.midiout) {
+			BGM_Channel.midiout->Reset();
+		}
+		BGM_Channel.midiout.reset();
```
fmatthew5876, line 40: Can you call `Reset()` in the midiout destructor? Then we don't need this messy cleanup code.

src/audio_generic.cpp
```diff
@@ -62,31 +67,49 @@ void GenericAudio::BGM_Play(const std::string& file, int volume, int pitch, int
...
 
 void GenericAudio::BGM_Pause() {
+	LockMidiOutMutex();
 	for (auto& BGM_Channel : BGM_Channels) {
-		if (BGM_Channel.decoder) {
+		if (BGM_Channel.decoder || BGM_Channel.midiout) {
+			if (BGM_Channel.midiout) {
+				BGM_Channel.midiout->Pause();
+			}
 			BGM_Channel.paused = true;
```
fmatthew5876, line 80: I would abstract this behind a `SetPaused()` function in `BGM_Channel`

src/audio_midiout.cpp
```diff
@@ -0,0 +1,387 @@
...
+#include "output.h"
+#ifdef _WIN32
+#include "platform/windows/midiout_device_win32.h"
+#elif __APPLE__
+#include "platform/macos/midiout_device_coreaudio.h"
+#endif
+
+class GenericMidiOut : public MidiOut, public midisequencer::output {
+public:
+	GenericMidiOut(MidiOutDevice* device);
```
fmatthew5876, line 31: This is taking ownership of the device, so change this to take `std::unique_ptr<MidiOutDevice>` by value and remove the calls to release() you have in other places where you call this.

src/audio_midiout_device.h
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	virtual void SendSysExMessage(const void* data, size_t size) {
+		(void)data;
+		(void)size;
+	}
+
+	virtual void SendMidiReset() {}
+
+	virtual bool IsOK() {
```
fmatthew5876, line 49: virtual calls are slow. It may be slightly more performant to store a `bool isok` in this class and let the child classes change the bool.

src/audio_sdl.h
```diff
@@ -27,6 +28,13 @@ class SdlAudio : public GenericAudio {
 
 	void LockMutex() const override;
 	void UnlockMutex() const override;
+	void LockMidiOutMutex() const override;
+	void UnlockMidiOutMutex() const override;
+
+	bool midiout_thread_exit = false;
```
fmatthew5876, line 34: `std::atomic<bool>`? Also, replace assignments with explicit calls to `load` and `store` to make it obvious that multi-threaded synchronization is happening.

src/platform/macos/midiout_device_coreaudio.h
```diff
@@ -0,0 +1,48 @@
...
+	void SendSysExMessage(const void* data, size_t size) override;
+	void SendMidiReset() override;
+
+	bool IsOK() override {
+		return graph != NULL;
+	}
+
+private:
+	AudioUnit midi_out;
+    AUGraph graph;
```
fmatthew5876, line 43: whitespace

src/audio_midiout_device.h
```diff
@@ -0,0 +1,54 @@
...
+#include <cstdint>
+#include <cstddef>
+
+ /**
+  * Manages sequencing MIDI files and emitting MIDI events to a MIDI out
+  * device.
+  */
+class MidiOutDevice {
+public:
+	virtual ~MidiOutDevice() = default;
```
fmatthew5876, line 30: Copy and move operations `= delete`

src/midisequencer.cpp
```diff
@@ -246,6 +260,33 @@ namespace midisequencer{
...
 
+    float sequencer::get_start_skipping_silence() {
+        for (auto i = messages.begin(); i != messages.end(); ++i) {
+            // If we find Loop Start before the first NoteOn, just start there
+            if (is_loop_start(i->message)) {
+                float time = i->time;
+                // RPG_RT always rewinds "a little"
+                // This amount is based on the tempo, and this 2100000 divisor
+                // I determined experimentally.
+                time = std::max(0.0f, time - (i->tempo / 2100000.0f));
```
fmatthew5876, line 271: Leaving a note here, maybe we can confirm exact numbers from RE tools? Or maybe it's too hard if it's buried in harmony.dll?

src/midisequencer.cpp
```diff
@@ -246,6 +260,33 @@ namespace midisequencer{
         }
     }
 
+    float sequencer::get_start_skipping_silence() {
+        for (auto i = messages.begin(); i != messages.end(); ++i) {
```
fmatthew5876, line 264: C++11 foreach loop

src/audio_sdl.h
```diff
@@ -27,6 +28,13 @@ class SdlAudio : public GenericAudio {
...
+private:
+	SDL_Thread *midiout_thread = nullptr;
+	SDL_mutex *midiout_mutex;
```
fmatthew5876, line 37: `= nullptr`

src/audio_midiout.h
```diff
@@ -0,0 +1,188 @@
...
+	 * Determines whether the MIDI sequence is finished.
+	 *
+	 * @return true sequence ended
+	 */
+	virtual bool IsFinished() const = 0;
+
+	/* Call this every ~1ms. 
+	 * @param delta Time in microseconds since last called. 
+	 */
+	virtual void Update(long long delta) = 0;
```
fmatthew5876, line 137: Why not use `std::chrono::microseconds` here? Then we're protected from unit bugs and you don't need the comment.

src/audio_midiout.cpp
```diff
@@ -0,0 +1,387 @@
...
+		midi_msg |= (((event_type << 4) & 0xF0) | (channel & 0x0F)) & 0x0000FF;
+		device->SendMidiMessage(midi_msg);
+	}
+}
+
+void GenericMidiOut::midi_message(int, uint_least32_t message) {
+	uint8_t event_type = (message & 0x0000F0) >> 4;
+	uint8_t channel = (message & 0x00000F);
+	uint8_t value1 = (message & 0x00FF00) >> 8;
+	uint8_t value2 = (message & 0xFF0000) >> 16;
```
fmatthew5876, line 325: It can be worth adding static inline helper functions to grab the event_type, channel, and values from the uint32_t. Then you don't need to copy paste the code.

src/audio_midiout.h
```diff
@@ -0,0 +1,188 @@
...
+	 * To do a fade out begin must be larger then end.
+	 * Call Update to do the fade.
+	 * Volume changes will not really modify the volume but are only helper
+	 * functions for retrieving the volume information for the audio hardware.
+	 *
+	 * @param begin Begin volume (from 0-100)
+	 * @param end End volume (from 0-100)
+	 * @param duration Fade duration in ms
+	 */
+	virtual void SetFade(int begin, int end, int duration) {
```
fmatthew5876, line 46: `chrono::milliseconds` for duration?

src/audio_midiout.cpp
```diff
@@ -0,0 +1,387 @@
...
+	mtime = mtime + ((delta / 1000000.0f) * tempo_multiplier);
+
+	if (IsFinished() && looping) {
+		mtime = seq->rewind_to_loop();
+		reset_tempos_after_loop();
+	}
+}
+
+bool GenericMidiOut::Seek(std::streamoff offset, std::ios_base::seekdir origin) {
+	if (offset == 0 && origin == std::ios_base::beg) {
```
fmatthew5876, line 297: What if I seek to the beginning correctly using `cur` or `end`?

src/audio_sdl.cpp
fmatthew5876: Let's re-evaluate, even if `std::thread` is not well supported what about `std::mutex`? The `lock_guard` and `unique_lock` RAII facilities are really great to ensure no deadlocks from leaks.

src/audio_generic.cpp
```diff
@@ -62,31 +67,49 @@ void GenericAudio::BGM_Play(const std::string& file, int volume, int pitch, int
...
 	for (auto& BGM_Channel : BGM_Channels) {
 		BGM_Channel.stopped = true; //Stop all running background music
-		LockMutex();
-		BGM_Channel.decoder.reset();
-		UnlockMutex();
+		if (BGM_Channel.midiout) {
+			BGM_Channel.midiout->Reset();
+			BGM_Channel.midiout.reset();
+		} else if (BGM_Channel.decoder) {
+			LockMutex();
```
fmatthew5876, line 107: In this function you lock the midi mutex, and then you lock the audio mutex. Just below here,  in `BGM_PlayedOnce`, you first lock the audio mutex and then the midi mutex.

Different lock orders of multiple mutex can result in a deadlock. 

Now given that only player main thread will ever call these 2 functions, it's safe by accident.

This is another reason in favor of `std::mutex` as the standard library has facilities to lock mutiple mutex in guaranteed order.

https://en.cppreference.com/w/cpp/thread/lock

fmatthew5876:
Given that @tyrone-sudeium has done this hard work to produce an RPG_RT compatible midi sequencer, should we use this as the base of the rest of our midi libraries instead of fmmidi?

Could not Fluidlite, fluidsynth etc,, be implemented as `MidiDevice` using this API? Then we have one sequencer we can fix bugs and maintain for RPG_RT compatibility, and stack any manner of synthesizer libraries underneath as needed for the various platforms.

src/audio_sdl.cpp
fmatthew5876: It occurs to me also that other than using the thread API, there is nothing SDL specific in this entire PR. I wonder if the management of this thread even belongs here?

Ghabry:
Thanks for opening the PR. Will think about ways to unify the logic but your code is a good base :+1: (and there shouldn't be conflicts for a while as my audio work is finished)

To explain the ``std::thread`` and ``std::mutex`` situation: They are broken on 3ds, Wii, emscripten and (I think?) Vita. So for the non-major targets only the Switch implements them. The good news is here that on the slow platforms (emscripten can't support it because SharedArrayBuffers are off for spectre mitigations) this isn't useful anyway. Sooo tif you want to use ``std::thread`` the realtime MIDI stuff must be properly abstracted to avoid a ifdef-mess in ``AudioGeneric``.

The motivation for spawning a thread here is that you need a resolution of 1ms. With my FillBuffer approach I use for the Midi Decoders the resolution is the buffer size, which is ~4ms. So for fast MIDI tracks the notes are off by up to 3 ms. This PR fixes this for the "real Midi devices" but it would be also very useful for the softsynths FluidSynth and FmMidi (not useful for WildMidi but wouldn't hurt - is unaffected by this issue).

src/audio_midiout.h
tyrone-sudeium: This `SetFade` is passed the value directly from the RPG Maker game's parameters, and it's also consistent with how `AudioDecoder` works. I'm not opposed to using a better C++ type here but if we do we'd want to adopt it across the board instead of just in this one place, and I'm not sure this is the PR for that...

src/audio_midiout.cpp
tyrone-sudeium: I'm not actually convinced this function is ever used or called. Visual Studio can't find any references to it, and when I comment it all out it still builds. Might just remove this entirely

Ghabry: ``Seek`` is currently not really supported by any ``AudioDecoder``. This is more about future proof, just wasn't needed by now. The only case where a Seek happens right now is in ``Rewind()`` when the end of the file is reached (for BGM looping).

src/midisequencer.cpp
Ghabry: I couldn't figure out the harmony code yet but yeah this should be verified in RE tools.

tyrone-sudeium: I have no experience in doing this myself but if any one of you smart folk work it out, I'd be very keen to hear how close my experimentally-derived value was ;)

src/audio_midiout_device.h
tyrone-sudeium: So I admit this is butting up against the limits of my C++ knowledge. Before EasyRPG the last C++ I ever did was C++03 so half of these words are foreign to me. Despite that, I've had a red-hot go, let me know how I did

Ghabry:
When we agree on the general implementation and style here the next steps are:

1. Decide how to do threading best
2. "Merge" MidiOut and AudioDecoder (common base class) to reduce code duplication. Imo the only real difference is "Update" vs "FillBuffer", otherwise the interface is almost the same. I could do this as I'm more familiar with the codebase.

src/audio_midiout_device.h
fmatthew5876:
```
MidiOutDevice& operator=(const MidiOutDevice&) = delete;
MidiOutDevice& operator=(MidiOutDevice&&) = delete;
```

src/audio_midiout.h
fmatthew5876: Agree, let's not refactor the whole interface. This is good as it is.

tyrone-sudeium:
One thing I haven't handled yet is: If there's an `easyrpg.soundfont` present, I would _assume_ we should make the highest priority MIDI renderer Fluidlite, right? At the moment on Windows and macOS, if we don't compile sdl_mixer, it'll always use MidiOut. I think eventually we'd want to deprecate sdl_mixer entirely so we'll need to handle this. Does this order of precedence sound reasonable?

1. Fluidlite (if `easyrpg.soundfont` available, or otherwise configured)
2. MidiOut (if on supported platform)
3. WildMidi (if patches available)
4. FmMidi (last resort)

Ghabry:
this is also open for far too long now but I haven't forgotten it.
My working order is right now

1. FS 2 (DirectoryTree)
2. Translations
3. Midi

Don't think I will have time for this before christmas :thinking:

tyrone-sudeium:
That's probably for the best anyway @Ghabry - if you needed me for anything with this PR I wouldn't be able to pick it up until the New Year anyway since I'm currently interstate and don't have a suitable computer with me!

Ghabry:
I will rebase this on the FS2 PR and then lets try to get this in :+1:  Native MIDI increases the user experience alot (and reduces dependencies if you don't want / need them).

What I also want to get working here is ALSA Midi support. That one looks like a good, simple example: https://ccrma.stanford.edu/~craig/articles/linuxmidi/alsa-1.0/

I hope the ALSA 1.0 API is still valid though.

fdelapena:
On GNU/Linux and other Unices, [RtMidi](http://www.music.mcgill.ca/~gary/rtmidi/index.html) seems useful to take a look on how to implement it there. Or for a more future proof, PipeWire control ports seems the trend, but it will take some time to get properly documented.

Ghabry:
@tyrone-sudeium
Did the MIDI output work for you on windows? I hear nothing and I'm not sure if the problem is on my side.

tyrone-sudeium:
@Ghabry I developed it on Windows first so it definitely worked for me at the time ...

Ghabry:
The bug must be in Resume or Pause. When I remove them the BGM works 👍

tyrone-sudeium:
Probably an edge case with the whole channel volumes thing. I couldn't think of any other way to "pause" a MIDI device since all the notes get stuck on if you just stop sending it messages, so I settled on sending every channel a volume zero message. When you resume it sends each channel a midi volume message of whatever the value in `channel_volumes` is. If these somehow end up as all zeroes that would probably cause this?

There's another bug I know about that I wanted to mention as well, that only happens on Macs. If you change the audio device (such as by attaching or detaching a bluetooth headset -- something that happens an awful lot in the Apple ecosystem thanks to the AirPods) MIDI basically stops working until the BGM changes. I'm not entirely sure why. I've more or less configured the `MusicDevice` the same way SDL Mixer does, and Apple's documentation explicitly calls out `kAudioUnitSubType_DefaultOutput` as being what you provide to have it automatically support audio device changes. There must be some undocumented behaviour with `MusicDevice` or an edge case in AUGraph I'm not handling. SDL Mixer doesn't have the issue so I should probably crawl through their implementation a bit more to see what we're missing.

src/audio_sdl.cpp
```diff
@@ -125,4 +175,16 @@ void SdlAudio::UnlockMutex() const {
...
+	if (SDL_LockMutex(midiout_mutex) != 0) {
+		Output::Debug("SDL_LockMutex failure: {}", SDL_GetError());
+	}
+}
+
+void SdlAudio::UnlockMidiOutMutex() const {
+	if (SDL_UnlockMutex(midiout_mutex) != 0) {
+		Output::Debug("SDL_UnlockMutex failure: {}", SDL_GetError());
+	}
+}
```
Ghabry, line 188: I cannot think of any platform with MIDI support that has no working std::thread or std::mutex.
Therefore will simplify most of the logic and move it directly into AudioGeneric (or some new class like AudioGenericMidiThread).

Ghabry:
Refactored now the Midi thread to use std::thread and std::mutex. So all the boring stuff is done now.

Now "only" cleanup remaining and testing all ours synths.

tyrone-sudeium:
I've synced up this PR with your branch, will check it out when I get a chance, though it's been a while so I think I've got to rebuild dependencies and whatnot too

tyrone-sudeium:
MIDI fades are definitely broken now; the code between audio and MIDI for `setFade` looked reasonably similar but they worked quite differently. My old implementation mimicked what RPG_RT did by sending volume messages across all the channels every ~10ms. This one might need to be virtual and overridden by `AudioDecoderMidi` in order to setup the variables correctly, if you'd like to keep that implementation working, otherwise maybe there's some way to express this algorithm using only the same variables set by the base class??

Ghabry:
I did not sync the branch yet because I was afraid of the build errors xD.

Yeah I still have to recheck some stuff. I noted that the Fade code looked quite different. Maybe I will make this pure virtual again and implement it in both parent classes...

Ghabry:
@tyrone-sudeium
I readded now your fade code.
But everything of native Midi needs careful testing again.

Also removed SDL2 mixer and OpenAL directly.
They are half-broken and unmaintained. I do not want to invest any further time into them.


Open problem:
The default priority of the various options.
Currently Native Midi wins over every other library. Likely not what one wants...

configure.ac
```diff
@@ -188,27 +181,21 @@ if test "yes" != "$silent"; then
...
-		test -z "$LIBMPG123_LIBS" -a -n "$SDLMIXER_LIBS" && \
-			echo "    Using SDL_mixer instead, this may result in noise or crashes for some MP3s."
+		echo "  -MP3 (libmpg123): $with_libmpg123"
 		echo "  -midi (libwildmidi):       $with_libwildmidi"
 		echo "  -midi (FluidSynth):        $with_fluidsynth"
 		echo "  -midi (FluidLite):         $with_fluidlite"
 		echo "  -ogg Vorbis (vorbisfile):  $with_oggvorbis"
 		echo "  -opus (opusfile):          $with_opus"
-		echo "  -improved WAV (sndfile):   $with_libsndfile"
+		echo "  -WAV (sndfile):   $with_libsndfile"
```
carstene1ns, line 196: missing spaces.

Ghabry:
The swimming pools are open again since Saturday and it is hot outside... but now that VFS is in I will try to get this finished ASAP :sweat_smile:

Ghabry:
To verify @tyrone-sudeium 

Via Twitter:
"EasyRPG's MIDI playback doesn't read the last 1 tick note off just before the loop."

Ghabry:
Added now broken ALSA Midi and completely broken libretro MIDI code.

Maybe someone can figure out why it is broken.

(Also lacks some error handling :))

Ghabry:
On further investigation I guess my "Relative Mode" for asound is the problem. I cannot see anyone using this.

For some reason one must provide the time of the MIDI event, either in ticks, absolute or relative values. I simply passed "0" there, but this does not properly work.

A good moment to refactor everything (MIDI) time related to ``std::chrono``. Should also make MIDI slightly faster due to the move from float to int :).

src/midisequencer.cpp
```diff
@@ -73,8 +73,15 @@ namespace midisequencer{
     float sequencer::rewind_to_loop()
     {
         position = loop_position;
+        if (position != messages.begin()) {
+            position--;
```
tyrone-sudeium, line 77: > Via Twitter:
> "EasyRPG's MIDI playback doesn't read the last 1 tick note off just before the loop."

Was that twitter user running this branch? I remember experiencing this back when I was working on #2245 but I was pretty sure I fixed it in this branch already with this line

Ghabry: No this was reported by someone using the web player on atsumaru. So is a bug in master. Was just not sure if this PR fixes this

Ghabry:
I just noticed that ALSA has a function ``snd_seq_event_output_direct`` for doing unqueued, direct output. So only libretro with there shitty API needs the time...

I also found the reason now why ALSA has stuck notes: The bit-masking for the channel in ``SendMessageToAllChannels``  is incorrect. Not all channels received the "all sound off". After fixing this ALSA works great :+1:

Ghabry:
ALSA and libretro MIDI should now work properly.

This PR still lacks some final polish (error handling :sweat: ) but it can be already carefully tested now. It is feature complete :+1:

Ghabry:
This is now finished code wise - reviewing & testing can begin 👍

src/audio_generic.cpp
```diff
@@ -182,11 +194,46 @@ bool GenericAudio::PlayOnChannel(BgmChannel& chan, Filesystem_Stream::InputStrea
 		return false;
 	}
 
+	// Midiout is only supported on channel 0 because this is an exclusive resource
+	if (chan.id == 0 && GenericAudioMidiOut::IsSupported(filestream)) {
+		chan.decoder.reset();
+
+		// FIXME: Try Fluidsynth and WildMidi first
```
fdelapena, line 201: Is this fixme related to how to reorder priority when games provide soundfonts/patches?

src/audio_decoder_midi.cpp
```diff
@@ -0,0 +1,405 @@
...
+#include "audio_decoder_midi.h"
+#include "midisequencer.h"
+#include "output.h"
+
+using namespace std::chrono_literals;
+
+constexpr int AudioDecoderMidi::midi_default_tempo;
+
+// 1 ms of MIDI message resolution for a 44100 Hz samplerate
+constexpr int samples_per_play = 512;
```
fdelapena, line 29: If the sample rate is guaranteed  to be the same for current settings, it could be based on sample rate setting somehow to prevent precision loss when playing at 22050 Hz or less.

src/audio_decoder_midi.h
```diff
@@ -106,45 +146,66 @@ class GenericMidiDecoder : public AudioDecoder, midisequencer::output {
...
+	bool paused = false;
+	float volume = 0;
 	bool loops_to_end = false;
 
+	int fade_steps = 0;
+	double fade_volume_end = 0;
+	double delta_volume_step = 0;
+	std::chrono::microseconds last_fade_mtime = std::chrono::microseconds(0);
+
+	int frequency = 44100;
```
fdelapena, line 182: Could this use EP_MIDI_FREQUENCY or is it unrelated?

src/platform/linux/midiout_device_alsa.cpp
```diff
@@ -0,0 +1,199 @@
...
+				dst_client_name = client_name;
+				dst_port = snd_seq_port_info_get_port(port_info);
+				dst_port_name = snd_seq_port_info_get_name(port_info);
+				candidate_found = true;
+			}
+		}
+	}
+
+	if (!candidate_found) {
+		Output::Debug("ALSA: No suitable client found");
```
fdelapena, line 66: Because ALSA is not just being used for the sequencer on Linux internally, it could be useful to tell this is about the sequencer for these too.

src/platform/libretro/midiout_device_libretro.cpp
```diff
@@ -0,0 +1,82 @@
...
+	midi_out.flush();
+}
+
+void LibretroMidiOutDevice::SendMidiReset() {
+	unsigned char gm_reset[] = {0xF0, 0x7E, 0x7F, 0x09, 0x01, 0xF7};
+	SendSysExMessage(gm_reset, sizeof(gm_reset));
+}
+
+std::string LibretroMidiOutDevice::GetName() {
+	return "libretro Midi";
```
fdelapena, line 77: Nit: Midi -> MIDI (there are 29 matches for non-symbol names).

src/midisequencer.cpp
```diff
@@ -99,6 +109,13 @@ namespace midisequencer{
     {
         return getc(static_cast<std::FILE*>(fp));
     }
+    static bool is_loop_start(uint_least32_t msg) {
+        // If the message matches the de facto standard MIDI loop instruction:
+        // Which is a Control Change on any channel with Value 111
```
fdelapena, line 114: Is this tested on channels different than 1 from RPG_RT?

What about multiple control changes, does RPG_RT pick the first or the last?

src/midisequencer.cpp
```diff
@@ -246,6 +263,33 @@ namespace midisequencer{
...
+                std::chrono::microseconds time = msg.time;
+                // RPG_RT always rewinds "a little"
+                // This amount is based on the tempo, and this 2100000 divisor
+                // I determined experimentally.
+                time = std::max(0us, std::chrono::microseconds(static_cast<int>(time.count() - (msg.tempo / 2.1f))));
+                return time;
+            } else if ((msg.message & 0xFF) == 0xF0) {
+                // SysEx message. RPG_RT doesn't skip silence if there's a SysEx
+                // message in the beginning, so neither should we...
+                return 0us;
```
fdelapena, line 279: When it says in the beginning, is from 1:1:000? What if the SysEx message exists at the 1st second and the song begins with notes at the 2nd second? will it pause 1 or 2 seconds?

src/audio_resampler.h
```diff
@@ -67,16 +67,51 @@ class AudioResampler : public AudioDecoder {
...
+	 * @param end End volume (from 0-100)
+	 * @param duration Fade duration in ms
+	 */
+	void SetFade(int begin, int end, std::chrono::milliseconds duration) override;
+
 	/**
 	 * Wraps the seek function of the contained decoder
-	 * @note If the seek function of the wrapped decoder is 
+	 * @note If the seek function of the wrapped decoder is
 	 *	somewhat corelated to time the offset is not influenced by the resampling ratio
```
fdelapena, line 115: Nit: This existed before, but tab character nesting note could use spaces after the asterisk for better aligning.

src/audio_midi.h
```diff
@@ -36,27 +36,39 @@
...
-	 * be implemented:
-	 *
-	 * When the processing of Midi messages is supported:
-	 * - OnMidiMessage
-	 * - OnSysExMessage (nice to have)
-	 * - OnMetaEvent (optional)
+	 * The library must implement the following commands:
+	 * - SendMidiMessage
+	 * - SendSysExMessage (nice to have)
 	 * - OnMidiReset
```
fdelapena, line 64: SendMidiReset

src/audio_decoder_midi.cpp
```diff
@@ -0,0 +1,405 @@
...
+			// Throw away all tempo data after the loop point
+			auto rit = std::find_if(tempo.rbegin(), tempo.rend(), [&](auto& t) { return t.mtime <= mtime; });
+			auto it = rit.base();
+			if (it != tempo.end()) {
+				tempo.erase(it, tempo.end());
+			}
+
+			// Bit of a hack, prevent stuck notes
+			// TODO: verify with a MIDI event stream inspector whether RPG_RT does this?
+			// FIXME synth->all_note_off();
```
fdelapena, line 203: Is this already checked and fixed with sysex gm reset or similar somewhere else?

tyrone-sudeium: Turned out to be unnecessary. This is testing my memory a bit since it's now been almost a whole year, but RPG_RT does nothing of the sort, and the stuck notes I was experiencing (if i remember correctly) was mostly because the initial implementation skipped past messages that had the _same time code_ as the loopstart.

src/midisequencer.cpp
tyrone-sudeium: It won't pause at all. If there's any SysExs before the first NoteOn it'll give up. Since it's been a while I can't remember if I definitively verified this behaviour against RPG_RT but this is my understanding of how it behaves. I recall from my youth that harmony would aggressively abandon attempts to skip silence when it encounters any SysExs before any notes play.

tyrone-sudeium:
Doing some exploratory testing on macOS...

Defaulting back to Fluidsynth works OK when putting `easyrpg.soundfont` in the game dir. Though it's a bit clunky because Finder will rename it to `easyrpg.soundfont.sf2` if the original file had `.sf2` as the extension, and it won't show you that it's doing that. Thanks for being "helpful" there, Apple.

I'm noticing a crash when the first MIDI track upon loading a save tries to loop. Apparently it's in `AudioDecoderMidi::meta_event` somewhere, possibly the assert? I'll track it down when I have more time to get my Mac builds working again.

I noticed that `Field 2.mid` from the RTP doesn't loop correctly, when it loops there's snare hit that shouldn't be there.

src/audio_generic.cpp
Ghabry: Yes, right now it is Fluidsynth, Wildmidi, Native, FmMidi.

FmMidi is kinda unused when Native is available. There is a configuration option needed.

src/audio_decoder_midi.h
Ghabry: good find. This must be ``EP_MIDI_FREQUENCY``

src/midisequencer.cpp
```diff
@@ -79,8 +79,9 @@ namespace midisequencer{
         if (position != messages.begin()) {
             position--;
         }
-        return position;
+        return loop_position;
```
Ghabry, line 82: oh, my bad. I broke this when I did the chrono refactor. This was ``loop_position->time`` before :/

Ghabry: btw do we even need the rest of the function? ``position `` is a local variable and dead code.

Ghabry:
There was a bug in the silence skipping when there was tempo-data encountered. This broke the tick and the sample calculation.

Also WildMidi learned now how to skip initial silence and I can confirm that the looping works :+1:

tyrone-sudeium:
I learned something fairly cursed today about win32 midi out. [According to the documentation](https://docs.microsoft.com/en-us/windows/win32/api/mmeapi/nf-mmeapi-midioutlongmsg), whether `midiOutLongMsg` is synchronous or asynchronous is apparently left up to the MIDI driver for the current device, and if it's _asynchronous_, then what I'm doing here (calling `midiOutUnprepareHeader` immediately after `midiOutLongMsg`) will always result in a failure and small memory leak. When it's asynchronous you have to poll the `midiOutUnprepareHeader` call until it results in `MMSYSERR_NOERROR` 🤦.

No idea how relevant this is in 2021, since I'm reasonably sure these days you can't even provide your own MIDI drivers any more, everything goes through Microsoft's built-ins now, and the behaviour in Windows 10 is these calls act synchronous.

src/midisequencer.cpp
tyrone-sudeium: RPG_RT definitely allows any channel. I will verify the behaviour of RPG_RT when there’s multiple loop starts found sometime this weekend 👍

Ghabry:
More to fix: Fluidsynth changed the function parameter in there VIO interface from long to fluid_long_long_t

src/midisequencer.cpp
tyrone-sudeium: Confirmed our behaviour matches RPG_RT here: they'll (and we) use the last loop start found (when sorting by time)

src/audio_decoder_midi.h
Ghabry: actually this value is not used, at least when the caller is well-behaving and does "SetFormat"

src/audio_decoder_midi.cpp
Ghabry: Fixed this now with a branch for the common values (22050 and 11025). For 44100 vs 48000 the difference doesn't really matter.

Ghabry:
no idea how I calculated the 512 for the samples. This number is totally wrong. It must be around 64 for ~1.5 ms. Fixes this now.

Ghabry:
This should finally address everything...

CMakeLists.txt
```diff
@@ -635,24 +639,20 @@ endif()
 
 # Sound system to use
 if(${PLAYER_TARGET_PLATFORM} STREQUAL "SDL2")
-	set(PLAYER_AUDIO_BACKEND "Auto" CACHE STRING "Audio system to use. Options: Auto SDL2_mixer SDL2 OpenAL OFF")
-	set_property(CACHE PLAYER_AUDIO_BACKEND PROPERTY STRINGS Auto SDL2 SDL2_mixer OpenAL OFF)
+	set(PLAYER_AUDIO_BACKEND "Auto" CACHE STRING "Audio system to use. Options: Auto SDL2 OFF")
```
carstene1ns, line 642: "Auto" could be removed here, since only choice is SDL.

carstene1ns (approved):
Works for me.

fdelapena (approved):

Ghabry:
jenkins: test this please

src/midisequencer.cpp
Ghabry: I create a new issue for this

Ghabry:
Jenkins: test this please

carstene1ns:
Unfortunately 3a19bb66203e7 was not enough, there are still traces of SDL_mixer around. Is only dead code, since the `HAVE_SDL_MIXER` definition should be missing, however, should still be deleted.

https://github.com/EasyRPG/Player/search?q=have_sdl_mixer

carstene1ns:
Another thing broken: fmmidi fallback mode (autotools build). Will prepare a patch for that.

PR #2304, opened 2020-08-26
Fix self-target skills outside of battles

rueter37:
Outside of battles the first party member is able to use self-target skills on the other party members too. This is because the ActorTarget scene checks for a negative index to lock the selection. But the first party member has the index zero and things like "negative zero" doesn't exist. This PR aims to fix this by shifting the party member index.


fmatthew5876:
I tested and confirmed the bug on master, confirmed the fix in this PR.

Ghabry:
Jenkins: test this please

Ghabry (approved):

Issue #804, opened 2016-03-07
Emscripten: Update HTML shell file from newer releases


carstene1ns:

Ghabry:
I diffed the latest emscripten shell against ours and there isn't really anything new.

Except for this bugfix here: https://github.com/emscripten-core/emscripten/commit/1fc1c90941936c140827e101dce1576c60811c3d#diff-8fd398136d2435d1c53386be85f6ff73 
Maybe solves the keyboard focus issue we have in some browsers

PR #2307, opened 2020-08-27
Backport emscripten-core/emscripten#7631 for focus on iframes

fdelapena:
Multiple games on sites like rmarchiv itch.io and similar will benefit on this.
Additinally, remove the default 'Downloading...' text to prevent user confusion when robots show web content preview.

Closes: #804.


Ghabry (commented):
looks fine but can't be tested right now because of https://github.com/EasyRPG/buildscripts/issues/83

fdelapena:
This is ready. Tested with Firefox and Chromium, automatic canvas resizing on windowed mode seems to work, though full screen toggle goes black for some (likely unrelated) reason.

Ghabry:
Imo removing downloading is not the way to go. You could call it "Loading game..." and then use open graph standard to get a better link preview.

fdelapena:
The "Downloading..." is set just later from JavaScript, so the experience doesn't change in practice. It could even contain a <noscript> there with meaningful Player details. Even the noscript gets overwritten from JS. Unfortunately not every preview out there uses open graph, including some search engines.

Ghabry (commented):
No idea how but your media queries (?) broke the fullscreen function.

You could also try with ``Module.requestFullscreen`` I replaced this with Module['canvas'].requestFullscreen because the emscripten solution was broken (it works by wrapping the canvas in a div and requesting fullscreen from the div)

Ghabry:
There is this ``OBJECT_DEPENDS`` stuff in the CMake file where pre and post JS are specified. Just add the shell there and it will work

```
set_source_files_properties("src/main.cpp" PROPERTIES OBJECT_DEPENDS "${PLAYER_JS_PREJS};${PLAYER_JS_POSTJS};${PLAYER_JS_SHELL}")
```.

Ghabry (approved):

PR #2308, opened 2020-08-28
Fix healing display in RPG Maker 2003 battles

### LINKED ISSUE #1680: Battle Algorithm clamping is wrong for 2k3 battles
The way we clamp damage is designed for 2k battles but this is broken for 2k3.

To see this, start a 2k3 battle and have an actor use a healing magic on someone who has full HP. In RPG_RT, you'll see some positive healing number in green text based on the skill power. In Player, you'll see 0.

This is because we clamp healing to `(max hp - cur hp)` to satisfy how the 2k battle system works.

This is one immediate example. There are probably several others.

rueter37:
This PR tries to fix #1680. It checks if RPG Maker 2003 is used and if it's true, then the clamping on this->hp in game_battlealgorithm.cpp gets disabled when a healing spell or healing item is used. This makes the jumping number always show the amount of HP healed. The function SetHp in game_actor.cpp makes sure that the maximum HP doesn't get exceeded despite the disabled clamping.

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -1232,8 +1232,13 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 
 			effect = Utils::Clamp(effect, 0, MaxDamageValue());
 
-			if (skill.affect_hp)
-				this->hp = std::max<int>(0, std::min<int>(effect, GetTarget()->GetMaxHp() - GetTarget()->GetHp()));
+			if (skill.affect_hp) {
+				if (Player::IsRPG2k3()) {
+					this->hp = effect;
+				} else {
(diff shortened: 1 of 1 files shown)


Ghabry:
Thanks. Would be good to verify the behaviour also for other numbers.

Not sure if AGI etc. can be displayed but I guess the same bug happens with

- [ ] MP

?

rueter37:
As far as I know, the jumping numbers in RPG Maker 2003 battles are only displayed if HP are affected.

fdelapena:
Jenkins: add to whitelist

Ghabry:
hm MP recovery really shows no numbers. How lame :-1:

Ghabry (approved):

PR #2310, opened 2020-08-30
Battle 2k3: Add support for flipped battle animations

### LINKED ISSUE #1564: Battle 2k3: Flip assets if facing the other direction option
Flips the assets automatically to the enemy when attacking instead of always facing to the left.

btw, which Chunk is this?

system2

rueter37:
This PR takes on a more difficult task and fixes #1564. The changes look more intrusive so it is possible that changes are required on this PR. Anyway, this basically adds a optional invert parameter on the BattleAnimationBattle constructor, the ShowBattleAnimation method and the PlayAnimation methods which makes the animations display mirrored if the parameter is set to true. The new parameter gets used in Scene_Battle_Rpg2k3 only because this feature is used in RPG Maker 2003 only. Some particularities about the flipped animations: Animations get flipped only if the "Flip assets if facing the other direction" flag is set. If this condition is met then the animations will be flipped if either the battler sprite of the source is flipped or the source is belonging to the enemy party (XOR-Operation). On reflected spells the second animation will be flipped if either the battler sprite of the original target is flipped or the original target is belonging to the enemy party (XOR-Operation).

### DIFF
--- a/src/battle_animation.cpp
+++ b/src/battle_animation.cpp
@@ -108,8 +108,10 @@ void BattleAnimation::DrawAt(Bitmap& dst, int x, int y) {
 		}
 
 		SetX(cell.x + x);
+		if (invert) SetX(x - cell.x);
 		SetY(cell.y + y);
 		int sx = cell.cell_id % 5;
+		if (invert) sx = 4 - sx;
 		int sy = cell.cell_id / 5;
 		int size = animation.large ? 128 : 96;
 		SetSrcRect(Rect(sx * size, sy * size, size, size));
@@ -122,6 +124,7 @@ void BattleAnimation::DrawAt(Bitmap& dst, int x, int y) {
 		SetOpacity(255 * (100 - cell.transparency) / 100);
 		SetZoomX(cell.zoom / 100.0);
 		SetZoomY(cell.zoom / 100.0);
+		SetFlipX(invert);
 		Sprite::Draw(dst);
 	}
 
@@ -274,9 +277,10 @@ void BattleAnimationMap::ShakeTargets(int /* str */, int /* spd */, int /* time 
 
 /////////
 
-BattleAnimationBattle::BattleAnimationBattle(const lcf::rpg::Animation& anim, std::vector<Game_Battler*> battlers, bool only_sound, int cutoff_frame) :
+BattleAnimationBattle::BattleAnimationBattle(const lcf::rpg::Animation& anim, std::vector<Game_Battler*> battlers, bool only_sound, int cutoff_frame, bool set_invert) :
 	BattleAnimation(anim, only_sound, cutoff_frame), battlers(std::move(battlers))
 {
+	invert = set_invert;
 }
 
 void BattleAnimationBattle::Draw(Bitmap& dst) {
--- a/src/battle_animation.cpp
+++ b/src/battle_animation.cpp
@@ -107,8 +107,7 @@ void BattleAnimation::DrawAt(Bitmap& dst, int x, int y) {
 			continue;
 		}
 
-		SetX(cell.x + x);
-		if (invert) SetX(x - cell.x);
+		SetX(invert ? x - cell.x : cell.x + x);
 		SetY(cell.y + y);
 		int sx = cell.cell_id % 5;
 		if (invert) sx = 4 - sx;
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -250,12 +250,7 @@ void Game_BattleAlgorithm::AlgorithmBase::PlayAnimation(bool on_original_targets
 		anim_targets.push_back(*current_target);
 	} while (TargetNextInternal());
 
-	Game_Battle::ShowBattleAnimation(
-		GetAnimation()->ID,
-		anim_targets,
-		false,
-		-1,
-		invert);
+	Game_Battle::ShowBattleAnimation(GetAnimation()->ID, anim_targets, false, -1, invert);
 	has_animation_played = true;
 
 	current_target = old_current_target;
@@ -763,7 +758,6 @@ void Game_BattleAlgorithm::AlgorithmBase::TargetFirst() {
 	if (party_target) {
 		if (IsReflected()) {
 			party_target->GetBattlers(original_targets);
-			current_original_target = original_targets.begin();
 			targets.clear();
 			source->GetParty().GetActiveBattlers(targets);
 		} else {
@@ -773,7 +767,6 @@ void Game_BattleAlgorithm::AlgorithmBase::TargetFirst() {
 	} else {
 		if (IsReflected()) {
 			original_targets.push_back(GetTarget());
-			current_original_target = original_targets.begin();
 			targets.clear();
 			targets.push_back(source);
 		}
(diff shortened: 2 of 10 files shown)


src/game_battlealgorithm.cpp
```diff
@@ -815,6 +823,14 @@ bool Game_BattleAlgorithm::AlgorithmBase::OriginalTargetsSet() const {
 	return (original_targets.size() > 0);
 }
 
+Game_Battler* Game_BattleAlgorithm::AlgorithmBase::GetOriginalTarget() const {
+	if (current_original_target == original_targets.end()) {
+		return NULL;
```
fmatthew5876, line 828: `nullptr`

src/game_system.cpp
```diff
@@ -574,3 +574,7 @@ void Game_System::IncFrameCounter() {
 Color Game_System::GetBackgroundColor() {
 	return bg_color;
 }
+
+const bool& Game_System::GetInvertAnimations() {
```
fmatthew5876, line 578: Primitive types like ints and bools should be returned by value, not `const ref`. References are less efficient for anything smaller than 2 words.

src/scene_battle_rpg2k3.cpp
```diff
@@ -837,7 +837,15 @@ bool Scene_Battle_Rpg2k3::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBas
 
 	if (play_reflect_anim) {
 		play_reflect_anim = false;
-		action->PlayAnimation();
+		if (Game_System::GetInvertAnimations()) {
+			invert_animation = false;
+			if (action->OriginalTargetsSet()) {
+				invert_animation = (action->GetOriginalTarget()->IsDirectionFlipped() ^ (action->GetOriginalTarget()->GetType() == Game_Battler::Type_Enemy));
```
fmatthew5876, line 843: Put this xor logic behind an inline function so it's not copy pasted. Give it a meaningful name so we know what it's trying to do without needing comments.

src/scene_battle_rpg2k3.cpp
```diff
@@ -924,11 +932,17 @@ bool Scene_Battle_Rpg2k3::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBas
 				Sprite_Battler::LoopState_WaitAfterFinish);
 		}
 
+		if (Game_System::GetInvertAnimations()) {
+			invert_animation = (action->GetSource()->IsDirectionFlipped() ^ (action->GetSource()->GetType() == Game_Battler::Type_Enemy));
```
fmatthew5876, line 936: Storing state in the scene adds complexity, makes the code harder to follow, and makes it easier to introduce bugs. Instead, I would suggest re-computing this boolean flag directly using the source and target instead of storing and retrieving this `invert_animation` flag.

fmatthew5876:
> On reflected spells the second animation will be flipped if either the battler sprite of the original target is flipped or the original target is belonging to the enemy party

What if the spell is targeting the enemy party and some but not all enemies are flipped? Which enemy is used to determine if the animation will flip?

src/battle_animation.cpp
```diff
@@ -108,8 +108,10 @@ void BattleAnimation::DrawAt(Bitmap& dst, int x, int y) {
 		}
 
 		SetX(cell.x + x);
+		if (invert) SetX(x - cell.x);
```
fmatthew5876, line 111: Minor performance nit: You're calling `SetX()` twice here. `Sprite::SetX()` is currently inlined so right now it doesn't matter, but later if `SetX()` ever changed to be out of line this would be slower. 

Lets make it a little more future proof by computing the new x once and then calling `SetX()`. You could use a ternary operator or create a temporary variable and assign it.

src/game_battlealgorithm.cpp
```diff
@@ -252,20 +252,23 @@ void Game_BattleAlgorithm::AlgorithmBase::PlayAnimation(bool on_original_targets
 
 	Game_Battle::ShowBattleAnimation(
 		GetAnimation()->ID,
-		anim_targets);
+		anim_targets,
+		false,
+		-1,
+		invert);
```
fmatthew5876, line 258: Doesn't have to be done in this PR, but when you get this level of mess it's time to convert these defaulted parameters to a `BattleAnimationArgs` struct.

src/game_battlealgorithm.cpp
```diff
@@ -766,6 +773,7 @@ void Game_BattleAlgorithm::AlgorithmBase::TargetFirst() {
 	} else {
 		if (IsReflected()) {
 			original_targets.push_back(GetTarget());
+			current_original_target = original_targets.begin();
```
fmatthew5876, line 776: Do we have to store this iterator? Can't you change `GetOriginalTarget()` to `GetFirstOriginalTarget()` and return the first one, or change it to `GetOriginalTargets()` and return a reference to the whole vector and let the caller decide what to do with it.

rueter37:
I have added a commit now which tries to take care of the issues you mentioned. I have added a CheckAnimFlip function which puts the XOR-Operation in one place. I hope I didn't miss anything important.

src/game_battlealgorithm.cpp
```diff
@@ -815,6 +816,10 @@ bool Game_BattleAlgorithm::AlgorithmBase::OriginalTargetsSet() const {
 	return (original_targets.size() > 0);
 }
 
+Game_Battler* Game_BattleAlgorithm::AlgorithmBase::GetFirstOriginalTarget() const {
+	return *original_targets.begin();
```
fmatthew5876, line 820: If you're going to return a pointer, check for empty first and return null if no original targets. Otherwise, return a `Game_Battler&` and add an `assert(!original_targets.empty())`

rueter37: This is probably a dumb question but should I still make the change what this function returns? Because you have successfully made tests and I'm afraid I will break something if I try to change this. And maybe I missed something but I don't quite understand the assert suggestion. For me it looks like an assertion gets triggered when original targets are set.

fmatthew5876: This is a very minor change, and converting pointer to reference any bugs will be compiler errors.

Just be sure to get the assert correct. `assert(!original_targets.empty())` means crash the program if I call this function and the targets are empty. asserts are only enabled in debug mode so you can quickly test a debug build.

Ghabry: Currently your code will crash when there are no ``original_targets``. This means you make an assumption here that this is not supposed to happen. An ``assert`` is nice to mark impossible situations - When the assert is triggered it means the code has a bug. (note that user input should never trigger asserts, for user generated problems there should be more sane error handling)

Using ``&`` instead of ``*`` here makes sense because this indicates to the caller "The return value is always a valid ``Battler``". For ``*`` the caller must expect ``NULL/nullptr``.

rueter37: I somehow could not think straight for the last few hours. But now I went for the first variant in @fmatthew5876 's comment, returning a pointer and doing the check for empty. I have pushed a new commit.

fmatthew5876:
Use `nullptr` instead of `NULL`. After that, this is ready to be merged!

rueter37:
Added a new commit which contains the change you requested.

fmatthew5876 (approved):

fmatthew5876:
I'm ok to merge this anytime.

Thank you for your contribution

Ghabry:
```
scene_battle_rpg2k3.cpp:1489:40: warning: ^ has lower precedence than ==; == will be evaluated first [-Wparentheses]
                return battler->IsDirectionFlipped() ^ battler->GetType() == Game_Battler::Type_Enemy;
```
Well this is what we want, but warning should be silenced by someone.

PR #2311, opened 2020-08-31
Fade BGM on new game and load game

fmatthew5876:
800 ms confirmed with RE


Ghabry (commented):
Not tested yet but this could cause problems when Title and Map have the same BGM (track does not play)

fmatthew5876:
> Not tested yet but this could cause problems when Title and Map have the same BGM (track does not play)

I tested this case and in RPG_RT, the track fades out and then plays again on the map. Seems to work ok in player too.

Should we add a BGMStop after transition out just to be extra safe in case of audio subsystem problems?

Ghabry (approved):
``Game_System`` handles this through a ``stopping`` flag when fading so that ``Play`` starts the track again even when it is the same.
No bug here as already handled.

PR #2312, opened 2020-08-31
Battle algorithm changes

rueter37:
This PR fixes some minor bugs I came across in battles:

- States which recover HP and SP: HP and SP recovery is not restricted by the current amount of the mentioned stats.
- Random variance: In RPG Maker 2000 the variance calculation is even performed on zero damage. I know from playing RPG Maker 2000 games before that attacks sometimes do 1 damage despite too high defense or full magic immunity.
- Random variance of healing skills outside of battles: The old variance formula was still applied so it got replaced by the VarianceAdjustEffect function.
- Skills which do absorbing: Absorbing attack, defense, spirit and agility is only supported in RPG Maker 2003. In RPG Maker 2000 the stats only get decreased.
- Skill accuracy: Accuracy checks are performed on every affected stat. These checks are performed on skills which affects allies too. And on skills which hit the enemy special checks are performed: If affect_hp is set and the check fails, the whole skill fails. With affect_sp it is the same. And if both affect_hp and affect_sp are set then the whole skill fails if both checks fail.
- Revival outside of battles: If a dead actor gets revived outside of battles, he gained 1 HP too much. Moreover skills which cure death only (with affect_hp unset) work only on dead actors even outside of battles.

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -667,7 +667,7 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 	if (GetAffectedAttack() != -1) {
 		int atk = GetAffectedAttack();
 		target->ChangeAtkModifier(IsPositive() ? atk : -atk);
-		if (IsAbsorb() && Player::IsRPG2k3()) {
+		if (IsAbsorb() && Player::IsRPG2k3Updated()) {
 			atk = std::max<int>(0, std::min<int>(atk, std::min<int>(source->MaxStatBattleValue(), source->GetBaseAtk() * 2) - source->GetAtk()));
 			source->ChangeAtkModifier(atk);
 		}
@@ -676,7 +676,7 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 	if (GetAffectedDefense() != -1) {
 		int def = GetAffectedDefense();
 		target->ChangeDefModifier(IsPositive() ? def : -def);
-		if (IsAbsorb() && Player::IsRPG2k3()) {
+		if (IsAbsorb() && Player::IsRPG2k3Updated()) {
 			def = std::max<int>(0, std::min<int>(def, std::min<int>(source->MaxStatBattleValue(), source->GetBaseAtk() * 2) - source->GetAtk()));
 			source->ChangeDefModifier(def);
 		}
@@ -685,7 +685,7 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 	if (GetAffectedSpirit() != -1) {
 		int spi = GetAffectedSpirit();
 		target->ChangeSpiModifier(IsPositive() ? spi : -spi);
-		if (IsAbsorb() && Player::IsRPG2k3()) {
+		if (IsAbsorb() && Player::IsRPG2k3Updated()) {
 			spi = std::max<int>(0, std::min<int>(spi, std::min<int>(source->MaxStatBattleValue(), source->GetBaseAtk() * 2) - source->GetAtk()));
 			source->ChangeSpiModifier(spi);
 		}
@@ -694,7 +694,7 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 	if (GetAffectedAgility() != -1) {
 		int agi = GetAffectedAgility();
 		target->ChangeAgiModifier(IsPositive() ? agi : -agi);
-		if (IsAbsorb() && Player::IsRPG2k3()) {
+		if (IsAbsorb() && Player::IsRPG2k3Updated()) {
 			agi = std::max<int>(0, std::min<int>(agi, std::min<int>(source->MaxStatBattleValue(), source->GetBaseAtk() * 2) - source->GetAtk()));
 			source->ChangeAgiModifier(agi);
 		}
@@ -1004,7 +1004,7 @@ bool Game_BattleAlgorithm::Normal::Execute() {
 			}
 		}
 
-		if ((Player::IsRPG2k() && !Player::IsRPG2kUpdated()) || effect > 0) {
+		if (Player::IsLegacy() || effect > 0) {
 			effect = Game_Battle::VarianceAdjustEffect(effect, 4);
 		}
 
@@ -1228,7 +1228,7 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 
 			effect *= GetTarget()->GetAttributeMultiplier(skill.attribute_effects);
 
-			if ((Player::IsRPG2k() && !Player::IsRPG2kUpdated()) || effect > 0) effect = Game_Battle::VarianceAdjustEffect(effect, skill.variance);
+			if (Player::IsLegacy() || effect > 0) effect = Game_Battle::VarianceAdjustEffect(effect, skill.variance);
 
 			effect = Utils::Clamp(effect, 0, MaxDamageValue());
 
@@ -1274,7 +1274,7 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 			}
 			effect *= GetTarget()->GetAttributeMultiplier(skill.attribute_effects);
 
-			if ((Player::IsRPG2k() && !Player::IsRPG2kUpdated()) || effect > 0) effect = Game_Battle::VarianceAdjustEffect(effect, skill.variance);
+			if (Player::IsLegacy() || effect > 0) effect = Game_Battle::VarianceAdjustEffect(effect, skill.variance);
 
 			effect = Utils::Clamp(effect, 0, MaxDamageValue());
 
@@ -1856,7 +1856,7 @@ bool Game_BattleAlgorithm::SelfDestruct::Execute() {
 	// Like a normal attack, but with double damage and always hitting
 	// Never crits, ignores charge
 	int effect = source->GetAtk() - GetTarget()->GetDef() / 2;
-	if ((Player::IsRPG2k() && !Player::IsRPG2kUpdated()) || effect > 0) {
+	if (Player::IsLegacy() || effect > 0) {
 		effect = Game_Battle::VarianceAdjustEffect(effect, 4);
 	}
 
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -1004,7 +1004,7 @@ bool Game_BattleAlgorithm::Normal::Execute() {
 			}
 		}
 
-		if (effect > 0) {
+		if ((Player::IsRPG2k() && !Player::IsRPG2kUpdated()) || effect > 0) {
 			effect = Game_Battle::VarianceAdjustEffect(effect, 4);
 		}
 
@@ -1228,7 +1228,7 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 
 			effect *= GetTarget()->GetAttributeMultiplier(skill.attribute_effects);
 
-			effect = Game_Battle::VarianceAdjustEffect(effect, skill.variance);
+			if ((Player::IsRPG2k() && !Player::IsRPG2kUpdated()) || effect > 0) effect = Game_Battle::VarianceAdjustEffect(effect, skill.variance);
 
 			effect = Utils::Clamp(effect, 0, MaxDamageValue());
 
@@ -1274,7 +1274,7 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 			}
 			effect *= GetTarget()->GetAttributeMultiplier(skill.attribute_effects);
 
-			effect = Game_Battle::VarianceAdjustEffect(effect, skill.variance);
+			if ((Player::IsRPG2k() && !Player::IsRPG2kUpdated()) || effect > 0) effect = Game_Battle::VarianceAdjustEffect(effect, skill.variance);
 
 			effect = Utils::Clamp(effect, 0, MaxDamageValue());
 
@@ -1835,7 +1835,7 @@ bool Game_BattleAlgorithm::SelfDestruct::Execute() {
 	// Like a normal attack, but with double damage and always hitting
 	// Never crits, ignores charge
 	int effect = source->GetAtk() - GetTarget()->GetDef() / 2;
-	if (effect > 0) {
+	if ((Player::IsRPG2k() && !Player::IsRPG2kUpdated()) || effect > 0) {
 		effect = Game_Battle::VarianceAdjustEffect(effect, 4);
 	}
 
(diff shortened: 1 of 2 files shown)


src/game_battler.cpp
```diff
@@ -455,7 +455,7 @@ int Game_Battler::ApplyConditions() {
 			}
 		}
 		else if(state.hp_change_type == state.ChangeType_gain) {
-			src_hp = std::min(source_hp, hp);
+			src_hp = hp;
 			if(src_hp < 0) {
```
fmatthew5876, line 464: This healing number also appears to be able to exceed the rm2k3 damage cap of 999. Make a recovery skill of 100% + 999 for an actor with 9999 hp. It'll heal for 10998

rueter37: I checked it with both RPG_RT and the changes made in this PR: Your example indeed heals for 10998 while healing skills are capped to 9999 (tested it with a healing skill which heals 10002) in both RPG_RT and this PR. So everything works correctly.

src/game_battlealgorithm.cpp
```diff
@@ -1004,7 +1004,7 @@ bool Game_BattleAlgorithm::Normal::Execute() {
 			}
 		}
 
-		if (effect > 0) {
+		if (Player::IsRPG2k() || effect > 0) {
```
fmatthew5876, line 1007: What is your basis for determining that 2k battles always add variance even if the damage is 0?
 
I'm looking at the 2k battle system in steam 2ke version in RE tools and in both normal attacks and skills I see it checking if the `damage > 0` before applying variance.

Are you using an older version of RPG_RT for your 2k testing? Which version are you using exactly?

src/game_battlealgorithm.cpp
```diff
@@ -647,7 +647,7 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 		int hp = GetAffectedHp();
 		int target_hp = target->GetHp();
 		target->ChangeHp(IsPositive() ? hp : -hp);
-		if (IsAbsorb()) {
+		if (IsAbsorb() && !IsPositive()) {
```
fmatthew5876, line 650: This looks incorrect. `IsPositive()` is a flag that is set for healing skills, where "healing skill" is defined as a skill targeting the same party as the source.

```
this->healing =
    skill.scope == lcf::rpg::Skill::Scope_ally ||
    skill.scope == lcf::rpg::Skill::Scope_party ||
    skill.scope == lcf::rpg::Skill::Scope_self;
```

src/game_battlealgorithm.cpp
```diff
@@ -676,7 +676,7 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 	if (GetAffectedDefense() != -1) {
 		int def = GetAffectedDefense();
 		target->ChangeDefModifier(IsPositive() ? def : -def);
-		if (IsAbsorb()) {
+		if (IsAbsorb() && !IsPositive() && Player::IsRPG2k3()) {
```
fmatthew5876, line 679: Confirmed that stats cannot be absorbed in 2k. I checked with RE and tested it in battle.

src/game_battlealgorithm.cpp
```diff
@@ -1232,24 +1232,24 @@ bool Game_BattleAlgorithm::Skill::Execute() {
...
+			if (skill.affect_attack && Utils::PercentChance(to_hit))
 				this->attack = std::max<int>(0, std::min<int>(effect, std::min<int>(GetTarget()->MaxStatBattleValue(), GetTarget()->GetBaseAtk() * 2) - GetTarget()->GetAtk()));
-			if (skill.affect_defense)
+			if (skill.affect_defense && Utils::PercentChance(to_hit))
 				this->defense = std::max<int>(0, std::min<int>(effect, std::min<int>(GetTarget()->MaxStatBattleValue(), GetTarget()->GetBaseDef() * 2) - GetTarget()->GetDef()));
-			if (skill.affect_spirit)
+			if (skill.affect_spirit && Utils::PercentChance(to_hit))
 				this->spirit = std::max<int>(0, std::min<int>(effect, std::min<int>(GetTarget()->MaxStatBattleValue(), GetTarget()->GetBaseSpi() * 2) - GetTarget()->GetSpi()));
-			if (skill.affect_agility)
+			if (skill.affect_agility && Utils::PercentChance(to_hit))
```
fmatthew5876, line 1252: Confirmed the `to_hit` is re-reolled for every stat which is affected by the skill. Nice catch!

src/game_battlealgorithm.cpp
```diff
@@ -1292,19 +1292,40 @@ bool Game_BattleAlgorithm::Skill::Execute() {
...
 				this->agility = std::max<int>(0, std::min<int>(effect, GetTarget()->GetAgi() - (GetTarget()->GetBaseAgi() + 1) / 2));
 
+			if (skill.affect_hp) {
+				if (skill.affect_sp) {
+					if (GetAffectedHp() == -1 && GetAffectedSp() == -1) {
+						this->success = false;
+						return this->success;
+					}
```
fmatthew5876, line 1313: Can you explain why this is needed? It's not mentioned in the commit.

Also, this is written badly, this entire nest of conditions can be replaced with a single if statement.
```
if ((skill.affect_hp && GetAffectedHp() == -1) || (skill.affect_sp && GetAffectedSp() == -1))
```

src/game_battlealgorithm.cpp
```diff
@@ -1366,7 +1387,7 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 		if (skill.affect_attr_defence) {
 			for (int i = 0; i < static_cast<int>(skill.attribute_effects.size()); i++) {
 				if (skill.attribute_effects[i] && GetTarget()->CanShiftAttributeRate(i + 1, IsPositive() ? 1 : -1)) {
-					if (!Utils::PercentChance(skill.hit))
+					if (!Utils::PercentChance(to_hit))
```
fmatthew5876, line 1390: Nice catch here

src/game_battler.cpp
```diff
@@ -326,7 +326,7 @@ bool Game_Battler::UseSkill(int skill_id, const Game_Battler* source) {
 		}
 		effect *= mul;
 
-		effect += (effect * Utils::GetRandomNumber(-skill->variance * 10, skill->variance * 10) / 100);
+		if (Player::IsRPG2k() || effect > 0) effect = Game_Battle::VarianceAdjustEffect(effect, skill->variance);
```
fmatthew5876, line 330: Checking `effect > 0` is correct, but again I don't see this being skipped for 2k

src/game_battler.cpp
```diff
@@ -341,11 +342,15 @@ bool Game_Battler::UseSkill(int skill_id, const Game_Battler* source) {
 					AddState(lcf::Data::states[i].ID, false);
 				}
 				else {
+					if (i == 0 && IsDead()) {
+						revived = 1;
+					}
```
fmatthew5876, line 347: I've confirmed this bug exists in master. I haven't tested your fix yet. Will do so after other comments addressed.

fmatthew5876 (changes requested):
Some of these are correct, some look wrong. See my comments.

src/game_battlealgorithm.cpp
rueter37: This !IsPositive() check is a preliminary work for an upcoming PR which handles negative attribute rates for RPG Maker 2003. This should stop the user of an absorbing skill healing himself if the absorption of the target fails due to "negative damage". This is because the IsPositive() flag gets inverted if the "negative damage" situations occurs.

src/game_battlealgorithm.cpp
rueter37: The reason for this is explained in the fifth bullet point of the description of this PR. Special accuracy checks are made if a skill targets enemies. If affect_hp without affect_sp is set, then the whole skill fails if the accuracy check for HP fails. If affect_sp without affect_hp is set, then the whole skill fails if the accuracy check for SP fails. If affect_hp and affect_sp are set, then the whole skill fails if both accuracy checks fail. If neither affect_hp nor affect_sp are set then the accuracy checks are performed normally.

src/game_battlealgorithm.cpp
rueter37: Sadly I cannot tell the exact version but it definitely is older than 1.50 (probably 1.10). I checked the RPG Maker 2000 changelog and for version 1.50 it is stated that special case with sometimes 1 damage despite calculated zero damage is fixed. So I have to change the check against RPG Maker 2000 versions before 1.50.

src/game_battler.cpp
rueter37: This "sometimes effect of 1 despite calculated effect of zero"-case even occurs on healing skills outside of battles. But here I have to change the check to RPG Maker 2000 versions older than 1.50 too.

src/game_battlealgorithm.cpp
Ghabry: if this is really only 1.5+ you can use https://github.com/EasyRPG/Player/blob/master/src/player.h#L180 ``IsRPG2kUpdated()``

src/game_battlealgorithm.cpp
Ghabry: because this is an enhancement not in RPG2k(3) better add a small comment here.

src/game_battlealgorithm.cpp
rueter37: I have tried out your proposal but unfortunately it doesn't work if both affect_hp and affect_sp are set. This makes the skill fail if one of both accuracy checks fail. But both checks need to fail to make the whole skill fail if both affect_hp and affect_sp are set.

src/game_battlealgorithm.cpp
fmatthew5876: I have a bunch of RPG_RT binaries here, if you can compute a hash of the RPG_RT exe your'e using you can compare it to these to find out the version. You can use for example `md5sum` program on linux.

 https://github.com/fmatthew5876/rpgrt/tree/master/rpg2000

rueter37: Thank you for providing the binaries! I did some tests and can tell that this "sometimes 1 damage oddity" occurs in RPG Maker 2000 versions 1.05, 1.07, 1.09 and the RPG_RT provided by "Vampires Dawn". The versions 1.51 and 1.61 do not have this oddity. But I came across another problem: How does EasyRPG check the binary version? The console output always tells (even then when the RPG_RT binary is version 1.05, 1.07 or 1.09):
`Debug: RPG2k >= v1.50 / RPG2k3 >= v1.05 detected`
`Debug: Engine configured as: 2k=true 2k3=false MajorUpdated=true Eng=false`

fmatthew5876: EasyRPG has a bunch of heuristics based on the LDB database and other things. We don't scan the RPG_RT.exe binary directly. Our heuristics are not perfect and sometimes get it wrong.

Thanks for this find. It indeed looks like it's possibly a change with older RPG_RT versions. I'll need to dig with RE tools to verify

src/game_battlealgorithm.cpp
fmatthew5876: If this is part of your implementation of negative attributes in a future PR, lets take it out of here and move it to there.

rueter37: The !IsPositive() checks on IsAbsorb() have been removed now.

src/game_battlealgorithm.cpp
rueter37: Added version check for RPG Maker 2000 version below 1.50 now.

src/game_battler.cpp
rueter37: Added version check for RPG Maker 2000 version below 1.50 now.

fmatthew5876:
For this one, because there are so many delicate changes here which are hard to review, I'm going to have to ask you to rebase the fixes into your original commits. Each commit here should be a standalone bug fix / feature addition with a useful commit message.

fmatthew5876:
Thanks, I'll review this again and test soon.

Btw, we're pretty active on irc and discord. We often use it to discuss development outside of git. You're welcome to join either one. They are bridged, so both are the same.

Ghabry:
The URLs are (pick any)

- #easyrpg in freenode
- https://riot.im/app/#/room/#easyrpg:matrix.org (basically open source discord)
- https://discord.gg/aRBtDAG

src/game_battlealgorithm.cpp
```diff
@@ -1004,7 +1004,7 @@ bool Game_BattleAlgorithm::Normal::Execute() {
 			}
 		}
 
-		if (effect > 0) {
+		if ((Player::IsRPG2k() && !Player::IsRPG2kUpdated()) || effect > 0) {
```
fmatthew5876, line 1007: So it looks like this special case also occurs in the old versions of rm2k3 v1.0.3.0.

So we should change this to `Player::IsLegacy() || effect > 0`

src/game_battlealgorithm.cpp
```diff
@@ -667,7 +667,7 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 	if (GetAffectedAttack() != -1) {
 		int atk = GetAffectedAttack();
 		target->ChangeAtkModifier(IsPositive() ? atk : -atk);
-		if (IsAbsorb()) {
+		if (IsAbsorb() && Player::IsRPG2k3()) {
```
fmatthew5876, line 670: Absorb stats not supported in 2k3 v1.0.3.0

Change this to `IsAbsorb() && Player::IsRPG2k3Updated()`

src/game_battler.cpp
```diff
@@ -326,7 +326,7 @@ bool Game_Battler::UseSkill(int skill_id, const Game_Battler* source) {
 		}
 		effect *= mul;
 
-		effect += (effect * Utils::GetRandomNumber(-skill->variance * 10, skill->variance * 10) / 100);
+		if ((Player::IsRPG2k() && !Player::IsRPG2kUpdated()) || effect > 0) effect = Game_Battle::VarianceAdjustEffect(effect, skill->variance);
```
fmatthew5876, line 330: `Player::IsLegacy()` again.

rueter37:
Added a new commit which contains the requested changes.

fmatthew5876 (approved):
I tested these and they work. All changes also confirmed in RE.

So I approve

fmatthew5876:
One other thing I noticed, this `effect > 0` for variance means variance won't be applied when negative attributes invert the damage? 

It looks that way from RPG_RT, which would be yet another 2k3 bug..

rueter37:
For skills this is true. I know from playing "Vampires Dawn II" the enemy "Gelber Drache" sometimes uses the skill "Gelbes Feuer" which is a fire-type skill which has a power of 400, spirit influence set to 4 and variance set to 3. If their skill get reflected and they hit themselves with this skill they always get healed by 200 without any changes. Their fire-type attribute is set to "E" which is set to -50%. My (still not published) negative attribute PR handles this case as well.

fmatthew5876:
Regarding negative attributes, I'm handling part of that in #2320.

Issue #2264, opened 2020-07-08
UI and text offset observations, comparisons


Tayruu:
#### Player platform:
Windows, version 0.6.2.1
RPG Maker 2000 1.61 English release

#### Case studies
As I've observed sometimes some of the UI elements don't completely match RPG_RT's UI, I thought I'd go through and check for the differences at hand. It's not always full-proof because of the font differences, but I hope it's close enough to try and get things as close as possible.

**Title screen, main menu screen, party/actor battle menus**
- Scrolling stops at end of menu. In RPG_RT, holding down arrow scrolls endlessly.
- Scrolling might be _slightly_ too slow.

**Item screen**
(RPG_RT / EasyRPG / Overlap)
![](https://i.imgur.com/1ew9qJh.png)
- Item quantity indicator (":"/"x") unknown if offset
- Item quantity is -2px (compared to RPG_RT pos)
- Selector +4px too long
- Right-column elements -4px
  - This also causes subsequent values to be further offset

**Skills screen**
![](https://i.imgur.com/T9lcpx4.png)
- Same as above for skill elements
- HP in active actor bar -6px
  - This also affects subsequent values (current, "/", max)

**Equip screen**
![](https://i.imgur.com/kWS4l0y.png)
- Same as above
- ">" glyph +3px (though since this apparently changes for JP locale, dunno if this would be fiddly to match)
- Scrolling stops at end of equip slots. In RPG_RT, holding down arrow scrolls endlessly.

**Save screen**
![](https://i.imgur.com/AYHaKqA.png)
- Missing space between "File" and slot number.
- Missing scroll arrows
- Arrow key can be tapped at end of list to scroll to the other end. This behaviour doesn't happen in RPG_RT.

**End screen**
![](https://i.imgur.com/DuUPlOq.png)
Now *this* is a fun one. I was boggling as to why when testing my own game why there was extra space to the side of the header. On further testing, it seems that this text for whatever reason renders spaces smaller than in other places.
Compare a string with more spaces, to the text in a message box:
![](https://i.imgur.com/PGlKFCn.png)
![](https://i.imgur.com/FX1kwLY.png)

**Message box**
![](https://i.imgur.com/8VjIEgw.png)
- Selector +4px too long

**Shop screen**
![](https://i.imgur.com/MgLoO2y.png)
- Actor offset +4px downwards
- Equipment stat change icon animates 1-2-3-4-3-2, or something? In RPG_RT it animates 1-2-3-4
[...]

Issue #2216, opened 2020-05-14
Multi-column UI windows should have a gap between the selector


fmatthew5876:
![Untitled](https://user-images.githubusercontent.com/1198466/81994538-168ce200-9616-11ea-873b-2509edb6bd89.png)

Using the same game in both Player and RPG_RT. Notice the difference in width of the selector bar. In RPG_RT, if you flip the bar left and right you'll see a gap in the middle.

If Player, there is no gap. Our bar covers exactly half of the window size.

There are a lot of other small differences here which I think are just due to fonts?

This also lists several UI tweaks, but I'm not sure how many are due to font differences vs actual pixel positioning. 

- [x] Fix selector bar in item / skills #2313
- [ ] Fix face position in main menu. Note differences between 2k / 2k3 and party target window
- [x] Skill window costs text is off by 2 pixels #2313
- [x] Save file name is wrong. Half space between filetext and number #2313
- [x] Save file selector bar is too short #2313
- [ ] Check all other screens for position fixes

#2313

carstene1ns:
I think we do not really need to be pixel perfect here, because of the font, the margin is simply different.
If add the gap, needs repositioning the texts and may cut off long lines or such.

Ghabry:
Kinda related: There is a reverse engenering framework for python (using js for scripting) called https://frida.re. This allows arbitrary function hooking anywhere independent of the calling convention. Saw this used by one of the Google Project Zero members (taviso).

Never used this before but here is a simple example hooking ExtTextOutA:
```
frida-trace -n RPG_RT.exe -i ExtTextOutA -X gdi32full.dll

onEnter: function (log, args, state) {
  log("ExtTextOutA(" + args[1] + ", " + args[2] + ", " + args[5].readAnsiString() + ")");
  args[5].writeAnsiString("X");
},
```

![Unbenannt-fs8](https://user-images.githubusercontent.com/1331889/82121897-ac4c7c80-9790-11ea-8872-72d94ea7389f.png)

fmatthew5876:
![equip](https://user-images.githubusercontent.com/1198466/82123027-21a25800-9765-11ea-9318-1c22d15406ce.png)
![exit](https://user-images.githubusercontent.com/1198466/82123028-223aee80-9765-11ea-87fb-c87b6e12d050.png)
![file](https://user-images.githubusercontent.com/1198466/82123030-22d38500-9765-11ea-85ec-eae6a23891d4.png)
[...]

PR #2313, opened 2020-08-31
UI changes (1/2)

rueter37:
This PR and #2361 provides partial fixes for #2216 and #2264. In #2216 the things mentioned in the checkboxes 1, 3, 4 and 5 get fixed. One major change is the addition of an optional "halfwidthspace" parameter in the Window_Base which is set to true by default. Setting it to false makes the help window draw full width spaces which is required in the ActorTarget scene for the skill/item name and in the EndGame scene for the confirmation message (this is handled in part 2: #2361). The second major change is the addition of a gap between the menu items in multi-column selection windows (this is handled in this part).

More changes in this part:
- The item count/skill cost is now displayed on the right end of the selector in the Item and Skill windows.
- Endless scrolling has been implemented.
- Shops: Change animation behavior of the comparison indicators and move prices to the right end.

Things mentioned in #2264 not covered by both PRs:
- Equipment scene (for context see #1660, and I personally prefer it that way too)
- Arrows in the Savegame window (this is a special case because the arrows are currently only implemented in the window class and in RPG_RT the arrows are drawn outside of any window in the Savegame window).
- Scrolling speed
- Text shadows

### DIFF
--- a/src/window_item.cpp
+++ b/src/window_item.cpp
@@ -123,11 +123,8 @@ void Window_Item::DrawItem(int index) {
 		bool enabled = CheckEnable(item_id);
 		DrawItemName(*item, rect.x, rect.y, enabled);
 
-		std::stringstream ss;
-		ss << number;
 		Font::SystemColor color = enabled ? Font::ColorDefault : Font::ColorDisabled;
-		contents->TextDraw(rect.x + rect.width - 28, rect.y, color, "x");
-		contents->TextDraw(rect.x + rect.width - 6, rect.y, color, ss.str(), Text::AlignRight);
+		contents->TextDraw(rect.x + rect.width - 24, rect.y, color, fmt::format(":{:3d}", number));
 	}
 }
 
--- a/src/window_selectable.cpp
+++ b/src/window_selectable.cpp
@@ -124,15 +124,13 @@ void Window_Selectable::Update() {
 	Window_Base::Update();
 	if (active && item_max > 0 && index >= 0) {
 		if (Input::IsRepeated(Input::DOWN) || Input::IsTriggered(Input::SCROLL_DOWN)) {
-			if (index < item_max - column_max || (column_max == 1 &&
-				(Input::IsTriggered(Input::DOWN) || Input::IsTriggered(Input::SCROLL_DOWN)))) {
+			if (index < item_max - column_max || column_max == 1) {
 				Game_System::SePlay(Game_System::GetSystemSE(Game_System::SFX_Cursor));
 				index = (index + column_max) % item_max;
 			}
 		}
 		if (Input::IsRepeated(Input::UP) || Input::IsTriggered(Input::SCROLL_UP)) {
-			if (index >= column_max || (column_max == 1 &&
-				(Input::IsTriggered(Input::UP) || Input::IsTriggered(Input::SCROLL_UP)))) {
+			if (index >= column_max || column_max == 1) {
 				Game_System::SePlay(Game_System::GetSystemSE(Game_System::SFX_Cursor));
 				index = (index - column_max + item_max) % item_max;
 			}
--- a/src/window_selectable.cpp
+++ b/src/window_selectable.cpp
@@ -64,12 +64,9 @@ int Window_Selectable::GetPageItemMax() {
 
 Rect Window_Selectable::GetItemRect(int index) {
 	Rect rect = Rect();
-	rect.width = (contents->GetWidth()) / column_max - 4;
+	rect.width = (width / column_max - 16);
+	rect.x = (index % column_max * (rect.width + 16));
 	rect.height = 12;
-	rect.x = index % column_max * rect.width;
-	if (rect.x > 0){
-		rect.x += 8;
-	}
 	rect.y = index / column_max * 16 + 2;
 	return rect;
 }
@@ -103,14 +100,8 @@ void Window_Selectable::UpdateCursorRect() {
 		SetTopRow(row - (GetPageRowMax() - 1));
 	}
 
-	if (column_max > 1){
-		cursor_width = (width / column_max - 16) + 12;
-		x = (index % column_max * cursor_width) - 4 ;
-	}
-	else{
-		cursor_width = (width / column_max - 16) + 8;
-		x = (index % column_max * (cursor_width + 16)) - 4;
-	}
+	cursor_width = (width / column_max - 16) + 8;
+	x = (index % column_max * (cursor_width + 8)) - 4;
 
 	int y = index / column_max * 16 - oy;
 	SetCursorRect(Rect(x, y, cursor_width, 16));
--- a/src/window_skill.cpp
+++ b/src/window_skill.cpp
@@ -84,10 +84,7 @@ void Window_Skill::DrawItem(int index) {
 		bool enabled = CheckEnable(skill_id);
 		int color = !enabled ? Font::ColorDisabled : Font::ColorDefault;
 
-		std::stringstream ss;
-		ss << costs;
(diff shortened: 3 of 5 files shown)


fmatthew5876:
I have not looked at your changes yet, but when fixing text positioning be very careful about fonts. Our fonts are not 100% matching RPG_RT in all cases, and therefore the text can sometimes appear in the wrong position when it may not be.

rueter37:
Did the rebase as requested.

src/window_skill.cpp
```diff
@@ -86,8 +86,8 @@ void Window_Skill::DrawItem(int index) {
 
 		std::stringstream ss;
 		ss << costs;
-		contents->TextDraw(rect.x + rect.width - 28, rect.y, color, "-");
-		contents->TextDraw(rect.x + rect.width - 6, rect.y, color, ss.str(), Text::AlignRight);
+		contents->TextDraw(rect.x + rect.width - 24, rect.y, color, "-");
+		contents->TextDraw(rect.x + rect.width, rect.y, color, ss.str(), Text::AlignRight);
```
fmatthew5876, line 90: RPG_RT uses printf, we should do the same for compatibility and to cleanup the code.

Change this to a single TextDraw call with `fmt::format("-{:3d}", costs)`

src/window_item.cpp
```diff
@@ -126,8 +126,8 @@ void Window_Item::DrawItem(int index) {
 		std::stringstream ss;
 		ss << number;
 		Font::SystemColor color = enabled ? Font::ColorDefault : Font::ColorDisabled;
-		contents->TextDraw(rect.x + rect.width - 28, rect.y, color, "x");
-		contents->TextDraw(rect.x + rect.width - 6, rect.y, color, ss.str(), Text::AlignRight);
+		contents->TextDraw(rect.x + rect.width - 24, rect.y, color, "x");
+		contents->TextDraw(rect.x + rect.width, rect.y, color, ss.str(), Text::AlignRight);
```
fmatthew5876, line 130: All versions of RPG_RT I checked use `:` not `x`

Change this to a single TextDraw call with `fmt::format(":{:3d}", number)`

Ghabry: About which scene are we talking here?

Unbenannt

fmatthew5876: I tried 2k3 dynrpg, 2ke steam, and 2k3e steam

fmatthew5876:
For the skill and item window, RPG_RT initializes the following. This happens in @ DynRPG 004C63DC
```
x = 0
y = 32
height = 208
width = 320
cursorX = 8
cursorY = 8
jumpX = width / 2 = 160
jumpY = 16
cursorWidth = jumpX - 16 = 144
cursorHeight = 16
```

Then to calculate the cursor rect DynRPG @ 004C677C
```
left = cursorX + jump X * (index  % choicesPerRow) - 4;
top = cursorY + jumpY * (index / choicesPerRow - scrollOffset)
right = cursorWidth + left + 8
bottom  = cursorHeight + top
```

fmatthew5876:
@rueter37 Can you please make a new PR for the window selection stuff?

There we can figure those out and get them in, and then do the rest of this after.

* Fix skill cost
* Fix item cost
* Add gap in multi-column
* Endless scrolling

rueter37:
I'm sorry for being dumb but I don't quite understand. Which commits exactly should I move into an other PR?

PS: I will move the last two commits (the ones related to 2k3) into the scene-battle2k3-changes branch because they fit better there.

src/window_skill.cpp
rueter37: Changed.

src/window_item.cpp
rueter37: Changed.

Ghabry: We can argue a while about one char. For menu scene implementations we have some freedom: We can do enhancements. the marker for "times you have something" is in all languages I know the "X" and not a ":". Not 100% sure about Japanese. They have lots of counter kanji

rueter37: I personally prefer `x` over `:` but I came across a case where `x` looks ugly: If an item uses the full item string length then it looks like the `x` belongs to the item name. So I went for @fmatthew5876's suggestion to use `:` instead of `x`.

rueter37:
@fmatthew5876 I have readed your comment and changed this PR according to your suggestion. I hope I did that right.

src/window_item.cpp
fmatthew5876: Yet another example of a custom EasyRPG database chunk

src/window_selectable.cpp
```diff
@@ -133,15 +133,13 @@ void Window_Selectable::Update() {
...
-				(Input::IsTriggered(Input::DOWN) || Input::IsTriggered(Input::SCROLL_DOWN)))) {
+			if (index < item_max - column_max || column_max == 1) {
 				Game_System::SePlay(Game_System::GetSystemSE(Game_System::SFX_Cursor));
 				index = (index + column_max) % item_max;
 			}
 		}
 		if (Input::IsRepeated(Input::UP) || Input::IsTriggered(Input::SCROLL_UP)) {
-			if (index >= column_max || (column_max == 1 &&
-				(Input::IsTriggered(Input::UP) || Input::IsTriggered(Input::SCROLL_UP)))) {
+			if (index >= column_max || column_max == 1) {
```
fmatthew5876, line 142: For some reason, this feature doesn't work in the actor battle command window in 2k3 only. It works everywhere else I tried it.

That being said, this is an RPG_RT bug I don't care to emulate.

fmatthew5876 (approved):

fdelapena (approved):

src/window_selectable.cpp
```diff
@@ -64,11 +64,11 @@ int Window_Selectable::GetPageItemMax() {
 
 Rect Window_Selectable::GetItemRect(int index) {
 	Rect rect = Rect();
-	rect.width = (contents->GetWidth()) / column_max - 4;
+	rect.width = (contents->GetWidth()) / column_max - 8;
 	rect.height = 12;
 	rect.x = index % column_max * rect.width;
 	if (rect.x > 0){
-		rect.x += 8;
+		rect.x += 16;
```
fmatthew5876, line 71: Pardon me for reviewing this after lack of sleep. Took another look and this actually appears wrong.

If you take a screenshot of the item window and zoom in, you'll the item text in EasyRPG is to the right by 1 whole pixel.

Also this doesn't actually work well if the `column_max > 2`

rueter37: Fixed.

fmatthew5876 (approved):

fmatthew5876:
This looks good to me now, thanks!

Ghabry:
@fmatthew5876 
Is there any issue for tracking "Settings"?

Because that the cursor gets stuck at the top/bottom instead of endless scrolling is actually a feature (as you can see in the diff the endless scrolling was implemented by deleting code). It is convenient when the cursor becomes "stuck" because this makes selecting the first or last entry very simple: Just hold the arrow key.

fmatthew5876:
> @fmatthew5876
> Is there any issue for tracking "Settings"?
> 
> Because that the cursor gets stuck at the top/bottom instead of endless scrolling is actually a feature (as you can see in the diff the endless scrolling was implemented by deleting code). It is convenient when the cursor becomes "stuck" because this makes selecting the first or last entry very simple: Just hold the arrow key.

We can make the endless scroll a boolean flag in the window class, and conditionally disable the feature. Most of the time it's useful be to be able to endless scroll, but maybe not in all cases.

Ghabry (approved):

PR #2320, opened 2020-09-05
Refactor Battle Algorithm effect and hit rate calculations

fmatthew5876:
This PR moves battle algo calculation logic into it's own namespace as a set of standalone free functions with minimal dependencies.

Goals:
- [x] Migrate battle algorithm logic into separate free functions that can be called from multiple contexts such as: battle, menu, events, auto battle, and unit tests
- [x] Unit tests around all edge cases in battle algos
- [x] Allow implementation of auto battle in a follow up PR
- [x] Restructure the code to be able to have confidence and converge on algo correctness
- [x] Support weapon modes, so we can implement correct dual wield in 2k3
- [x] Eliminate `Main_Data::game_data.actors` #1954

This also implements 2k3 negative attributes using logic taken from RE. It implements calculating a negative effect, but does not yet pass it through the battle system.

TODO:
- [x] Write unit tests for all edge cases
- [x] Verify each unit test by testing in RPG_RT
- [x] Implement weapon modes for 2k3 dual wield
- [x] Verify differences between actor / monster normal attack in RPG_RT including row
- [x] Recheck for differences in menu vs battle skills
- [x] Verify behavior in 2k3 dynrpg
- [x] Verify behavior in 2ke
- [x] Verify behavior in 2k legacy
- [x] Verify behavior in 2k3 legacy

DynRPG Addresses:
------------------

| address | class | function | Comments |
| -- | -- | -- | -- |
| 004B71AC | Actor | GetBaseAtk | level + mod + all equipment |
| 004B7408 | Actor | GetBaseDef | level + mod + all equipment |
| 004B74B8 | Actor | GetBaseInt | level + mod + all equipment |
| 004B7568 | Actor | GetBaseAgi | level + mod + all equipment |
| 004BFB28 | Battler | GetAtk | base + battle_mod + states |
| 004BFBF4 | Battler | GetDef | base + battle_mod + states |
| 004BFCC0 | Battler | GetInt | base + battle_mod + states |
| 004BFD5C | Battler | GetAgi | base + battle_mod + states |
| 004B760C | Actor | GetBaseAgiForUsedWeapon | level + mod + used weapon + all armor + battle mod (not states ❗ ) |
| 004B7250 | Actor | GetBaseAtkForUsedWeapon | level + mod + used weapon + all armor + battle mode + additional clamp(1, 9999) ❗ |
| 004B9834 | SceneBattle | ApplyNormalAttack | Always used by monster, sometimes used by actor - has states inflict bug for actor attacks |
| 0049B048 | SceneBattle | ApplyNormalWeaponAttack | Used only for actors attack |
| 004C0B2C | Battler | CalcNormalAttackDamage | Always used by monster attack, sometimes used by actor attack. For actors, always uses combined atk and agi from both weapons. Used for actor auto battle ranking. |
| 004B9834 | Actor | CalcNormalWeaponAttackDamage | Used only for actors attack - has back attack row bug not present in other version |
| 0049C644 | SceneBattle | ApplySkillEffect | 
| 004A20D8 | SceneMenu | ApplySkillEffect |
| 004C0DBC | Battler | ComputeSkillPower | Used for actor auto battle ranking |
| 0049B7E4 | SceneBattle | ProcessActionSelfDestruct |

List of RPG_RT bugs discovered ❗ 
---------------------------------

* When the battle condition is "back" all actor attacks against enemies are treated as if they were in the back row.
* When actor attack uses `ApplyNormalAttack` path, state infliction percentages are taken as the max rate of both weapons, instead of the just the weapon which is performing the attack.
* When actor uses a normal attack with a weapon, the actor's AGI used to calculate the hit probability ignores states which can double or half AGI.
* In 2k3, RPG_RT will still check if `failure_message = 3` and run physical evasion checks. There is no way to set this flags anymore in 2k3 editor. What about skills ported from 2k?
* In physical skill evasion checks, 2k3 will not check for states with `avoid_attacks` or `row`.
* When calculating the rank for skill use in auto battle, RPG_RT highly values skills which can revive dead allies. However RPG_RT does not check `invert_state` flag, and so it would prefer to use a skill which inflicts death on the party.
* 2k3 protects against divide by zero in the normal attack hit rate agi calculation. 2k does not. However in both cases, the agi value used is already clamped to >= 1..

### DIFF
--- a/src/algo.cpp
+++ b/src/algo.cpp
@@ -28,7 +28,7 @@ namespace Algo {
 
 bool IsRowAdjusted(lcf::rpg::SaveActor::RowType row, lcf::rpg::System::BattleCondition cond, bool offense) {
 	return (cond == lcf::rpg::System::BattleCondition_surround
-			|| (row == (1 - offense)
+			|| (row != offense
 				&& (cond == lcf::rpg::System::BattleCondition_none || cond == lcf::rpg::System::BattleCondition_initiative))
 			|| (row == offense
 				&& (cond == lcf::rpg::System::BattleCondition_back))
@@ -103,15 +103,14 @@ int CalcNormalAttackToHit(const Game_Battler &source,
 int CalcSkillToHit(const Game_Battler& source, const Game_Battler& target, const lcf::rpg::Skill& skill) {
 	auto to_hit = skill.hit;
 
+	// RPG_RT BUG: rm2k3 editor doesn't let you set the failure message for skills, and so you can't make them physical type anymore.
+	// Despite that, RPG_RT still checks the flag and run the below code?
 	if (skill.failure_message != 3
 		   || (skill.scope != lcf::rpg::Skill::Scope_enemy && skill.scope != lcf::rpg::Skill::Scope_enemies)) {
 		return to_hit;
 	}
 
-	// RPG_RT BUG: rm2k3 editor doesn't let you set the failure message for skills, and so you can't make them physical type anymore.
-	// Despite that, RPG_RT still checks the flag and run the below code?
-	// FIXME: Verify if skills ported from 2k retain this flag and exercise the evasion logic in 2k3?
-	// RPG_RT BUG: RPG_RT does not check for "EvadesAllPhysicaAttacks() states here
+	// RPG_RT BUG: RPG_RT 2k3 does not check for "EvadesAllPhysicaAttacks() states here
 
 	// If target has Restriction "do_nothing", the attack always hits
 	if (!target.CanAct()) {
--- a/src/algo.cpp
+++ b/src/algo.cpp
@@ -19,8 +19,8 @@ bool IsRowAdjusted(const Game_Actor& actor, lcf::rpg::System::BattleCondition co
 		   );
 }
 
-int CalcNormalAttackToHit(const Game_Battler &source, const Game_Battler &target, lcf::rpg::System::BattleCondition cond) {
-	auto to_hit = source.GetHitChance();
+int CalcNormalAttackToHit(const Game_Battler &source, const Game_Battler &target, int weapon, lcf::rpg::System::BattleCondition cond) {
+	auto to_hit = source.GetHitChance(weapon);
 
 	// If target has rm2k3 state which grants 100% dodge.
 	if (target.EvadesAllPhysicalAttacks()) {
@@ -37,12 +37,12 @@ int CalcNormalAttackToHit(const Game_Battler &source, const Game_Battler &target
 
 	// Stop here if attacker ignores evasion.
 	if (source.GetType() == Game_Battler::Type_Ally
-		&& static_cast<const Game_Actor&>(source).AttackIgnoresEvasion()) {
+		&& static_cast<const Game_Actor&>(source).AttackIgnoresEvasion(weapon)) {
 		return to_hit;
 	}
 
 	// AGI adjustment.
-	to_hit = 100 - (100 - to_hit) * (1.0f + (float(target.GetAgi()) / float(source.GetAgi()) - 1.0f) / 2.0f) ;
+	to_hit = 100 - (100 - to_hit) * (1.0f + (float(target.GetAgi()) / float(source.GetAgi(weapon)) - 1.0f) / 2.0f) ;
 
 	// If target has physical dodge evasion:
 	if (target.GetType() == Game_Battler::Type_Ally
@@ -84,7 +84,7 @@ int CalcSkillToHit(const Game_Battler& source, const Game_Battler& target, const
 
 	// Stop here if attacker ignores evasion.
 	if (source.GetType() == Game_Battler::Type_Ally
-		&& static_cast<const Game_Actor&>(source).AttackIgnoresEvasion()) {
+		&& static_cast<const Game_Actor&>(source).AttackIgnoresEvasion(Game_Battler::kWeaponAll)) {
 		return to_hit;
 	}
 
@@ -100,9 +100,9 @@ int CalcSkillToHit(const Game_Battler& source, const Game_Battler& target, const
 	return to_hit;
 }
 
-int CalcCriticalHitChance(const Game_Battler& source, const Game_Battler& target) {
+int CalcCriticalHitChance(const Game_Battler& source, const Game_Battler& target, int weapon) {
 	// FIXME: Make this function return int 0 to 100.
-	auto crit_chance = static_cast<int>(source.GetCriticalHitChance() * 100.0);
+	auto crit_chance = static_cast<int>(source.GetCriticalHitChance(weapon) * 100.0);
 	if (target.GetType() == Game_Battler::Type_Ally && static_cast<const Game_Actor&>(target).PreventsCritical()) {
 		crit_chance = 0;
 	}
@@ -132,10 +132,11 @@ int AdjustDamageForDefend(int dmg, const Game_Battler& target) {
 
 int CalcNormalAttackEffect(const Game_Battler& source,
 		const Game_Battler& target,
+		int weapon,
 		bool is_critical_hit,
 		bool apply_variance,
 		lcf::rpg::System::BattleCondition cond) {
-	const auto atk = source.GetAtk();
+	const auto atk = source.GetAtk(weapon);
 	const auto def = target.GetDef();
 
 	// Base damage
@@ -149,12 +150,9 @@ int CalcNormalAttackEffect(const Game_Battler& source,
 	}
 
 	// Attacker weapon attribute adjustment
-	const lcf::rpg::Item* weapons[2] = { nullptr, nullptr };
 	if (source.GetType() == Game_Battler::Type_Ally) {
-		weapons[0] = static_cast<const Game_Actor&>(source).GetWeapon();
-		weapons[1] = static_cast<const Game_Actor&>(source).Get2ndWeapon();
+		Attribute::ApplyAttributeNormalAttackMultiplier(dmg, static_cast<const Game_Actor&>(source), target, weapon);
 	}
-	Attribute::ApplyAttributeMultiplier(dmg, weapons[0], weapons[1], target);
 
 	// Defender row adjustment
 	if (Player::IsRPG2k3()) {
@@ -202,7 +200,7 @@ int CalcSkillEffect(const Game_Battler& source,
 
 	effect = std::max(0, effect);
 
-	effect = Attribute::ApplyAttributeMultiplier(effect, skill, target);
+	effect = Attribute::ApplyAttributeSkillMultiplier(effect, target, skill);
 
 	// FIXME: RPG_RT 2k3 doesn't apply variance if negative attribute flips damage
 	if (apply_variance && (effect > 0 || Player::IsLegacy())) {
--- a/src/algo.cpp
+++ b/src/algo.cpp
@@ -1,3 +1,19 @@
+/*
+ * This file is part of EasyRPG Player.
+ *
+ * EasyRPG Player is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * EasyRPG Player is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
+ */
 #include "algo.h"
 #include "game_battler.h"
 #include "game_actor.h"
--- a/src/algo.cpp
+++ b/src/algo.cpp
@@ -42,6 +42,8 @@ int CalcNormalAttackToHit(const Game_Battler &source, const Game_Battler &target
 	}
 
 	// AGI adjustment.
+	// NOTE: RPG_RT 2k3 has a bug where if the source is an actor with a selected weapon, the agi
+	// calculation calls a bespoke function which doesn't consider states which can cause half or double AGI.
 	to_hit = 100 - (100 - to_hit) * (1.0f + (float(target.GetAgi()) / float(source.GetAgi(weapon)) - 1.0f) / 2.0f) ;
 
 	// If target has physical dodge evasion:
--- a/src/algo.cpp
+++ b/src/algo.cpp
@@ -19,6 +19,15 @@ bool IsRowAdjusted(const Game_Actor& actor, lcf::rpg::System::BattleCondition co
 		   );
 }
 
+static int CalcToHitAgiAdjustment(int to_hit, const Game_Battler& source, const Game_Battler& target, int weapon) {
+	// NOTE: RPG_RT 2k3 has a bug where if the source is an actor with a selected weapon, the agi
+	// calculation calls a bespoke function which doesn't consider states which can cause half or double AGI.
+	const float src_agi = std::max(1, source.GetAgi(weapon));
+	const float tgt_agi = target.GetAgi();
+
+	return 100 - (100 - to_hit) * (1.0f + (tgt_agi / src_agi - 1.0f) / 2.0f);
+}
+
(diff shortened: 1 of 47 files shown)


fmatthew5876:
This is now ready for review and testing.

I've written a comprehensive unit test suite for everything here.

This PR needs to be tested, in particular:
- [ ] Do each `tests/algo` unit test edge case in RPG_RT and verify the numbers match
- [ ] Do some basic testing of games

If any of you guys have time,  I could really use some help testing this stuff. Once we get this locked down, our core battle algorithm numbers will be correct and stable forever.

Ghabry:
For running the tests you must use the check target:

```
cmake --build . --target check (Cmake)

make check (autotools or cmake makefile generator)
```

rueter37:
Perfect, thank you! Using `--target check` worked for me.

fdelapena (approved):

rueter37:
Your test cases look good for me. I noticed two things:
- In the "SelfDestructEffect" test case, there are two tests called "150_0" and "150_0_var" which is confusing because the defender has 100 defense in the tests. It looks like the tests are supposed to be called "150_100" and "150_100_var".
- In the "NormalAttackEffect" test case, aren't the RPG Maker 2003 surround battle condition tests where the actor attacks an enemy supposed to do 35 instead of 47 damage? Because I know from RPG_RT testing with the surround battle condition if an actor attacks an enemy he receives the 25% damage bonus and the 25% damage penalty which means the damage gets reduced by 6.25% which means only 35 damage. 47 damage is with the bonus applied without penalty.

fmatthew5876:
> * In the "SelfDestructEffect" test case, there are two tests called "150_0" and "150_0_var" which is confusing because the defender has 100 defense in the tests. It looks like the tests are supposed to be called "150_100" and "150_100_var".

Thank you, thats a bug in the test name I will fix

> * In the "NormalAttackEffect" test case, aren't the RPG Maker 2003 surround battle condition tests where the actor attacks an enemy supposed to do 35 instead of 47 damage? Because I know from RPG_RT testing with the surround battle condition if an actor attacks an enemy he receives the 25% damage bonus and the 25% damage penalty which means the damage gets reduced by 6.25% which means only 35 damage. 47 damage is with the bonus applied without penalty.

Nice catch! This is actually another manifestation of the "2k3 back attack bug" which we should call "2k3 row bug". I'll investigate further and fix.

fmatthew5876:
Added a "fix" for the row "bug"

Also refactored a bit to improve the code and added to all the tests.

fmatthew5876:
I checked the battle algorithms from this PR against 2k legacy, 2k updated, 2ke, 2k3 legacy, and 2k3 dynrpg.

I discovered the battle adjusted stats can go up to 9999 in all versions of 2k. Other than that, everything appears the same except the legacy variance adjustment discovered by @rueter37 and already implemented here.

rueter37 (approved):

fmatthew5876:
I verified the unit test cases in RPG_RT. I did normal attacks with the various row configurations and use a debugger to get the damage value before variant.

I tested 0 variance skills. 

The numbers match up.

So this PR is done.

Ghabry:
Minor: You forgot our shiny header in some new files. (does not apply to files in tests/ where we always omit it).

Won't be able to look through this before the next weekend. This is just tooooo much and git diffs is of no help because most of the changes are "take everything from battle algorithm and paste it with changes somewhere else". :(

I reaaaaaaaaaaaaaaally hope that there will be never any reason to change the Api ever again in the classes where you added 3000 lines of unit tests. When somebody touches them they have to modify at least 100 unit tests. (or give up out of frustration)

fmatthew5876:
> I reaaaaaaaaaaaaaaally hope that there will be never any reason to change the Api ever again in the classes where you added 3000 lines of unit tests. When somebody touches them they have to modify at least 100 unit tests. (or give up out of frustration)

Well that's kind of the point. If you want to refactor the API, you have to make sure your code change passes all the tests. This gives us more confidence in refactors than any manual testing can.

That being said, I don't see too much refactoring in the future for these interfaces

src/attribute.h
```diff
@@ -0,0 +1,82 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef EP_ATTRIBUTE_H
+#define EP_ATTRIBUTE_H
+
+#include <lcf/rpg/fwd.h>
+#include <lcf/dbbitarray.h>
+#include <span.h>
```
Ghabry: > Add Attribute namespace for attribute algos 

shouldnt this be "span.h"?

src/attribute.cpp
```diff
@@ -0,0 +1,122 @@
...
+	if (n == 0) {
+		return effect;
+	}
+
+	return ApplyAttributeMultiplier(effect, Span<const lcf::DBBitArray*>(attribute_sets.data(), n), target);
+}
+
+int ApplyAttributeMultiplier(int effect, const lcf::rpg::Skill& skill, const Game_Battler& target) {
+	std::array<const lcf::DBBitArray*, 1> attribute_sets = { &skill.attribute_effects };
+	return ApplyAttributeMultiplier(effect, MakeSpan(attribute_sets), target);
```
Ghabry: > Add Attribute namespace for attribute algos 

Are spans out of initializer lists not allowed?

src/algo.cpp
```diff
@@ -0,0 +1,230 @@
...
+#include "player.h"
+#include <lcf/rpg/skill.h>
+
+#include <algorithm>
+
+namespace Algo {
+
+bool IsRowAdjusted(const Game_Actor& actor, lcf::rpg::System::BattleCondition cond, bool offense) {
+	return (cond == lcf::rpg::System::BattleCondition_surround
+			|| (actor.GetBattleRow() == (1 - offense)
```
Ghabry: > Use Algo library and port normal attack calculations to it 

``(1 - offense)``. That's a bool, how about ``actor.GetBattleRow() != offense``? Or whatever the intention is (have unit tests ;)). Is hard to read.

src/algo.cpp
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	return to_hit;
+}
+
+
+int CalcSkillToHit(const Game_Battler& source, const Game_Battler& target, const lcf::rpg::Skill& skill) {
+	auto to_hit = skill.hit;
+
+	if (skill.failure_message != 3
```
Ghabry, line 106: > Use Algo library and port normal attack calculations to it

Could you flag this as a RPG2k3 bug? (the skill message can't be modified)

src/algo.cpp
```diff
@@ -0,0 +1,230 @@
...
+	if (target.GetType() == Game_Battler::Type_Ally
+			&& static_cast<const Game_Actor&>(target).HasPhysicalEvasionUp()) {
+		to_hit -= 25;
+	}
+
+	return to_hit;
+}
+
+int CalcCriticalHitChance(const Game_Battler& source, const Game_Battler& target) {
+	// FIXME: Make this function return int 0 to 100.
```
Ghabry: > Use Algo library and port normal attack calculations to it

Can you explain this fixme?

src/algo.h
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * @return Success hit rate
+ */
+int CalcNormalAttackToHit(const Game_Battler& source, const Game_Battler& target, lcf::rpg::System::BattleCondition cond);
+
+/**
+ * Compute the hit rate for a skill
+ *
+ * @param source The source of the action
+ * @param target The target of the action
```
Ghabry, line 88: ``@param skill`` missing

src/game_actor.cpp
```diff
@@ -388,7 +388,42 @@ int Game_Actor::GetBaseMaxSp() const {
...
+		if (item->type == lcf::rpg::Item::Type_weapon) {
+			if (!allow_weapon || (weapon != Game_Battler::kWeaponAll && weapon != slot)) {
+				continue;
+			}
+		} else if (IsArmorType(item)) {
+			if (!allow_armor) {
+				continue;
+			}
+		} else {
+			continue;
```
Ghabry, line 434: > Allow computing stats as if single weapon equipped 

Is it possible to have another type of equipment? Maybe ``assert`` here?

src/game_actor.cpp
```diff
@@ -1319,77 +1319,52 @@ const lcf::rpg::Item* Game_Actor::GetAccessory() const {
 	return nullptr;
 }
 
-bool Game_Actor::HasPreemptiveAttack() const {
-	auto* w1 = GetWeapon();
-	auto* w2 = Get2ndWeapon();
-	return (w1 && w1->preemptive) || (w2 && w2->preemptive);
+bool Game_Actor::HasPreemptiveAttack(int weapon) const {
+	bool rc = false;
+	ForEachEquipment<true, false>(GetWholeEquipment(), [&](auto& item) { rc |= item.preemptive; }, weapon);
```
Ghabry, line 1314: > Allow computing stats as if single weapon equipped 

What does ``rc`` mean?

src/game_actor.h
```diff
@@ -827,8 +828,8 @@ class Game_Actor final : public Game_Battler {
 
 	int GetBattleAnimationId() const override;
 
-	int GetHitChance() const override;
-	float GetCriticalHitChance() const override;
+	int GetHitChance(int weapon = kWeaponAll) const override;
+	float GetCriticalHitChance(int weapon = kWeaponAll) const override;
```
Ghabry: > Allow computing stats as if single weapon equipped

Sorry, but for all of them the doc comment was not updated.

From ``GetBaseAtk`` to ``AttackIgnoresEvasion``

Also in ``Game_battler.h`` and ``game_enemy.h``

src/game_battler.cpp
```diff
@@ -478,84 +478,46 @@ static int AffectParameter(const int type, const int val) {
...
-	int base_atk = GetBaseAtk(weapon);
-	int n = Utils::Clamp(base_atk, 1, MaxStatBaseValue());
-
-	for (int16_t i : GetInflictedStates()) {
-		// States are guaranteed to be valid
-		const lcf::rpg::State& state = *lcf::ReaderUtil::GetElement(lcf::Data::states, i);
-		if (state.affect_attack) {
-			n = AffectParameter(state.affect_type, base_atk);
-			break;
+static int AdjustParam(int base, int mod, int maxval, Span<const int16_t> states, bool lcf::rpg::State::*adj) {
```
Ghabry, line 472: > Refactor GetAtk() etc.. 

What is this ``bool lcf::rpg::State::*adj``? Looks like a Pointer to ``State::``... what?

src/game_battler.cpp
```diff
@@ -478,84 +478,46 @@ static int AffectParameter(const int type, const int val) {
...
+	auto value = Utils::Clamp(base + mod, 1, maxval);
+	bool half = false;
+	bool dbl = false;
+	for (auto i: states) {
+		const auto* state = lcf::ReaderUtil::GetElement(lcf::Data::states, i);
+		assert(state);
+		if (state->*adj) {
```
Ghabry, line 479: > Refactor GetAtk() etc..

Or here. ``state`` has no field ``adj``. I'm super confused.

src/game_actors.cpp
```diff
@@ -32,26 +32,24 @@ Game_Actors::Game_Actors() {
 Game_Actors::~Game_Actors() {
 }
 
-void Game_Actors::Fixup() {
+void Game_Actors::SetSaveData(std::vector<lcf::rpg::SaveActor> save) {
```
Ghabry, line 35: > Remove Main_Data::game_data.actors 

Intential copy?

src/game_battler.h
```diff
@@ -898,4 +906,8 @@ inline void Game_Battler::SetDirectionFlipped(bool flip) {
 	direction_flipped = flip;
 }
 
+inline int Game_Battler::GetBaseAttributeRate(int attribute_id) const {
+	return 2;
```
Ghabry, line 941: > Refactor Battler attributes 

Does this have a use? It is overoaded for both actor and enemy.
Also needs a ``// C - default`` comment

Ghabry (changes requested):
.

src/algo.cpp
Ghabry: I think this doesn't apply anymore. Saw this in a later commit commented.

src/attribute.cpp
```diff
@@ -0,0 +1,149 @@
...
+	return effect;
+}
+
+int ApplyAttributeNormalAttackMultiplier(int effect, const Game_Battler& source_battler, const Game_Battler& target, Game_Battler::Weapon weapon) {
+	if (source_battler.GetType() != Game_Battler::Type_Ally) {
+		return effect;
+	}
+	auto& source = static_cast<const Game_Actor&>(source_battler);
+
+	std::array<const lcf::DBBitArray*, 2> attribute_sets;
```
Ghabry, line 125: Clang tidy complaints that you access uninitialized memory. Though this looks safe (maybe initialize it to make it happy?).

src/algo.cpp
fmatthew5876: I got rid of it later

src/algo.cpp
fmatthew5876: It was added later, but it should go above here to make it easier to read.

src/game_actors.cpp
```diff
@@ -22,59 +22,50 @@
...
+	for (size_t i = 0; i < data.size(); i++) {
+		data[i] = Game_Actor(i + 1);
+	}
 }
 
-void Game_Actors::Init() {
-	data.resize(lcf::Data::actors.size());
-	for (size_t i = 1; i <= data.size(); i++)
-		GetActor(i)->Init();
+Game_Actors::~Game_Actors() {
```
Ghabry, line 32: Is now empty. Can be defaulted?

Ghabry:
Your refactoring finally solved the "Invalid actor" warnings on shutdown printed into the logfile :+1: .

src/game_actor.cpp
fmatthew5876: return code

src/algo.h
```diff
@@ -0,0 +1,173 @@
...
+ * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef EP_ALGO_H
+#define EP_ALGO_H
+
+#include <lcf/rpg/fwd.h>
+#include <lcf/rpg/system.h>
+#include <lcf/rpg/saveactor.h>
+#include <game_battler.h>
```
Ghabry, line 24: ``<>`` -> ``""``

src/game_battler.cpp
fmatthew5876: It's a pointer to data member, points to `State::affect_atack` or `State::affect_defense` etc.. Essentially lets me write one function for all 4 parameters without needing templates.

src/game_actors.cpp
fmatthew5876: Yes, internally we move the `SaveActor` objects.

The new model for loading save data will be that you move it.

src/game_battler.h
fmatthew5876: Removed this function

src/attribute.cpp
fmatthew5876: Added initializer

src/game_actor.h
fmatthew5876: Added

src/game_actors.cpp
fmatthew5876: Removed

src/algo.h
fmatthew5876: Fixed

Ghabry:
There are now virtual function calls in the constructors:
Afaik this is not allowed? But somehow it works (last time I did this in MSVC I got a "pure virtual called"?! Was this changed in a later standard?

Actor:
```cpp
if (GetLevel() > 0) {
	SetHp(GetMaxHp()); <- this
	SetSp(GetMaxSp()); <- this
	SetExp(exp_list[GetLevel() - 1]);
}
```

Enemy
```cpp
hp = GetMaxHp();
sp = GetMaxSp();
```

Game_Battler:
```
ResetBattle();
```

Ghabry:
Also non-standard it seems: (but I let you decide)

On ``GetBaseAtk`` etc. "Clang-Tidy: Default arguments on virtual or override methods are prohibited"

Issue #2219, opened 2020-05-15
Add configurable engine features


fmatthew5876:
Instead of querying the engine version, we can create a giant feature struct with flags for the feature differences between engines. 

Examples:
* has_unlocked_picture_patch
* dual_wield_behavior - combined hit (2k), multiple hits (2k3)
* supports_reversed_states
* supports_bottom_transparency
* key_input_proc_style
* battle_system_type - 2k, 2k3
* max_actor_hp - 999 (2k), 9999 (2k3), custom value
* max_variable_value

etc..

This could be used to implement features from third party patches. It would also allow game developers to customize these for EasyRPG games.

I would add some macro and/or template magic to transform these features into compile time values. This would allow custom games to hard code their behavior and not pay any runtime costs for unsupported behaviors.

- [ ] Implement full feature set
- [ ] Add helper functions to generate feature structs from all mainstream RPG_RT versions.
- [ ] Add support for third party patch detection to enable patch features
- [ ] Add support for compile time feature set for performance

rueter37:
Because this issue is about adding features, would it make sense to add some of the following features in the LCF database (I implemented them myself, you can check it out in my [experimental player branch](https://github.com/rueter37/player/tree/experimental) and [experimental liblcf branch](https://github.com/rueter37/liblcf/tree/experimental)):

- BattleCommands option 200: Default ATB mode
- BattleCommands option 201: Enable in-battle row command
- BattleCommands option 202: Sequential order battling
- Skill option 200: Ignore reflect state
- Item option 200: Maximum item count in inventory
- System option 200: Alternative EXP curve
- System option 201: Use RPG Maker 2003 menu
- System option 202: Enable auto battle
- System option 203: Maximum actor HP value
- System option 204: Maximum enemy HP value
- System option 205: Maximum damage value
- System option 206: Maximum EXP value
- System option 207: Maximum level
- System option 208: Fix double negative effect bug
- System option 209: Enable stat absorbing
- System option 210: Enable certain RPG Maker 2003 features
- System option 211: Maximum number of savegame slots
- System option 212: Default maximum item count in inventory

carstene1ns:
A part of this has been implemented here: https://github.com/EasyRPG/Player/pull/2697/commits/d6584cb52e79a0e4cfa5259fbfed672f0d02c99d

Issue #680, opened 2015-12-18
Embric of Wulfhammer's Castle: "Row" battle command is missing


Siltala:
With bug #601 having been fixed, most of the commands are now correctly shown and usable in git. However, one of the commands, 'Row' is still missing.

Steps to reproduce:
- Download game ( http://aeresland.net/EmbricDELUXE_2.2.2.7z )
- Download my savegame from http://www.heavylobster.com/files/Save03.lsd
- Take one step north to enter combat
- Select 'strategize'. The 'Row' command, which is present on RPG_RT, is missing from the list. Note that even on RPG_RT, you need to scroll down in the list to actually see it.

Ghabry:
Row is a special command that everybody has. Is not implemented yet because i never found it useful.

Ghabry:
Asked cherry about the "Row" command but he hasn't figured this out yet.
So feel free to suggest what the "Row" command shall do :D

carstene1ns:
Just leaving these two links here:
https://rpgmaker.net/articles/371/
https://forums.rpgmakerweb.com/index.php?threads/do-you-remember-the-row-thing.20014/

What I remember/think of it:
- Basically you have front and back row.
- Not all actors can be in the front row, because it would be back row then ;-)
- So back row is default
- Front row "protects" back row, i.e. monsters target first row with higher chance (60/40 or 70/30 ?)
- speculation: some things work better depending on the row
  - from back row: recovering HP?, preparing a special attack?
  - from front row: higher chance of critical hit?

Ghabry:
I remember one non RPG Maker game that used a row system (Pier solar, don't play it, didn't age well) and in this game the front-row completely protected the back row from normal attacks.

The 60/40 makes sense, maybe 70/30 would be better, otherwise your healer is still hitted to often :D.

Ghabry:
If I understood this correctly the row does the following:
Physical attacks get a 25% bonus (or you receive 25% extra damage when):
- Battle Type is Normal or Initative and you are in the front row
- Battle Type is Back Attack, row doesn't matter
- Battle Type is Sourround attack and row is back (???, probably a bug because you can't change the row in Sourround and Pincer type battles)

Pincer is always 0%.

So this is mega boring.

rueter37:
I can confirm that the damage from physical attacks is increased by 25% if the source and the target are in the front row. From my tests with RPG_RT it looks like that being in the back row doesn't reduce the probability being targeted by an enemy.

[...]

PR #2321, opened 2020-09-05
Battle 2k3: Implement battle row damage and accuracy modifiers

rueter37:
This PR implements the damage and accuracy modifiers for normal attacks which depend on the battle condition and the row of the actors in RPG Maker 2003.

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -75,6 +75,23 @@ static inline int ToHitPhysical(Game_Battler *source, Game_Battler *target, int 
 		to_hit -= 25;
 	}
 
+	// Row adjustment (RPG2k3 only)
+	if (Player::IsRPG2k3()) {
+		if (Game_Battle::GetBattleCondition() == lcf::rpg::System::BattleCondition_none || Game_Battle::GetBattleCondition() == lcf::rpg::System::BattleCondition_initiative) {
+			if (target->GetType() == Game_Battler::Type_Ally && static_cast<Game_Actor*>(target)->GetBattleRow() == Game_Actor::RowType::RowType_back) {
+				to_hit -= 25;
+			}
+		}
+		if (Game_Battle::GetBattleCondition() == lcf::rpg::System::BattleCondition_back) {
+			if (target->GetType() == Game_Battler::Type_Enemy || (target->GetType() == Game_Battler::Type_Ally && static_cast<Game_Actor*>(target)->GetBattleRow() == Game_Actor::RowType::RowType_front)) {
+				to_hit -= 25;
+			}
+		}
+		if (Game_Battle::GetBattleCondition() == lcf::rpg::System::BattleCondition_surround) {
+			to_hit -= 25;
+		}
+	}
+
 	return to_hit;
 }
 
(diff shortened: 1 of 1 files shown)


src/game_battlealgorithm.cpp
```diff
@@ -987,10 +1004,42 @@ bool Game_BattleAlgorithm::Normal::Execute() {
...
 		int effect = (source->GetAtk() / 2 - GetTarget()->GetDef() / 4);
 
+		// Row attacker adjustments (RPG2k3 only)
+		if (Player::IsRPG2k3()) {
+			if (Game_Battle::GetBattleCondition() == lcf::rpg::System::BattleCondition_none || Game_Battle::GetBattleCondition() == lcf::rpg::System::BattleCondition_initiative) {
+				if (source->GetType() == Game_Battler::Type_Ally && static_cast<Game_Actor*>(source)->GetBattleRow() == Game_Actor::RowType::RowType_front) {
+					effect = 125 * effect / 100;
+				}
+			}
+			if (Game_Battle::GetBattleCondition() == lcf::rpg::System::BattleCondition_surround) {
```
fmatthew5876, line 1014: You forgot the back/pincer  back row case

rueter37: This is because the pincer battle condition doesn't do any changes at all. More details about that in my comments in #680.

fmatthew5876: As we discussed, you're right about pincer. But this is still missing back attack case.

rueter37: Whoops, after doing some retests you are right about the back attack case. I will fix this in the next commit.

rueter37:
Added a commit which fixes the handling of back battle conditions.

fdelapena (approved):

src/game_battlealgorithm.cpp
```diff
@@ -1031,7 +1036,7 @@ bool Game_BattleAlgorithm::Normal::Execute() {
 				}
 			}
 			if (Game_Battle::GetBattleCondition() == lcf::rpg::System::BattleCondition_back) {
-				if (source->GetType() == Game_Battler::Type_Ally && static_cast<Game_Actor*>(source)->GetBattleRow() == Game_Actor::RowType::RowType_front || target->GetType() == Game_Battler::Type_Ally && static_cast<Game_Actor*>(target)->GetBattleRow() == Game_Actor::RowType::RowType_front) {
+				if (target->GetType() == Game_Battler::Type_Enemy || target->GetType() == Game_Battler::Type_Ally && static_cast<Game_Actor*>(target)->GetBattleRow() == Game_Actor::RowType::RowType_front) {
```
fmatthew5876, line 1039: This `Type_Enemy` thing looks like a bug? Also lets smash all this into one commit so it's easier to review

rueter37: The back attack condition is a special case. If the player attacks an enemy the damage is always reduced by 25% regardless of row. If the player is in the front row he gets the 25% damage bonus but it still means the the damage is reduced by 6.25%. That were the results from my tests with RPG_RT.

rueter37:
Did the squash as requested.

src/game_battlealgorithm.cpp
fmatthew5876: You're right. I checked again.

As mentioned in the issue, there are 2 normal attack code paths. One appears to be taken for normal, initiative, and surround. The other for back and pincer.

The first code path only does row adjustements on the attacker / defender if it's an actor. The second one does not.

Enemies are always in the front row. So for back you always miss more often and hit for less. I need to verify this more, it looks like RPG_RT bugs because this behavior does not make a lot of sense.

Also we need to verify when actors attack other attacks via confuse status.

rueter37:
For handling the confused status a change in the back attack condition was necessary. I assume now you want everything in one commit so I amended the commit and force-pushed the changes.

src/game_battlealgorithm.cpp
rueter37: I have tested the confused status cases and in this cases the modifiers are applied as well. I only had to check in back attack condition if source and target is an ally because the 25% damage bonus is not applied if an ally attacks an other ally or himself due to confusion.

fmatthew5876: Also lets put parens around the second condition to make the logic more clear

`target->GetType() == Game_Battler::Type_Enemy || (target->GetType() == Game_Battler::Type_Ally && static_cast<Game_Actor*>(target)->GetBattleRow() == Game_Actor::RowType::RowType_front)`

rueter37:
Sorry, my bad. Had to force-push it again because I forgot the add the changed file before the commit amending.

src/game_battlealgorithm.cpp
rueter37: Parenthesis have been added.

rueter37:
And once again a force push because I forgot the second pair of parenthesis... I really must stop that because permanent force pushes are considered bad practice.

fmatthew5876 (approved):
This looks good to me now. I'm working on a comprehensive refactor of Normal Attack to support multiple weapons as well as auto battle in #2320 

Once we have that, we should do comprehensive testing on the new code end to end to try to achieve perfection

fmatthew5876:
Open question:

This back row thing is definitely an RPG_RT bug in my opinion. Should we implement it to be compatible with any games maybe relying on this to tune difficulty? or fix it?


In `SceneBattle::ApplyNormalAttack` all 3 row logics (target evasion, source atk bonus, target def bonus) are surrounded by a query that checks whether the affected source or target battler is an actor.

In `SceneBattle::ApplyWeaponNormalAttack` the row logic is the same, but the actor checks are missing from all of them.

The `IsActor()` check is a virtual function call. My hypothesis is someone removed the actor checks (maybe for speed?) because the second function is only used when the source is already an actor. I think they removed this but accidentally forgot that the target can still be a monster.

Outside of that, the logic also doesn't make any sense.

rueter37:
I think we should go for the RPG_RT route regarding the back attack condition for compatibility sake. But your suggestion to fix this behavior would be a nice optional feature for #2219.

fmatthew5876:
Considering that some games may have tuned their difficulty mechanics to account for this behavior, I have to agree we should keep it, at least for now.

Ghabry:
Is the RPG_RT compatible nonsense-back-attack implemented or not?

Is Back Attack the same as Surprise Attack (thats the official English term)? https://twitter.com/EasyRPG/status/1282987664636088320/photo/1

btw @rueter37 no need to apologize for force pushes. They are only bad style when more than one person works on the branch.

fmatthew5876:
I don't like the 2k3 terminology, it's confusing.

none/init = default
back = surprise - this PR has RPG_RT compatible bug
surround = preemptive
pincers = pincers

rueter37:
I went for the RPG_RT compatible back attack condition.

fmatthew5876:
I noticed here that the `back attack bug` should only be triggered when the source is an actor and the target is an enemy. If enemy attacks enemy in back attack, the bug should not occur.

This is already fixed in #2320 and is a super rare edge case, so we can wait for that to get it addressed.

Issue #1486, opened 2018-11-06
Collection of RPG_RT bugs for compatibility


fmatthew5876:
There are various RPG_RT bugs out there which may be depended on by different games. I've created this issue to catalog them. At some point we may want to enable a compatibility mode switch for some of these bugs if we find games need them.

Please list your known RPG_RT bugs here:

* Break statements in nested loops don't obey scope. They always jump to the next `EndLoop`
* Medicine usability is ignored in rm2k3 battles.

CherryDT:
I have a huge list.

Maybe I should do what I wanted to do a while ago while discussing how to approach any possible future fixes: Open a GitHub repo for RM2k(3) but for issues only, similar to https://github.com/Microsoft/wsl

Thing is just that it's a huge amount of work.

Ghabry:
iirc the "Target select" window (where you select a party member to use a item/skill on in the menu) ignores the "Half SP cost" attribute and shows the full SP cost but when executing the value is correct (halved).

CherryDT:
> Flash saturation and Flash period (?) are not saved in the savegame.
> [Player/src/game_screen.h](https://github.com/EasyRPG/Player/blob/1d5aedccce4b2fbf0328bbb264b4631108cb9597/src/game_screen.h#L101)

Are you sure? It should be saved as SaveScreen.flash_current_level and SaveScreen.flash_time_left

Ghabry:
> Are you sure

No Im not, just did a code search.

This bug here: #1391 
Negative E values which allows "enemy attacks can heal" is broken across all RPG Maker versions.

fmatthew5876:
> 
> 
> > Flash saturation and Flash period (?) are not saved in the savegame.
> > [Player/src/game_screen.h](https://github.com/EasyRPG/Player/blob/1d5aedccce4b2fbf0328bbb264b4631108cb9597/src/game_screen.h#L101)
> 
> Are you sure? It should be saved as SaveScreen.flash_current_level and SaveScreen.flash_time_left

The easyrpg variables he refers to are for continuous flashing. You need copies of these 2 fields to tell the flasher how to repeat the flash. #1501

fmatthew5876:
Rm2k3 doesn't let you put all party members in the back using the Row command. At least 1 member must be in the front.

However, if you put some party members in the back row and then remove the front row members, you'll end up with a party with everyone in the back.

Ghabry:
[...]

PR #2322, opened 2020-09-06
Battle 2k3: Implement in-battle row command

### LINKED ISSUE #680: Embric of Wulfhammer's Castle: "Row" battle command is missing

rueter37:
This PR adds the in-battle row command used in RPG Maker 2003. Merging this PR and #2321 into master fixes #680.

### DIFF
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -526,6 +526,7 @@ void Scene_Battle_Rpg2k3::CreateBattleCommandWindow() {
 			}
 			++i;
 		}
+		commands.push_back(ToString(lcf::Data::terms.row));
 	}
 
 	command_window.reset(new Window_Command(commands, option_command_mov));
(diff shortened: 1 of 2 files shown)


src/scene_battle_rpg2k3.cpp
```diff
@@ -1277,6 +1284,30 @@ void Scene_Battle_Rpg2k3::SpecialSelected() {
...
+void Scene_Battle_Rpg2k3::RowSelected() {
+	// Switching rows is only possible if in back row or
+	// if at least 2 party members are in front row
+	int current_row = active_actor->GetBattleRow();
+	int front_row_battlers = 0;
+	if (current_row == Game_Actor::RowType::RowType_front) {
+		for (auto& actor: Main_Data::game_party->GetActors()) {
+			if (actor->GetBattleRow() == Game_Actor::RowType::RowType_front) front_row_battlers++;
+		}
+	}
```
fmatthew5876, line 1296: This case doesn't appear to handle back attacks etc..

I checked in RE tools and found that RPG_RT checks `row == mirrored` to mean front `row != mirrored` where you say back.

fmatthew5876:
For bonus points, are you bug compatible with this?

https://github.com/EasyRPG/Player/issues/1486#issuecomment-441373701

😄

rueter37:
Okay, that thing mentioned in [#1486](https://github.com/EasyRPG/Player/issues/1486#issuecomment-441373701) is a really crazy one. I have tried to reproduce that bug in RPG_RT without avail. Probably I'm missing something obvious. But do we want to emulate that bug? I'm pretty sure the implementation in this PR doesn't.

src/scene_battle_rpg2k3.cpp
rueter37: Did the change as requested and tested it. I have amended the commit and force pushed it because I assume you want this in one commit.

fmatthew5876:
I don't care about the bug it's dumb and better we fix it. 

I suspect it's because you reset the actor position after processing row command. RPG_RT just adds or subtracts 24 from x.

rueter37:
Then we are all set!

fmatthew5876 (approved):

src/scene_battle_rpg2k3.cpp
```diff
@@ -1461,8 +1492,14 @@ void Scene_Battle_Rpg2k3::ActionSelectedCallback(Game_Battler* for_battler) {
...
 	if (for_battler->GetType() == Game_Battler::Type_Ally) {
-		const lcf::rpg::BattleCommand* command = static_cast<Game_Actor*>(for_battler)->GetBattleCommands()[command_window->GetIndex()];
-		for_battler->SetLastBattleAction(command->ID);
+		int index = command_window->GetIndex();
+		// Row command always uses the last index
+		if (index < command_window->GetRowMax() - 1) {
+			const lcf::rpg::BattleCommand* command = static_cast<Game_Actor*>(for_battler)->GetBattleCommands()[index];
+			for_battler->SetLastBattleAction(command->ID);
+		} else {
+			for_battler->SetLastBattleAction(-1);
```
Ghabry, line 1501: This must be verified because the last battle action is a battle page start condition (or was it a condition in a conditional branch?). So this value must match.
Even if the row command can't be picked there through the editor I bet there is at least one game that patched the value and checks for it (see Change Vehicle position with -1)

Ghabry: -1 is indeed the Row command, just tested this via LDB-patching :thinking:

fmatthew5876: I'm not sure it's actually -1. The logic I've seen in RPG_RT is usually something like this:

```
auto* batcmd = GetBattleCommandByIndex(window->currentchoice);
if (batcmd) {
  switch(batcmd->type) {
    // Handle each command
  }
} else {
  // Handle row command
}
```

What if you try -2 or a large value greater than the number of configured battle commands?

Ghabry: I only tested -1 and -2. -2 didn't activate my event page.

fdelapena (approved):

src/scene_battle_rpg2k3.cpp
Ghabry: Figured it out by now: That's (kinda) a bug. Only kinda because this needs external tools to be utilized:

The ``GetBattleCommandByIndex`` function checks if the cmd idx is ``>= 6`` and if yes it returns ``-1``.
Otherwise it resolves the battle command of the actor. For the row this is ``0``.

Conclusion:
When an actor has 6 battle commands then row has an action ID of -1, otherwise it has an action ID of 0.

@rueter37 can you implement this behaviour? (``SetLastBattleAction(-1)`` for "6 cmds``, else ``SetLastBattleAction(-1)``)

rueter37: Added the LastBattleAction changes.

PR #2323, opened 2020-09-06
Fix HSL to RGB overflow error

### LINKED ISSUE #2266: Red/Green visual glitching with Monster hues
#### Player platform:
Windows, Player 0.6.2.1

#### Issue
image
If you set an RTP Slime enemy to have a red hue, specifically the third notch (swatch 7), it will cause them to render in a dissolve-blend-like effect with their original palette.

Upon further investigation, this issue possibly affects all Red/Green changes. Testing the same on "People3" (blue-haired NPC) and "DragonA" produces the same problem. It also happens with "Dark Elf" on swatch 8.
image

#### Demonstration
The following images represent troops of slimes for each swatch, the first and last in a screen repeating a previous set.
RPG_RT:
image

Player:
image

rueter37:
This PR fixes #2266. If the hue value in the HSL_to_RGB function was greater than or equal to 1536, the color was not converted. Using modulo 1536 on the hue value fixes this because the hue ranges between 0 and 1536 in this function.

### DIFF
--- a/src/bitmap_hslrgb.h
+++ b/src/bitmap_hslrgb.h
@@ -56,7 +56,7 @@ static inline void HSL_to_RGB(const int& h, const int& s, const int& l,
 
 static inline void HSL_adjust(int& h, int& s, int& l, int hue) {
 	h += hue;
-	if (h > 0x600) h -= 0x600;
+	if (h >= 0x600) h -= 0x600;
 	if (s > 0xFF) s = 0xFF;
 	l = (l > 0xFF) ? 0xFF : (l < 0) ? 0 : l;
 }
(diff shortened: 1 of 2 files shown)


fmatthew5876:
Going to have to defer to @Ghabry on this one. If he doesn't see anything wrong with the fix than I approve.

Ghabry (changes requested):
I didn't write this function and it was never touched since 6 years ^^'

Well, the 1536 (0x600) gives some hint about the problem. Thanks for figuring this out.

Imo the real problem is in the function

```cpp
static inline void HSL_adjust(int& h, int& s, int& l, int hue) {
	h += hue;
	if (h > 0x600) h -= 0x600; // <- this must be changed to >=
	if (s > 0xFF) s = 0xFF;
	l = (l > 0xFF) ? 0xFF : (l < 0) ? 0 : l;
}
```

Your modulo is only needed for the ``h = 0x600`` case. For all other values there is now an additional mod operation.
To avoid unnecessary calculations the ``>`` to ``>=`` change is better.

rueter37:
Did the requested change now. And thank you for the invitation to the team, very appreciated! :smile:

Ghabry (approved):
I nuked the first commit and added a warning fix commit

fmatthew5876 (approved):

PR #2326, opened 2020-09-10
Add Rand::API + allow locking the RNG

fmatthew5876:
This PR allows you to lock the rng to always return a single value. In order to maintain accurate semantics for callers to random, the value is clamped.

So for example:

```
Rand::LockRandom(5);
assert(Rand::GetRandomNumber(0, 10) == 5);

Rand::LockRandom(65);
assert(Rand::GetRandomNumber(0, 10) == 10);

Rand::LockRandom(-65);
assert(Rand::GetRandomNumber(0, 10) == 0);
```

My use case for this is to be able to accurately unit test battle variance in #2320, but I imagine we can also use this for the regression test system.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -210,6 +210,8 @@ add_library(${PROJECT_NAME}
 	src/player.cpp
 	src/player.h
 	src/point.h
+	src/rand.cpp
+	src/rand.h
 	src/rect.cpp
 	src/rect.h
 	src/registry.h
--- a/Makefile.am
+++ b/Makefile.am
@@ -203,6 +203,8 @@ libeasyrpg_player_a_SOURCES = \
 	src/point.h \
 	src/game_quit.cpp \
 	src/game_quit.h \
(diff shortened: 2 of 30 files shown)


fmatthew5876:
I'd rather get this in sooner deal with rebasing #2208  so I can unblock #2320 unit tests using this.

Unless you're planning to merge #2208 soon?

fdelapena (approved):

fmatthew5876:
Anot

fmatthew5876:
Another option here. I can wait for character PR for this one. We can let #2320 go in first and I can fix the variance unit tests after. So I am fine waiting.

fmatthew5876:
Rebased this. Also there was a bug in one of the character unit tests in master that it was using `MoveType_Random` and so it was sometimes passing and sometimes failing depending on RNG. Since this PR is approved and likely to merge soon. I put the fix here.

fmatthew5876:
jenkins: test this please

PR #2328, opened 2020-09-10
Battle 2k3: Battle animation sprites changes

rueter37:
This PR will be updated as soon as #2399 is merged into master. The planned changes are in [rueter37:battleanimation-sprites-new](https://github.com/rueter37/Player/tree/battleanimation-sprites-new).

Depends on: #2399 because of the new sprite_actor class and changes in the sprite_battler class made there.

This PR does the following fixes (after the update):
- Placement of battlers: The battlers were placed wrong (too much distance between each other and partially and full off-screen placing starting with the third party member) if battle animation sprites were used. BattleCharset sprites are working fine.
- Displaying state afflictions: While BattleCharset sprites are showing its state afflictions at the beginning of the battle, the battle animation sprite battlers did so not until their first turn.
- Battle animation sprites are flipped now if the flip flag is set.
- BattleAnimation height: Battle animations are displayed on the right height if the target is a battle animation sprite.
- Battler flashes: The battler flashes didn't work if battle animation sprites were used. Again BattleCharset sprites fully work here.
- Battler shake effects: The battler shake effects didn't work if battle animation sprites were used. Again BattleCharset sprites fully work here.
- ~~Screen flashes: The screen flashes didn't work if battle animation sprites were used. Again BattleCharset sprites fully work here.~~ (#2399 fixes this)

### DIFF
--- a/src/battle_animation.cpp
+++ b/src/battle_animation.cpp
@@ -73,7 +73,8 @@ void BattleAnimation::Update() {
 	UpdateScreenFlash();
 	UpdateTargetFlash();
 
-	SetFlashEffect(Main_Data::game_screen->GetFlashColor());
+	Color screenflash = Main_Data::game_screen->GetFlashColor();
+	if (screenflash.red > 0 || screenflash.green > 0 || screenflash.blue > 0 || screenflash.alpha > 0) SetFlashEffect(Main_Data::game_screen->GetFlashColor());
 
 	frame++;
 }
@@ -209,7 +210,7 @@ void BattleAnimation::UpdateScreenFlash() {
 void BattleAnimation::UpdateTargetFlash() {
 	int r, g, b, p;
 	UpdateFlashGeneric(target_flash_timing, r, g, b, p);
-	FlashTargets(r, g, b, p);
+	if (r > 0 || g > 0 || b > 0 || p > 0) FlashTargets(r, g, b, p);
 }
 
 // For handling the vertical position.
--- a/src/battle_animation.cpp
+++ b/src/battle_animation.cpp
@@ -204,13 +204,13 @@ void BattleAnimation::UpdateFlashGeneric(int timing_idx, int& r, int& g, int& b,
 void BattleAnimation::UpdateScreenFlash() {
 	int r, g, b, p;
 	UpdateFlashGeneric(screen_flash_timing, r, g, b, p);
-	if (r > 0 || g > 0 || b > 0 || p > 0) Main_Data::game_screen->FlashOnce(r, g, b, p, 0);
+	if (!overwrite_screen_effects) Main_Data::game_screen->FlashOnce(r, g, b, p, 0);
 }
 
 void BattleAnimation::UpdateTargetFlash() {
 	int r, g, b, p;
 	UpdateFlashGeneric(target_flash_timing, r, g, b, p);
-	if (r > 0 || g > 0 || b > 0 || p > 0) FlashTargets(r, g, b, p);
+	if (!overwrite_screen_effects) FlashTargets(r, g, b, p);
 }
 
 // For handling the vertical position.
@@ -333,4 +333,8 @@ void BattleAnimation::SetIgnoreYOffset(bool ignoreyoffset) {
 
 void BattleAnimation::SetInvert(bool inverted) {
 	invert = inverted;
+}
+
+void BattleAnimation::SetOverwriteScreenEffects(bool overwrite) {
+	overwrite_screen_effects = overwrite;
 }
--- a/src/battle_animation.cpp
+++ b/src/battle_animation.cpp
@@ -330,3 +330,7 @@ void BattleAnimation::SetFrame(int frame) {
 void BattleAnimation::SetIgnoreYOffset(bool ignoreyoffset) {
 	ignore_y_offset = ignoreyoffset;
 }
+
+void BattleAnimation::SetInvert(bool inverted) {
+	invert = inverted;
+}
--- a/src/battle_animation.cpp
+++ b/src/battle_animation.cpp
@@ -203,7 +203,7 @@ void BattleAnimation::UpdateFlashGeneric(int timing_idx, int& r, int& g, int& b,
 void BattleAnimation::UpdateScreenFlash() {
 	int r, g, b, p;
 	UpdateFlashGeneric(screen_flash_timing, r, g, b, p);
-	Main_Data::game_screen->FlashOnce(r, g, b, p, 0);
+	if (r > 0 || g > 0 || b > 0 || p > 0) Main_Data::game_screen->FlashOnce(r, g, b, p, 0);
 }
 
 void BattleAnimation::UpdateTargetFlash() {
--- a/src/battle_animation.cpp
+++ b/src/battle_animation.cpp
@@ -294,7 +294,7 @@ void BattleAnimationBattle::Draw(Bitmap& dst) {
 	for (auto* battler: battlers) {
 		const Sprite_Battler* sprite = Game_Battle::GetSpriteset().FindBattler(battler);
 		int offset = 0;
-		if (sprite && sprite->GetBitmap()) {
+		if (sprite && (sprite->GetBitmap() || (sprite->IsUsingAnimationAsSprite() && !ignore_y_offset))) {
 			offset = CalculateOffset(animation.position, sprite->GetHeight());
 		}
 		DrawAt(dst, battler->GetBattlePosition().x, battler->GetBattlePosition().y + offset);
@@ -327,3 +327,6 @@ void BattleAnimation::SetFrame(int frame) {
 	this->frame = frame;
 }
 
+void BattleAnimation::SetIgnoreYOffset(bool ignoreyoffset) {
+	ignore_y_offset = ignoreyoffset;
+}
--- a/src/battle_animation.h
+++ b/src/battle_animation.h
@@ -86,6 +86,13 @@ public:
 	 **/
 	void SetIgnoreYOffset(bool ignoreyoffset);
 
+	/**
+	 * Set if the animation overwrites screen effects
+	 *
+	 * @param overwrite if the animation overwrites screen effects
+	 **/
+	void SetOverwriteScreenEffects(bool overwrite);
+
 protected:
 	BattleAnimation(const lcf::rpg::Animation& anim, bool only_sound = false, int cutoff = -1);
 
@@ -110,6 +117,7 @@ protected:
 	bool only_sound = false;
 	bool invert = false;
 	bool ignore_y_offset = false;
+	bool overwrite_screen_effects = false;
 };
 
 // For playing animations on the map.
--- a/src/battle_animation.h
+++ b/src/battle_animation.h
@@ -73,6 +73,13 @@ public:
 	int GetAnimationCellHeight() const;
 
 	/**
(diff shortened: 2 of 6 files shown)


src/sprite_battler.cpp
```diff
@@ -264,19 +264,20 @@ bool Sprite_Battler::IsIdling() {
 }
 
 int Sprite_Battler::GetWidth() const {
-	if (animation) {
-		return animation->GetWidth();
-	}
+	if (animation) return 48;
```
fmatthew5876, line 268: Why assume 48? If we ever support custom large animation formats this would break.

Why doesn't `animation->GetWidth()` work here?

src/sprite_battler.cpp
```diff
@@ -264,19 +264,20 @@ bool Sprite_Battler::IsIdling() {
...
 int Sprite_Battler::GetHeight() const {
-	if (animation) {
-		return animation->GetHeight();
-	}
+	if (animation) return 48;
 	return Sprite::GetHeight();
 }
 
+bool Sprite_Battler::IsUsingAnimationAsSprite() {
+	if (animation) return true;
```
fmatthew5876, line 278: You can shorten this while function to `return animation != nullptr;`

src/scene_battle_rpg2k3.cpp
```diff
@@ -383,6 +383,15 @@ void Scene_Battle_Rpg2k3::CreateUi() {
 	} else {
 		SetupSystem2Graphics();
 	}
+
+	std::vector<Game_Battler*> battlers;
+	Main_Data::game_party->GetActiveBattlers(battlers);
+	for (std::vector<Game_Battler*>::const_iterator it = battlers.begin(); it != battlers.end(); ++it) {
```
fmatthew5876, line 389: Use a C++11 for each loop here and ditch the iterators.

src/battle_animation.cpp
```diff
@@ -203,7 +203,7 @@ void BattleAnimation::UpdateFlashGeneric(int timing_idx, int& r, int& g, int& b,
 void BattleAnimation::UpdateScreenFlash() {
 	int r, g, b, p;
 	UpdateFlashGeneric(screen_flash_timing, r, g, b, p);
-	Main_Data::game_screen->FlashOnce(r, g, b, p, 0);
+	if (r > 0 || g > 0 || b > 0 || p > 0) Main_Data::game_screen->FlashOnce(r, g, b, p, 0);
```
fmatthew5876, line 207: This needs further investigation.

Battle animations when run actually cancel screen flash effect each frame. 

Try an event like this on the map

```
FlashScreen(31,31,31,31,12sec)
Wait 1.0sec
Show BattleAnimation on player
```

This happens for map and battle animations.

For these battler animations, probably different update logic is used to prevent this. I haven't deep dived into battler animations in RE tools yet as it's super complicated and kind of the final frontier of the 2k3 battle system.

If I get some time I can try to look more how this works.

fmatthew5876:
I feel like we may possibly need a third `BattleAnimationBattler` class in the `BattleAnimation` heirarchy to handle the different behaviors with flip, screen flash etc..

Not really sure as honestly I have no idea how the battle animation stuff works.

src/sprite_battler.cpp
rueter37: Because `animation->GetWidth()` returns 0 for some reason. Battle animations normally have a width and height of 96 but for some reason 48 is needed to place the actors correctly.

src/battle_animation.cpp
rueter37: This is hard to explain but it somehow looks like that battle animation sprites are "blocking" the flash functions. This `FlashOnce` is called every frame with all parameters set to 0. Adding this check above fixes the flashing in RPG Maker 2003 battles with battle animation sprites used but as you said this might break the flash functions in other situations.

src/sprite_battler.cpp
fmatthew5876: That's because the underlying graphical resources haven't been loaded yet, and so we don't . Take a look at `Scene_Battle_Rpg2k3` and the `Start()` and `Start2()` functions. Notice how we use this `AsyncNext` mechanism for `Start2`.

This exists because anytime some logic depends on loading resources, we have to wait for them to be loaded. We have a asynchronous IO system here because of platforms like emscripten, where data can take a couple frames to load. 

Basically what happens here is:

1. Screen Erased
2. Start() called - does setup logic
3. Start() registers Start2() to be called when all pending async IO is done
4. System waits for async files to be loaded
5. Start2() is called
6. TransitionIn and begin the scene.

All of the battler placement logic is done in Start2() for this reason, as it depends on battler sprites. Also notice at the end how we reset the battler sprite Z files.

TLDR; Do your animation initialization login in `Start2()`

src/battle_animation.cpp
fmatthew5876: > that battle animation sprites are "blocking" the flash functions.

Yes, and thats what I was saying. This is on purpose. I actually did some tests where I saved the game each frame and as soon as the battle animation starts you see it clobbering the flash screen chunks. It also works this way for map character flash chunks.

These battler battle animations probably work differently. So your proposed patch of filtering here would only be valid for battler battle animations, but not regular battle and map animations.

src/sprite_battler.cpp
fmatthew5876: >  for some reason 48 is needed to place the actors correctly.

It's the half width. 96 / 2 = 48 

The placement logic works the same way, RPG_RT hard codes 24. We do battler sprite width (48) / 2 = 24

src/sprite_battler.cpp
rueter37: Did the requested change.

src/scene_battle_rpg2k3.cpp
rueter37: Did the requested change as well.

Ghabry:
Here are still some uncleared issues concerning Battle animation sprite flasshing, right?

rueter37:
Yes, this is true. In my tests inside and outside of battles I did not encounter any regressions in relation to flashes but the initial question is if the way I changed that behavior is the right one or not.

fmatthew5876:
The flashing is currently broken. Do my test as mentioned by doing a slow flash on the screen and then running a battle animation with no flash effects before it finishes.

The battle animation should cancel the screen flash.

rueter37:
Welp, you are right. The screen flash no longer get canceled by the battle animation which clearly is a regression. Let's see if I find a way to work around this...

fmatthew5876:
I suggest the best way around this is probably to add a third `BattleAnimation` class type for these, which overrides the flash function and has the logic you wrote.

rueter37:
Canceling screen flashes through battle animations work again now.

Ghabry:
@rueter37 
Does this PR also fix the missing finger cursor? Somehow is not displayed anymore when targeting.

rueter37:
@Ghabry: Well, this PR doesn't do anything regarding the finger cursor.

src/battle_animation.h
```diff
@@ -76,6 +76,13 @@ class BattleAnimation : public Sprite {
 	 **/
 	void SetNoOffset(bool nooffset);
 
+	/**
+	 * Set if the animation is coming from a battle animation sprite
+	 *
+	 * @param battleanimationsprite
+	 **/
+	void SetBattleAnimationSprite(bool battleanimationsprite);
```
fmatthew5876, line 84: Lets give this a better name, like "SetOverwriteScreenEffects". From the point of view of the BattleAnimation class it's not about whether it's a sprite or not. It's if the screen effects get overwritten even frame.

It just so happens that BattleAnimationSprite use case uses this feature.

rueter37: Did the renaming.

src/sprite_battler.cpp
```diff
@@ -264,19 +268,19 @@ bool Sprite_Battler::IsIdling() {
...
-	}
+	if (animation) return 48;
 	return Sprite::GetWidth();
 }
 
 int Sprite_Battler::GetHeight() const {
-	if (animation) {
-		return animation->GetHeight();
-	}
+	if (animation) return 48;
```
Ghabry, line 276: More on this sprite stuff @rueter37 
First: ``GetWidth/GetHeight`` don't return any meaningful here in general because this is the dimension of the entire animation file, so not useful here.
You forgot to handle the case that it is a large file: https://github.com/EasyRPG/Player/blob/master/src/battle_animation.cpp#L50
Maybe add two new functions ``GetAnimationCellWidth/GetAnimationCellHeight`` that handle this. (yeah they return the same right now, but maybe changes later)

src/battle_animation.h
```diff
@@ -62,6 +62,27 @@ class BattleAnimation : public Sprite {
...
+	 * @param inverted if the animation is inverted
+	 **/
+	void SetInvert(bool inverted);
+
+	/**
+	 * Set if the animation ignores vertical offset
+	 *
+	 * @param nooffset if the animation ignores vertical offset
+	 **/
+	void SetNoOffset(bool nooffset);
```
Ghabry, line 77: What is this vertical offset? Can you tell me how to reproduce this?
Also a negation in a Set is confusing: ``SetOffset`` or because this only affects the vertical offset: ``SetIgnoreVerticalOffset`` and vertical is Y so it could also be ``SetIgnoreYOffset``

src/battle_animation.cpp
```diff
@@ -203,13 +204,13 @@ void BattleAnimation::UpdateFlashGeneric(int timing_idx, int& r, int& g, int& b,
...
 	UpdateFlashGeneric(screen_flash_timing, r, g, b, p);
-	Main_Data::game_screen->FlashOnce(r, g, b, p, 0);
+	if (!overwrite_screen_effects || r > 0 || g > 0 || b > 0 || p > 0) Main_Data::game_screen->FlashOnce(r, g, b, p, 0);
 }
 
 void BattleAnimation::UpdateTargetFlash() {
 	int r, g, b, p;
 	UpdateFlashGeneric(target_flash_timing, r, g, b, p);
-	FlashTargets(r, g, b, p);
+	if (!overwrite_screen_effects || r > 0 || g > 0 || b > 0 || p > 0) FlashTargets(r, g, b, p);
```
Ghabry, line 213: I find these entire ``r > 0`` changes (here and two times above) too controversial. Adding another conditional doesn't make this better.
Can we get rid of them for now or move them in another PR?

Ghabry (changes requested):
-

src/battle_animation.cpp
rueter37: Did some tests and came to the conclusion that these `r > 0` checks are not necessary. The `!overwrite_screen_effects` check is sufficient.

src/battle_animation.h
rueter37: This is difficult to explain but this vertical offset check is needed if the target is a battle animation sprite. In RPG_RT the position setting (top, center of bottom) of a battle animation is ignored if the target is a battle animation sprite. You can test this in EasyRPG (current master) in the game "Vampires Dawn II" fighting against the "Horanmagier (battle ID 62)". When he casts "Eisiger Entzug" you will notice that the animation is shown too far above.

src/sprite_battler.cpp
rueter37: Added the functions now.

rueter37:
Closing this PR because it is outdated and interferes with #2399.

PR #2329, opened 2020-09-10
Battle 2k3: Add negative attribute modifiers and allow state infliction by skills outside of battles

rueter37:
This PR implements the negative attribute modifiers used in RPG Maker 2003 which allows for damage absorbing (e.g. dragons absorbing fire-type skills) and the possibility for skills to inflict states outside of battles (this makes the "Regenerationstrank" in "Vampires Dawn II" work outside of battles). By the way, negative attribute modifiers even allow for "funny" things like taking damage through healing skills (but you cannot get killed this way, 1 HP remain - tested with RPG_RT). This is probably my "most insane" PR so I expect many changes needed. Feedback is very appreciated!

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -629,7 +629,7 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 		int hp = GetAffectedHp();
 		int target_hp = target->GetHp();
 		target->ChangeHp(IsPositive() ? hp : -hp);
-		if (IsAbsorb()) {
+		if (IsAbsorb() && !IsPositive()) {
 			// Only absorb the hp that were left
 			int src_hp = std::min(target_hp, hp);
 			source->ChangeHp(src_hp);
@@ -640,7 +640,7 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 		int sp = GetAffectedSp();
 		int target_sp = target->GetSp();
 		target->SetSp(target->GetSp() + (IsPositive() ? sp : -sp));
-		if (IsAbsorb()) {
+		if (IsAbsorb() && !IsPositive()) {
 			int src_sp = std::min(target_sp, sp);
 			source->ChangeSp(src_sp);
 		}
@@ -649,7 +649,7 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 	if (GetAffectedAttack() != -1) {
 		int atk = GetAffectedAttack();
 		target->ChangeAtkModifier(IsPositive() ? atk : -atk);
-		if (IsAbsorb() && Player::IsRPG2k3Updated()) {
+		if (IsAbsorb() && !IsPositive() && Player::IsRPG2k3Updated()) {
 			atk = std::max<int>(0, std::min<int>(atk, std::min<int>(source->MaxStatBattleValue(), source->GetBaseAtk() * 2) - source->GetAtk()));
 			source->ChangeAtkModifier(atk);
 		}
@@ -658,8 +658,8 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 	if (GetAffectedDefense() != -1) {
 		int def = GetAffectedDefense();
 		target->ChangeDefModifier(IsPositive() ? def : -def);
-		if (IsAbsorb() && Player::IsRPG2k3Updated()) {
-			def = std::max<int>(0, std::min<int>(def, std::min<int>(source->MaxStatBattleValue(), source->GetBaseAtk() * 2) - source->GetAtk()));
+		if (IsAbsorb() && !IsPositive() && Player::IsRPG2k3Updated()) {
+			def = std::max<int>(0, std::min<int>(def, std::min<int>(source->MaxStatBattleValue(), source->GetBaseDef() * 2) - source->GetDef()));
 			source->ChangeDefModifier(def);
 		}
 	}
@@ -667,8 +667,8 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 	if (GetAffectedSpirit() != -1) {
 		int spi = GetAffectedSpirit();
 		target->ChangeSpiModifier(IsPositive() ? spi : -spi);
-		if (IsAbsorb() && Player::IsRPG2k3Updated()) {
-			spi = std::max<int>(0, std::min<int>(spi, std::min<int>(source->MaxStatBattleValue(), source->GetBaseAtk() * 2) - source->GetAtk()));
+		if (IsAbsorb() && !IsPositive() && Player::IsRPG2k3Updated()) {
+			spi = std::max<int>(0, std::min<int>(spi, std::min<int>(source->MaxStatBattleValue(), source->GetBaseSpi() * 2) - source->GetSpi()));
 			source->ChangeSpiModifier(spi);
 		}
 	}
@@ -676,8 +676,8 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 	if (GetAffectedAgility() != -1) {
 		int agi = GetAffectedAgility();
 		target->ChangeAgiModifier(IsPositive() ? agi : -agi);
-		if (IsAbsorb() && Player::IsRPG2k3Updated()) {
-			agi = std::max<int>(0, std::min<int>(agi, std::min<int>(source->MaxStatBattleValue(), source->GetBaseAtk() * 2) - source->GetAtk()));
+		if (IsAbsorb() && !IsPositive() && Player::IsRPG2k3Updated()) {
+			agi = std::max<int>(0, std::min<int>(agi, std::min<int>(source->MaxStatBattleValue(), source->GetBaseAgi() * 2) - source->GetAgi()));
 			source->ChangeAgiModifier(agi);
 		}
 	}
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -1186,7 +1186,7 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 					this->hp *= 2;
 				}
 
-				if (IsAbsorb())
+				if (IsAbsorb() && !this->negative_effect)
 					this->hp = std::min<int>(hp, GetTarget()->GetHp());
 
 				if (!this->negative_effect && GetTarget()->GetHp() - this->hp <= 0) {
@@ -1209,7 +1209,11 @@ bool Game_BattleAlgorithm::Skill::Execute() {
 
 		if (skill.affect_sp && Rand::PercentChance(to_hit)) {
 			if (IsNegativeSkill()) {
-				this->sp = std::min<int>(effect, GetTarget()->GetSp());
+				if (!this->negative_effect) {
+					this->sp = std::min<int>(effect, GetTarget()->GetSp());
+				} else {
+					this->sp = effect;
+				}
 			} else {
 				int sp_cost = GetSource() == GetTarget() ? source->CalculateSkillCost(skill.ID) : 0;
 				if (!this->negative_effect) {
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -650,7 +650,7 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 		int atk = GetAffectedAttack();
 		target->ChangeAtkModifier(IsPositive() ? atk : -atk);
 		if (IsAbsorb() && !IsPositive() && Player::IsRPG2k3Updated()) {
-			atk = Utils::Clamp(std::min<int>(source->MaxStatBattleValue(), source->GetBaseAtk() * 2) - source->GetAtk(), 0, atk);
+			atk = Utils::Clamp(atk, 0, std::min<int>(source->MaxStatBattleValue(), source->GetBaseAtk() * 2) - source->GetAtk());
 			source->ChangeAtkModifier(atk);
 		}
 	}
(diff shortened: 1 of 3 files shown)


fmatthew5876:
I had a thought about this, let me know what you think.

I find this whole idea of `GetAffectedHp() < 0` meaning hp is not affected confusing. At a minimum we should wrap this in a function call and stop doing `< 0` inline everywhere.

My idea:

What if we had boolean flags for hp, sp, atk, def, spi, and agi to say whether or not the skill affects them, and then just use + or -  signed numbers to indicate healing vs damage for each one. Then there is no more `healing` or `IsPositive()` flag. Such a design also allows custom battle systems to extend and add skills which can affect each parameter individually.

I've always had an idea to try this approach, but haven't yet, so I don't know for sure if it'll end up better or worse.

src/game_battlealgorithm.cpp
```diff
@@ -1354,14 +1400,34 @@ bool Game_BattleAlgorithm::Skill::Execute() {
...
-				this->attack = std::max<int>(0, std::min<int>(effect, GetTarget()->GetAtk() - (GetTarget()->GetBaseAtk() + 1) / 2));
-			if (skill.affect_defense && Utils::PercentChance(to_hit))
-				this->defense = std::max<int>(0, std::min<int>(effect, GetTarget()->GetDef() - (GetTarget()->GetBaseDef() + 1) / 2));
-			if (skill.affect_spirit && Utils::PercentChance(to_hit))
-				this->spirit = std::max<int>(0, std::min<int>(effect, GetTarget()->GetSpi() - (GetTarget()->GetBaseSpi() + 1) / 2));
-			if (skill.affect_agility && Utils::PercentChance(to_hit))
-				this->agility = std::max<int>(0, std::min<int>(effect, GetTarget()->GetAgi() - (GetTarget()->GetBaseAgi() + 1) / 2));
+			if (skill.affect_attack && Utils::PercentChance(to_hit)) {
+				if (!this->healing) {
+					this->attack = std::max<int>(0, std::min<int>(effect, GetTarget()->GetAtk() - (GetTarget()->GetBaseAtk() + 1) / 2));
```
fmatthew5876: Since you're refactoring and testing here, this min/max junk should be replaced with `Utils::Clamp()`

rueter37: Did the replacements now.

rueter37:
Your idea sounds interesting. Being able to affect each parameter individually would be a nice addition for #2219. But I would like to hear what the others think about it.

fmatthew5876:
Regardless, I think the overall approach you took here probably makes sense, Any further enhancement can be a later PR.

Please also test carefully the 2k battle system. It's message based state machine heavily relies on the various battle algorithm flags

rueter37:
Did tests with the games "Vampires Dawn" (RPG Maker 2000), "Vampires Dawn II" (RPG Maker 2003) and some custom test cases and did not encounter any regressions.

fdelapena (approved):

fmatthew5876:
I saw this in the RE code, couldn't believe it, but then confirmed it's true...

There is a *massive* RPG_RT bug with negative attributes. ❗ 

Negative attribute HP healing is doubled in it's effect!

To see this, Create a skill that does exactly 50 damage, with fire element. Create an actor with -100% fire attribute.

Now start a battle with actor at 1 hp, and  have an enemy use the fire skill on the actor. 50 hp will be displayed in green pop up numbers, but actor HP will actually increase by 100!

The reason for this is because in the battle code, first they call `RPG_Actor::ApplyHpDamage()` unconditionally with a negative value. The function is meant to handle hp damage, but because the amount is negative, it actually recovers HP. Then it starts to spawn white popup numbers with a negative value.

Then they immediately check if the damage is negative and call `RPG_Actor::ApplyHpRecovery()` with the value flipped. This heals the actor again, and spawns green numbers with the effect amount, immediately overwriting the previous white numbers.

Here is the function with this bug ...

| Address | Class | Name |
| -- | -- | -- |
| 0049CFE8 | RPG_SceneBattle | ApplyHpDamageEffect |

rueter37:
The "double strength for negative effects" bug is emulated now.

rueter37:
@fmatthew5876 To avoid conflicts with your PR (#2320) I made a separate branch which implements the negative attribute modifiers on top of #2320. This task is already completed and my tests worked so far. So we can wait until #2320 is merged then I will force-push my new changes into this PR.

fmatthew5876:
You can rebase your PR now onto master, #2320 was merged. Once that's done I'll test and review this guy.

rueter37:
Did the rebase now, you can test and review it now.

src/game_battlealgorithm.cpp
```diff
@@ -925,12 +926,19 @@ bool Game_BattleAlgorithm::Normal::Execute() {
...
 		auto effect = Algo::CalcNormalAttackEffect(source, target, Game_Battler::WeaponAll, critical_hit, true, Game_Battle::GetBattleCondition(), true);
 		effect = Algo::AdjustDamageForDefend(effect, target);
-		// FIXME: Handle negative effect from attributes
+
+		// Handle negative effect from attributes
+		if (effect < 0) {
+			this->healing = true;
+			// RPG_RT bug: Negative effects have double strength
+			effect = -effect * 2;
+		}
```
fmatthew5876, line 935: Even if it's not used in `Normal`, we should for consistency set the `negative_effect` flag here in case something later wants to rely on it.

Also this `*2` to emulate the bug is not correct. The double affect happens only for HP, but not other stats. So if you want to emulate the bug you'll need to double the hp effect but not others.

rueter37: I don't think we should emulate the animated damage number bug for better user experience in my opinion.

src/game_battlealgorithm.cpp
```diff
@@ -1312,7 +1349,7 @@ void Game_BattleAlgorithm::Skill::Apply() {
 	AlgorithmBase::Apply();
 
 	for (auto& sa: shift_attributes) {
-		GetTarget()->ShiftAttributeRate(sa, healing ? 1 : -1);
+		GetTarget()->ShiftAttributeRate(sa, (healing ^ negative_effect) ? 1 : -1);
```
fmatthew5876, line 1352: This is one part where the code gets super confusing, when I first read this I thought you were reversing the attribute shift but you're not because both of these flags get set at the same time.

This inline xor everywhere is super ugly. Can you please wrap this in an inline helper function with a useful name so we can know whats going on.

Btw for this stuff you can push new commits instead of rebasing if thats easier.

src/game_battlealgorithm.cpp
fmatthew5876: I think I agree. We'll probably get "bug reports" about the effect being double, but we'll then just have to point out it's actually RPG_RT compatible!

rueter37: The double affect bug is now applied on HP only.

src/game_battlealgorithm.cpp
rueter37: Moved the XOR-Operations into own functions.

src/game_battlealgorithm.cpp
```diff
@@ -925,12 +926,24 @@ bool Game_BattleAlgorithm::Normal::Execute() {
...
+		}
+
 		effect = Utils::Clamp(effect, 0, MaxDamageValue());
 
 		this->hp = effect;
 
-		if (GetTarget()->GetHp() - this->hp <= 0) {
+		// RPG_RT bug: Negative effects affect HP double
+		if (this->negative_effect) {
+			this->hp *= 2;
```
fmatthew5876, line 943: One more edge case, the double effect bug *does not* happen for absorb hp. Only normal hp damage..

src/game_battlealgorithm.cpp
```diff
@@ -1135,68 +1148,91 @@ bool Game_BattleAlgorithm::Skill::Execute() {
...
+
+		if (IsNegativeSkill()) absorb = skill.absorb_damage;
+
+		if (skill.affect_hp && Utils::PercentChance(to_hit)) {
+			if (IsNegativeSkill()) {
 				this->hp = Algo::AdjustDamageForDefend(effect, *GetTarget());
 
+				// RPG_RT bug: Negative effects affect HP double
+				if (this->negative_effect) {
+					this->hp *= 2;
```
fmatthew5876, line 1169: One more edge case, the double effect bug *does not* happen for absorb hp. Only normal hp damage..

src/game_battlealgorithm.cpp
```diff
@@ -1135,68 +1148,91 @@ bool Game_BattleAlgorithm::Skill::Execute() {
...
 				}
+			} else {
+				if (!this->negative_effect) {
+					if (Player::IsRPG2k3()) {
+						this->hp = effect;
+					} else {
+						this->hp = Utils::Clamp(GetTarget()->GetMaxHp() - GetTarget()->GetHp(), 0, effect);
+					}
+				} else {
+					this->hp = Utils::Clamp(GetTarget()->GetHp() - 1, 0, effect);
```
fmatthew5876, line 1188: Confirmed this looks right, HP recovery code clamps (1, maxhp), so a negative healing effect cannot kill.

Pretty stupid really. You can't make a game that you can kill undead with healing spells, only weaken them to 1 hp..

src/game_battlealgorithm.cpp
```diff
@@ -1135,68 +1148,91 @@ bool Game_BattleAlgorithm::Skill::Execute() {
...
-			if (skill.affect_sp && Utils::PercentChance(to_hit)) {
+		if (skill.affect_sp && Utils::PercentChance(to_hit)) {
+			if (IsNegativeSkill()) {
 				this->sp = std::min<int>(effect, GetTarget()->GetSp());
+			} else {
+				int sp_cost = GetSource() == GetTarget() ? source->CalculateSkillCost(skill.ID) : 0;
+				if (!this->negative_effect) {
+					this->sp = Utils::Clamp(GetTarget()->GetMaxSp() - GetTarget()->GetSp() + sp_cost, 0, effect);
+				} else {
+					this->sp = effect;
```
fmatthew5876, line 1201: Don't you still need to reduce the SP recovery by the `sp_cost` here?

src/game_battler.cpp
```diff
@@ -266,17 +266,21 @@ bool Game_Battler::UseSkill(int skill_id, const Game_Battler* source) {
...
+
 		// Cure states
 		for (int i = 0; i < (int)skill->state_effects.size(); i++) {
 			if (skill->state_effects[i]) {
 				if (skill->reverse_state_effect) {
 					was_used |= !HasState(lcf::Data::states[i].ID);
-					// FIXME: This logic always called from menu, so we set false
-					// to allow_battle_states
-					AddState(lcf::Data::states[i].ID, false);
+					AddState(lcf::Data::states[i].ID, true);
```
fmatthew5876, line 283: Not sure how we missed this last time. You're right that this case allows you to inflict battle only states, even from the menu.

fmatthew5876:
This looks good to me. In particular I really like how `Skill::Execute` has changed from 2 overly long branches for positive and negative, to now flatter code which has a separate branch for each effect. It's more readable.

Once you address the few minor nits remaining, I can approve this.

src/game_battlealgorithm.cpp
```diff
@@ -1135,68 +1148,91 @@ bool Game_BattleAlgorithm::Skill::Execute() {
...
 					// Death state
 					lethal = true;
 					killed_by_dmg = true;
 				}
+			} else {
+				if (!this->negative_effect) {
+					if (Player::IsRPG2k3()) {
+						this->hp = effect;
+					} else {
+						this->hp = Utils::Clamp(GetTarget()->GetMaxHp() - GetTarget()->GetHp(), 0, effect);
```
fmatthew5876, line 1185: All these Clamps are written wrong, it should be (effect, 0, maxvalue)

src/game_battlealgorithm.cpp
rueter37: Fixed.

src/game_battlealgorithm.cpp
rueter37: Well, the code as is works for me. The else branch is taking SP damage from the healing skill because of the negative effect and in my tests it worked correctly.

src/game_battlealgorithm.cpp
rueter37: This is probably a dumb question but what value I'm supposed to use for maxvalue? Is it MaxDamageValue() or the maximum positive integer value?

rueter37:
I have added a commit which addresses two special cases which I must have overseen:
- If the source tries to absorb HP from a target and this fails due to negative damage, the amount healed of the target is no longer restricted by the targets current HP.
- If the target recovers SP due to negative damage, the amount recovered is no longer restricted by the targets current SP.

src/game_battlealgorithm.cpp
fmatthew5876: No you're right. This was a mistake on my part

src/game_battlealgorithm.cpp
fmatthew5876: The maxvalue depends on the context. You just need to flip the order of the arguments here. While what you wrote technically works, it doesn't read well.

The function signature is `Clamp(value, min, max)`

So here it would be:

```
Clamp(effect, 0, GetTarget()->GetMaxHp() - GetTarget()->GetHp())
```

rueter37: The order of the arguments have been corrected now.

fmatthew5876 (approved):
Looks good to me now, thanks!

fmatthew5876:
This one needs a rebase again

Ghabry:
sorry, this will be the next one that gets in :)

rueter37:
Did the rebase now.

Ghabry:
In general I let @fmatthew5876 handle any PRs that are algorithm related. I don't really want to dive into this

Issue #1586, opened 2018-12-19
Implement the RPG_RT Autobattle algorithm


fmatthew5876:
Figure out and implement the RPG_RT auto battle algorithm.

Using it for the `AutoBattle` command is a bit controversial. Using it for an actor with `AutoBattle` flag set however is not. If a game dev created an actor with auto battle, the game experience probably demands that actor does more than just attack random enemies. There can also be some smarts from RPG_RT we can reuse event if we don't want to change the `AutoBattle` command to use skills.

I did some testing and one thing I saw. I had an actor with no skills set on autobattle. He would always attack the first enemy. This is better AI than attacking randomly, as at least he works on killing 1 at a time instead of randomly spreading the damage.

Also I noticed he would attack a different enemy only if his weapon would kill that enemy in a single hit. Adjusting the enemy HP would change which target he chose.

That's all I've got for now.

fmatthew5876:
@CherryDT

Any help with this one?

CherryDT:
### Hero AI:

* If angry, attack random monster (regardless of whether the character would have the "attack all" modifier normally).
* Else if confused, attack random party member (regardless of whether the character would have the "attack all" modifier normally).
* Else:
  * Assign scores: (**see details below**)
    * Possible actions are basic attack and all skills that the character can use. For each action, scores are assigned for all possible targets (which can be enemies or allies, depending on the action), let's call these "secondary scores".
    * Then, primary scores are assigned to the action itself, based on the secondary scores of all possible targets for each action.
  * Find best option based on primary score
  * If best option is basic attack: If character has the "attack all" modifier, attack all. Otherwise, attack enemy with highest secondary score for the basic attack action.
  * Otherwise: If the skill has target "all enemies" or "all allies", target all enemies/allies. Otherwise, target the enemy/ally with the highest secondary score for this skill.

**How secondary scores are assigned:**

* Attack:
  * Score = Ratio (i.e. 0 to 1 - never >1!) of HP the attack would deduct from the target
  * If score is now 1, make it 1.5 (this prioritizes killing strikes)
  * If the target is the first alive one in the enemy group (based on array index), multiply the score by 1.5 and add an additional 0.5 at the end (this will make the character seem to attack the enemies more or less in order, unless there is a _much_ better option)
  * Multiply the score by 1.5 (this prefers basic attacks of skills)
  * If score is nonzero, add a random value between 0 and 1 (this adds variance)
* Skills:
[...]

PR #2331, opened 2020-09-12
Implement auto battle

fmatthew5876:
Fix: #1586 

This implements RPG_RT compatible auto battle. I've designed auto battle to support different auto battle algorithms. The end goal will be Player ships a few algos that you can choose from. Game developers customizing player could also add their own to the `AutoBattle::CreateAlgorithm` factory function.

This PR provides 3 algorithms:
* `RPG_RT` - RPG_RT compatible including bugs - this is the default
* `ATTACK` - RPG_RT, only physical attacks, bug fixes included
* `RPG_RT+` - RPG_RT with bug fixes - Also open to further improvements

For now I just added a `--autobattle-algo` command line parameter to select the algo, later this can probably be part of `Scene_Settings` etc.. Or maybe even a custom save chunk.

TODO
-------

- [x] More game play testing
- [x] Unit tests
- [x] Verify 2k3 dynrpg
- [x] Verify 2k3 legacy
- [x] Verify 2ke
- [x] Verify 2k updated
- [x] Verify 2k legacy
- [x] A better algo selection mechanism? Or punt till later?

DynRPG Addresses
-------------------

| Address | Class | Name | Comment |
| -- | -- | -- | -- |
| 004BA518 | Actor | SelectAutoBattleAction | Virtual function which runs the whole auto battle algorithm, everything is here |
| 004B9EF0 | Actor | GetAutoBattleSkillRank | Computes best ranking of skill against all possible targets |
| 004BA384 | Actor | GetAutoBattleAttackRank | Computes best rank of normal attack against all possible targets |
| 004BA0B8 | Actor | GetAutoBattleSkillEffectRank | Computes the rank of a skill against a single target |
| 004BA3E8 | Actor | GetAutoBattleAttackEffectRank | Computes the rank of a normal attack against a single target |

RPG_RT bugs ❗ 
------------------

* Normal attack ranking does not consider attack all. It should sum the ranks of all enemies
* Normal attack ranking does not consider dual attack
* Normal attack ranking doesn't accurately consider dual wield for 2k3
* Normal attack ranked does not consider sp cost of weapons
* Skill ranking uses variance, normal attack does not
* Revive skill ranking does not check for the `reverse_state` flag, and so a skill which inflicts death on an ally would be highly ranked and probably used on a dead ally.

### DIFF
--- a/resources/unix/bash-completion/easyrpg-player
+++ b/resources/unix/bash-completion/easyrpg-player
@@ -35,7 +35,7 @@ _easyrpg-player ()
       #_filedir '@(lmu|emu)'
       return
       ;;
-    # load map files
+    # Select autobattle algorithm
     --autobattle-algos)
       COMPREPLY=($(compgen -W "$autobattle_algos" -- $cur))
       return
--- a/resources/unix/bash-completion/easyrpg-player
+++ b/resources/unix/bash-completion/easyrpg-player
@@ -13,13 +13,14 @@ _easyrpg-player ()
   prev=${COMP_WORDS[COMP_CWORD-1]}
 
   # all possible options
-  ouropts='--battle-test --disable-audio --disable-rtp --enable-mouse --enable-touch \
+  ouropts='--autobattle-algo --battle-test --disable-audio --disable-rtp --enable-mouse --enable-touch \
            --encoding --engine --fps-limit --fps-render-window --fullscreen -h --help \
            --hide-title --load-game-id --new-game --no-vsync --project-path --record-input \
            --replay-input --save-path --seed --show-fps --start-map-id --start-party \
            --start-position --test-play --window -v --version'
   rpgrtopts='BattleTest battletest HideTitle hidetitle TestPlay testplay Window window'
   engines='rpg2k rpg2kv150 rpg2ke rpg2k3 rpg2k3v105 rpg2k3e'
+  autobattle_algos='RPG_RT RPG_RT+ ATTACK'
 
   # first list all special cases
   case $prev in
@@ -32,6 +33,11 @@ _easyrpg-player ()
     --start-map-id)
       # broken, disabled for now
       #_filedir '@(lmu|emu)'
+      return
+      ;;
+    # load map files
+    --autobattle-algos)
+      COMPREPLY=($(compgen -W "$autobattle_algos" -- $cur))
       return
       ;;
     # load save files
--- a/src/autobattle.cpp
+++ b/src/autobattle.cpp
@@ -45,13 +45,13 @@ constexpr decltype(AttackOnly::name) AttackOnly::name;
 constexpr decltype(RpgRtImproved::name) RpgRtImproved::name;
 
 std::unique_ptr<AlgorithmBase> CreateAlgorithm(StringView name) {
-	if (name == RpgRtImproved::name) {
+	if (Utils::StrICmp(name, RpgRtImproved::name) == 0) {
 		return std::make_unique<RpgRtImproved>();
 	}
-	if (name == AttackOnly::name) {
+	if (Utils::StrICmp(name, AttackOnly::name) == 0) {
 		return std::make_unique<AttackOnly>();
 	}
-	if (name != RpgRtCompat::name) {
+	if (Utils::StrICmp(name, RpgRtCompat::name) != 0) {
 		static bool warned = false;
 		if (!warned) {
 			Output::Debug("Invalid AutoBattle algo name `{}' falling back to {} ...", name, RpgRtCompat::name);
--- a/src/autobattle.cpp
+++ b/src/autobattle.cpp
@@ -1,0 +1,376 @@
+/*
+ * This file is part of EasyRPG Player.
+ *
+ * EasyRPG Player is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * EasyRPG Player is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
+ */
+#include "autobattle.h"
+#include "game_actor.h"
+#include "game_enemy.h"
+#include "game_enemyparty.h"
+#include "game_party.h"
+#include "game_battlealgorithm.h"
+#include "game_battle.h"
+#include "algo.h"
+#include "player.h"
+#include "output.h"
+#include "rand.h"
+#include <lcf/reader_util.h>
+#include <lcf/data.h>
+
+namespace AutoBattle {
+
+#ifdef EP_DEBUG_AUTOBATTLE
+template <typename... Args>
+static void DebugLog(const char* fmt, Args&&... args) {
+	Output::Debug(fmt, std::forward<Args>(args)...);
+}
+#else
+template <typename... Args>
+static void DebugLog(const char*, Args&&...) {}
+#endif
+
+constexpr decltype(RpgRtCompat::name) RpgRtCompat::name;
+constexpr decltype(AttackOnly::name) AttackOnly::name;
+constexpr decltype(RpgRtImproved::name) RpgRtImproved::name;
+
+std::unique_ptr<AlgorithmBase> CreateAlgorithm(StringView name) {
+	if (name == RpgRtImproved::name) {
+		return std::make_unique<RpgRtImproved>();
+	}
+	if (name == AttackOnly::name) {
+		return std::make_unique<AttackOnly>();
+	}
+	if (name != RpgRtCompat::name) {
+		static bool warned = false;
+		if (!warned) {
+			Output::Debug("Invalid AutoBattle algo name `{}' falling back to {} ...", name, RpgRtCompat::name);
+			warned = true;
+		}
+	}
+	return std::make_unique<RpgRtCompat>();
+}
+
+void AlgorithmBase::SetAutoBattleAction(Game_Actor& source) {
+	vSetAutoBattleAction(source);
+	if (source.GetBattleAlgorithm() == nullptr) {
+		source.SetBattleAlgorithm(std::make_shared<Game_BattleAlgorithm::NoMove>(&source));
+	}
+}
+
+void RpgRtCompat::vSetAutoBattleAction(Game_Actor& source) {
+	SelectAutoBattleActionRpgRtCompat(source, Game_Battle::GetBattleCondition());
+}
+
+void AttackOnly::vSetAutoBattleAction(Game_Actor& source) {
+	SelectAutoBattleAction(source, Game_Battler::WeaponAll, Game_Battle::GetBattleCondition(), false, false, false, false);
+}
+
+void RpgRtImproved::vSetAutoBattleAction(Game_Actor& source) {
+	SelectAutoBattleAction(source, Game_Battler::WeaponAll, Game_Battle::GetBattleCondition(), true, false, false, false);
+}
+
+double CalcSkillHealAutoBattleTargetRank(const Game_Actor& source, const Game_Battler& target, const lcf::rpg::Skill& skill, bool apply_variance, bool emulate_bugs) {
+	assert(skill.type == lcf::rpg::Skill::Type_normal || skill.type >= lcf::rpg::Skill::Type_subskill);
+	assert(skill.scope == lcf::rpg::Skill::Scope_self || skill.scope == lcf::rpg::Skill::Scope_ally || skill.scope == lcf::rpg::Skill::Scope_party);
+
+	const double src_max_sp = source.GetMaxSp();
+	const double tgt_max_hp = target.GetMaxHp();
+	const double tgt_hp = target.GetHp();
+
+	if (target.GetHp() > 0) {
+		// Can the skill can heal the target?
+		if (!skill.affect_hp) {
+			return 0.0;
+		}
+
+		const double base_effect = Algo::CalcSkillEffect(source, target, skill, apply_variance);
+		const double max_effect = std::min(base_effect, tgt_max_hp - tgt_hp);
+
+		auto rank = static_cast<double>(max_effect) / static_cast<double>(tgt_max_hp);
+		if (src_max_sp > 0) {
+			const double cost = source.CalculateSkillCost(skill.ID);
+			// Note: RPG_RT 2ke only uses integer division for cost / src_max_sp here.
+			rank -= cost / src_max_sp / 8.0;
+			rank = std::max(rank, 0.0);
+		}
+		return rank;
+	}
+
+	// Can the skill revive the target?
+	if (skill.state_effects.size() > 1 && skill.state_effects[0]) {
+		// BUG: RPG_RT does not check the reverse_state_effect flag to skip skills which would kill party members
+		if (emulate_bugs || !skill.reverse_state_effect) {
+			return static_cast<double>(skill.power) / 1000.0 + 1.0;
+		}
+	}
+	return 0.0;
+}
+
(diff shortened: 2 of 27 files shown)


NovemberJoy:
In its current state, the timing for auto-battle is off by enough to grant a second, player-controlled turn to AI characters after their standard turn. However, the issue was strangely intermittent - in one case, it happened with two out of three characters, with the third never getting an extra turn. After further testing, it appears to be dependent on stats or more likely animation timing somehow.

src/autobattle.cpp
```diff
@@ -0,0 +1,339 @@
+#include "autobattle.h"
```
Ghabry, line 1: No header love :(

src/autobattle.cpp
```diff
@@ -0,0 +1,339 @@
...
+		return std::make_unique<RpgRtImproved>();
+	}
+	if (name == AttackOnly::name) {
+		return std::make_unique<AttackOnly>();
+	}
+	return std::make_unique<RpgRtCompat>();
+}
+
+void AlgorithmBase::SetAutoBattleAction(Game_Actor& source) {
+	vSetAutoBattleAction(source);
```
Ghabry, line 41: Is this some convention to use ``v``?
I have a similiar "call forwarder to virtual" in my unfinished Filesystem code.

There I use ``XXXXX`` and for virtual ``XXXXImpl``.

Opinion?

src/autobattle.h
```diff
@@ -0,0 +1,193 @@
...
+/**
+ * Calcuate the auto battle total effectiveness rank of using a normal attack.
+ *
+ * @param source the user of the skill
+ * @param weapon Which weapon to use in the calculation
+ * @param cond the battle condition
+ * @param apply_variance If true, apply variance to the damage
+ * @param emulate_bugs Emulate all RPG_RT bugs for accuracy
+ */
+double CalcNormalAttackAutoBattleRank(const Game_Actor& source, Game_Battler::Weapon weapon, const lcf::rpg::System::BattleCondition cond, bool apply_variance, bool emulate_bugs);
```
Ghabry, line 158: "Clang-Tidy: Parameter 'cond' is const-qualified in the function declaration; const-qualification of parameters only has an effect in function definitions" whatever that means ;)

src/autobattle.cpp
```diff
@@ -0,0 +1,339 @@
...
+			return 0.0;
+		}
+
+		const auto base_effect = Algo::CalcSkillEffect(source, target, skill, apply_variance);
+		const auto max_effect = std::min(base_effect, tgt_max_hp - tgt_hp);
+
+		auto rank = static_cast<double>(max_effect) / static_cast<double>(tgt_max_hp);
+		if (src_max_sp > 0) {
+			const auto cost = source.CalculateSkillCost(skill.ID);
+			rank -= cost / src_max_sp * 0.125;
```
Ghabry, line 79: ``cost / src_max_sp`` int div in floating point. Intentional?

src/autobattle.cpp
```diff
@@ -0,0 +1,339 @@
...
+	const auto tgt_hp = target.GetHp();
+
+	const auto base_effect = Algo::CalcSkillEffect(source, target, skill, apply_variance);
+	rank = static_cast<double>(std::min(base_effect, tgt_hp)) / static_cast<double>(tgt_hp);
+	if (rank == 1.0) {
+		rank = 1.5;
+	}
+	if (src_max_sp > 0) {
+		const auto cost = source.CalculateSkillCost(skill.ID);
+		rank -= cost / src_max_sp * 0.25;
```
Ghabry, line 115: also here

src/autobattle.cpp
```diff
@@ -0,0 +1,339 @@
...
+	// RPG_RT BUG: Normal damage variance is not used
+	// Note: RPG_RT does not do the "2k3_enemy_row_bug" when computing autobattle ranks.
+	auto base_effect = Algo::CalcNormalAttackEffect(source, target, weapon, is_critical_hit, apply_variance, cond, false);
+	// RPG_RT BUG: Dual Attack is ignored
+	if (!emulate_bugs && source.HasDualAttack(weapon)) {
+		base_effect *= 2;
+	}
+	const auto tgt_hp = target.GetHp();
+
+	auto rank = static_cast<double>(std::min(base_effect, tgt_hp) / tgt_hp);
```
Ghabry, line 202: also here

Ghabry (commented):
-

src/autobattle.cpp
fmatthew5876: This is the non-virtual idiom NVI
http://www.gotw.ca/publications/mill18.htm

If I were a more disciplined NVI folllower, I would do this for all virtual functions. I like the `v` prefix, but I don't really care that much what naming scheme is used.

This is also a good way to fix the clang-tidy warnings on `GetBaseAtk()` etc..

src/autobattle.cpp
fmatthew5876: all bugs, thanks!

src/autobattle.h
fmatthew5876: You should never add const to pass by value parameter in a declaration because it does nothing.

Ghabry: oh I didn't see that this is "by value". Its getting late. Right, makes sense.

Ghabry:
More Todos:
The autobattle options must be added to the manpage and to the "--help" output and needs config file support.

For later (editor, pls):
Imo this belongs in LDB for a default game value and needs a new event command "Select Battle Algorithm" which is stored in LSD.

fmatthew5876:
I created `Game_ConfigPlayer` for this and other related future items, and wired it up. Added the manpages etc.. too.

I don't think we need anything else for this iteration. Now this just has to be tested.

fmatthew5876:
I checked all the RPG_RT versions and found something really strange. Every version of the engine uses double division for maxhp, maxsp, etc.. math.

*except* 2ke. For some reason I can't grasp, it looks like 2ke developers actually decided to go into the auto battle code and switch to using integer division.

Other than that there are no differences. 

Given the amount of randomness here and the pain required to support both integer and double (templates), I'm inclined to leave it with double division.

Ghabry:
Maybe @cherrydt as 2ke developer can give some insight here?
Could be also some strange compiler behaviour maybe.

CherryDT:
Weird indeed, I didn't change anything. How is the behavior in the Japanese Value!+ version 1.52? This is the version that I changed to become 1.60. Unlike 2k3e, I actually changed the code and compiled from source - it should be the same compiler (Delphi 6) but maybe originally some different compiler flags were used or something (although I'd expect those settings to be stored in the project file)... If it changed only between 1.52 and 1.60 then it's a bit of a mystery...

fmatthew5876:
I got an RPG_RT binary from here https://tkool.jp/products/trial.html which I believe is 1.5.2 JP?

I checked that one and it is using double math.

CherryDT:
I went to check what's going on there, and there is no change in Git for the relevant file. So it must be before I got the code. I wonder if it's indeed a subtle compiler difference or something like that because the calculation (if that's the one you meant, that is) looks like this (in pseudocode):

```c++
x = min(p, maxHp - hp) / maxHp;
```

`x` is a double, but `p`, `maxHp` and `hp` are integers. Now I wonder if maybe the compiler version I used does this using integer division because all terms in the formula are integers, while the compiler version used by the original devs used double division because the result of the calculation is assigned to a double variable? Not sure.

But actually that doesn't sound right either, because I checked the help file in my Delphi version and it says that `/` should _always_ do double division (as opposed to `div` which does integer division):

image

🤷‍♂️

fmatthew5876:
Yes, that is the division 

> I went to check what's going on there, and there is no change in Git for the relevant file. So if anything _was_ changed, it must have happened before I got the code.
> 
> I wonder if it's indeed a subtle compiler difference or something like that because the calculation (if that's the one you meant, that is) looks like this (in pseudocode):
> 
> ```c++
> x = min(p, maxHp - hp) / maxHp;
> ```
> 
> `x` is a double, but `p`, `maxHp` and `hp` are integers. Now I wonder if maybe the compiler version I used does this using integer division because all terms in the formula are integers, while the compiler version used by the original devs used double division because the result of the calculation is assigned to a double variable? Not sure.

Yes that is the division I'm talking about, that and the `cost / maxSp` that happens for skills.

I guess this one will remain a mystery..

fmatthew5876:
Sorry but it looks like this was a red herring. I was looking at the wrong disassembly. It indeed appears that double division is being used by all version, including 2ke.

Sorry about that

rueter37:
I have checked out your auto battle implementation and I like it so far! But I noticed some weird behavior in RPG Maker 2003 battles: In auto battle mode, after an actor executed his automatic selected action, the game switches to manual battle mode and even grants the actor an extra turn (automatic and manual action in the same turn despite ATB being zero after the automatic action).

fmatthew5876:
Thanks, I fixed the auto battle 2k3 bug you mentioned.

As far as I'm concerned this PR is finished. I will follow up with unit tests and any related bug fixes in #2326

src/autobattle.h
```diff
@@ -0,0 +1,192 @@
...
+ * @param source the user of the skill
+ * @param target the target of the skill
+ * @param cond the battle condition
+ * @param apply_variance If true, apply variance to the damage
+ * @param emulate_bugs Emulate all RPG_RT bugs for accuracy
+ */
+double CalcNormalAttackAutoBattleTargetRank(const Game_Actor& source, const Game_Battler& target, Game_Battler::Weapon weapon, lcf::rpg::System::BattleCondition cond, bool apply_variance, bool emulate_bugs);
+
+/**
+ * Calcuate the auto battle total effectiveness rank of using a normal attack.
```
fdelapena, line 149: Calculate typo (5 found)

fdelapena (approved):

src/autobattle.h
fmatthew5876: Fixed, thanks!

Ghabry:
Jenkins: test this please

fmatthew5876:
I've added a start for some unit tests. This is as far as I will go on this PR.

src/scene_battle.cpp
```diff
@@ -89,7 +90,9 @@ void Scene_Battle::Start() {
 		return;
 	}
 
-	Output::Debug("Starting battle {} ({})", troop_id, troop->name);
+	autobattle_algo = AutoBattle::CreateAlgorithm(Player::player_config.autobattle_algo.Get());
+
+	Output::Debug("Starting battle {} ({}) : autobattle={}", troop_id, troop->name, autobattle_algo->GetName());
```
Ghabry, line 95: isn't it a bit verbose to print the constant autobattle algo everytime a battle starts?

btw is the algorithm name case sensitive or can I also specify "attack" on the command line?

fmatthew5876: My original design was printing this once on startup in it's own logfile line. I was thinking of future extensibility where you could change the algo at runtime, but since we don't have that yet this is a bit overkill.

The algo name is case sensitive, I can fix that

Ghabry:
Sorry I wanted to review this and comparing with RPG_RT on the weekend but I'm away from my dev-station until wednesday so this has to wait a bit longer.

src/scene_battle.cpp
fmatthew5876: Added the case insensitive name compare. I left the log message in as it seems to me it'll be better for debugging when the algo is changeable in scene settings etc..

That being said if you feel strongly I'll remove it.

src/autobattle.cpp
```diff
@@ -0,0 +1,376 @@
...
+			return 0.0;
+		}
+
+		const double base_effect = Algo::CalcSkillEffect(source, target, skill, apply_variance);
+		const double max_effect = std::min(base_effect, tgt_max_hp - tgt_hp);
+
+		auto rank = static_cast<double>(max_effect) / static_cast<double>(tgt_max_hp);
+		if (src_max_sp > 0) {
+			const double cost = source.CalculateSkillCost(skill.ID);
+			// Note: RPG_RT 2ke only uses integer division for cost / src_max_sp here.
```
Ghabry, line 103: sure? You checked the assembly?

resources/unix/bash-completion/easyrpg-player
```diff
@@ -34,6 +35,11 @@ _easyrpg-player ()
       #_filedir '@(lmu|emu)'
       return
       ;;
+    # load map files
```
Ghabry, line 38: load map files?

src/autobattle.cpp
```diff
@@ -0,0 +1,376 @@
...
+		if (!skill.affect_hp) {
+			return 0.0;
+		}
+
+		const double base_effect = Algo::CalcSkillEffect(source, target, skill, apply_variance);
+		const double max_effect = std::min(base_effect, tgt_max_hp - tgt_hp);
+
+		auto rank = static_cast<double>(max_effect) / static_cast<double>(tgt_max_hp);
+		if (src_max_sp > 0) {
+			const double cost = source.CalculateSkillCost(skill.ID);
```
Ghabry, line 102: For ``CalculateSkillCost`` RPG_RT appears to ignore the "Half SP" (or I missed it). Maybe also emulate this bug?

src/autobattle.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+	const double base_effect = Algo::CalcSkillEffect(source, target, skill, apply_variance);
+	// Note: RPG_RT 2ke only uses integer division for effect / tgt_hp here.
+	rank = std::min(base_effect, tgt_hp) / tgt_hp;
+	if (rank == 1.0) {
+		rank = 1.5;
+	}
+	if (src_max_sp > 0) {
+		const double cost = source.CalculateSkillCost(skill.ID);
+		// Note: RPG_RT 2ke only uses integer division for cost / src_max_sp here.
```
Ghabry, line 141: same two comments apply here

src/autobattle.cpp
```diff
@@ -0,0 +1,376 @@
...
+double CalcSkillHealAutoBattleTargetRank(const Game_Actor& source, const Game_Battler& target, const lcf::rpg::Skill& skill, bool apply_variance, bool emulate_bugs) {
+	assert(skill.type == lcf::rpg::Skill::Type_normal || skill.type >= lcf::rpg::Skill::Type_subskill);
+	assert(skill.scope == lcf::rpg::Skill::Scope_self || skill.scope == lcf::rpg::Skill::Scope_ally || skill.scope == lcf::rpg::Skill::Scope_party);
+
+	const double src_max_sp = source.GetMaxSp();
+	const double tgt_max_hp = target.GetMaxHp();
+	const double tgt_hp = target.GetHp();
+
+	if (target.GetHp() > 0) {
+		// Can the skill can heal the target?
```
Ghabry, line 92: remove 2nd can

src/player.cpp
```diff
@@ -1212,6 +1215,11 @@ R"(EasyRPG Player - An open source interpreter for RPG Maker 2000/2003 games.
       --start-map-id N     Overwrite the map used for new games and use.
                            MapN.lmu instead (N is padded to four digits).
                            Incompatible with --load-game-id.
+      --autobattle-algo A  Which AutoBattle algorithm to use.
+                           Possible options:
+                            RPG_RT     - The default RPG_RT compatible algo, including RPG_RT bugs
+                            RPG_RT+    - The default RPG_RT compatible algo, with bug fixes
+                            ATTACK     - RPG_RT+ but only physical attacks, no skills
```
Ghabry, line 1222: The lines are too long (>80) but this is a problem in many parts of the help output

src/scene_battle.cpp
Ghabry: well keep it but please get rid of the whitespace before the ":". (Also imo a comma or a semi-colon would fit better here.)

Ghabry (changes requested):
some remarks, but imo the algorithms are correct

src/autobattle.cpp
fmatthew5876: Left over comment I forgot to remove

src/autobattle.cpp
fmatthew5876: Nice catch, you're right this is another RPG_RT bug

src/player.cpp
fmatthew5876: Would like to punt until we have a wholistic answer to this one

fmatthew5876:
Addressed review comments. I also found and fixed an off by 1 bug in `CalculateSpCost()`.

Ghabry (approved):

PR #2332, opened 2020-09-12
Fix enemy default state rank

rueter37:
The default state rank of an enemy is only "B" instead of "C". All the other default ranks (enemy attribute rank, player state rank and player attribute rank) are already correct.

### DIFF
--- a/src/game_enemy.cpp
+++ b/src/game_enemy.cpp
@@ -59,7 +59,7 @@ int Game_Enemy::MaxStatBaseValue() const {
 }
 
 int Game_Enemy::GetStateProbability(int state_id) const {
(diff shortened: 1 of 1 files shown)


fmatthew5876:
Nice catch!

fmatthew5876 (approved):

fdelapena (approved):

Ghabry (approved):
Confirmed, the chunk is not written for the B-case

Ghabry:
@fmatthew5876 
Also liblcf should be checked if we handle this correctly.

fmatthew5876:
I believe it does, considering that we tested reading and writing an empty database.

PR #2333, opened 2020-09-12
Battle: Fix output text of item drops

rueter37:
I stumbled upon an oddity in relation with the output text of enemy item drops: After the enemy party gets defeated and they drop items the output says funny things like "U received!" instead of the actual item name. Changing `item->name` to `item->name.c_str()` fixes the output but I would like to hear your opinions if this is the right way to fix this.

### DIFF
--- a/src/scene_battle_rpg2k.cpp
+++ b/src/scene_battle_rpg2k.cpp
@@ -1496,7 +1496,7 @@ void Scene_Battle_Rpg2k::PushItemRecievedMessages(PendingMessage& pm, std::vecto
 	for (std::vector<int>::iterator it = drops.begin(); it != drops.end(); ++it) {
 		const lcf::rpg::Item* item = lcf::ReaderUtil::GetElement(lcf::Data::items, *it);
 		// No Output::Warning needed here, reported later when the item is added
-		StringView item_name = item ? item->name : "??? BAD ITEM ???";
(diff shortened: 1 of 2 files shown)


Ghabry (approved):
I think this is a missed case of the StringView refactor.
Better check also for other unnecessary c_str calls.

fmatthew5876:
Thanks for finding this. The bug and it's fix make no sense to me though, the bug here might be much deeper issues. I will investigate

fmatthew5876:
Ouch.. so I wasn't aware of this gotcha with `string_view`:

https://stackoverflow.com/questions/56634198/why-does-stdstring-view-create-a-dangling-view-in-a-ternary-expression

I tested this case with address sanitizer and it is memory corruption. I really hate this because as far as I can tell there is no compile time way to catch these kinds of bugs :(

The slightly better fix is this one:

```
StringView item_name = item ? ToStringView(item->name) : "??? BAD ITEM ???";
```

fmatthew5876:
I did a quick grep through the code, the only other place this bug appears is in `game_battleralgorithm.cpp` in several places.

rueter37:
Your proposal worked for me. I amended the commit and force-pushed it.

fmatthew5876 (approved):

fmatthew5876:
So the other ones in battlealgorithm are ok because both types are `DBString`

It's only when you mix DBString with something else like `const char*` this happens :/

fdelapena (approved):

PR #2339, opened 2020-09-16
Remove SaveActor Fixup and Setup - depends on liblcf #399

fmatthew5876:
Depends on: https://github.com/EasyRPG/liblcf/pull/399

DynRPG Addresses
---------------------

| Address | Class | Function | Comment |
| -- | -- | -- | -- |
| 004B5330 | Actor | Reset | Resets all fields to default values |
| 004B6DD8 | Actor | SetTitle |
| 004B7004 | Actor | SetFace |
| 004B6D34 | Actor | SetName |
| 004B6E58 | Actor | GetCharset |
| 004B6E94 | Actor | GetCharsetId |
| 004B6EBC | Actor | GetCharsetTransparency |
| 004B6D88 | Actor | GetTitle |
| 004B6FDC | Actor | GetFaceId |
| 004B6FA0 | Actor | GetFaceName |
| 004B6CE4 | Actor | GetName |
| 004B5A48 | Actor | InitSaveChunk | Maps save chunk ids to fields

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -76,7 +76,11 @@ Game_Actor::Game_Actor(int actor_id) {
 	MakeExpList();
 	SetBattlePosition(GetOriginalPosition());
 
-	ChangeLevel(dbActor->initial_level, nullptr);
+	data.level = 0;
+	if (dbActor->initial_level > 0) {
+		// For games like COLORS: Lost Memories which use level 0, don't change level because it'll clamp to 1.
+		ChangeLevel(dbActor->initial_level, nullptr);
+	}
 	SetHp(GetMaxHp());
 	SetSp(GetMaxSp());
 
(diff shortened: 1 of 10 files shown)


fmatthew5876:
Rebased and retested this, including the minor battle test cleanup.

This is ready

fdelapena (approved):

src/game_actor.cpp
```diff
@@ -58,62 +58,66 @@ Game_Actor::Game_Actor(int actor_id) {
...
-	RemoveInvalidData();
+	MakeExpList();
+	SetBattlePosition(GetOriginalPosition());
 
-	if (GetLevel() > 0) {
-		SetHp(GetMaxHp());
-		SetSp(GetMaxSp());
-		SetExp(exp_list[GetLevel() - 1]);
-	}
+	ChangeLevel(dbActor->initial_level, nullptr);
```
Ghabry, line 78: Needs a test for COLORS: Lost Memories (there is a comment in this file somewhere) which uses Level 0.

Ghabry:
btw the only reason why the title scene invokes the battle test is because newer engines do it this way.

fmatthew5876:
> btw the only reason why the title scene invokes the battle test is because newer engines do it this way.

Given that many things could potentially depend on the title scene existing in the stack, this is a safer way to do it anyway.

src/game_actor.cpp
fmatthew5876: Retested and fixed a bug with this. Thanks for raising it.

Ghabry (approved):

Ghabry:
jenkins: test this please

PR #2342, opened 2020-09-16
Fix crash caused by IsReflect

### LINKED ISSUE #2340: rm2k - Game shutdown after switch skill in a battle

---

Hi! I'm developing a game in rpg maker 2000 and I use to test it in EasyRPG. 

About 2 days ago, while I was playing the game on my cel phone, during a battle event, I used a skill that is activated by a switch (it shows a battle animation) and the game just shutdown!

The most weird is that it wasn't happening before! Apparently It only happens when the player use a skill that is triggered by a switch during a battle. 

I'm not sure if it is being caused by some EasyRPG update, but I noticed that the program layout is a bit different and some days ago, this bug wasn't happening.. maybe it is.

I'd really apreciate if it could be give some attention to this.

Thanks.

rueter37:
This PR fixes #2340. The IsReflect() method caused crashes because it tried to access targets even when they doesn't exist (Switch skills never have targets).


src/game_battlealgorithm.cpp
```diff
@@ -1428,6 +1428,8 @@ bool Game_BattleAlgorithm::Skill::IsReflected() const {
 			if (party_target) {
 				std::vector<Game_Battler*> targetlist;
 				party_target->GetActiveBattlers(targetlist);
+				// Return false if no targets exist
+				if (targetlist.size() == 0) return false;
```
fmatthew5876, line 1432: Is this needed? The loop below will already not iterate anything if the list is empty

src/game_battlealgorithm.cpp
```diff
@@ -1439,6 +1441,8 @@ bool Game_BattleAlgorithm::Skill::IsReflected() const {
 					}
 				}
 			} else {
+				// Return false if no target exist
```
fmatthew5876, line 1444: How about just `else if (GetTarget() != nullptr)` and then fall through if no targets

src/game_battlealgorithm.cpp
rueter37: You are right, this check is unnecessary.

src/game_battlealgorithm.cpp
rueter37: Your proposal works for me. I changed the code according to your suggestion.

rueter37:
Updated the PR.

fmatthew5876 (approved):

fdelapena (approved):

Ghabry (approved):

PR #2343, opened 2020-09-16
States: Damage and recovery is rounded down instead of up

rueter37:
If an actor get damaged or recovered due to a state, the damage/recovery value is rounded down instead of up. This means that the damage/recovery value can even be zero if the damage/recovery percentage is set to a low value. Moreover a bug got fixed in SP damage/recovery: The SP damage/recovery percentage uses now the maximum SP instead of the maximum HP for the calculation.

### DIFF
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -384,8 +384,8 @@ int Game_Battler::ApplyConditions() {
 	for (int16_t inflicted : GetInflictedStates()) {
 		// States are guaranteed to be valid
 		lcf::rpg::State& state = *lcf::ReaderUtil::GetElement(lcf::Data::states, inflicted);
-		int hp = state.hp_change_val + (int)(std::ceil(GetMaxHp() * state.hp_change_max / 100.0));
-		int sp = state.sp_change_val + (int)(std::ceil(GetMaxHp() * state.sp_change_max / 100.0));
+		int hp = state.hp_change_val + (GetMaxHp() * state.hp_change_max / 100);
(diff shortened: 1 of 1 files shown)


Ghabry (commented):
@fmatthew5876 
Please check if you see anything related to Delphi rounding here

src/game_battler.cpp
```diff
@@ -384,8 +384,8 @@ int Game_Battler::ApplyConditions() {
 	for (int16_t inflicted : GetInflictedStates()) {
 		// States are guaranteed to be valid
 		lcf::rpg::State& state = *lcf::ReaderUtil::GetElement(lcf::Data::states, inflicted);
-		int hp = state.hp_change_val + (int)(std::ceil(GetMaxHp() * state.hp_change_max / 100.0));
-		int sp = state.sp_change_val + (int)(std::ceil(GetMaxHp() * state.sp_change_max / 100.0));
+		int hp = state.hp_change_val + (GetMaxHp() * state.hp_change_max / 100.0);
+		int sp = state.sp_change_val + (GetMaxSp() * state.sp_change_max / 100.0);
```
fmatthew5876, line 388: There is no rounding done here by RPG_RT so this change is correct. However, RPG_RT also does integer division. Please change `100.0` to `100` and this will be correct.

rueter37: Did the change and updated this PR.

fmatthew5876 (approved):

fdelapena (approved):

PR #2349, opened 2020-09-23
Actor parameter cleanups / optimizations - prep for battle algo refactor

fmatthew5876:
~~Depends on #2339~~

Testing
----------

- [x] Battle algo hp damage
- [x] Menu skill healing
- [x] state hp and sp walking
- [x] event hp damage non-lethal
- [x] event hp damage lethal
- [x] state battle damage
- [x] terrain damage

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -628,11 +628,11 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 	if (GetAffectedHp() != -1 && !was_dead) {
 		int hp = GetAffectedHp();
 		int target_hp = target->GetHp();
-		target->ChangeHp(IsPositive() ? hp : -hp);
+		target->ChangeHp(IsPositive() ? hp : -hp, true);
 		if (IsAbsorb() && !IsPositive()) {
 			// Only absorb the hp that were left
 			int src_hp = std::min(target_hp, hp);
-			source->ChangeHp(src_hp);
+			source->ChangeHp(src_hp, true);
 		}
 	}
 
@@ -685,7 +685,7 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 	if (IsPositive() && was_dead && !target->IsDead()) {
 		if (GetAffectedHp()) {
 			int hp = GetAffectedHp();
-			target->ChangeHp(hp - 1);
+			target->ChangeHp(hp - 1, true);
 		}
 	}
 }
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -917,104 +917,105 @@ bool Game_BattleAlgorithm::Normal::Execute() {
 	auto crit_chance = Algo::CalcCriticalHitChance(source, target, Game_Battler::WeaponAll);
 
 	// Damage calculation
-	if (Rand::PercentChance(to_hit)) {
-		if (Rand::PercentChance(crit_chance)) {
-			critical_hit = true;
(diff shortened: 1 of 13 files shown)


fmatthew5876:
Tested everything. Once #2339 goes in this is ready

src/game_battler.cpp
```diff
@@ -659,3 +665,31 @@ void Game_Battler::Flash(int r, int g, int b, int power, int frames) {
...
+	return spi_modifier - prev;
+}
+
+int Game_Battler::ChangeAgiModifier(int modifier) {
+	const auto prev = agi_modifier;
+	const auto base = GetBaseAgi();
+	SetAgiModifier(Utils::Clamp(agi_modifier + modifier, base / 2, base * 2));
+	return agi_modifier - prev;
+}
+
```
rueter37, line 695: I have tested your changes and noticed weird behaviors in relation to stat changes: As soon a stat gets changed (no matter if buff or debuff), the affected stat is then at least 1.5x the base value which makes debuffs no longer work. The clamping makes the modifier at least half of the base value, this explains the 1.5x issue. The modifier must be in the range between `-base / 2` and `+base` because unchanged stats have a modifier of 0 and the calculation of a stat is `base + modifier`.

rueter37 (commented):
The changes look good so far. But I noticed some weird behavior in relation to stat changes (see comment) and the changes from #2351 are missing. (But to be fair the review request came earlier than the update in #2351)

src/game_battler.cpp
fmatthew5876: Fixed and added unit tests

src/game_battlealgorithm.cpp
```diff
@@ -918,103 +918,104 @@ bool Game_BattleAlgorithm::Normal::Execute() {
 
 	// Damage calculation
 	if (Rand::PercentChance(to_hit)) {
-		if (Rand::PercentChance(crit_chance)) {
-			critical_hit = true;
-		}
+		this->success = false;
+		return this->success;
+	}
```
rueter37, line 923: Because `this->success` is set to false here after a successful accuracy check, normal attacks almost always miss now (tested it myself). It looks like the check is supposed to be inverted, using `!Rand::PercentChance(to_hit)` fixes this.

fmatthew5876: Fixed

src/game_battlealgorithm.cpp
```diff
@@ -627,22 +627,21 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
...
+			source->ChangeHp(-hp, false);
 		}
 	}
 
 	if (GetAffectedSp() != -1) {
 		int sp = GetAffectedSp();
-		int target_sp = target->GetSp();
-		target->SetSp(target->GetSp() + (IsPositive() ? sp : -sp));
+		sp = IsPositive() ? sp : -sp;
+		sp = target->SetSp(sp);
```
rueter37, line 641: Using `target->SetSp(sp)` unfortunately breaks SP changes. Changing this to `target->ChangeSp(sp)` fixes this. After this, everything should work then and I will approve this PR.

fmatthew5876: That was a rebase error, fixed.

rueter37 (approved):

fmatthew5876:
jenkins: test this please

fdelapena (approved):

Ghabry:
I invoked 3ds and Vita PR builders here because the clamping could fail to compile in newlib

PR #2350, opened 2020-09-23
States: Allow states simultaneously with death

### LINKED ISSUE #2327: Allow states simultaneously with death
In player whenever we add death state we clear all other states. In addition, our battle algorithms are setup to abort early if death is added.

This behavior is incorrect. 

RPG_RT actually lets you have other states simultaneously with death. This is however uncommon, because death always has priority 100, and so any other state which is allowed to coexist must have priority 90 or higher.

Simply create another state with priority=100 and do this event code:
```
AddState: OtherState
AddState: Death
RemoveState:Death
```

The actor will still have OtherState

- [ ] Adding states ignores death
- [ ] Skills will abort all effects if the damage component kills the target

rueter37:
This PR fixes #2327. If an actor gets the death state every already inflicted state with a priority of >= 91 no longer gets removed. Moreover skills and weapons adds states only if the target does not get killed by the damage and weapons can remove states in RPG Maker 2003 now.

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -975,7 +975,7 @@ bool Game_BattleAlgorithm::Normal::Execute() {
 			// Conditions caused / healed by weapon.
 			if (source.GetType() == Game_Battler::Type_Ally) {
 				auto& ally = static_cast<Game_Actor&>(source);
-				const bool is2k3 = Player::IsRPG2k();
+				const bool is2k3 = Player::IsRPG2k3();
 				auto* weapon1 = ally.GetWeapon();
 				auto* weapon2 = ally.Get2ndWeapon();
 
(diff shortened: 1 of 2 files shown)


src/game_battlealgorithm.cpp
```diff
@@ -975,7 +975,7 @@ bool Game_BattleAlgorithm::Normal::Execute() {
 			// Conditions caused / healed by weapon.
 			if (source.GetType() == Game_Battler::Type_Ally) {
 				auto& ally = static_cast<Game_Actor&>(source);
-				const bool is2k3 = Player::IsRPG2k();
+				const bool is2k3 = Player::IsRPG2k3();
```
fmatthew5876, line 978: Ouch, nice catch!

src/game_battlealgorithm.cpp
```diff
@@ -1304,39 +1303,41 @@ bool Game_BattleAlgorithm::Skill::Execute() {
...
 		// Conditions:
-		bool heals_states = IsPositiveSkill() ^ (Player::IsRPG2k3() && skill.reverse_state_effect);
-		for (int i = 0; i < (int) skill.state_effects.size(); i++) {
-			if (!skill.state_effects[i])
-				continue;
-			auto state_id = i + 1;
+		// If the target gets killed by damage, do not add or remove states
+		if (!lethal) {
+			bool heals_states = IsPositiveSkill() ^ (Player::IsRPG2k3() && skill.reverse_state_effect);
+			for (int i = 0; i < (int) skill.state_effects.size(); i++) {
```
fmatthew5876, line 1309: Change to `static_cast<int>()`. Some compilers warn about C casts

fmatthew5876 (approved):

src/game_battlealgorithm.cpp
rueter37: Did the change and updated this PR.

PR #2351, opened 2020-09-23
Battle: Absorb only HP if the target gets killed by the absorption

rueter37:
If a target gets killed by an absorption skill, then only the HP are absorbed.

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -640,7 +640,7 @@ void Game_BattleAlgorithm::AlgorithmBase::Apply() {
 		int sp = GetAffectedSp();
 		int target_sp = target->GetSp();
 		target->SetSp(target->GetSp() + (IsPositive() ? sp : -sp));
-		if (IsAbsorb() && !IsPositive()) {
+		if (IsAbsorb() && !IsPositive() && !IsKilledByDamage()) {
 			int src_sp = std::min(target_sp, sp);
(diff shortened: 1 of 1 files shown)


fmatthew5876 (approved):

Ghabry:
Slowly this should be refactored..... except for the stat type this is 4 times the same code which becomes more complex with each PR.

fmatthew5876:
> Slowly this should be refactored..... except for the stat type this is 4 times the same code which becomes more complex with each PR.

It's coming in my next PR, don't worry :)

fmatthew5876:
@rueter37 Are you certain that absorption even works at all for atk, def, spi, and agi?

I'm looking at RE and I don't see this at all. I only see absorb logic for HP and SP.

I checked both dynrpg and 2k3e.

fmatthew5876:
Ok I just tested this myself. 

1. Attack enemy ~ 260 dmg
2. Use a skill which absorbs 9999 atk
3. Attack enemy ~ 260 dmg
4. Use skill on self which raises atk by 9999
5. Attack enemy ~ 750 dmg

So clearly absorb does not work for parameters outside of hp and sp
I used a skill which absorbs 9999 atk.

rueter37:
Did some retests and confirmed you are right. Attack, defense, spirit and agility really cannot be absorbed even in RPG Maker 2003. I have updated this PR.

fmatthew5876 (approved):

PR #2354, opened 2020-09-25
RPG Maker 2003 battle scene changes

rueter37:
This PR will be updated as soon as #2399 and #2361 are merged into master. The planned changes are in [rueter37:scene-battle2k3-changes-new](https://github.com/rueter37/Player/tree/scene-battle2k3-changes-new).

Depends on: #2399 because #2361 depends on it and #2361 because of changes in the battle window made there.

This PR does the following changes (after the update):

- The status window components have been rearranged to match the RPG_RT style.
- The transparent windows use an alpha of 160 instead of 128 now.
- The gauge display is more accurate now but still needs work. Getting this to 100% accuracy requires a pixman expert. Picture scaling is a general problem as stated in #446.
- In traditional style battles two bugs have been fixed: The SP window got printed over when a skill which costs SP has been used and the HP value didn't change its color if the HP reached critical or zero.
- In alternative style battles the gauges are now shown transparent if the window is opaque.
- The ally cursor is always drawn 40 pixels above the Y-Position of the battler. (Tested with RPG_RT, but correct me if this is wrong nevertheless)
- ~~In the alternative and gauge battle types it is now possible to switch between actors with full ATB gauge.~~ (Fixed in #2399)
- ~~In the traditional battle type the player is forced to battle manually now.~~ (Fixed in #2399)
- ~~The ally cursor is only shown in the "Select_Actor" and "Select_AllyTarget" states.~~ (Fixed in #2399)
- ~~Actors with a "do nothing" restriction due to a state cannot do an action with full ATB anymore.~~ (Fixed in #2399)

### DIFF
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -268,6 +268,12 @@ void Scene_Battle_Rpg2k3::Update() {
 					Game_Battle::UpdateAtbGauges();
 				}
 
+				if (state == State_SelectCommand) {
+					if (active_actor->GetSignificantRestriction() != lcf::rpg::State::Restriction_normal) {
+						SetState(State_SelectActor);
+					}
+				}
+
 				if (state != State_SelectEnemyTarget) {
 					int old_state = state;
 
@@ -1645,7 +1651,7 @@ void Scene_Battle_Rpg2k3::SelectNextActor() {
 
 	int i = 0;
 	for (auto* battler: actors) {
-		if (battler->IsAtbGaugeFull() && !battler->GetBattleAlgorithm() && battle_actions.empty()) {
+		if (battler->IsAtbGaugeFull() && !battler->GetBattleAlgorithm() && battle_actions.empty() && battler->GetSignificantRestriction() != lcf::rpg::State::Restriction_do_nothing) {
 			actor_index = i;
 			active_actor = static_cast<Game_Actor*>(battler);
 
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -1850,7 +1850,7 @@ void Scene_Battle_Rpg2k3::ShowNotification(std::string text) {
 		return;
 	}
 	help_window->SetVisible(true);
-	help_window->SetText(std::move(text));
+	help_window->SetText(std::move(text), Text::AlignLeft, false);
 }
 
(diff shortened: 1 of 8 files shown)


fmatthew5876:
Does it need to depend on the other window changes? It's very unclear if those are correct or not and would take some time to verify.

Ghabry:
Okay this is a battle PR I'm interested in so will also give some feedback later :)

Initial warning:
This "enemy gets the turn while in the menu" can have many subtile race conditions because enemy state changes while navigating in the menu which the code was not designed for!

rueter37:
@fmatthew5876: About the dependencies: This PR depends on the commits 4d41077, 932302e and c2da261 from #2313.

Ghabry:
when it only depends on a few window commits you can "git cherry-pick COMMIT-ID" them instead.

rueter37:
This PR no longer depends on #2313 now.

src/scene_battle_rpg2k3.cpp
```diff
@@ -376,7 +376,7 @@ void Scene_Battle_Rpg2k3::CreateUi() {
 	}
 
 	if (lcf::Data::battlecommands.battle_type != lcf::rpg::BattleCommands::BattleType_traditional) {
-		int transp = lcf::Data::battlecommands.transparency == lcf::rpg::BattleCommands::Transparency_transparent ? 128 : 255;
+		int transp = lcf::Data::battlecommands.transparency == lcf::rpg::BattleCommands::Transparency_transparent ? 160 : 255;
```
fmatthew5876, line 381: Hard to be 100% sure because my tool doesn't work well for this code. But 160 appears correct.

fmatthew5876:
There is *a lot* going on here. In order to make this more manageable, could you please split this into 3 PRs. This way we can review and finalize the stuff that is easier to confirm and get it in, while not getting blocked on the more difficult items.

1. Notification window timing and message support. See #2357 for more info on how to do this correctly.
2. Implement ATB active mode
3. The rest of the stuff here, is probably ok to stay in this PR for now.

fdelapena (approved):

rueter37:
I think we can close my PR here in favor of #2399 because the battle scenes get fully rewritten there.

fmatthew5876:
I have not reviewed the PR in detail. There may be things here that I have not fixed yet.

rueter37:
Closing this PR because it is outdated and interferes with #2399.

PR #2356, opened 2020-09-25
Fix compilation with SDL_mixer 1.2

gameblabla:
This is a fix for compiling sdl_audio_mixer.cpp against SDL_Mixer 1.2.
Otherwise, this will happen (with GCC 10.2)
[https://github.com/EasyRPG/Player/files/5285677/errors.txt](https://github.com/EasyRPG/Player/files/5285677/errors.txt)

I had to do this because using ints instead of size_t will break SDL_Mixer 2.0 instead.

This commit is mostly useful with GCC 10+ and SDL 1.2 (like a modern linux distro or a custom toolchain like mine or this one https://github.com/WerWolv/rg350_buildroot_gcc10 for opendingux).

However, i realize that this might not be ideal and could make the source code clunky if we keep doing stuff like that.

### DIFF
--- a/src/audio_sdl_mixer.cpp
+++ b/src/audio_sdl_mixer.cpp
@@ -50,7 +50,11 @@ namespace {
 		return -1;
 	}
 
+#if SDL_MAJOR_VERSION>1
 	Sint64 SDLCALL vio_seek(struct SDL_RWops * context, Sint64 offset, int whence) {
+#else
+	int SDLCALL vio_seek(struct SDL_RWops * context, int offset, int whence) {
+#endif
 		auto stream = reinterpret_cast<Filesystem_Stream::InputStream*>(context->hidden.unknown.data1);
 
 		stream->seekg(offset, Filesystem_Stream::CSeekdirToCppSeekdir(whence));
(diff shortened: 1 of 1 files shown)


Ghabry:
Thanks.
Please note that we are close to deprecating SDL mixer. The only remaining use for it was native Midi on Windows and macOS. 
I recommend using SDL instead.
Especially the SDL1 Mixer is extremely bad. Without the recommended audio libraries (mpg123, speexdsp etc) half of the music will play too slow, too fast, is glitched or crashes often (MP3).

gameblabla:
> I recommend using SDL instead.
> Especially the SDL1 Mixer is extremely bad. Without the recommended audio libraries (mpg123, speexdsp etc) half of the music will play too slow, too fast, is glitched or crashes often (MP3).
Honestly i would have loved to but disabling SDL_mixer breaks support for WAV files, despite being compiled against MPG123, LIBSAMPLERATE and LIBSNDFILE and those being enabled.
(i even checked manually by removing those from LDFLAGS to make sure they do try to link against them and they do)

EasyRPG complains about the sound format not being supported in the command line instead of trying to play them back.
Should i just open a bug report about it instead of trying to push this fix ?

Ghabry:
This sounds odd. Can you provide me with the list of defines?
Maybe you are using ``-DWANT_FASTWAV``? This one makes decoding of some WAVs fail.

gameblabla:
> This sounds odd. Can you provide me with the list of defines?
> Maybe you are using `-DWANT_FASTWAV`? This one makes decoding of some WAVs fail.
Hmm, well now it works after disabling it even though it didn't work before when i tried to disable that... (maybe i was missing samplerate)
I was using those defines :
`-DGCW_ZERO -DHAVE_FREETYPE=1 -DHAVE_LIBSAMPLERATE=1 -DHAVE_LIBSNDFILE=1 -DHAVE_MPG123=1  -DSUPPORT_AUDIO=1 -DWANT_FMMIDI=1
`

I guess i should close this ?

Ghabry:
I could accept the patch but I don't consider it very useful anymore. Feel free to close it.

The define list looks fine. By now we also added "FluidLite" (FluidSynth)  (-DHAVE_FLUIDLITE).
You can put a easyrpg.soundfont (sf2) in the game directory.

PR #2359, opened 2020-09-26
Allow Haiku build

kwyxz:
This allows building the EasyRPG core on Haiku.

### DIFF
--- a/src/platform.cpp
+++ b/src/platform.cpp
@@ -21,6 +21,16 @@
 #include <cassert>
 #include <utility>
 
+#ifndef DT_UNKNOWN
+#define DT_UNKNOWN 0
+#endif
+#ifndef DT_REG
+#define DT_REG DT_UNKNOWN
+#endif
+#ifndef DT_DIR
+#define DT_DIR DT_UNKNOWN
+#endif
+
 Platform::File::File(std::string name) :
 #ifdef _WIN32
 		filename(Utils::ToWideString(name))


Ghabry:
I guess the problem is that DT_DIR is not defined or what is the compiler error?

Please do it like for PSP2: add the guard to line 163 and add your code to the function at line 186.

You are calling stat again. Is really necessary? Please check the directory entry structure of the haiku system if it provides anything useful concerning the file type.

Ghabry (changes requested):
-

kwyxz:
> I guess the problem is that DT_DIR is not defined or what is the compiler error?

The issue is d_type only exists on Linux and is not POSIX. I'll look at how it's done for PSP2.

Ghabry:
@kwyxz 

Can you try this much simpler patch? It should make the compiler happy.
In theory this should work. We already have a codepath that calls ``stat`` when ``.d_type`` is missing.

```
diff --git a/src/platform.cpp b/src/platform.cpp
index 181058ca..0c105b80 100644
--- a/src/platform.cpp
+++ b/src/platform.cpp
@@ -21,6 +21,16 @@
 #include <cassert>
 #include <utility>
 
+#ifndef DT_UNKNOWN
+#define DT_UNKNOWN 0
+#endif
+#ifndef DT_REG
+#define DT_REG DT_UNKNOWN
+#endif
+#ifndef DT_DIR
+#define DT_DIR DT_UNKNOWN
+#endif
+
 Platform::File::File(std::string name) :
 #ifdef _WIN32
 		filename(Utils::ToWideString(name))
```

kwyxz:
Thank you for the suggestion, it builds fine.

Ghabry:
Jenkins: Test this please

Ghabry (approved):
I squashed the other non-relevant commits away and only left the last working one.

fdelapena (approved):

fmatthew5876 (approved):

Ghabry:
@kwyxz 
I picked this one for the next "stable" release (0.6.2.3 https://github.com/EasyRPG/Player/tree/0-6-2-stable). When it is released in ca. one week you can use it (but of course you can also continue using the "master" branch)

Issue #2175, opened 2020-04-27
OFF Missing battle background transparent and player must attack in turn


KKha-11:
#### Player platform:
Windows, ...
 

#### Describe the issue in detail :

- [x]  1. Missing battle background transition. When fighting Japhet will change background from pink to red, but when using EasyRPG background turn from pink to black 

- [x]  2. Characters attack in turn, usually when 'all characters' turn is refilled' we can choose any character to atttack first, but in easyRPG when that come I have to choose the batter first then the add-ons to attack

- [x]  3.  Normal: Allies and enimies' turn will automatically refill overtime no matter what. But in easyrpg, their  turns only refill when all members in the team have done attacking.

![image](https://user-images.githubusercontent.com/63910516/80335729-641bf780-887f-11ea-9b1c-6466e08f3c17.png)
[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/4537591/easyrpg_log.txt)

fmatthew5876:
ATB Active mode has some quirks.

In wait mode, ATB will only accumulate when you're in the party window. Which means you're waiting for your actors to be able to do something.

In Active mode, ATB will accumulate in every battle scene state except for the initial "Fight/Auto/Escape" window.
However, battlers will only attack when you're in the party window or an actors command window.

This means that while in items, skills, targeting enemies, or targeting allies, atb will accumulate but no battler will act until you exit out of those states.

Ghabry:
> Missing battle background transition. When fighting Japhet will change background from pink to red, but when using EasyRPG background turn from pink to black

This is already fixed since 0.6.2. Everything is resolved. Closing :+1:

PR #2360, opened 2020-09-26
Implement ATB active mode

rueter37:
This PR implements the ATB active mode. This fixes bullet point 3 of #2175.

### DIFF
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -260,6 +260,17 @@ void Scene_Battle_Rpg2k3::Update() {
 	}
 
 	switch (state) {
+		case State_SelectEnemyTarget:
+		case State_SelectAllyTarget:
+			if (previous_state == State_SelectSkill || previous_state == State_SelectItem) {
+				break;
+			}
+			// fallthrough
+		case State_SelectCommand:
+			if (Game_System::GetAtbMode() == lcf::rpg::SaveSystem::AtbMode_atb_wait) {
+				break;
+			}
+			// fallthrough
 		case State_SelectActor:
 		case State_AutoBattle: {
 			if (!IsWindowMoving()) {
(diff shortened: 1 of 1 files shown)


src/scene_battle_rpg2k3.cpp
```diff
@@ -261,29 +261,39 @@ void Scene_Battle_Rpg2k3::Update() {
 
 	switch (state) {
 		case State_SelectActor:
-		case State_AutoBattle: {
+		case State_AutoBattle:
+		case State_SelectCommand:
+		case State_SelectEnemyTarget:
+		case State_SelectAllyTarget: {
+			if (Game_System::GetAtbMode() == lcf::rpg::SaveSystem::AtbMode_atb_wait && (state == State_SelectCommand || state == State_SelectEnemyTarget || state == State_SelectAllyTarget)) break;
+			if ((state == State_SelectEnemyTarget || state == State_SelectAllyTarget) && (previous_state == State_SelectSkill || previous_state == State_SelectItem)) break;
```
Ghabry, line 269: You check "case" values in an if and then break out of the switch.
I suggest putting them before ``State selectactor`` and then just ``// fallthrough``

rueter37: I'm probably too stupid but how I'm supposed to do the fallthrough?

Ghabry: To fallthrough just don't use break. It is just convention (compilers warn about it) that you must use ``// fallthrough`` (a comment) when the fallthrough was intended.

Something like this.

```
case State_SelectEnemyTarget:
case State_SelectAllyTarget:
  if (previous_state == State_SelectSkill || previous_state == State_SelectItem) {
    break;
  }
  // fallthrough
case State_SelectCommand:
  if (Game_System::GetAtbMode() == lcf::rpg::SaveSystem::AtbMode_atb_wait) {
    break;
  }
  // fallthrough
case State_SelectActor:
case State_AutoBattle:
 /* more code */
```

rueter37: Thank you for your input! It helps me understanding the switch-case-operation a bit better. I have updated the PR.

Ghabry (approved):
Imo this is an anti feature but works for me.
Tested attacking while being killed through a skill, worked without crashing.
For later commits there shouldbe a checklist what RPG_RT does at these edge cases:

- Killed while starting normal attack
- Killed while opening Skill Window
- Game Over while opening Item window
etc.

fdelapena (approved):

fmatthew5876 (approved):
Code looks ok, if @Ghabry tested this one then it's good enough for me.

Issue #1277, opened 2017-10-07
Minor battle bugs in OFF


rsclifford:
Please fill in the following fields before submitting an issue:
#### Name of the game:
OFF (v1.6.1, English 2.0 Update)

#### Player platform:
Windows

#### Attach files (as a .zip archive or link them)
[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/1366434/easyrpg_log.txt)
[Save01.zip](https://github.com/EasyRPG/Player/files/1366433/Save01.zip)

#### Describe the issue in detail and how to reproduce it:
Two bugs I've run into with battles in OFF:

- [x] The "Purification in progress" message at the beginning of battles doesn't appear #2364
- [x] The characters don't animate when using physical attacks: ex. when Alpha uses a physical attack, it should rush at the targeted enemy, collide, and fly back into position. (since 0.6.3, #2503)

To reproduce:
- Load the attached save
- Walk around until you get into a battle
- Use a physical attack

carstene1ns:
Again, thanks for the bug report.
Did a quick search for the term `battle_start`. It is currently not used, so this is not implemented yet.
We only have a part of the possible animations already, some are missing. Did not check what is used here, but high chance this is also not implemented yet.

Ghabry:
Rush and collide could be #1269

carstene1ns:
Patch for the battle start message:

```patch
diff --git a/src/scene_battle_rpg2k3.cpp b/src/scene_battle_rpg2k3.cpp
index e301cb9a..1a0304fc 100644
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -394,6 +394,11 @@ void Scene_Battle_Rpg2k3::SetState(Scene_Battle::State new_state) {
 
 	switch (state) {
 	case State_Start:
+		if (!Data::terms.battle_start.empty()) {
+			help_window->SetVisible(true);
+			message_timer = 120;
+			help_window->SetText(Data::terms.battle_start);
+		}
 		break;
 	case State_SelectOption:
 		options_window->SetVisible(true);
```
Problem is, that the message is not displayed long enough... :/

Ghabry:
You can do it in the switch-case in ProcessActions, there is the state transition to State_SelectOption.

carstene1ns:
I had it there before, however, the window was not shown at all then.

fmatthew5876:
[...]

PR #2364, opened 2020-09-27
Battle 2k3: Add battle start message and frame delays

### LINKED ISSUE #2357: Battle2k3 Notification window + frame delays
Just like in 2k battle, the 2k3 battle notifications have a mechanic where there are 2 times. There is a 2k3 battle "Wait" function which takes 2 parameters.

The first value is an unconditional wait. RPG_RT will always wait this number of frames. The second value is a conditional wait. RPG_RT will wait for those frames unless you press the decision key or the shift key. Finally, during this wait if you hold the escape key, it will pause the battle until you release the key.

For example if a wait is (10, 80), it will wait 10 frames, and then 80 additional frames, which can be skipped if decision key is pressed, or delayed indefinately if escape is held.

Notification Messages
----------------

The notifcation window is what pops up and dissapears for battle start, skills etc.. This window is followed by a `Wait(t1, t2)` which dictates how long it is displayed. While you can use the input keys to affect the wait, you cannot interact directly with this message window.

Battle End Messages
--------------

This window shows victory, defeat,  exp, gold, and items. This is a regular message window, with a single line. It responds to key input like any other regular message box does. You progress using the decision key and wait is *not* used here.

Input Keys
--------------

All input keys react to whether the key down or up state, not the event of key pressing.

| Input | Action |
| -- | -- |
| Decision | Immediately cancels the window after the first wait frames have passed |
| Shift | Same as decision |
| Escape | The notification window is kept open and the battle scene is paused for as long as you hold the escape key. The frame timer however still ticks down to 0. |

Waits and notifications
-------------------

| Battle Phase | Message | Waits | Details |
| -- | -- | -- | -- |
| Init | Start | (10, 80) | On battle start if the start message is not empty |
| Init | BattleInitiative | (30, 70) | Immediately after battle start. If condition is initiative, surround, or normal and the first_strike flag is set |
| Init | | (30, 70) | If BattleInitiative message was not displayed, and the condition is back or pincers, does a wait with no message|
| Victory | | (30, 30) | After playing victory music, but before displaying the battle awards (exp, gold, item, etc..) the game pauses for 30 frames with no message. 
| Defeat | | (60, 60) | After playing game over music, but before displaying defeat message, the game pauses for 60 frames with no message. |
| EndBattleScene | | (30, 30) | Only in battle test mode! After the battle ends and the screen is erased. If battle test mode, RPG_RT waits 30 frames while screen is erased before exiting. I guess this is to make it so the RPG_RT window doesn't disappear immediately after transitioning out, maybe the application exit a bit more smooth. |
| EscapeFailure | EscapeFailure | (10, 30) | If actor escape command failed |
| EscapeSuccess | | (10, 30) | When processing escape action, waits immediately after playing flee sound effect |
| Action | "Double Attack " | (10, 40) | When a monster does a double attack, before the action is processed |
| Action | "Defend" | (10, 40) | When a monster defends, before the action is processed |
| Action | "Observe" | (10, 40) | When a monster observers, before the action is processed |
| Action | "Charge Up" | (10, 40) | When a monster charges, before the action is processed |
| Action | "Self-Destruct" | (10, 40) | When a monster self destructs, before the action is processed |
| Action | "Escape" | (10, 40) | When a monster escapes, before the action is processed |
| Action | Skill name | (15, 50) | When a skill or skill item is used |
| Action | Item Name | (10, 40) | When a non-skill item is used |

DynRPG Address:

| Address | Class | Function | Comment |
| -- | -- | -- | -- |
|004974C4 | SceneBattle | Wait | Has all above described wait mechanics |

rueter37:
This PR tries to fix #2357 and bullet point 1 from #1277. It adds the battle start and initiative messages and the frame delays in RPG Maker 2003 battles.

### DIFF
--- a/src/scene_battle.cpp
+++ b/src/scene_battle.cpp
@@ -682,6 +682,11 @@ void Scene_Battle::EndBattle(BattleResult result) {
 
 	Scene::Pop();
 
+	// For RPG_RT compatibility, wait 30 frames if a battle test ends
+	if (Game_Battle::battle_test.enabled) {
+		Scene::instance->SetDelayFrames(30);
+	}
+
 	if (on_battle_end) {
 		on_battle_end(result);
 		on_battle_end = {};
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -247,18 +247,6 @@ void Scene_Battle_Rpg2k3::FaceTarget(Game_Actor& source, const Game_Battler& tar
 }
 
 void Scene_Battle_Rpg2k3::Update() {
-	// FIXME: RPG_RT sets initial directions on start, displays any
-	// battle start messages, and then monsters may turn to face the party
-	// and the party turns to face the monsters.
-	// This only happens once on battle start. Until we refactor 2k3 state machine,
-	// we emulate the behavior with this bool.
-	// Monster directions never change during battle, but actors can.
-	if (!initial_directions_updated) {
-		UpdateEnemiesDirection();
-		UpdateActorsDirection();
-		initial_directions_updated = true;
-	}
-
 	switch (state) {
 		case State_SelectEnemyTarget:
 		case State_SelectAllyTarget:
@@ -273,7 +261,7 @@ void Scene_Battle_Rpg2k3::Update() {
 			// fallthrough
 		case State_SelectActor:
 		case State_AutoBattle: {
-			if (!IsWindowMoving()) {
+			if (battle_action_wait == 0 && !IsWindowMoving()) {
 
 				if (battle_actions.empty()) {
 					Game_Battle::UpdateAtbGauges();
@@ -775,13 +763,40 @@ void Scene_Battle_Rpg2k3::ProcessActions() {
 		return;
 	}
 
+	const bool wait = !CheckWait();
+
+	if (wait) {
+		return;
+	}
+
+	if (escape_initiated) {
+		EndBattle(BattleResult::Escape);
+		return;
+	}
+
 	if (!battle_actions.empty()) {
 		auto* battler = battle_actions.front();
 		if (!battle_action_pending) {
 			// If we will start a new battle action, first check for state changes
 			// such as death, paralyze, confuse, etc..
 			PrepareBattleAction(battler);
+			auto* action = battler->GetBattleAlgorithm().get();
 
+			if (action->GetSource()->GetType() == Game_Battler::Type_Enemy && combo_repeat == 1) {
+				std::string notification = action->GetStartMessage();
+
+				ShowNotification(notification);
+				if (!notification.empty()) {
+					if (action->GetType() == Game_BattleAlgorithm::Type::Skill) {
+						SetWait(15, 50);
+					} else {
+						SetWait(10, 40);
+					}
+				}
+			}
+
+			battle_action_pending = true;
+			return;
 		}
 		auto* alg = battler->GetBattleAlgorithm().get();
 		battle_action_pending = true;
@@ -798,14 +813,35 @@ void Scene_Battle_Rpg2k3::ProcessActions() {
 		}
 	}
 
-	if (help_window->IsVisible() && message_timer > 0) {
-		message_timer--;
-		if (message_timer <= 0)
+	if (help_window->IsVisible()) {
+		if ((state == State_SelectActor || state == State_SelectCommand || state == State_AutoBattle) && battle_action_wait == 0) {
 			help_window->SetVisible(false);
+		}
 	}
 
 	switch (state) {
 		case State_Start:
+			if (start_message_shown == 0) {
+				start_message_shown = 1;
+				if (!lcf::Data::terms.battle_start.empty()) {
+					ShowNotification(ToString(lcf::Data::terms.battle_start));
+					SetWait(10, 80);
+				}
+				return;
+			}
+			if (start_message_shown == 1) {
+				start_message_shown = 2;
+				const auto cond = Game_Battle::GetBattleCondition();
+				if (!lcf::Data::terms.special_combat.empty() && (cond != lcf::rpg::System::BattleCondition_none || first_strike)) {
+					if (cond == lcf::rpg::System::BattleCondition_initiative || cond == lcf::rpg::System::BattleCondition_surround || (cond == lcf::rpg::System::BattleCondition_none && first_strike)) {
+						ShowNotification(ToString(lcf::Data::terms.special_combat));
+					}
+					SetWait(30, 70);
+				}
+				return;
+			}
+			UpdateEnemiesDirection();
+			UpdateActorsDirection();
 			SetState(State_SelectOption);
 			break;
 		case State_SelectActor:
@@ -899,8 +935,14 @@ bool Scene_Battle_Rpg2k3::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBas
 
 		action->TargetFirst();
 
-		if (combo_repeat == 1) {
-			ShowNotification(action->GetStartMessage());
+		if (action->GetSource()->GetType() == Game_Battler::Type_Ally && combo_repeat == 1) {
+			if (action->GetType() == Game_BattleAlgorithm::Type::Skill) {
+				ShowNotification(action->GetStartMessage());
+				SetWait(15, 50);
+			} else if (action->GetType() == Game_BattleAlgorithm::Type::Item) {
+				ShowNotification(action->GetStartMessage());
+				SetWait(10, 40);
+			}
 		}
 
 		if (!action->IsTargetValid()) {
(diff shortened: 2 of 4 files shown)


src/scene_battle_rpg2k3.cpp
```diff
@@ -760,13 +748,47 @@ void Scene_Battle_Rpg2k3::ProcessActions() {
...
 			// If we will start a new battle action, first check for state changes
 			// such as death, paralyze, confuse, etc..
 			PrepareBattleAction(battler);
+			auto* action = battler->GetBattleAlgorithm().get();
+
+			if (action->GetSource()->GetType() == Game_Battler::Type_Enemy && combo_repeat == 1) {
+				std::string notification = "";
+
+				if (action->GetType() == Game_BattleAlgorithm::Type::Normal && action->GetRepeat() == 2) {
+					notification = "Double Attack";
```
fmatthew5876, line 769: Embed this logic into `Normal::GetStartMessage()`

src/scene_battle_rpg2k3.cpp
```diff
@@ -760,13 +748,47 @@ void Scene_Battle_Rpg2k3::ProcessActions() {
...
+				} else {
+					notification = action->GetStartMessage();
+				}
+				ShowNotification(notification);
+				if (action->GetType() == Game_BattleAlgorithm::Type::Skill) {
+					SetWait(15, 50);
+				} else {
+					if (action->GetType() == Game_BattleAlgorithm::Type::Item
+						|| (action->GetType() == Game_BattleAlgorithm::Type::Normal && action->GetRepeat() == 2)
```
fmatthew5876, line 778: You can wrap all this in `if (!notification.empty())` and then only check if skill then (15,50) else (10,40)

src/scene_battle_rpg2k3.cpp
```diff
@@ -783,14 +805,32 @@ void Scene_Battle_Rpg2k3::ProcessActions() {
...
+					ShowNotification(ToString(lcf::Data::terms.battle_start));
+					SetWait(10, 80);
+				}
+				return;
+			}
+			if (start_message_shown == 1) {
+				start_message_shown = 2;
+				if (!lcf::Data::terms.special_combat.empty() && (Game_Battle::GetBattleCondition() == lcf::rpg::System::BattleCondition_initiative || Game_Battle::GetBattleCondition() == lcf::rpg::System::BattleCondition_surround || (Game_Battle::GetBattleCondition() == lcf::rpg::System::BattleCondition_none && first_strike))) {
+					ShowNotification(ToString(lcf::Data::terms.special_combat));
+					SetWait(30, 70);
```
fmatthew5876, line 828: This `Wait(30, 70)` also occurs for back and pincer attack, but with no message displayed.

It does not occur for a normal battle without first strike.

To see this, press and hold escape immediately after the start message box disappears.

src/scene_battle_rpg2k3.cpp
```diff
@@ -884,8 +924,16 @@ bool Scene_Battle_Rpg2k3::ProcessBattleAction(Game_BattleAlgorithm::AlgorithmBas
...
 		action->TargetFirst();
 
-		if (combo_repeat == 1) {
-			ShowNotification(action->GetStartMessage());
+		if (action->GetSource()->GetType() == Game_Battler::Type_Ally && combo_repeat == 1) {
+			if (action->GetType() == Game_BattleAlgorithm::Type::Skill) {
+				ShowNotification(action->GetStartMessage());
+				SetWait(15, 50);
+			} else {
+				if (action->GetType() == Game_BattleAlgorithm::Type::Item) {
```
fmatthew5876, line 932: use `else if` to reduce scopes

src/scene_battle_rpg2k3.cpp
```diff
@@ -1318,10 +1370,18 @@ void Scene_Battle_Rpg2k3::Escape(bool force_allow) {
 
 	SetState(State_SelectActor);
 	ShowNotification(ToString(lcf::Data::terms.escape_failure));
+	SetWait(10, 30);
 }
 
 bool Scene_Battle_Rpg2k3::CheckWin() {
 	if (Game_Battle::CheckWin()) {
+		if (!win_wait_elapsed) {
```
fmatthew5876, line 1378: This and the corresponding lose logic looks incorrect.

We still call `Scene_Battler::Update()` every frame which tries to run the event interpreter in `Game_Battle::UpdateEvents()`. Since you haven't "finished" the battle, in theory it appears to me the battle could end, then an event runs while this wait is processing which revives an enemy, and then the battle continues.

That being said, this part of the battle scene logic for 2k and 2k3 needs a refactor. So I am ok accepting your change here as is and fixing this part later.

src/scene_battle_rpg2k3.cpp
```diff
@@ -1318,10 +1370,18 @@ void Scene_Battle_Rpg2k3::Escape(bool force_allow) {
 
 	SetState(State_SelectActor);
 	ShowNotification(ToString(lcf::Data::terms.escape_failure));
+	SetWait(10, 30);
```
fmatthew5876, line 1373: There is also a `Wait(10, 30)` on escape success with no message, immediately after playing the sound effect.

fmatthew5876:
One you didn't implement was the 30 frame delay after the screen is erased but before the application exits in battle test mode.

This should be doable with our scene delay functionality.

In `Scene_Battle::EndBattle()` right after the call to `Scene::Pop()` you should be able to add
```
if (is_battle_test) {
  Scene::instance()->SetSceneDelay(30);
}
```

Add a comment saying why we do this (RPG_RT compatibility)

src/scene_battle_rpg2k3.cpp
rueter37: Changed.

src/scene_battle_rpg2k3.cpp
rueter37: Changed.

src/scene_battle_rpg2k3.cpp
rueter37: Fixed.

src/scene_battle_rpg2k3.cpp

src/scene_battle_rpg2k3.cpp
rueter37: Added.

rueter37:
Added the battle test mode end delay.

src/scene_battle.cpp
```diff
@@ -682,6 +682,11 @@ void Scene_Battle::EndBattle(BattleResult result) {
 
 	Scene::Pop();
 
+	// For RPG_RT compatibility, wait 30 frames if a battle test ends
+	if (Game_Battle::battle_test.enabled) {
+		Scene::instance.get()->SetDelayFrames(30);
```
fmatthew5876, line 687: Minor: `.get()` is not needed here.

src/scene_battle_rpg2k3.cpp
```diff
@@ -783,14 +793,34 @@ void Scene_Battle_Rpg2k3::ProcessActions() {
...
+				start_message_shown = 1;
+				if (!lcf::Data::terms.battle_start.empty()) {
+					ShowNotification(ToString(lcf::Data::terms.battle_start));
+					SetWait(10, 80);
+				}
+				return;
+			}
+			if (start_message_shown == 1) {
+				start_message_shown = 2;
+				if (!lcf::Data::terms.special_combat.empty() && (Game_Battle::GetBattleCondition() != lcf::rpg::System::BattleCondition_none || first_strike)) {
```
fmatthew5876, line 814: `auto cond = Game_Battle::GetBattleCondition()` then use cond everywhere. Not only is this faster in the case the function is not inlined, it makes the code more compact and easier to read.

src/scene_battle_rpg2k3.cpp
```diff
@@ -1312,16 +1352,25 @@ void Scene_Battle_Rpg2k3::Escape(bool force_allow) {
 	if (force_allow || TryEscape()) {
 		// There is no success text for escape in 2k3
 		Game_System::SePlay(Game_System::GetSystemSE(Game_System::SFX_Escape));
+		SetWait(10, 30);
 		EndBattle(BattleResult::Escape);
```
fmatthew5876, line 1356: `EndBattle()` pops the scene, so this wait never actually gets exercised.

src/scene_battle_rpg2k3.cpp
rueter37: Changed.

src/scene_battle.cpp
rueter37: Fixed.

src/scene_battle_rpg2k3.cpp
Ghabry: yeah this will continue running the battle after a victory which will have side effects because it lies about the state.

Ghabry (commented):
Besides this Win/Lose issue. Is there any other review comment left here that wasn't considered yet? If not imo this can be approved after fighting some test battles.

fmatthew5876:
I think it's fine from a code review perspective, but yes I agree needs some testing by someone other than the PR author who is comfortable signing off on this.

Ghabry (approved):
Like everything in battle this is hard to proof correct but for me it works.

fmatthew5876 (approved):
Good enough for me until the 2k3 battle refactor

PR #2371, opened 2020-10-03
SP accuracy check on enemy change

rueter37:
The SP accuracy check on enemies even checks if the target has SP present at all. The check always fails if the targets current SP is zero.


fdelapena (approved):

fmatthew5876:
This and many other edge cases are now covered by #2399

PR #2372, opened 2020-10-03
Battle 2k: Fix monster sprite display on revival

rueter37:
If a monster gets revived on RPG Maker 2000 his sprite was still invisible. This PR fixes this.

### DIFF
--- a/src/scene_battle_rpg2k.cpp
+++ b/src/scene_battle_rpg2k.cpp
@@ -953,6 +953,21 @@ bool Scene_Battle_Rpg2k::ProcessActionResults(Game_BattleAlgorithm::AlgorithmBas
 		return ProcessNextAction(BattleActionState_Death, action);
 	}
 
+	// Check if dead enemies get revived to make them visible again
+	if (battle_action_substate == eConditions
+			&& action->IsRevived())
+	{
+		if (action->GetTarget() != nullptr) {
+			auto* target = action->GetTarget();
+			if (target->GetType() == Game_Battler::Type_Enemy) {
+				auto* target_sprite = Game_Battle::GetSpriteset().FindBattler(target);
+				if (target_sprite) {
+					target_sprite->DetectStateChange();
+				}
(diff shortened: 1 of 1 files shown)


src/scene_battle_rpg2k.cpp
```diff
@@ -953,6 +953,22 @@ bool Scene_Battle_Rpg2k::ProcessActionResults(Game_BattleAlgorithm::AlgorithmBas
...
+	// Check if dead enemies get revived to make them visible again
+	if (battle_action_substate == eConditions
+			&& action->IsRevived())
+	{
+		if (action->GetTarget() != nullptr) {
+			auto* target = action->GetTarget();
+			if (target->GetType() == Game_Battler::Type_Enemy) {
+				auto* target_sprite = Game_Battle::GetSpriteset().FindBattler(target);
+				if (target_sprite) {
+					target_sprite->SetForcedAlive(true);
```
fmatthew5876, line 965: This "forced alive" thing was an ugly hack to allow the 2k battle system to "kill" the monster and animate it's death later.

Do we need to set this flag here or is `DetectStateChange` enough?

rueter37: Using `DetectStateChange` is indeed sufficient. I have updated this PR now.

fdelapena (approved):

fmatthew5876:
For some reason, this patch works when you kill an enemy with an attack. However when I inflict "death" on an enemy using a battle event on turn 0, the enemy remains invisible after revivied.

fmatthew5876 (approved):
I will leave a note in the issue to solve the event death. This is more complicated, also this battler sprite stuff needs a refactor

rueter37:
Adding `target_sprite->SetVisible(true)` solves this. Oops, I was too late...

fmatthew5876:
Lets be careful about just adding `SetVisible(true)`, that can break something else.

PR #2373, opened 2020-10-03
Skills which targets allies ignore hidden targets

rueter37:
If an enemy uses a skill which targets all his allies then his hidden allies are no longer targeted anymore.

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -1106,6 +1106,11 @@ bool Game_BattleAlgorithm::Skill::IsTargetValid() const {
 
 	if (skill.scope == lcf::rpg::Skill::Scope_ally ||
 		skill.scope == lcf::rpg::Skill::Scope_party) {
+		// Ignore hidden targets
+		if (GetTarget()->IsHidden()) {
+			return false;
+		}
+
 		if (GetTarget()->IsDead()) {
 			// Cures death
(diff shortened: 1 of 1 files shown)


Ghabry:
Is this a regression?
Take a look at #2107 first commit https://github.com/EasyRPG/Player/pull/2107/commits/192943d89d9f0891d0663820782c08e9a5ec87fe
Why isn't that change sufficient here?

rueter37:
It is because if the `skill.scope` is `Scope_ally` or `Scope_party` then the `return GetTarget()->Exists()` is never reached because it gets caught by `return true`. We have no regression here, the problem mentioned here already existed before #2107 was merged.

Ghabry:
ah okay, so this is an edge-case we didn't consider.

@rueter37 
Could you please stop for the next hours to push any commits onto Github? (at least not into branches where you have open PRs)

Want to release Player 0.6.2.3 today which fixes the absorb crash. As these releases are created via backporting commits (https://github.com/EasyRPG/Player/tree/0-6-2-stable). I must recompile liblcf and this will take a long time when anybody pushes commits because this will make the builders busy.
Thank you!

rueter37:
Acknowledged. :+1:

fmatthew5876 (approved):

fdelapena (approved):

PR #2378, opened 2020-10-04
Enemy AI changes: Change weighting of priorities and add special check

### LINKED ISSUE #1831: Verify correctness of `Game_Enemy::ChooseRandomAction`
The RPG Maker 2003 Steam help file says this:

> Process for Choosing Attack Patterns
>
> ?The action that the enemy character will take will be chosen based on the following rules.
>
> 
>
> 1.Of all attack patterns, those that fit the ?conditions? will be chosen. If there are multiple relevant conditions, (a) the one with the highest priority and (b) anything within 9 priority points of (a) will be considered.
>
>2.The importance of each attack pattern is calculated based on the effectiveness of the action during the current situation in battle. These values will be mutually compared among the attack patterns that were considered, and converted into a choice probability (the percentage that each pattern comprises out of the total importance value).
>
>3.Using the choice probabilities, the attack pattern will be randomly decided.

This sounds like there is some smart logic here like the autobattle algo which evaluates the effectiveness of the action to weigh the action's probability. Our code does a randomized selection weighed by rating only.

How confident are we in the algo we use?

@CherryDT any help here?

rueter37:
This PR tries to fix #1831.

This one changes the weighting of the priorities: The action with the highest priority always gets a weight of 10 and actions with a priority of (highest priority - 9) gets a weight of 1. The scaling between is linear. This means if we have two actions, one with a
priority of 50 and one with a priority of 45, the first action gets a 10/15 chance and the second action a 5/15 chance to get selected. Moreover a special check has been added if the enemy uses a skill which only heals states without affecting anything else: He actually checks if there is an ally which is affected by at least one state which get healed by the skill. If this is not the case, then the action gets discarded.

### DIFF
--- a/src/game_enemy.cpp
+++ b/src/game_enemy.cpp
@@ -22,6 +22,7 @@
 #include "game_battle.h"
 #include "game_enemy.h"
 #include "game_party.h"
+#include "game_enemyparty.h"
 #include "game_switches.h"
 #include <lcf/reader_util.h>
 #include "output.h"
@@ -218,10 +219,11 @@ const lcf::rpg::EnemyAction* Game_Enemy::ChooseRandomAction() {
 
 	int total = 0;
 	for (auto it = valid.begin(); it != valid.end();) {
-		if (actions[*it].rating < highest_rating - 9) {
+		const lcf::rpg::EnemyAction& action = actions[*it];
+		if (action.rating < highest_rating - 9 || (action.kind == action.Kind_skill && !IsSkillUsableForAI(action.skill_id))) {
 			it = valid.erase(it);
 		} else {
-			total += actions[*it].rating;
+			total += 10 - (highest_rating - action.rating);
 			++it;
 		}
 	}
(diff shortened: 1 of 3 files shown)


fmatthew5876:
This should go after #2380

rueter37:
This PR is still intact - a rebase is not needed after merging #2380. The `IsSkillUsable` function remains unchanged because the AI makes an extra check besides `IsSkillUsable` which is put into an own function (which almost always returns true except for skills which heals states only without affecting anything else).

src/game_enemy.cpp
```diff
@@ -244,3 +247,65 @@ const lcf::rpg::EnemyAction* Game_Enemy::ChooseRandomAction() {
 	return nullptr;
 }
 
+bool Game_Enemy::IsSkillUsableForAI(int skill_id) const {
```
fmatthew5876, line 250: RPG_RT does this by calling a function called `RPG_Battler::IsSkillEffectiveOn` which does a set of checks on each battler. The logic you have here is close but missing some edge cases. Also I need to do more testing to be sure.

fmatthew5876:
You've made some good improvements and found some bugs here, but to be honest this is another one like autobattle where it can never be perfect with black box testing. We really need to use RE tools to discover all the edge cases.

EnemyAI is next on my list to do as I also want to make a standalone factory function for it with customizable algos, just like autobattle.

rueter37:
Okay, for now I wait for further directions from you regarding this PR to avoid merge conflicts. :+1:

rueter37:
@fmatthew5876 Would it make sense to close this PR in favor of #2381? Because you have already added the priority weight fixes in your PR and you are planning to implement the remaining AI fixes.

fmatthew5876:
Yes I think so, unfortunately

PR #2380, opened 2020-10-05
Refactor Skill and Item usability and fix some incompatibilities

fmatthew5876:
Taken from RE disassembly, which is a bit messy and hard to read here. Needs to be tested to verify.

Tests
-----------

Skill Usability Tests
- [x] Test all 3 Escape conditions
- [x] Test all 3 teleport conditions
- [x] Test switch skill + field and battle flags
- [x] Test states which limit physical and magical rate
- [x] Test all 5 scopes
- [x] Test affect none, hp only, mp only, both hp and mp
- [x] Test heals state which persists after battle
- [x] Test heals state which ends after battle
- [x] Test skill which has 1 physical attribute with and w/o equipment
- [x] Test skill which has 2 physical attribute with and w/o equipment
- [x] Test skill which shifts physical attribute with and w/o equipment
- [x] Test all above as skill in map / menu
- [x] Test all above as skill in battle
- [x] Test all above as skill item in map / menu
- [x] Test all above as skill item in battle
- [x] Test all above as equipment invoke skill item in map / menu
- [x] Test all above as equipment invoke skill item in battle

Item Usability Tests:
- [x] Medicine usability
- [x] Common and Equipment usability
- [x] Skill book
- [x] Seed
- [x] Switch item - with field and battle activiation flags
- [x] Live party member which can use item required for normal and subskill item skills but not escape, teleport, or switch

RPG_RT bugs ❗ 
--------------

See here: https://github.com/EasyRPG/Player/issues/1486#issuecomment-703376641

### DIFF
--- a/src/game_party.cpp
+++ b/src/game_party.cpp
@@ -374,8 +374,7 @@ bool Game_Party::IsSkillUsable(int skill_id, const Game_Actor* target, bool from
 		return !Game_Battle::IsBattleRunning() && Main_Data::game_system->GetAllowEscape() && Main_Data::game_targets->HasEscapeTarget() && !Main_Data::game_player->IsFlying();
 	} else if (skill->type == lcf::rpg::Skill::Type_teleport) {
 		return !Game_Battle::IsBattleRunning() && Main_Data::game_system->GetAllowTeleport() && Main_Data::game_targets->HasTeleportTargets() && !Main_Data::game_player->IsFlying();
-	} else if (skill->type == lcf::rpg::Skill::Type_normal ||
-		skill->type >= lcf::rpg::Skill::Type_subskill) {
+	} else if (Algo::IsNormalOrSubskill(*skill)) {
 		int scope = skill->scope;
 
 		if (Game_Battle::IsBattleRunning()) {
--- a/src/game_party.cpp
+++ b/src/game_party.cpp
@@ -255,10 +255,6 @@ bool Game_Party::IsItemUsable(int item_id, const Game_Actor* target) const {
 		return false;
 	}
 
-	if (data.party.size() == 0) {
-		return false;
-	}
-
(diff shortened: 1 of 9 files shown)


fmatthew5876:
I've tested all the edge cases listed in the original post with a test game. This is ready.

fdelapena (approved):

rueter37 (approved):

Ghabry:
@fmatthew5876 
To sum this up: What were the biggest bug fixes here? The checkbox-list is not very helpful for a blog post later because it doesn't differentiate between "worked before" and "works now" :)

src/game_party.cpp
```diff
@@ -255,10 +255,6 @@ bool Game_Party::IsItemUsable(int item_id, const Game_Actor* target) const {
 		return false;
 	}
 
-	if (data.party.size() == 0) {
```
Ghabry, line 257: Does removing of this check have any practical use? You can't open the item scene in the menu when the party is empty and there is no event command to use items.

fmatthew5876:
TBH I don't have a full list. I just decided to clean this up and make it match RPG_RT.

A few that come to mind:
- [x] The bugs in https://github.com/EasyRPG/Player/issues/1486#issuecomment-703376641
- [x] Physical attributes requirement for equipment only applies if skill doesn't shift attribute

Probably more things got fixed, but I didn't test master thoroughly to compare

src/game_party.cpp
fmatthew5876: Maybe not, but having it has no practical use either. It wasn't in RPG_RT so I took it out.

Ghabry:
hm okay. The next blog will be full of "we changed something but nobody really knows what it fixed" ^^' (looking at you, Interpreter 80 commits refactor)

PR #2383, opened 2020-10-06
Battle 2k3: Add run away animation on escape

### LINKED ISSUE #2147: Battle 2k3: Misc issues
- [x] Enemies update before actors

Start a battle with 1 actor and 1 enemy, both with the same AGI values. Then do an autobattle.

The enemies will attack first. This indicates that RPG_RT processes enemy actions before actors.

- [x] Actors don't animate run away when escape is performed

```
if (escapeAllowed )
  if (battler.mirrored )
      battler.x -= 6;
    else
      battler.x += 6;
  }
}
```

Looks like if you use a "Force Flee" event on the party on a battle where escape is not allowed, they may not run

- [x] Actors with a state restriction but full atb gauge (first strike etc..) are still able to take their turn.
    - [x] Test with recoverable and non-recoverable states

rueter37:
Depends on #2364

If the escape command is used in RPG Maker 2003 battles, the actors do a run away animation now. This fixes the second bullet point of #2147.

Fix #2147

### DIFF
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -299,6 +299,18 @@ void Scene_Battle_Rpg2k3::Update() {
 		default:;
 	}
 
+	if (escape_initiated) {
+		for (auto& actor: Main_Data::game_party->GetActors()) {
+			Point p = actor->GetBattlePosition();
+			if (actor->IsDirectionFlipped()) {
+				p.x -= 6;
+			} else {
+				p.x += 6;
+			}
+			actor->SetBattlePosition(p);
+		}
+	}
+
 	for (std::vector<FloatText>::iterator it = floating_texts.begin();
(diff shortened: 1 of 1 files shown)


Ghabry (changes requested):
The speed of there movement looks correct, I made a recording to verify this.

But: In RPG_RT they seem to play some kind of "walking animation". That changes every 3-4 frames.

rueter37:
The animation state is set to `AnimationState_WalkingLeft` or `AnimationState_WalkingRight` depending on the direction the actor flees. I'm pretty sure the walking animation works, I have reduced the speed from 6 to 1 in my test to verify that the animation works. 6 is pretty fast, it is possible to not notice the walking animation in this case.

Ghabry (approved):
yeah this time the animation worked for me, no idea what happened last time

fmatthew5876 (approved):

PR #2385, opened 2020-10-09
Cancel battle action if getting restriction state added by an event

### LINKED ISSUE #2169: Battle 2k: When an actor dies and is revived during the same turn the battle action is still executed
Found this while finding ways to crash:

Create a page that does
```
Condition Sw1 on
Alex: State Dead +
Alex: State Dead -
```

Let Brian (must have higher AGI than Alex) invoke a Skill that enables Switch 1.

Expected:
- Only Brians Switch skill is activated
- [Turn End]

Player:
- Brian activates switch skill
- Alex will do it's action as if he never died
- [Turn End]

rueter37:
This PR fixes #2169.

If a battler gets a restriction state added by an event his battle action gets canceled.

### DIFF
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -370,6 +370,9 @@ bool Game_Battler::AddState(int state_id, bool allow_battle_states) {
 	if (GetSignificantRestriction() != lcf::rpg::State::Restriction_normal) {
 		SetIsDefending(false);
 		SetCharged(false);
+		if (GetBattleAlgorithm() != nullptr) {
+			this->SetBattleAlgorithm(std::make_shared<Game_BattleAlgorithm::NoMove>(this));
+		}
 	}
 
 	return was_added;


src/game_battler.cpp
```diff
@@ -370,6 +370,9 @@ bool Game_Battler::AddState(int state_id, bool allow_battle_states) {
 	if (GetSignificantRestriction() != lcf::rpg::State::Restriction_normal) {
 		SetIsDefending(false);
 		SetCharged(false);
+		if (GetBattleAlgorithm() != nullptr) {
+			this->SetBattleAlgorithm(std::make_shared<Game_BattleAlgorithm::NoMove>(this));
+		}
```
fmatthew5876, line 375: Overall, this is the right approach. 

However, it's not clear whether this should be `NoMove` `Null` or if we just set the algo itself to `nullptr`. Need to test the behavior carefully in 2k and 2k3.

rueter37: At first I tried `nullptr` but this approach throwed `Invalid battle action` error messages.

fmatthew5876: Right, `nullptr` would mean it's up to the battle system to decide what to do, which would require more changes elsewhere.

fmatthew5876:
So the real difference between `Null` and `NoMove` is whether the 2k battle system will flash the actor and wait.

https://github.com/EasyRPG/Player/blob/master/src/scene_battle_rpg2k.cpp#L513
https://github.com/EasyRPG/Player/blob/master/src/scene_battle_rpg2k.cpp#L534

`NoMove` is like I will perform a battle action this round, which does nothing. `Null` literally is a placeholder for I will not perform any battle action this round.

Both of them still cause battlers to process state changes.

So in short, try to hit an enemy with a paralyze state before their turn, and then see if they flash or not. Make sure to test with no messages displayed as messages also cause flash to occur, even for null.

Ghabry:
Oh right. I forgot that there is also a null action. Thought you mean nullptr

rueter37:
@fmatthew5876 I did the test with RPG_RT now and if a paralyzed enemy is hit no flash occurs with state messages disabled. In EasyRPG the paralyzed enemy flashes on its turn but that already occurred before this PR.

fmatthew5876:
I'm ok with accepting this PR as it is, but please write up a new issue for flashing inconsistency with 2k battles. It may be more than just this PR.

rueter37:
Submitted the issue now (#2388).

fdelapena (approved):

fmatthew5876 (approved):

PR #2389, opened 2020-10-09
Battle 2k: Disable enemy flashing on no move restriction

### LINKED ISSUE #2388: Battle: Enemy flash inconsistency
In RPG Maker 2000 battles the enemy flashes on its turn even if he has a state with a no move restriction. In RPG_RT this flash doesn't occur.

Edit: RPG Maker 2003 battles are affected as well.

rueter37:
This PR fixes #2388.

If it is an enemy's turn in a RPG Maker 2000 battle and he has a state with a no move restriction, then he no longer flashes.

### DIFF
--- a/src/scene_battle_rpg2k.cpp
+++ b/src/scene_battle_rpg2k.cpp
@@ -511,7 +511,9 @@ bool Scene_Battle_Rpg2k::ProcessActionBegin(Game_BattleAlgorithm::AlgorithmBase*
 		}
 
 		if (action->GetType() != Game_BattleAlgorithm::Type::Null || show_message) {
-			SelectionFlash(action->GetSource());
+			if (action->GetSource()->CanAct()) {
+				SelectionFlash(action->GetSource());
+			}
 		}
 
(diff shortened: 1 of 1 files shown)


fmatthew5876:
It's been so long that I don't remember all the details of the original solution, but I'm pretty sure this is not right.

You're circumventing the purpose of `Null` battle algo here. This whole bit of logic needs to be confirmed in RE, otherwise we'll fix one case and break another.

rueter37:
@fmatthew5876 Since your PR (#2399) refactors the battle behaviors and scenes, it would probably make sense to close my PR here because #2388 is mentioned in your PR so I think the issue will be fixed by it.

fmatthew5876:
Lets leave this one up for now as a reminder because I want to look into it more carefully. It may be the case that we actually don't need both `NoMove` and `Null`.

fmatthew5876:
I checked and this change is correct, actually `Null` shouldn't even exists. I removed `Null` in #2399

PR #2395, opened 2020-10-11
Refactor directory access into DirectoryTree - FS [2/3]

Ghabry:
This is a huge refactor of the FileFinder.

To sum up:
- All general file access related stuff moved from FileFinder to ``DirectoryTree``. The game directory tree and specific functions such as ``FindImage`` are still managed by ``FileFinder``. Will stay this way until the next PR.
- The RTP redirection was moved to ``FileFinder_RTP``. Converted to a class and managed by ``Main_Data`` as this is game specific.
- StringView-ifyd stuff

It will not conclude my refactor though, next step after this are the Virtual File Systems

--------------

The ``DirectoryTree`` uses now unique instead of shared_ptr. With ``Subtree`` non-owning subviews of the tree (``DirectoryTreeView``) can be created.
E.g. the Game Browser can just pass a Subtreeview to the game it boots, this way there is no need anymore for enumerating twice on startup -> faster startup times on slow systems.
Though the ownership semantics are quite messy in Player so this is not possible yet and I postbone it to the next PR.

These subviews will be also useful for the Virtual Filesystems later.

--------------

TODO:

Retest:

- [x] Fix Windows
- [x] Game Browser Game booting
- [ ] Meta File (Game Import)
- [x] Extensions guessing
- [x] MP3 major game heuristic
- [x] Random game testing (Frozen Trigger works)
- [x] Translation support 
- [x] File not found debug log is now very common because there is only one find entry point that handles everything-
- [x] Fix ownership of RTP trees.
- [x] Documentation
- [x] Testing


Ghabry:
Surprisingly emscripten still works

Ghabry:
just a rebase. Translation support disabled for now while I think about a nice way to hook it into my API.

Ghabry:
This is now ready for review :).

Only things I want to do is write some doc comments and fix two pending problems:

- I can't get the "Save Import from previous game" (Scene Import) Importer to work
- Windows path handling (the backslash) is broken according to a unit test

Ghabry:
Obviously Windows was a one-line fix...

Ghabry:
If you don't mind I would like to postbone the "Scene Import" Fix for "FS [3/3]".
When I fix this now I very likely have to redo the fix for the VFS, so thats extra work.

The problem is:
The Scene requests a directory tree for the folder "..".

This has multiple problems:

- The path is non-canonical
- The directory tree of the game cannot access its parent folder because that is above the game root
- Have to invent an API in the VFS to request the parent VFS

tests/rtp.cpp
```diff
@@ -28,6 +39,7 @@ static std::shared_ptr<FileFinder::DirectoryTree> make_tree() {
 			{"カーソル1.wav", "カーソル1.wav"},
 			{"キャンセル1.wav", "キャンセル2.wav"}
 	};
+#endif
```
Ghabry, line 42: oh forgot to fix these

tests/filefinder.cpp
```diff
@@ -28,19 +28,62 @@ TEST_CASE("IsDirectory") {
...
+    CHECK(file == "file");
+
+    std::tie(path, file) = FileFinder::GetPathAndFilename("file");
+    CHECK(path == "");
+    CHECK(file == "file");
+
+    std::tie(path, file) = FileFinder::GetPathAndFilename("folder/folder2/file");
+    CHECK(path == "folder/folder2");
+    CHECK(file == "file");
+}
```
Ghabry, line 87: whitespace instead of tabs...

src/directory_tree.cpp
```diff
@@ -0,0 +1,215 @@
...
+		const std::string name = dir.GetEntryName();
+		Platform::FileType type = dir.GetEntryType();
+
+		static bool has_fast_dir_stat = true;
+		bool is_directory = false;
+		if (has_fast_dir_stat) {
+			if (type == Platform::FileType::Unknown) {
+				has_fast_dir_stat = false;
+			} else {
+				is_directory = type == Platform::FileType::Directory;
```
fmatthew5876, line 123: Put parens around the expression

src/directory_tree.cpp
```diff
@@ -0,0 +1,291 @@
...
+	if (!dir) {
+		Output::Debug("Error opening dir {}: {}", full_path, ::strerror(errno));
+		return nullptr;
+	}
+
+	while (dir.Read()) {
+		const std::string name = dir.GetEntryName();
+		Platform::FileType type = dir.GetEntryType();
+
+		static bool has_fast_dir_stat = true;
```
fmatthew5876, line 117: Whats the reason for this bool?

fmatthew5876, line 117: This dance with the static bool is confusing, is it really necessary? Why not just say 
```
if (type == Directory) {
  is_directory = true;
} else if (type == Unknown) {
  is_directory = FileFinder::IsDirectory(...);
}
```

src/directory_tree.cpp
```diff
@@ -0,0 +1,291 @@
...
+			} else {
+				is_directory = type == Platform::FileType::Directory;
+			}
+		}
+
+		if (!has_fast_dir_stat) {
+			is_directory = FileFinder::IsDirectory(FileFinder::MakePath(full_path, name), true);
+		}
+
+		if (name == "." || name == "..") {
```
fmatthew5876, line 131: Move this check up so you're not doing filesystem calls on `.` and `..`

src/directory_tree.cpp
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+	Platform::Directory dir(full_path);
+	if (!dir) {
+		Output::Debug("Error opening dir {}: {}", full_path, ::strerror(errno));
+		return nullptr;
+	}
+
+	while (dir.Read()) {
+		const std::string name = dir.GetEntryName();
```
fmatthew5876, line 114: `const auto&` ?

src/directory_tree.cpp
```diff
@@ -0,0 +1,291 @@
...
+	dir_cache[dir_key] = fs_path;
+
+	DirectoryListType fs_cache_entry;
+
+	for (auto& entry : entries) {
+		fs_cache_entry.emplace(std::make_pair(make_key(entry.name), entry));
+	}
+	fs_cache.emplace(dir_key, fs_cache_entry);
+
+	dir_it = dir_cache.find(dir_key);
```
fmatthew5876, line 159: `dir_it` is unused?

src/directory_tree.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+DirectoryTreeView DirectoryTree::Subtree(std::string sub_path) {
+	return DirectoryTreeView(this, std::move(sub_path));
+}
+
+std::string DirectoryTree::FindFile(StringView filename, Span<StringView> exts) const {
+	return FindFile({ ToString(filename), exts });
+}
+
+std::string DirectoryTree::FindFile(StringView directory, StringView filename, lcf::Span<StringView> exts) const {
```
fmatthew5876, line 172: `Span` instead of `lcf::Span`

src/directory_tree.cpp
```diff
@@ -0,0 +1,291 @@
...
+	if (!entries) {
+		if (args.file_not_found_warning) {
+			Output::Debug("Cannot find: {}/{}", dir, name);
+		}
+		return "";
+	}
+
+	std::string dir_key = make_key(dir);
+	auto dir_it = dir_cache.find(dir_key);
+	assert(dir_it != dir_cache.end());
```
fmatthew5876, line 210: `dir_it` not used except for in the assert, I would wrap it all in `NDEBUG` so the `dir_cache` lookup isn't done in release build

src/directory_tree.cpp
```diff
@@ -0,0 +1,291 @@
...
+	return FileFinder::MakePath(root, subpath);
+}
+
+StringView DirectoryTree::GetRootPath() const {
+	return root;
+}
+
+DirectoryTreeView::DirectoryTreeView(const DirectoryTree* tree, std::string sub_path) :
+	tree(tree), sub_path(std::move(sub_path)) {
+	valid = tree->ListDirectory(this->sub_path) != nullptr;
```
fmatthew5876, line 253: parens

src/directory_tree.h
```diff
@@ -0,0 +1,266 @@
...
+
+		Entry(std::string name, FileType type) : name(std::move(name)), type(type) {}
+	};
+
+	/** Argument struct for more complex find operations */
+	struct Args {
+		/** File relative to the current tree to search */
+		std::string path;
+		/** File extensions to append to the filename when searching */
+		lcf::Span<StringView> exts;
```
fmatthew5876, line 63: `Span`

src/directory_tree.h
```diff
@@ -0,0 +1,266 @@
...
+	const DirectoryTree* tree = nullptr;
+	std::string sub_path;
+	bool valid = false;
+};
+
+inline DirectoryTreeView::operator bool() const noexcept {
+	return valid;
+}
+
+inline bool operator<(const DirectoryTree::Entry& l, const DirectoryTree::Entry& r) {
```
fmatthew5876, line 250: It's annoying, but we should provide the other operators. They can all be written in terms of `operator<` so you don't need to duplicate code.

src/fileext_guesser.cpp
```diff
@@ -31,36 +31,43 @@ namespace {
...
-			if (ext != "exe" && ext != "ini") {
-				candidates.push_back({item.second, ext, FileFinder::GetFileSize(FileFinder::FindDefault(dir, item.second))});
+	const auto* entries = tree.ListDirectory();
+	if (entries) {
+		for (const auto &item: *entries) {
+			if (item.second.type != DirectoryTree::FileType::Regular) {
+				continue;
+			}
+
+			if (item.first.length() == RtPrefix.length() + 3 && lcf::ToStringView(item.first).starts_with(RtPrefix)) {
```
fmatthew5876, line 45: `lcf::StringView` -> `StringView`

src/filefinder_rtp.cpp
```diff
@@ -0,0 +1,308 @@
...
+		search_paths.push_back(std::move(tree));
+
+		if (hit_info.empty()) {
+			Output::Debug("The folder does not contain a known RTP!");
+		}
+
+		// Only consider the best RTP hits (usually 100% if properly installed)
+		float best = 0.0;
+		for (const auto& hit : hit_info) {
+			float rate = (float)hit.hits / hit.max;
```
fmatthew5876, line 180: `static_cast<float>(hit.hits) / hit.max` to make the precedence clear

src/directory_tree.cpp
Ghabry: As this was copied over from FileFinder the code is likely 3 or more years old.
Your solution should work and looks better, thanks.
(The directory type in dirent is an extension and some filesystems reports Unknown here, will also add a comment)

src/directory_tree.h
Ghabry: added the other ops

fdelapena (approved):

Ghabry:
Because this is a big change lets keep this open for 7 more days until 6th of Feb and when there is no further feedback: Merge.

Will already use the code here as a base for the final filesystem step :)

fmatthew5876 (approved):

rueter37:
I'm late for giving feedback but I noticed two oddities after this got merged into master: The compiler (gcc 9.3) complains over `#include <memory>` missing in `directory_tree.h` and RTP files are not getting loaded if certain game folders are missing. You can reproduce this in "TestGame-2000" and "TestGame-2003". In "TestGame-2000" the "Backdrop" and "Monster" folders are missing which makes the background and the monster sprites not being loaded in battle even if the RTP is installed. "TestGame-2003" is even more extreme: Because of every folder missing there, basically no assets are being loaded. But this is probably a rare edge case because most games have every asset folder present (even if they are RTP-reliant).

Ghabry:
Thanks for the reports. I will hotfix them next week.

rueter37:
One more oddity: Savegames are not put into the game folder if you save a game. The save file is put into the same directory from where you have started EasyRPG (you can reproduce it using the command line).

Ghabry:
I didn't notice that bug but this is fixed in PR 3 #2455

rueter37:
Good to know, thanks! :+1:

Ghabry:
Well this shows that it was a good idea to do the DirectoryTree before the VFS. One can find bugs early and the VFS uses the Tree.

The RTP one is dumb: Is caused by this line (abort when dir is missing)
https://github.com/EasyRPG/Player/blob/master/src/directory_tree.cpp#L191-L192

I must restructure the ifs so it falls through to the rtp handler.

Issue #1367, opened 2018-05-25
Harvest Moon - A new Life: Executes LS.dat and SaveCount.dat via Ineluki Key Patch


Ghabry:
Today in "Abusing RPG_RT": [Harvest Moon a new Live](https://rmarchiv.tk/games/1092)

Describe the issue in detail and how to reproduce it:

This game uses ExecProgram of Inelukis Key Patch to invoke the two programs LS.dat and SaveCount.dat.

**LS.dat** (UPX compressed):
This program takes one argument (a memory address, always 0x46f16c), uses FindWindow for getting a handle to RPG_RT and then uses WriteProcessMemory to write a single byte (0x75) to that address.

No idea what the purpose of this write is but it changes
push offset 0x46F189 to
push offset 0x467589

Maybe @CherryDT can help ^^. Probably related to automatic savegame loading.

**SaveCount.dat:**

This program counts the number of files that end on "lsd" and writes the following text in Sounds/Saves.script.wav:

```
[execute]
Action = MidiTickFunction
Command = output
next=Saves

[Saves]
Action = AddOutPut
Value = X
next=

;by Stephan Fischer
```

Where X is the amount of Savegames found.

The script is later used to decide wheter a "load game" option shall be displayed.

stavrossk:
I wish this game was translated into English...

carstene1ns:
Tried this game some year(s) ago, it was item switching all the time (because of unimplemented Ineluki), so unplayable.

CherryDT:
This is one of the later versions of Mirokus original auto enter patch if I'm not mistaken. Sonic Party 2 by Simon does the same thing.

I am not totally sure but as far as I remember, LS.dat is used to make the "load game" option selected by default on the title screen. The game is then reset and the "auto enter" on the title screen will therefore open the load menu. But I never fully explored this. I believe there is a wrapper EXE (?) which listens to F12 and resets this patch then, possibly.

CherryDT:
But the instruction change that you wrote doesn't seem right, I'd expect it to change a 74 to a 75.

Ghabry:
Well the auto enter patch doesn't work obviously but my Ineluki branch has now an "emulation" for this:

https://github.com/Ghabry/easyrpg-player/blob/ineluki/src/game_ineluki.cpp#L32-L38

CherryDT:
[...]

PR #2397, opened 2020-10-12
Implement (most of) Ineluki Key Patch

### LINKED ISSUE #1181: Ineluki Key Patch
DL: http://cherrytree.at/cms/download/?did=3

"This famous tool (which isn’t really a patch but a DLL replacement) allows full keyboard and mouse input as well es executing external programs from an RPG Maker event. It also provides MP3 support, which is very complicated and buggy, however."

Basicly a harmony.dll replacement with "scripts" written as INI files in Music folder...

Files are of format ".script.wav". Player already detects them and reports "Nope!"

Ghabry:
Can we agree on that this fixes #1181 and  #1367 even though there are more problems here?
Imo there can be new issues opened when they occur.

For #1181 not all ExecProgram calls are supported.
#1367 also has a patches exe to skip the title screen and load the first save, not related to Ineluki.

Limitations:
- ``ExecProgram`` not supported except for some specific programs that are HLE'ed
- ``MciProgram`` not supported
- Only alphanumeric and some special keys supported. This doesn't map properly to the key codes because Ineluki was written with a German keyboard in mind
- Cheat support does not reset when wrong key is pressed, needs some kind of input callback API, polling is ugly and highly inefficient here.

Problems:
- Mouse support needs ``--enable-mouse`` argument and this binds clicks and wheel to confirm and other stuff. So it conflicts with the patch
- WSAD conflict. We use them for walking but in RPG_RT they are not mapped.

Ideas are welcome about how to resolve the problems.

Tests:
- https://easyrpg.org/play/pr2397/?game=ineluki&enable-mouse
- https://easyrpg.org/play/pr2397/?game=ineluki_autorun&enable-mouse - Tests that the autorun.script is executed on startup
- https://easyrpg.org/play/pr2397/?game=calm_falls MP3 link file support in the web player (vs. before https://easyrpg.org/play/?game=calm_falls )

Best for testing are the two samples at the bottom:
- Diablo-Like Mouse control (click to walk)
- Tasten Test (Key Test) - Prints the Key ID

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -844,6 +844,8 @@ void Player::ResetGameObjects() {
 	Game_Clock::ResetFrame(Game_Clock::now());
 
 	Main_Data::game_system->ReloadSystemGraphic();
+
+	Input::ResetMask();
 }
 
 static bool DefaultLmuStartFileExists(const FileFinder::DirectoryTree& dir) {
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -809,7 +809,7 @@ void Player::CreateGameObjects() {
 
 	ResetGameObjects();
 
-	Main_Data::game_ineluki->ExecuteAutorunScript();
+	Main_Data::game_ineluki->ExecuteScriptList(FileFinder::FindDefault("autorun.script"));
 }
 
 void Player::ResetGameObjects() {
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -54,6 +54,7 @@
 #include "game_map.h"
 #include "game_message.h"
 #include "game_enemyparty.h"
+#include "game_ineluki.h"
 #include "game_party.h"
 #include "game_player.h"
 #include "game_switches.h"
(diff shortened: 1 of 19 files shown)


Ghabry:
Also added Ineluki MP3 Link file support to the web Player:

https://easyrpg.org/play/pr2397/?game=calm_falls

The title and the first map music are already links.

src/player.cpp
```diff
@@ -370,7 +371,13 @@ void Player::Update(bool update_scene) {
 	}
 
 	if (update_scene) {
+		if (Main_Data::game_ineluki) {
+			Main_Data::game_ineluki->Update();
+		}
+
 		Scene::instance->Update();
+
+		IncFrame();
```
Ghabry, line 380: oh bad rebase, must remove this IncFrame. sry

Ghabry:
This is ready for testing. The "Problems" should be discussed but otherwise I'm done here.

fdelapena (approved):

fdelapena:
Works for me :+1:. Regarding the need of adding the mouse arg and the WASD keys conflict, perhaps a flag when ineluki's key patch is detected somehow to toggle the mouse and keys behaviour could be a possible approach to handle this.

Ghabry:
WSAD and the mouse issue is solved:

I invented a new feature "key masking". The Input system can be told which keys are never reported as pressed.

I mask now WSAD or the Left/Right Mouse when the key/mouse support of Ineluki is requested.

This also allows now to always process mouse events in SDL2: When the ``mouse_flag (--enable-mouse)`` is not set then the input system masks the mouse now instead. (The ``mouse_flag`` maps left click to Confirm and scroll to Page up/down)

Issue #2396, opened 2020-10-12
CMake doesn't install bash completion scripts on Linux


akien-mga:
#### Name of the game:
n/a

#### Player platform:
Linux, packaging EasyRPG Player for Mageia

#### Attach files (as a .zip archive or link them)
n/a

#### Describe the issue in detail and how to reproduce it:
I'm packaging EasyRPG Player for the Mageia Linux distribution.

Since upgrading to 0.6.x and switching the buildsystem from autotools to CMake, the bash-completion script in `resources/unix/bash-completion/easyrpg-player` is no longer installed by `make install`.

I'd suggest making CMake install it on Linux as autotools used to.

Ghabry:
Yeah that's an oversight. Thanks for the report :)

Ghabry:
@akien-mga 
Do you know if there is any accepted standard convention in CMake to install the manpage?

Can already see the can of worms opening here because the manpage folder depends on the operating system...

akien-mga:
Well this is about the bash completion file, which are typically installed in `/usr/share/bash-completion/completions/`. I don't know if that can depend on the distro, but bash-completion has a FAQ entry on this: https://github.com/scop/bash-completion/blob/master/README.md#faq

`pkg-config --variable=completionsdir bash-completion` can be used to query the system's bash-completion folder. Maybe there's a CMake way to query that easily. Otherwise I'd say hardcoding `CMAKE_INSTALL_PREFIX/share/bash-completion/completions/` and waiting to see if anyone complains is a good way to do it too :)

As for multiple operating systems, I guess it's uncommon to have bash on Windows, and macOS defaults to zsh IIRC. Might be good to look into whether Windows and macOS can consume bash-completion scripts when configured to use this shell, but otherwise making the installation of the script Linux-specific should also be OK IMO.

Ghabry:
oh right bash completion. No idea why I researched manpages.

The FAQ also mentions a bundled CMake script file. Will see if I can get something working.

Yeah I will limit it to UNIX and make some testing on Apple. Makes no sense for Windows.

PR #2398, opened 2020-10-12
CMake: Install bash-completion and add man and doxygen status message

Ghabry:
Hardcoded the completion dir to "${CMAKE_INSTALL_DATADIR}/bash-completion/completions" as suggested in #2396 otherwise this is a mess because the one returned by bash is an absolute directory which makes "install" fail without sudo...

For Unix I always install it, for all other platforms I depend on bash-completion being installed. This way one can also get the completion on Msys2 or macOS if this makes any sense ^^

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -853,7 +853,8 @@ endif()
 # Handled as extra target "man"
 set(MAN_NAME easyrpg-player.6)
 find_program(A2X_EXECUTABLE NAMES a2x a2x.py)
-if(NOT A2X_EXECUTABLE STREQUAL "A2X_EXECUTABLE-NOTFOUND")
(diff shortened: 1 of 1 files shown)


Ghabry:
For emscripten, Windows and macOS I get:

```
-- Manual page: Unavailable
-- Bash completion: Unavailable
-- Doxygen: Unavailable
```
Which is good.

Local I get

```
-- Manual page: Generated
-- Bash completion: Available
-- Doxygen: Available (target "doc")
```

@akien-mga
Can you check if this patch works for you? Thanks.

akien-mga:
The patch works well for me (applied on top of 0.6.2.3) :+1:

fdelapena (approved):

fmatthew5876 (approved):

Issue #2335, opened 2020-09-13
2k3 Battle: Player waits for animations, RPG_RT doesn't


NovemberJoy:
Name of the game: Captain Novolin RPG (under debug conditions)

Player platform:
Ubuntu 20.04 GNU/Linux

Game download:
- [Here.](https://drive.google.com/file/d/1_0hS0zeV9QoAbik5LgERwGsPnFIAXB2N/view?usp=sharing)

Attach files (as a .zip archive or link them)
 - [Save14.zip](https://github.com/EasyRPG/Player/files/5213907/Save14.zip)
 - [easyrpg_log.txt](https://github.com/EasyRPG/Player/files/5213888/easyrpg_log.txt)

Describe the issue in detail and how to reproduce it:
Load the provided save file and walk into an enemy. Once Dreamer-001's turn comes, choose Brace (Defend). In RPG_RT, the battle continues while the animation plays, while Player waits until it finishes before continuing. This happens with other animations as well, but to a much lesser extent due to said animations not being anywhere near as long. For context, this character is usually under AI control, which never guards.

fmatthew5876:
This game is a good test candidate for 2k3 battle system with battle animation sprites.

My suspicion here is how we generally wait for battle animations to finish. We probably only want to wait for normal battle animations and not sprite battle animations.

rueter37:
In RPG_RT, it depends on the battle action. If an enemy starts a normal attack, the game doesn't wait for the battle animation, the damage is done immediately (You see the battle animation and the damage number appear at the same time). But if an enemy uses a skill, the game waits for the battle animation. And player actions always wait for the battle animation.

Issue #1681, opened 2019-03-10
2k3 Death animations loop in EasyRPG


NovemberJoy:
#### Player platform:
Windows 10 64-bit/Ubuntu 18.10 GNU/Linux

#### Attach files (as a .zip archive or link them)
[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/2949060/easyrpg_log.txt)
https://drive.google.com/open?id=1_0hS0zeV9QoAbik5LgERwGsPnFIAXB2N (this is the game, use save slot 11 for demonstration purposes)

#### Describe the issue in detail and how to reproduce it:
In battle, if any character in the player's party dies from any cause, their death animation will constantly loop, sound and all, in a way that RPG_RT doesn't.

PR #2399, opened 2020-10-12
Battle System Rewrite

### LINKED ISSUE #1496: Broken Animations (Battle)
**Game:** Alter A.I.L.A. Genesis

**Platform:** Windows/Android, ...
**Version:** v0.5.4 (Nov. 9) / Nightly Build (Nov. 10)

Attached a .gif of the issue, as well as the easyrpg_log.txt. The issue should be universal to every battle in the game, since all the battle animations were broken from the November 9th update onward. Also, I noticed that none of my attacks were showing the amount of damage dealt.

brokenanimation
[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/2568986/easyrpg_log.txt)

### LINKED ISSUE #1727: Battle 2k3 - dual wield and double attack not implemented
In 2k, dual wield is treated as a combined single attack with 2 weapons.

- [x] In 2k3, dual wield is actually 2 separate attacks. 
- [x] If the first hit kills the enemy, the second hit will randomly select a 2nd target.
- [x] If both weapons have double attack, you will attack 4 times. Potentially killing up to 4 enemies if all the blows are lethal.
- [x] If one weapon has double attack and the other not, you will attack 3 times.
- [x] Weapons designed to attack multiple times and attack all targets at the same time are also unimplemented, only attacking a single target one time.
- [x] Attack all weapons still allow you to select a target, and then it hits all enemies
- [x] Ignore evasion flag is treated separately for each weapon
- [x] Weapon sp cost is deducted for every hit in 2k3
- [x] Correct CBA right and left hand animations for each weapon. Including when actor is mirrored.
- [x] When combo is applied, actor will do all combo hits with first weapon, and then all combo hits with second weapon

We don't do any of this. Instead it looks like we just emulate the 2k behavior of doing a single combined hit.

For a double attack, we don't select a 2nd target if the first one kills.

We also don't even show the 2 weapon animations in 2k3, only the first one.

Open Questions:
- [x] How is hit chance determined for attack? For each weapon individually or max()
- [x] How is hit chance determined for physical skills?
- [x] Ignore evasion flag?

### LINKED ISSUE #2153: Memory of Moshimo: Monster "Turn End Processing" (ターン終了処理") displayed
#### Name of the game:
Memory of Moshimo
http://vipkohaku2013.web.fc2.com/ent/18.html

https://cloud.mastergk.de/f/5818bfeadfd54cd68aad/?dl=1

Start a battle test.
Because of bad sanity checks (out of bounds monster access via (*Main_Data::game_party)[target]) the first battles crash.
Use Battle 16. At first a slime is displayed (Correct), then a bat (Wrong).

This bat is called "ターン終了処理" (Turn End Processing) and must be invisible.

Another case of "You never notice this without knowing the language" :)

### LINKED ISSUE #2330: Battle2k3: Implement row for actors with no battle commands
Looks like there's an edge case we forgot, which is in the code:

https://github.com/EasyRPG/Player/blob/master/src/scene_battle_rpg2k3.cpp#L1472

Maybe this conditional just needs to be deleted now? 
Needs testing

### LINKED ISSUE #2347: Game "The doritos" 's Weapon glitch

---
 - The `easyrpg_log.txt` log file (you can find it in the game  #directory)
[Doritos.zip](https://github.com/EasyRPG/Player/files/5251249/Doritos.zip)
[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/5251251/easyrpg_log.txt)

### LINKED ISSUE #2388: Battle: Enemy flash inconsistency

### LINKED ISSUE #2391: Enemies who get death state inflicted by events stay invisible forever
When an event adds deaths state, the enemy is supposed to disappear immediately instead of fade out. We emulate this by calling `SetVisible(false)` but the enemy death animation logic uses `SetOpacity()` and so the visible flag never goes back to true.

This only happens for an event adding the state, it does not occur for damage.

This enemy animation stuff needs a refactor. 

RPG_RT just has 3 timers for enemy animations
* blink timer - for damage
* death timer - for death
* explode timer - for explode

The way we do it we try to reuse the CBA animation states. This might be workable but I'm afraid that later it will interfere with CBA implementation

### LINKED ISSUE #2394: Segmentation fault when an enemy kills himself due to a reflected spell
This issue is a post-0.6.2.3 regression and occurs on current master.

When an enemy casts a spell and his spell gets reflected and he gets killed by his spell, a segmentation fault is triggered.

### LINKED ISSUE #2406: Correct Battle2k Update Sequence
Each frame, RPG_RT runs a function called UpdateBattleState, which does the following. We should emulate how this works and call it the same way each frame regardless of what state the battle state machine is in.

Battle 2k
----------

2ke: RPG_SceneBattle::UpdateBattleState `00479350`

- [ ] Update Screen Effects
- [ ] Update Monster blink/death/explode timers
- [ ] Update Monster battle animations
- [ ] Update windows
- [ ] Update message window
- [ ] Run the interpreter for 1 frame
    - [ ] does not handle game over scene calls
    - [ ] does handle title scene calls, as that throws an exception
- [ ] Update timers
    - [ ] End battle if timer aborts
- [ ] Check for and call debug scene

### LINKED ISSUE #2418: Battle2k3: Incorrect battler placement on party change
Write a battle event which adds someone to your party. Use automatic placement.

In RPG_RT, when the battler is removed, the existing battlers don't change position. When a battler is added, all of the parties' positions are reset.

In Player, adding or removing has no effect.

### LINKED ISSUE #2442: 2k3 level up message isn't shown (Regression?)
Name of the game: 
Captain Novolin RPG

Player platform:
Ubuntu GNU/Linux 20.04, Funkey-OS 2.0.0

Game download:
[Here.](https://drive.google.com/file/d/1_0hS0zeV9QoAbik5LgERwGsPnFIAXB2N/view?usp=sharing)

Attach files (as a .zip archive or link them)
[Save01.zip](https://github.com/EasyRPG/Player/files/5849491/Save01.zip) (Just in case the Save01 with the game download isn't the right one.)
[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/5849494/easyrpg_log.txt)

Describe the issue in detail and how to reproduce it:
In the provided save file, walk to the right to the cutscene trigger and fight the battle given, wiping the battlefield clean with Missile Barrage. Once you do, assuming he survives, Capt Novolin will gain 22 EXP and a level, with the increase in his max HP visible upon the start of the next fight, but the "level up" message and raising of statistics won't occur during the victory messages.

### LINKED ISSUE #2443: 2k3: Gauge battle style's MP gauge is full when current and max is 0
Name of the game: 
Shyouzen 2

Download link: 
[Here.](https://drive.google.com/file/d/1h7qIVVTlXPHAxr3d7KkkoGqxPR6P3dqu/view?usp=sharing)

Player platform:
Ubuntu GNU/Linux 20.04, Funkey-OS 2.0.0

Attach files (as a .zip archive or link them)
Savegame is included in the download as Save01.lsd.
[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/5860113/easyrpg_log.txt)

Describe the issue in detail and how to reproduce it:
Walk through the area and enter an encounter. All battlers except Shyouzen have 0 MP and 0 max MP - RPG_RT draws these as empty, while Player draws them as full.

### LINKED ISSUE #2449: 2k3 Battle waits for animation when a BattleAnimation actor recovers from a "No Move" state and uses a skill
Name of the game: 
Shyouzen 2

Download link: 
[Here.](https://drive.google.com/file/d/1h7qIVVTlXPHAxr3d7KkkoGqxPR6P3dqu/view?usp=sharing)

Player platform:
Ubuntu GNU/Linux 20.04, Funkey-OS 2.0.0

Attach files (as a .zip archive or link them)
[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/5900019/easyrpg_log.txt)
Save game is included with the download.

Describe the issue in detail and how to reproduce it:
Load the provided save and walk through the area until you encounter the next enemy group. If they use Hypersonic and it hits Shyouzen, it causes a short-term stun that wears off within one turn, but the animation won't return to normal afterwards. If you then select a "Phaze" move when his turn comes around, the Player will wait until his entire idle animation plays before the fight continues. Strangely, it doesn't have the same effect if you select the "Use Weapon" (Attack) command.

### LINKED ISSUE #2476: Battle2k3: Vampires Dawn II should not play enemy attack sound effect - unknown RPG_RT binary version
When playing "Vampires Dawn II" (RPG Maker 2003 game) with RPG_RT that the enemy_attack_se effect is not played when an enemy attacks. I have checked the LDB database of the game - the enemy_attack_se is set to attack2 with volume and tempo set to 100% and the sound file exists in the Sound folder of the game. You see the enemy flashing once and the attack animation played on the target actor (with its own sound). But the enemy_attack_se effect is not played.

https://www.vampiresdawn.org/vd-2-ancient-blood/

I investigated further and this looks like an RPG_RT.exe binary version difference. If I replace the VD 2 RPG_RT with dynrpg binary, the sound effect plays.

Both the VD 2 RPG_RT.exe, advocate, and dynrpg binaries have FileVersion 1.0.8.0. This game appears to be using advocate, but the hash is different.

```
98149c681fa4f087d770e913a9d58e1f  RPG Maker 2003/Vampires Dawn II - Ancient Blood/RPG_RT.exe
```

```
fdd68cee936ea75c3929b91daa934532  ../../rpgrt/rpg2003/RPG_RT.2003.1.0.3.0.advocate.eng.exe
6b0e33b4e9af49f52e7db66a91174ee3  ../../rpgrt/rpg2003/RPG_RT.2003.1.0.3.0.jp.exe
ea8edc01fea5e776719528c9e6bdaf60  ../../rpgrt/rpg2003/RPG_RT.2003.1.0.8.0.advocate.eng.32bit_color.exe
eed8e0fb848521fcd1699b156e1cc50e  ../../rpgrt/rpg2003/RPG_RT.2003.1.0.8.0.advocate.eng.dynrpg.exe
ff0fb3b085047f93646832a762d2a883  ../../rpgrt/rpg2003/RPG_RT.2003.1.0.8.0.advocate.eng.exe
c87407cb68c51e87e2e19c3e5ac0b416  ../../rpgrt/rpg2003/RPG_RT.2003.1.0.9.1.fdelapena.spa.cleartype.exe
fa15d283766197c0e8833c48b69a60fc  ../../rpgrt/rpg2003/RPG_RT.2003.1.0.9.1.unknown.eng.exe
ad0e8f0dd16ca6cf88de1db55b78d840  ../../rpgrt/rpg2003/RPG_RT.2003.1.1.2.0.eng.exe
9373d938e1b79db88a6ddac9640bee98  ../../rpgrt/rpg2003/RPG_RT.2003.1.1.2.1.eng.exe
```

I took a peak at the VD 2 binary in RE tools and it is indeed different, very different at least in the logic that triggers the enemy SE effect.

In DynRPG the enemy flash and attack sound effect happens unconditionally in:

```
ProcessBattleAction (0049A04C)
SetupMonsterAction (0045D7D1)
GetEnemyAttackSE (0048BC84)
```

This is also the same for 2k3e

In the VD 2 binary:
*  `ProcessBattleAction (0049A04C)` just does `SetFlash (9004C1178)` inline
* The call to `GetEnemyAttackSE 0048BC84` happens in `RPG_Monsters::UpdateBattle (004BEE88)` during animation processing

This behavior also occurs in the 1.0.3.0 version of RPG_RT.exe. I checked the one I have an its Resource FileVersion says 1.0.4.0

So in summary, VD 2 `RPG_RT.exe` appears to be the same version as advocate, but at least has some legacy RPG_RT.exe behavior w.r.t. the battle system.

So in order to fix this in Player we need to:
- [ ] Identify exactly which RPG_RT version is used for VD 2
- [ ] Make Player detect it properly (currently detects it as MajorUpdated)
- [ ] Modify battle system code to not play the enemy SE effect for this unknown version

fmatthew5876:
The goal here is perfect emulation of all battle algorithm corner cases as well as trying as much as possible to simplify the battle algos

Bugs fixed:
- [x] Refactor 2k battle system to apply each effect separately at the right time
- [x] Console message timing around death in 2k battle scene
- [x] Move 2k HP healing clamping logic out of algo and into 2k battle system.
- [x] Removal of "forced alive" hack from `Sprite_Battler`
- [x] Correct handling of what other effects are and are not allowed when an algo kills the target
- [x] Physical state recovery does not happen for absorb hp
- [x] Messages do not happen when absorb hp does 0 damage
- [x] Correct handling of success or failure - all edge cases
- [x] Refactor and fix crashing bugs in reflect code
- [x] Add weapon support to normal attack
- [x] BattleAlgo use `unique_ptr`
- [x] Use signed values for hp, sp, atk, def, spi, agi. 0 means no effect. + means healing, - means damage
- [x] Make battle algo framework easier to extend for custom engines
- [x] Battle2k3: Attack all weapons still require targeting enemies first
- [x] Battle2k3: Actor should change direction with attack all weapons - move this "actor facing" logic to be triggered after using the monster target window
- [x] Refactor enemy sprite and animation timers to match RPG_RT behavior better
- [x] Refactor and cleanup actor sprites to allow later CBA implementation
- [x] RPG_RT accurate monster levitation algo
- [x] Battle action retargeting
- [x] Clean up battle algo animation logic
- [x] Correct timing for battle algo animations vs event animations vs usage messages.
- [x] Battle algo behavior when actors added to party
- [x] Battle algo behavior when actors removed from party
- [x] Switch skill / item vs action switches and 2k3 interpreter execution
- [x] Correct sequencing of update state: screen, monster animations, actor animations, windows, message, interpreter, timers, debug scene call
- [x] Correct handling of event calling game over, doesn't happen every frame, happens when events refreshed
- [x] Item vs Skill using messages
- [x] Message no target ??? - self, switch. ??
- [x] When skill is used by enemy, on a single target, another `IsSkillEffectiveOn()` check is performed and if it fails, the skill action is cancelled the same way it would be cancelled if the user didn't have enough SP. This is only performed for enemies and not actors.
- [x] Refactor 2k battle system to match RPG_RT
- [x] Correct Active and Wait ATB behavior
- [x] Fix charged normal attacks

Dual Wield - #1727 
- [x] Allow actor weapon double attack
- [x] Allow enemy double attack
- [x] Allow CBA multiple hits
- [x] Dual Wield attack with separate weapons and their stats and multiple hits
- [x] retargeting multiple hits if target ties
- [x] Single hit weapon + Attack all weapon

EasyRPG Battle System Extensions:
- [x] Battle Algo support arbitrary number of battle animations, not just 2 for dual wield in 2k battle system
- [x] Battle Algo support arbitrary number of start messages
- [x] Battle2k supports reflect state
- [x] Support absorbing all parameters. And support absorbing each one separately. Support added to 2k battle system for messages.

Issues:
- [x] Fix #1496
- [x] Fix #2330
- [x] Fix #2391
- [x] Fix #2394
- [x] Fix #2388
- [x] Fix #2347
- [x] Fix #2153
- [x] Fix #1727
- [x] Fix #2406
- [x] Fix #2418
- [x] Fix #2442
- [x] Fix #2443
- [x] Fix #2449
- [x] Fix #2476

TBD:
- [x] Verify 2k battle monster flashing behavior
- [x] Execute: check states with death inflicted by damage
- [x] Cancel to option window and event runs and makes all actors not controllable
- [x] Skill window actor index when actor added/removed from party
- [x] Test 2k3 battle turns for actor, monster, and inventory
- [x] Test if negative attribute affects atk, def, spi, agi
- [x] Test if attribute flipped damage is capped
- [x] Double check absorb with negative hp and sp
- [x] No battle animation is shown if an enemy uses a revival skill on his dead ally.

Manual Testing:
- [x] Retest #1677

Unit tests:
- [x] Basic Battle algorithm function
- [x] Simple algos: None, Defend, Observe, DoNothing, Escape
- [x] SelfDestruct algo
- [x] NormalAttack algo
- [x] Skill Algo
- [x] Item Algo 

2ke Addresses
---------------

| Address | Class | Function | Notes |
| -- | -- | -- | -- |
| 0047AFEC | SceneBattle | ProcessBattleAction | Performs usability checks and then dispatches to executing basic actions, skill, item etc.. |

DynRPG Addresses
---------------

Remaining issues
-------------

- [x] If an enemy gets killed due to direct death state infliction, the death message appears twice.
- [x] If a state gets canceled due to a physical attack, then after showing the state cancel message the state cancel message disappears but the attack message is still shown for a brief period.
- [x] If an enemy gets killed in the same turn where he gets a bad state recovered due to a physical attack or physical skill, the state recovery message gets displayed despite the enemy being dead.
- [x] If a target gets a state (like silence) which blocks skill use in the same turn before he acts he is still able to use a skill if he planned to do so.
- [x] https://github.com/EasyRPG/Player/pull/2399#issuecomment-739244662
- [x] https://github.com/EasyRPG/Player/pull/2399#issuecomment-739245525
- [x] The actor plays the pose animation and the battle animation in sequential order. In RPG_RT these are played at the same time.
- [x] The row battle command text is hardcoded to "Row". In RPG_RT the row term is used.
- [x] The enemy attack sound is played on enemy turn (RPG_RT uses this only in RPG Maker 2000 battles).
- [x] The ally selection flash is only half as fast than in RPG_RT (Waiting 30 frames instead of 15 between two flashes).
- [x] Battle animation sprites get invisible if they use a skill which affects the whole ally party.
- [x] Enemies doesn't play the death animation (disappearing with sound effect) if they get killed by direct death state infliction.
- [x] The enemies' blink effect, actors damage animation and the damage sound effects are not played on zero damage. In RPG_RT, the blink effects, damage animations and sound effects are played if the damage is greater or equal to zero and no absorb skill was used.
- [x] No healing number is displayed on the source if he successfully uses a HP absorb skill.
- [x] No damage/healing number is displayed if actors or enemies get their HP changed by an event.
- [x] Skills that unset switches don't unset the switch.
- [x] In Shyouzen Vs The Retards II (thanks for providing the link @NovemberJoy) when you battle the bats it is possible that one of the bats use the "Hypersonic" skill (skill which stuns the entire party for a short period) two times in a row in the same turn. When that happens, the battle hangs (regardless of manual or automatic battling). 
- [x] Automatic battling is broken in RPG Maker 2003 battles: After one automatic battle action has been performed, the battle mode switches to manual battling. This can always be reproduced and occurs independent of the ATB mode.
- [x] When getting a no move restriction state (except of death) while the ATB gauge is full, the actor loses his turn (the ATB gauge gets empty). In RPG_RT the ATB gauge remains full which means the actor is able to do an action as soon the restriction state is canceled. This special case can occur if you play on ATB active mode or if the enemy (which casts the restriction state skill) and the actor have the same agility values.


fmatthew5876:
Update on progress. This has been going for quite a long time now.

- [x] Battle Algos: 99% - I've got all the algos covered by unit tests and the new algorithm class interface looks to be just about where I want it to implement all of 2k, 2k3, and be extensible for custom battle systems.
- [x] 2k Battle System State Machine: 99% - This refactor is basically done, minus any existing bugs. 
- [x] 2k3 Battle System State Machine: 60% - There are still several cases I need to figure out in terms of what happens on what frame. The disassembly for this stuff is awful.
- [x] CBA: 10% - Not really planning to implement all of this, but I at least want to clean up the battle sprite code and separate everything out so in a clean way so that it can be implemented.


The 2k battle system can be tested now. However other changes I make in sprites or other areas could affect it and require retesting later.

fmatthew5876:
I've split out the battler sprites into `Sprite_Actor` and `Sprite_Enemy`. This should be ready now for someone to go in and implement the rest of CBA. I'm not going to go that far.

Last thing are a few more timing and compatibility fixes for 2k3 battle, then this PR is done.

src/scene_battle_rpg2k.cpp
```diff
@@ -120,267 +354,546 @@ void Scene_Battle_Rpg2k::SetState(Scene_Battle::State new_state) {
...
-	default:
-		break;
+
+	if (scene_action_substate == eFailure) {
+		SetState(State_Battle);
+		NextTurn();
+
+		CreateEnemyActions();
+		CreateExecutionOrder();
+		//Game_Battle::RefreshEvents();
```
fdelapena, line 888: debug leftover?

src/scene_battle_rpg2k3.cpp
```diff
@@ -520,722 +501,1778 @@ void Scene_Battle_Rpg2k3::CreateBattleTargetWindow() {
...
+			case lcf::rpg::BattleCommands::BattleType_traditional:
+				ResetWindows(false);
+				command_window->SetVisible(false);
+				target_window->SetVisible(true);
+				break;
+			case lcf::rpg::BattleCommands::BattleType_alternative:
+				ResetWindows(true);
+				status_window->SetVisible(true);
+				command_window->SetVisible(true);
+				//command_window->SetIndex(-1);
```
fdelapena, line 1414: intended for extending it if ever implemented for enemy targets?

fmatthew5876:
Addressed the review comments, thanks!

rueter37:
I noticed two oddities in RPG Maker 2000 battles while testing this PR:
- If a skill is used, the second message is always shown even if the `using_message2` attribute isn't set. In "Vampires Dawn" the output looks like the following:
`Valnar zaubert Lichtstrahl !`
`Valnar`
- If a target gets a state inflicted which he already has, the `message_affected` message is used instead of the `message_already` message so the wrong or even an empty message (with only the name of the target) is shown.

fmatthew5876:
Thanks, those are now fixed

fmatthew5876:
The 2k battle system is done. The 2k3 battle system still has more work to be done with CBA and there are probably some micro timing fixes that could go in to.

I'm finished with this PR and work on the 2k3 battle system. I will fix any regressions that come out of this but do not plan to do more major work on the battle system.

rueter37:
More oddities I encountered while testing:
- If a skill lowers the targets SP, an attack lowered message is displayed.
- Defending no longer works if hit by a skill. The target takes full damage.
- No battle animation is shown if an enemy uses a revival skill on his dead ally.
- (Not sure if this is related to this PR) The AI tries to use revival skills if hidden but no dead allies are present. If no hidden allies are present the AI behaves correctly.

Ghabry:
Design question (you don't have to implement it, is just for future work):
An extension I have in my mind since years is: Enemies should support the same animations that Actors support.
Is this enhancement easy to achive with your codebase or do you see any road blocks?

fmatthew5876:
> Design question (you don't have to implement it, is just for future work):
> An extension I have in my mind since years is: Enemies should support the same animations that Actors support.
> Is this enhancement easy to achive with your codebase or do you see any road blocks?

I split up the sprite classes in an attempt to achieve better frame accuracy. With the code the way it is now, it is harder to achieve that than it was in the old code. 

All that being said, the sprites use inheritance, so the idea is that a custom battle system could override and make it's own sprite class for monsters. I plan on doing this for hh3 as I want monster animations.

fmatthew5876:
I fixed the sp and defend issues. The battle animation one is a bit more involved to get right. I will address that one soon.

fmatthew5876:
Nevermind that, revive issue fixed also.

Thanks @rueter37 ! Let me know if you find anything else.

rueter37:
Two more oddities:
- If an enemy gets killed due to direct death state infliction, the death message appears twice.
- If a state gets canceled due to a physical attack, then after showing the state cancel message the state cancel message disappears but the attack message is still shown for a brief period.

Ghabry:
That's not algorithm related but there is a visual regression:

When selecting an enemy in the "gauge battle mode" there is now an empty help window shown.

UBSan reports an out of range integer for flash:

```
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior easyrpg-player/src/transform.h:43:4 in 
easyrpg-player/src/flash.h:71:58: runtime error: 256 is outside the range of representable values of type 'unsigned char'
    #0 0x55dbcc79190d in Flash::MakeColor(int, int, int, double) easyrpg-player/src/flash.h:71:58
    #1 0x55dbcc7a0ea6 in Game_Battler::GetFlashColor() const easyrpg-player/src/game_battler.h:891:9
    #2 0x55dbcc7b3a9e in Sprite_Enemy::Draw(Bitmap&) easyrpg-player/src/sprite_enemy.cpp:121:24
```

Ghabry:
About the ATB mode. There is a regression here:

In master and RPG_RT when an actor has "Berserk" (Auto attack because of state) and you have the "Command window" of a different actor open then the Berserk'ed actor will wait with the attack until you chose an action.
In your PR the Berserk actor will do attacks regardless whether the command window is open or not.

Ghabry:
Though in good news: I just found a heap-use-after-free after using a self-skill that inflicts a state. THis appears to be fixed here :+1:

rueter37:
Did some tests with RPG Maker 2003 battles out of curiosity ([because the RPG Maker 2003 part is not done yet](https://github.com/EasyRPG/Player/pull/2399#issuecomment-720892575)) and found the following oddities besides of the bugs @Ghabry mentioned:
- The actor plays the pose animation and the battle animation in sequential order. In RPG_RT these are played at the same time.
- The row battle command text is hardcoded to "Row". In RPG_RT the row term is used.
- The enemy attack sound is played on enemy turn (RPG_RT uses this only in RPG Maker 2000 battles).
- The ally selection flash is only half as fast than in RPG_RT (Waiting 30 frames instead of 15 between two flashes).
- Battle animation sprites get invisible if they use a skill which affects the whole ally party.
- Enemies doesn't play the death animation (disappearing with sound effect) if they get killed by direct death state infliction.
- The enemies' blink effect, actors damage animation and the damage sound effects are not played on zero damage. In RPG_RT, the blink effects, damage animations and sound effects are played if the damage is greater or equal to zero and no absorb skill was used.
- No healing number is displayed on the source if he successfully uses a HP absorb skill.
- No damage/healing number is displayed if actors or enemies get their HP changed by an event.

Please correct me if some of my observations are wrong or if I missed something.

To be fair, most of these oddities are probably due to the fact that the RPG Maker 2003 battle system rewrite is not done yet.

But aside of the oddities, the RPG Maker 2000 and 2003 battles work pretty good now! Outstanding work @fmatthew5876! :+1:

fmatthew5876:
I fixed and retested issues with battle 2k message timing. I've added the remaining bug list to the original post and will be going thru them one by one.

fmatthew5876:
@rueter37 Can you please elaborate on this one?

* The enemy attack sound is played on enemy turn (RPG_RT uses this only in RPG Maker 2000 battles).

I checked and the enemy attack sound always happens when enemies act in 2k3, but not in 2k. My PR seems to be correct.

Or did I miss an edge case?

rueter37:
I know from playing "Vampires Dawn II" (RPG Maker 2003 game) with RPG_RT that the `enemy_attack_se` effect is not played when an enemy attacks. I have checked the LDB database of the game - the `enemy_attack_se` is set to `attack2` with volume and tempo set to 100% and the sound file exists in the Sound folder of the game. You see the enemy flashing once and the attack animation played on the target actor (with its own sound). But the `enemy_attack_se` effect is not played.

fmatthew5876:
>  If a target gets a state (like silence) which blocks skill use in the same turn before he acts he is still able to use a skill if he planned to do so.

I found an extension to this. If you make an event which silences and then immediately cures silence the skill is still cancelled. This happens in the code which inflicts states.

Whenever a state is inflicted, RPG_RT will check if the current action is a skill, and if it is, check if it's usable. If it's not usable, the action is cancelled.

fmatthew5876:
> The enemies' blink effect, actors damage animation and the damage sound effects are not played on zero damage. In RPG_RT, the blink effects, damage animations and sound effects are played if the damage is greater or equal to zero and no absorb skill was used.

@rueter37 It looks like the blink always happens, even on 0 damage. The actor damage pose however does not happen on 0 damage.

Ghabry:
Jenkins: test this please

rueter37:
> @rueter37 It looks like the blink always happens, even on 0 damage. The actor damage pose however does not happen on 0 damage.

@fmatthew5876 Thanks for the heads up! I must have most likely overseen that one with the actor damage pose while testing with RPG_RT.

Ghabry:
Report by @NovemberJoy

Skills that unset switches don't unset the switch.

NovemberJoy:
When an actor using BattleAnimations recovers from a "No Move" state, the battle stops progressing until their standard animation completes, similar to #2335

NovemberJoy:
When a BattleAnimation actor recovers from a "No Move" state, battle pauses until their neutral animation plays, similar to #2335 . Currently borked in master as well, seems like a regression.

rueter37:
I have closed my PRs #2328 and #2354 because they are outdated and interfered with this PR. They contained battle fixes in relation the battle animation sprites and some UI changes.

Below a list of RPG Maker 2003 battle issues I tried to fix:

Battle animation sprite related issues:
- Placement of battlers: The battlers are placed wrong (too much distance between each other and partially and full off-screen placing starting with the third party member) if battle animation sprites are used. BattleCharset sprites are working fine.
- Displaying state afflictions: While BattleCharset sprites are showing its state afflictions at the beginning of the battle, the battle animation sprite battlers don't do so until their first turn.
- Battle animation sprites are not flipped if the flip flag is set.
- Battle animation height: Some battle animations are not displayed on the right height if the target is a battle animation sprite (depends on the height setting of the battle animation).
- Battler flashes: The battler flashes don't work if battle animation sprites are used. Again BattleCharset sprites fully work here.
- Battler shake effects: The battler shake effects don't work if battle animation sprites are used. Again BattleCharset sprites fully work here.

UI related issues:
- Transparent windows use an alpha of 160 in RPG_RT.
- In traditional style battles the SP window got printed over when a skill which costs SP has been used and the HP value didn't change its color if the HP reached critical or zero.
- In alternative style battles the gauges are shown transparent if the window is opaque in RPG_RT.
- The ally cursor is always drawn 40 pixels above the Y-Position of the battler. (Tested with RPG_RT, but correct me if this is wrong nevertheless)

@fmatthew5876: Unfortunately I don't know if fixing those issues is outside the scope of your PR. Please tell me if this is the case.

fmatthew5876:
I plan to fix any issues in 2k battle system. I will only fix regressions from master in 2k3 + a few things if they are easy to do.

I do not plan on finishing 2k3 in this PR.

I suggest rebasing your prs and reviving your work after this goes in.

rueter37:
Thanks for the information! I will reopen my PRs after your PR is merged in.

rueter37:
The bug where battle actions which unset switches doesn't unset the switch is most likely caused by an oversight in the [Enemy AI code, lines 165-172](https://github.com/EasyRPG/Player/blob/master/src/enemyai.cpp#L165-L172). Changing the code in [line 170](https://github.com/EasyRPG/Player/blob/master/src/enemyai.cpp#L170) to `algo->SetSwitchDisable(action.switch_off_id)` probably fixes this.

fmatthew5876:
@rueter37 
 > Battle animation sprites get invisible if they use a skill which affects the whole ally party.

I can't reproduce this. Do you have a test case or specific example?

fmatthew5876:
> When a BattleAnimation actor recovers from a "No Move" state, battle pauses until their neutral animation plays, similar to #2335 . Currently borked in master as well, seems like a regression.

Since this is broken in master. I'm leaving it as out of scope. Please write a github issue.

I want to get this PR wrapped up and merged asap.

rueter37:
@fmatthew5876 you can reproduce the battle animation sprites getting invisible in the game "Vampires Dawn II". Use the skill "Heilender Regen" (Skill which heals the entire party) in battle and the caster disappears until he starts another action.

fmatthew5876:
Do you have a savegame where I can quickly reproduce it?

rueter37:
Sadly I haven't got a savegame. I always did it with the test play mode, starting on map ID 45 with party member IDs 1, 3 and 2. Then over the debug menu I add the item with the ID 271 to the inventory and use it on Alaine (this learns her the Skill "Heilender Regen"). And then I start a battle over the debug menu (Troop ID 1, Seemonster). Let Alaine use the "Heilender Regen" skill and see her disappear.

rueter37:
@fmatthew5876 I have now added a savegame where you can reproduce the issue with the battle animation sprites getting invisible upon using a skill which affects the whole player party: [Save01.zip](https://github.com/EasyRPG/Player/files/5905819/Save01.zip). Run into one of the monsters and let Alaine use the "Heilender Regen" skill. She will disappear afterwards until she starts another battle action. One possible fix for the issue is to change [line 257 of the sprite actor code](https://github.com/fmatthew5876/Player/blob/batalgo/src/sprite_actor.cpp#L257) to `if (animation || idling_anim != anim_state)` (At least it works for me).

rueter37:
While testing this PR I came across a serious bug in RPG Maker 2003 battles which makes the battle hang: In [Shyouzen Vs The Retards II](https://drive.google.com/file/d/1h7qIVVTlXPHAxr3d7KkkoGqxPR6P3dqu/view?usp=sharing) (thanks for providing the link @NovemberJoy) when you battle the bats it is possible that one of the bats use the "Hypersonic" skill (skill which stuns the entire party for a short period) two times in a row in the same turn. When that happens, the battle hangs (regardless of manual or automatic battling). I have a savegame ([Save02.zip](https://github.com/EasyRPG/Player/files/5920073/Save02.zip)) where the bug can be quickly reproduced: Run right into the black blob. This starts the battle against the bats. If one of them cast the "Hypersonic" skill, the bug mentioned above can occur.

rueter37:
Automatic battling is broken in RPG Maker 2003 battles: After one automatic battle action has been performed, the battle mode switches to manual battling. This can always be reproduced and occurs independent of the ATB mode.

rueter37:
When getting a no move restriction state (except of death) while the ATB gauge is full, the actor loses his turn (the ATB gauge gets empty). In RPG_RT the ATB gauge remains full which means the actor is able to do an action as soon the restriction state is canceled. This special case can occur if you play on ATB active mode or if the enemy (which casts the restriction state skill) and the actor have the same agility values.

rueter37:
@fmatthew5876 I want to assist you to get your PR merged in so I made some [commits](https://github.com/rueter37/Player/tree/batalgo-changes) which you can put into your branch to fix the remaining issues mentioned in the thread. Probably the quality of my code is too bad but I hope that my commits are at least helpful to give pointers to get the issues fixed.

By the way, as soon as your PR is merged in, I plan to create PRs to make RPG Maker 2003 battles even better:
- PR which fixes the issues mentioned by @NovemberJoy (#1681, #2335, #2442, #2443, #2449)
- PR which fixes issues in relation to battle animation sprites (successor of #2328)
- PR which makes changes to the user interface in RPG Maker 2003 battles (successor of #2354)
- PR which partially implements CBA (Yes, I remember that [joke comment](https://easyrpg.org/irc/log/2020/2020-08-28.html) from @Ghabry :smile:)

fmatthew5876:
I've fixed the outstanding issues

@rueter37 Thanks for those commits, I was able to use some of them and others inspired the cause of the issue. I had some questions about your commits.

* Battle 2k: Fix state messages - What is this fixing exactly? Do you have a test case? I tested these 2 scenarios and they both work fine without your patch:
    * Enemy has a state with 100% physical recovery - Weapon damage kills enemy
    * Enemy has a state with 100% physical recovery - Weapon damage does not kill, but weapon inflicts death with 100%.

Both cases Player does the same as RPG_RT. Is there another test case I missed?

* VD 2 enemy attack sound not played

So the SE doesn't occur for legacy 2k3 and I fixed that. But for VD 2 itself the issue is more complicated. Punted to #2476 

* Battle 2k3: Change sound effects on damage

Do you have a test case for this? When I look at RPG_RT my code seems right, but maybe I missed something.


I don't plan to do anymore here unless some serious regressions are found. We gotta get this PR merged. Apologies for the very long delay.

fdelapena (approved):

rueter37:
I have tested the state messages again and while they are fixed on normal physical attacks, the state recover message still appears sometimes if you kill an enemy using a physical skill (`failure_message` set to 3). I have tested it by putting the enemy to sleep and then using a physical skill which kills the enemy by damage. Sometimes after the enemy killed message the state recover message (enemy awakes) appears.

rueter37:
The "Change sound effects on damage" commit has been made to stop the damage sound being played on negative damage. I recall from playing "Vampires Dawn II" with RPG_RT that the damage sound effect is not played on negative damage.

fmatthew5876:
> I have tested the state messages again and while they are fixed on normal physical attacks, the state recover message still appears sometimes if you kill an enemy using a physical skill (`failure_message` set to 3). I have tested it by putting the enemy to sleep and then using a physical skill which kills the enemy by damage. Sometimes after the enemy killed message the state recover message (enemy awakes) appears. Because of this I made this "Fix state messages" commit.

I just tried with a physical skill and still can't reproduce it. Do you have a test case? Or at least a screen capture of the skill?

rueter37:
Here is a screenshot:
Screenshot_20210313_202158
After I killed the last sleeping slime with the attack "Special blow", the "Slime is now awake" message appeared.

You can reproduce this with this [modified version of TestGame-2000](https://github.com/EasyRPG/Player/files/6135405/TestGame-2000.zip): Start in debug mode a battle against the slimes and then use "Ultra Sleep" to make the slimes sleep. Then use "Special blow" on a sleeping slime. Then sometimes this message appears like in the screenshot.

fmatthew5876:
Thanks, I was able to reproduce it! Nice catch.

This is actually a battle algo bug and not a message bug. States should not be recovered if the skill or weapon kills the target. Normal attack already had this logic but skill did not.

fmatthew5876:
@rueter37 You should really hop on irc or discord and chat with us!

rueter37:
I can confirm that the state messages are fixed now! :+1:

About IRC chat: I tried once or twice (cannot remember) but I'm lacking confidence...

fmatthew5876:
You've done a lot of work helping us test and providing some bug fixes. We can communicate better on chat and it will help things move faster. Come on over and say hello, we welcome you.

rueter37:
I think we can mark #2443 and #2449 fixed by your PR as well.

rueter37 (approved):

rueter37:
Perfect, I did more testing and can confirm that the remaining bugs are fixed! :+1: This gets my approval.
The last open question is the [Change sound effects on damage commit.](https://github.com/EasyRPG/Player/pull/2399#issuecomment-798770474)

Ghabry:
Added #1496, this looks fixed
Is possible that some animations have CBA but at least the damage display and the animations I tested look perfect now.

src/window_message.h
```diff
@@ -128,6 +128,12 @@ class Window_Message: public Window_Selectable {
 	/** @return the last async operation */
 	AsyncOp GetAsyncOp() const;
 
+	/** @return the number of lines per page */
+	int GetMaxLinesPerPage() const;
+
+	/** @return the number of lines per page */
```
Ghabry, line 134: should be "@param lines"

Ghabry (commented):
This one is obviously impossible to review. I tested a few games with previously fixed bugs. None of them broke.

But I have a few more games on the list I want to retest.

In the interest of getting this out quickly to the Android "testers" I grant a review/testing window until **22 March**.

rueter37:
Found a small bug while doing more testing: Dead actors gain EXP after a victory in RPG Maker 2003 battles. RPG Maker 2000 battles work correctly, dead actors don't gain EXP there. I can provide a fix for this: https://github.com/rueter37/player/tree/batalgo-2k3-exp.

Ghabry:
I still have the feeling that this is more unresponsive that the RPG_RT system: When an enemy attacks I cannot interact with the command menu but in RPG_RT I can. But I guess @rueter37 is already aware of this in one of the patches? ^^

rueter37:
Sorry, I must have missed this one. Probably because I'm always playing on ATB Wait Mode.

fmatthew5876:
> I still have the feeling that this is more unresponsive that the RPG_RT system: When an enemy attacks I cannot interact with the command menu but in RPG_RT I can. But I guess @rueter37 is already aware of this in one of the patches? ^^

Fixing that requires another big refactor. The problem is battle actions happen in `State_Battle` while UI control is in one of the Ui navigation states. So you can't have both at the same time.

What we really need is to decouple these 2 things so that the battle state machine and the UI state machine can run concurrently.

This is even more annoying because half of the state machine is baked into `Scene_Battle` and shared across 2k and 2k3..

Thats not getting fixed in this PR

Ghabry (approved):
Well I couldn't find a game that is broken by this. Amazing work.

ro6qtqmwt3y41

PR #2400, opened 2020-10-12
More cmake fixes

Ghabry:
Lets keep this open for a while.
libretro has a new shiny buildbot, so I can finally do proper testing... Bet there will be more commits.


Ghabry:
My last patch for libretro was in November, so its fine now to upstream this stuff.

Still works. The FPS counter is funny, maybe broken Vsync, but not my problem, is Retroarch. :)

Screenshot_20210114_211535-fs8

fmatthew5876 (approved):

fdelapena (approved):

CMakeLists.txt
```diff
@@ -473,17 +475,28 @@ elseif(${PLAYER_TARGET_PLATFORM} STREQUAL "switch")
...
 	message(FATAL_ERROR "Invalid target platform")
 endif()
 
+if(3DS)
+	target_compile_definitions(${PROJECT_NAME} PUBLIC _3DS)
+elseif(VITA)
+	target_compile_definitions(${PROJECT_NAME} PUBLIC PSP2)
+elseif(NSWITCH)
+	target_compile_definitions(${PROJECT_NAME} PUBLIC __SWITCH__)
+endif()
```
carstene1ns, line 488: I dislike here that libretro dictates this. At least for 3DS we will need `if(3DS OR N3DS)` to stay compatible with devkitpro toolchain file or do something custom.

src/platform_clock.h
```diff
@@ -22,7 +22,10 @@
 #include <type_traits>
 #include <thread>
 
-#if defined(_3DS)
+#if defined(USE_LIBRETRO)
+#include "platform/libretro/libretro_clock.h"
+using Platform_Clock = LibretroClock;
```
carstene1ns, line 27: Just asking because the commit message does not show the intention clearly:
We want to always use the libretro clock, even if we have a "better" one for some platforms?

carstene1ns (approved):

CMakeLists.txt
Ghabry: I must improve the commit message. I don't plan 3DS support for libretro because our optimized port already runs badly.

This is actually not dictated by libretro. They hate CMake and only use there ugly makefiles. I just added this for consistency.

Must have looked in the wrong toolchain file. Will fix this, thanks

src/platform_clock.h
Ghabry: That's a good question. I have to think about it

Ghabry: I decided to only use the libretro clock where we use the "default clock"

CMakeLists.txt
Ghabry: fixed

PR #2413, opened 2020-10-27
Add snapcraft recipe

lotharsm:
Last weekend, I added the EasyRPG Player to the Snap Store (https://snapcraft.io/easyrpg-player) in order to make current EasyRPG Player easily available on Ubuntu and other Linux distributions.

For reference, this is the recipe used to build the snap package.

I'm planning to keep this package up-to-date in the future (since I really like the EasyRPG player) and backport all changes into this repository.

However, if you prefer to publish this under your own name/organisation, please ping me and I'll arrange a transfer of the Snap ownership (yes, it's a bit unfortunate that the Snap Store calls me "Developer" rather than "Publisher").

### DIFF
--- a/builds/snap/snapcraft.yaml
+++ b/builds/snap/snapcraft.yaml
@@ -38,6 +38,9 @@ parts:
     plugin: autotools
     source: https://github.com/EasyRPG/liblcf
     source-type: git
+    build-environment:
+      - CXXFLAGS: "$CXXFLAGS -fuse-ld=gold -flto=$(nproc) -ffunction-sections -fdata-sections"
+      - LDFLAGS:  "$LDFLAGS  -fuse-ld=gold -flto=$(nproc) -Wl,--gc-sections"
     autotools-configure-parameters:
       - --prefix=/usr
       - --disable-static
@@ -117,6 +120,9 @@ parts:
     plugin: cmake
     source: https://github.com/Mindwerks/wildmidi
     source-type: git
+    build-environment:
+      - CXXFLAGS: "$CXXFLAGS -fuse-ld=gold -flto=$(nproc) -ffunction-sections -fdata-sections"
+      - LDFLAGS:  "$LDFLAGS  -fuse-ld=gold -flto=$(nproc) -Wl,--gc-sections"
     cmake-parameters:
       - -DCMAKE_BUILD_TYPE=Release
       - -DBUILD_SHARED_LIBS=ON
--- a/builds/snap/snapcraft.yaml
+++ b/builds/snap/snapcraft.yaml
@@ -114,7 +114,6 @@ parts:
       - libwayland-egl1-mesa
       - libwildmidi-config
       - libxmp4
-      - locales-all
 
   libwildmidi:
     plugin: cmake
@@ -135,5 +134,3 @@ parts:
     plugin: make
     build-packages:
       - libglib2.0-dev
-    stage-packages:
-      - libglib2.0-bin
--- a/builds/snap/snapcraft.yaml
+++ b/builds/snap/snapcraft.yaml
@@ -44,11 +44,10 @@ parts:
     autotools-configure-parameters:
       - --prefix=/usr
       - --disable-static
+      - --disable-tools
     override-build: |
       git fetch
       snapcraftctl build
-      strip --strip-all $SNAPCRAFT_PART_INSTALL/usr/bin/lcf2xml
-      strip --strip-all $SNAPCRAFT_PART_INSTALL/usr/bin/lcfstrings
       strip --strip-all $SNAPCRAFT_PART_INSTALL/usr/lib/liblcf.so.0.0.0
     build-packages:
       - libexpat1-dev
@@ -75,7 +74,7 @@ parts:
       if [ "${last_committed_tag}" != "${last_released_tag}" ]; then
         git fetch
         git checkout -f "${last_committed_tag}"
-        snapcraftctl set-version $(git -C ../src tag --list | head -n1)
+        snapcraftctl set-version $(git -C ../src tag --list | tail -n1)
       else
         snapcraftctl set-version $(git describe | sed 's/desc\///')
       fi
@@ -93,7 +92,6 @@ parts:
       - libsndfile1-dev
       - libsdl2-dev
       - libsdl2-mixer-dev
-      - libsdl2-ttf-dev
       - libspeexdsp-dev
       - libvorbis-dev
       - libxmp-dev
@@ -108,9 +106,9 @@ parts:
       - libglu1-mesa
       - libopusfile0
       - libpixman-1-0
+      - libpng16-16
       - libsdl2-2.0-0
       - libsdl2-mixer-2.0-0
-      - libsdl2-ttf-2.0-0
       - libsndfile1
       - libspeexdsp1
       - libvorbis0a
--- a/builds/snap/snapcraft.yaml
+++ b/builds/snap/snapcraft.yaml
@@ -47,6 +47,9 @@ parts:
     override-build: |
       git fetch
       snapcraftctl build
+      strip --strip-all $SNAPCRAFT_PART_INSTALL/usr/bin/lcf2xml
+      strip --strip-all $SNAPCRAFT_PART_INSTALL/usr/bin/lcfstrings
+      strip --strip-all $SNAPCRAFT_PART_INSTALL/usr/lib/liblcf.so.0.0.0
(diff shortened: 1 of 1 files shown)


Ghabry:
Jenkins: Test this please

Sorry for the late response.
Thanks for the PR. Personally I'm not interested in maintaining a Snap repository, so I'm fine with a 3rd-party maintaining this (we already provide packages through OBS, though they are a bit outdated because our maintainer for this not around anymore)

This is not invasive and we already have lots of (stale...) makefiles in the build folder so I don't see a reason against this here.

@lotharsm Could you also provide a snapfile for the stable version? Then I could cherry-pick this into the stable-branch - just for reference, as I don't plan another stable release.

fdelapena (approved):

builds/snap/snapcraft.yaml
```diff
@@ -0,0 +1,133 @@
...
+      - locales-all
+
+  libwildmidi:
+    plugin: cmake
+    source: https://github.com/Mindwerks/wildmidi
+    source-type: git
+    cmake-parameters:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DBUILD_SHARED_LIBS=OFF
+      - -DWANT_PLAYER=ON
```
carstene1ns, line 121: - Player component is not needed (also, no alsa library)
- Reason why this library should be static?
- `CMAKE_INSTALL_PREFIX=/usr` should be set to avoid `/usr/local` mess.

builds/snap/snapcraft.yaml
```diff
@@ -0,0 +1,133 @@
...
+  /usr/share/midi/freepats:
+    bind: $SNAP/usr/share/midi/freepats
+
+parts:
+  liblcf:
+    plugin: autotools
+    source: https://github.com/EasyRPG/liblcf
+    source-type: git
+    autotools-configure-parameters:
+      - --prefix=/usr
```
carstene1ns, line 42: Add `--disable-static` to save 275M install size.

builds/snap/snapcraft.yaml
```diff
@@ -0,0 +1,133 @@
...
+  easyrpg-player:   
+    plugin: autotools
+    after: [libwildmidi,liblcf,desktop-glib-only]
+    source: https://github.com/EasyRPG/Player
+    source-type: git
+    autotools-configure-parameters:
+      - --prefix=/usr
+    override-build: |
+      last_committed_tag="$(git tag --list | tac | head -n1)"
+      trimmed_tag="$(echo $last_committed_tag | sed 's/desc\///' | sed 's/git//' | sed 's/^v//')"
```
carstene1ns, line 62: These should be simplified.
No need to reverse: `git tag --list | tac | head -n1` → `git tag --list | tail -n1`
No real need to trim anything here, since our tags are only numeric (loosely based on semver)

builds/snap/snapcraft.yaml
```diff
@@ -0,0 +1,133 @@
...
+      - libsndfile1-dev
+      - libspeexdsp-dev
+      - libvorbis-dev
+      - libxmp-dev
+
+    stage-packages:
+      - freepats
+      - libfreetype6
+      - libharfbuzz0b
+      - libmpg123-dev
```
carstene1ns, line 95: `-dev` package should be changed to "normal" one.

carstene1ns (changes requested):
Good work on this so far! :+1:

However, as I am an oldschool guy, the download and extracted sizes bug me a bit.
The snap content is almost 800M and download size 163M. We can save a bit bandwith and disk space (some pointers in the detailed review).

builds/snap/snapcraft.yaml
```diff
@@ -0,0 +1,133 @@
...
+      - libfmt-dev
+      - libfreetype6-dev
+      - libharfbuzz-dev
+      - libmpg123-dev
+      - libopusfile-dev
+      - libpixman-1-dev
+      - libpng-dev
+      - libsdl2-dev
+      - libsdl2-mixer-dev
+      - libsdl2-ttf-dev
```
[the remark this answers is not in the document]
lotharsm: libsdl seems to be at least a dependency at build time:
```
checking for SDL... no
checking for SDL... no
configure: error: Package requirements (sdl) were not met:

No package 'sdl' found
```

builds/snap/snapcraft.yaml
lotharsm: Fixed.

builds/snap/snapcraft.yaml
lotharsm: Fixed too.

builds/snap/snapcraft.yaml
lotharsm: Thanks a lot, I can even re-use this for some other snaps :+1:

builds/snap/snapcraft.yaml
lotharsm: Ah, right - thanks!

lotharsm:
Thanks @carstene1ns for the suggestions!

The download size for the snap itself is now 127M, the total installed size with all dependencies (this includes the environment brought in by snapd itself) is now at 527M.

I'll try to bring it down further, but I think we are pretty close to the limit here.

lotharsm:
Okay, the download size is now down to 91M with an installed size of 248M. This is it, I think.

builds/snap/snapcraft.yaml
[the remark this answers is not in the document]
carstene1ns: What I meant was that we do not depend on the ttf library, so `libsdl2-ttf-dev` and `libsdl2-ttf-2.0-0` should be removed.

builds/snap/snapcraft.yaml
```diff
@@ -0,0 +1,139 @@
...
+      - CXXFLAGS: "$CXXFLAGS -fuse-ld=gold -flto=$(nproc) -ffunction-sections -fdata-sections"
+      - LDFLAGS:  "$LDFLAGS  -fuse-ld=gold -flto=$(nproc) -Wl,--gc-sections"
+    autotools-configure-parameters:
+      - --prefix=/usr
+      - --disable-static
+    override-build: |
+      git fetch
+      snapcraftctl build
+      strip --strip-all $SNAPCRAFT_PART_INSTALL/usr/bin/lcf2xml
+      strip --strip-all $SNAPCRAFT_PART_INSTALL/usr/bin/lcfstrings
```
carstene1ns, line 51: Is it possible to launch the tools from the snap directly? Otherwise I would suggest to disable them alltogether.
(`--disable-tools` configure switch)

builds/snap/snapcraft.yaml
```diff
@@ -0,0 +1,139 @@
...
+      - --prefix=/usr
+    override-build: |
+      last_committed_tag="$(git tag --list | tail -n1)"
+      last_released_tag="$(snap info easyrpg-player | awk '$1 == "latest/beta:" { print $2 }')"
+      # If the latest tag from the upstream project has not been released to
+      # beta, build that tag instead of master.
+      if [ "${last_committed_tag}" != "${last_released_tag}" ]; then
+        git fetch
+        git checkout -f "${last_committed_tag}"
+        snapcraftctl set-version $(git -C ../src tag --list | head -n1)
```
carstene1ns, line 78: oversight: `head` → `tail`

builds/snap/snapcraft.yaml
```diff
@@ -92,14 +95,13 @@ parts:
...
       - libharfbuzz0b
-      - libmpg123-dev
+      - libmpg123-0
       - libstdc++6
       - libgl1-mesa-dri
       - libgl1-mesa-glx
       - libglu1-mesa
       - libopusfile0
       - libpixman-1-0
-      - libpng16-16
```
carstene1ns: libpng should stay, even if there is an indirect dependency by pixman or such.

builds/snap/snapcraft.yaml
lotharsm: Currently, the tools are not available - the only command that's exposed right now is the `easyrpg-player` command itself. I'll disable them for now and furthermore think about how to best integrate them at a later point in time.

builds/snap/snapcraft.yaml
lotharsm: Thanks, fixed!

carstene1ns (approved):
:+1:

carstene1ns:
I know this is not necessary, but anyway:

jenkins: test this please.

PR #2423, opened 2020-11-25
Update Emscripten shell with ES6 flavour, icon, fixes and more

johannschopplich:
Hello there!

first and foremost: What a great project you have built over the years. Astonishing. 

I was deploying some RM2k games and looked deeper into the HTML/JS part. I've beautified the shell HTML file slightly and would be happy if the following changes sound good to you, too:

**Changes incorporated**:
- prevent pinching, zooming and other gestures on touch-input based devices by setting `touch-action` to `none`
- remove unnecessary meta tags for HTML5 environments
- use button instead of input element to improve semantics
- essential CSS button resets
- embed SVG icon `expand` from [Teenyicons](https://teenyicons.com/):
  grafik
- use `image-rendering: crisp-edges` as fallback, since Firefox supports `crisp-edges` but not `pixelated` and Chrome supports `pixelated` but not `crisp-edges` (Reference: https://caniuse.com/?search=pixelated)
- `type="text/javascript"` can be omitted in HTML5
- remove dead code around non-existing `output` element
- since WASM comes to use, we can safely assume ES6+ support in the executing browsers, so let's simplify some functions, e.g.:
```diff
-printErr: function(text) {
-  if (arguments.length > 1) text = Array.prototype.slice.call(arguments).join(' ');
-    console.error(text);
-},
+printErr: (...args) => {
+  console.error(...args);
+},
```

**In use**
- You can test the changes made here: https://realtroll.de/player/?game=die-reise-ins-all

****

**Notes**:
- I've made intendation consistent and therefore diffing the HTML file is not appealing visually. I'm sorry. 
- I've tried all day to get the emscripten buildchain up'n'running. I've downloaded the emscripten artifacts, tried to replicate the CMake command used on the Jenkins server but had no luck. Where can I find instructions to build it (on macOS)? Thanks in advance!
- The emscripten compiler introduces some syntax errors while optimizing the final `index.html`. E.g. the file begins with `!doctypehtml>`, where a space is missing. There are further bugs down the road. Maybe we can soften the optimization level of `emcc`? I had no luck to try the buildchain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,8 @@ EasyRPG Player makes use of the following 3rd party software:
 * [Dirent] interface for Microsoft Visual Studio -
   Copyright (c) 2006-2012 Toni Ronkko, provided under the MIT license
 
+* [Teenyicons] - Copyright (c) 2020 Anja van Staden, provided under the MIT license
+
 [liblcf]: https://github.com/EasyRPG/liblcf
 [vcpkg]: https://github.com/Microsoft/vcpkg
 [#easyrpg at irc.freenode.net]: https://kiwiirc.com/nextclient/#ircs://irc.freenode.net/#easyrpg?nick=rpgguest??
@@ -171,3 +173,4 @@ EasyRPG Player makes use of the following 3rd party software:
 [FMMidi]: http://unhaut.x10host.com/fmmidi/
 [PicoJSON]: https://github.com/kazuho/picojson
 [Dirent]: https://github.com/tronkko/dirent
+[Teenyicons]: https://github.com/teenyicons/teenyicons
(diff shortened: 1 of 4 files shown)


Ghabry:
Jenkins: Test this please

Thank you for your contribution!

Could you also take a look at the default emscripten shell? https://github.com/emscripten-core/emscripten/blob/master/src/shell.html

Our shell was based on it so maybe some of your enhancements also apply to this shell? Would be good to upstream the relevant parts, then all can profit from it.

johannschopplich:
Thanks for the note! I took a look at the base shell. Indeed, some optimizations can actually be transferred.

Since the shell for EasyRPG has been amended anyway, I hope that the pull request is nevertheless possible.

Ghabry (commented):
Looks good to me.
Just one thing: Because the new svg is MIT licensed: Can you add in the ``3rd party software`` section of the Player repository the Teenyicons?

johannschopplich:
Good catch. Have done so. 🙂

Ghabry:
Jenkins: test this please

Ghabry (approved):

fmatthew5876 (approved):

PR #2426, opened 2020-12-15
Add touch input support for web shell

johannschopplich:
This pull requests adds buttons to the player shell for touch input capable devices. The following keyboard inputs are bound:
- Arrow up
- Arrow right
- Arrow down
- Arrow left
- Enter
- Escape

Thus, RM2k games are now playable on e.g. mobile devices. For RM2k3 games it get's trickier, since Shift and Numpad keys are possibly used as well. The web player is unusable on touch-based devices at the moment, so in my opinion this is a step up even if some keys are still missing. Do you think further buttons shall be added? Speaking for myself, I'd like to keep the layout as clean as possible. 

You may see and test it in action here: https://realtroll.de/player/?game=el-dorado

el-dorado-easy-rpg

**Further features**:
- Buttons are highlighted when touched.
- Operating the d-pad while holding down also works. Inspired by @pulsejet's [lines](https://github.com/pulsejet/mkxp-web/blob/262a2254b684567311c9f0e135ee29f6e8c3613e/extra/js/dpad.js#L45) from his [mkxp-web](https://github.com/pulsejet/mkxp-web/) project. Thank you!

**Notes**: 
- Implementation is future-proof: `key` and `code` keyboard event properties are used along deprecated `keyCode` (like `13` for enter) and `which`. Later ones are still required for emscripten to interpret the keyboard events.
- SVG for the d-pad is hand-crafted, therefore no license to add to the readme.
- Media query `@media all and …` has been reduced to `@media …`, since type `all` is implied by defining a media query in the first place.

### DIFF
--- a/resources/emscripten/emscripten-pre.js
+++ b/resources/emscripten/emscripten-pre.js
@@ -1,45 +1,100 @@
-Module.EASYRPG_GAME = ""
+// Note: The `Module` context is already initialized as an
+// empty object by emscripten even before the pre script
+Module = {
+  EASYRPG_GAME: "",
 
-function parseargs () {
-    var tmp = [];
-    var result = [];
-    var items = window.location.search.substr(1).split("&");
+  preRun: [],
+  postRun: [],
 
-    // Store saves in subdirectory `Save`
-    result.push("--save-path");
-    result.push("Save");
+  print: (...args) => {
+    console.log(...args);
+  },
 
-    for (var index = 0; index < items.length; index++) {
-        tmp = items[index].split("=");
+  printErr: (...args) => {
+    console.error(...args);
+  },
 
-        if (tmp[0] === "project-path" || tmp[0] === "save-path") {
-            // Filter arguments that are set by us
-            continue;
-        }
+  canvas: (() => {
+    const canvas = document.getElementById('canvas');
 
-        // Filesystem is not ready when processing arguments, store path to game
-        if (tmp[0] === "game" && tmp.length > 1) {
-            Module.EASYRPG_GAME = tmp[1].toLowerCase();
-        }
+    // As a default initial behavior, pop up an alert when webgl context is lost
+    // See http://www.khronos.org/registry/webgl/specs/latest/1.0/#5.15.2
+    canvas.addEventListener('webglcontextlost', event => {
+      alert('WebGL context lost. You will need to reload the page.');
+      event.preventDefault();
+    }, false);
 
-        result.push("--" + tmp[0]);
+    return canvas;
+  })(),
 
-        if (tmp.length > 1) {
-            var arg = decodeURI(tmp[1]);
-            // Split except if it's a string
-            if (arg.length > 0) {
-                if (arg.slice(0) === '"' && arg.slice(-1) === '"') {
-                    result.push(arg.slice(1, -1));
-                } else {
-                    var spl = arg.split(" ");
-                    result = result.concat(spl);
-                }
-            }
-        }
+  setStatus: text => {
+    if (!Module.setStatus.last) Module.setStatus.last = {
+      time: Date.now(),
+      text: ''
+    };
+
+    if (text !== Module.setStatus.text) {
+      document.getElementById('status').innerHTML = text;
+    }
+  },
+
+  totalDependencies: 0,
+
+  monitorRunDependencies: left => {
+    Module.totalDependencies = Math.max(Module.totalDependencies, left);
+    Module.setStatus(left ? `Preparing... (${Module.totalDependencies - left}/${Module.totalDependencies})` : 'Downloading game data...');
+  }
+};
+
+/**
+ * Parses the current location query to setup a specific game
+ */
+function parseArgs () {
+  const items = window.location.search.substr(1).split("&");
+  let result = [];
+
+  // Store saves in subdirectory `Save`
+  result.push("--save-path");
+  result.push("Save");
+
+  for (let i = 0; i < items.length; i++) {
+    const tmp = items[i].split("=");
+
+    if (tmp[0] === "project-path" || tmp[0] === "save-path") {
+      // Filter arguments that are set by us
+      continue;
     }
 
-    return result;
+    // Filesystem is not ready when processing arguments, store path to game
+    if (tmp[0] === "game" && tmp.length > 1) {
+      Module.EASYRPG_GAME = tmp[1].toLowerCase();
+    }
+
+    result.push("--" + tmp[0]);
+
+    if (tmp.length > 1) {
+      const arg = decodeURI(tmp[1]);
+      // Split except if it's a string
+      if (arg.length > 0) {
+        if (arg.startsWith('"') && arg.endsWith('"')) {
+          result.push(arg.slice(1, -1));
+        } else {
+          result = [...result, ...arg.split(" ")];
+        }
+      }
+    }
+  }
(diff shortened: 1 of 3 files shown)


Ghabry:
Jenkins: test this please <3

This sounds awesome

johannschopplich:
Is it possible to place files along the shell HTML and copy them over after building? I'd like to outsource styles into `index.css` and the touch-input script block into `controls.js` or something like that. The shell file gets confusing since its line number has grown... 😄

Ghabry:
For your JavaScript you can modify emscripten-pre.js. this one is embedded.

For installing the CSS you must modify the CMakeLists at various places. Likely not something you want to figure out. But when you commit the CSS file I can write a patch for you.

Ghabry:
PR build is ready:

https://easyrpg.org/play/pr2426

https://easyrpg.org/play/pr2426/?game=uid

https://easyrpg.org/play/pr2426/?game=ib

johannschopplich:
Wonderful news, good to know. No rush, let's put that into another pull request and leave this one as a pure feature request. 

I've just added one last commit to make buttons unselectable. In some cases when holding down for a while they got selected, what we want to avoid of course.  

PR build looks good! Nothing to add. 🙂

Ghabry:
What I noticed: Nothing major but for further enhancements:

Compared to the Android app the buttons are flipped. We have confirm left and cancel right. Though no idea why. This was based on the Gameboy but the Gameboy has confirm/A on the right, so we messed up here 😅.

In full screen mode the touch buttons disappear.

-------

There is no button editor like we have on Android. Games that needs Shift or the numpad (yume Nikki) are unplayable. Also no way to move buttons

fdelapena:
Great feature, thank you!

It seems there's a regression when the Player runs under an iframe (arrows not working after getting focus. This is good to keep it working because is used on some game hosting websites (itch.io, etc.).

Tests:

https://easyrpg.org/play/iframetest.html?path=master/?game=default
https://easyrpg.org/play/iframetest.html?path=pr2426/?game=default

fdelapena:
jenkins: ok to test

johannschopplich:
@Ghabry I intuitively inherited the Gameboy layout A/B button layouts as well. Didn't know it is switched on the Android ports (never tested). 😄 Should I chance them to match the Android player?

We could add a Shift button specifically for RM2k3 games. The engine gets logged anyway, so we could parse that before writing to the console. 
For the Numpad, I guess we'll find a solution on the future. 

@fdelapena Good catch. Will look into it.

Ghabry:
> I intuitively inherited the Gameboy layout A/B button layouts as well. Didn't know it is switched on the Android ports (never tested). smile Should I chance them to match the Android player?

Would be good. I guess most users are already used to the Android layout. At least I am. Pressed lots of times the wrong buttons in the web ui xD

> We could add a Shift button specifically for RM2k3 games. The engine gets logged anyway, so we could parse that before writing to the console.

That's not a bad idea. But also not generic enough, I think we will post-bone this. Now that the basic ui is there it will be possible to implement a settings ui based on it.

resources/emscripten/emscripten-pre.js
```diff
@@ -1,45 +1,100 @@
-Module.EASYRPG_GAME = ""
+// Note: The `Module` context is already initialized as an
+// empty object by emscripten even before the pre script
+Module = {
```
Ghabry, line 3: doesn't this assignment remove the ``Module.EASYRPG_FS`` from the pre-script?

Ghabry:
Jenkins: test this please

Ghabry:
@johannschopplich  I think you broke the touch input. I see the buttons but nothing happens

johannschopplich:
Might be. I'll fix it. Having trouble compiling emscripten from the source.

johannschopplich:
I've used the chance to slightly refactor the shell: All `Module`-related code has been moved from the HTML file to the pre-script. The script handling button simulation is inside the index.html. Resulting in better SOC. 

Also the CSS content has been expanded for increased readability. 

Will build locally and tackle the remaining points.

johannschopplich:
@fdelapena Arrow keys are working again inside iframe. Fullscreen button is visible again as well.

@Ghabry Touch input fixed. Escape and Enter reversed.

It's pretty easy to register new buttons with the new setup. Apart from some additional CSS, you just create a new element and set a `data-key`and `data-key-code`:

```html
<div id="apad-shift" data-key="Shift" data-key-code="16"></div>
```

The key and key code may be retrieved via tools like [keycode.info](https://keycode.info).

Ghabry (commented):
Looks now good to me. I will ask some testers if they all there devices work. Mobile is always a bit... special

Ghabry:
Got one bug report: The touch controls are not shown on Android Samsung Galaxy S9+.

Somehow all bugs are Samsung related. They patch/mess around in Android and randomly break stuff. -_-

johannschopplich:
@Ghabry Thanks, I've enhanced the device checks. I guess you're speaking of your Samsung Galaxy. Please test again. 

Touch controls will now show only on screens with primary touch support, while they are hidden if the device supports mouse or touchpad primarily.

Ghabry:
Nah is not my device. I don't buy expensive samsung hardware :) Waiting for the response.

Maybe a nice enhancement: Somebody else suggested: Just listen to a touch event and when one is received just show the buttons. Would work around this problem (and useful for devices that are not primary touch devices)

johannschopplich:
All right. Thanks.

Haha, I know. `('ontouchstart' in window)` works, but is ugly in comparison to a media query like`(hover: none) and (pointer: coarse)')`. The later one is [broadly supported](https://caniuse.com/mdn-css_at-rules_media_hover), therefore I'm wondering why it wouldn't work.
Let's try to make it clean. If it won't work and I won't find a solution, well, then we gotta use the ugly one.

Ghabry (approved):
Confirmed working on Samsung so gets my approval. thanks

johannschopplich:
@Ghabry Regarding buttons disappearing in fullscreen mode: Making that working is a whole new topic. Solely the canvas element get's expanded. Will have to find a way to overlay the buttons on top of the canvas in fullscreen mode. Mabye a wrapper around the canvas. Dunno. Leaving that for another pull request. 😉

rygos:
Yesterday I noticed another problem with iPhones, which was not the case at first.
When I press the control pad or the buttons, the player is highlighted. As if I were marking text. When I press the fullscreen button, the blue flashes only briefly. But the fullscreen button doesn't work for me anyway. But this is another story.

johannschopplich:
Text selection should be inhibited through additional `user-select: none` rules. I've [added them yesterday](https://github.com/EasyRPG/Player/pull/2426/commits/0a7985faa7a54822a2de481300bc4dda5ea4fc29). They work as intended on iOS for me. Could you please refresh the window some times, so we ensure the browsers uses the latest pr build? I guess it's cached and thus text selection still visible. Please retest with the latest build. If text selection still occurs, I'll look into it.

The fullscreen button is still selectable, good catch. Will fix that asap.

Yeah, fullscreen mode is unsupported in the Safari Mobile. It keeps on getting the new Internet Explorer... A pain for all web devs.

Ghabry:
@rygos can you retest with the latest version?

rygos:
Sorry for my late reply.

The Text Selection bug is fixed for me :)

fdelapena (approved):

PR #2428, opened 2020-12-17
Change random movement behavior

### LINKED ISSUE #2424: Regression: Random movement sometimes moves more than 1 tile
Create an event with ``Movement Type: Random`` and ``Frequency: 3``.

Run the game, sometimes the event will move more than one tile which is incorrect, they always only move 1 tile.

Thanks to Jan/Deep8 for reporting.

rueter37:
This PR fixes #2424. The cause where an event with movement behavior set the random moves more than one tile at a move frequency below 8 is because the random stop count multiplier sometimes returns a straight zero which means that the event will not wait before the next move. Changing the divisor from 5 (int) to 5.0 (double) stops the random stop count multiplier returning zero, which fixes the issue mentioned above. Move frequency 8 (uninterrupted movement) is not affected by #2424 and this PR because in this case the stop count is always zero regardless of the stop count multiplier.

### DIFF
--- a/src/game_event.cpp
+++ b/src/game_event.cpp
@@ -465,7 +465,8 @@ void Game_Event::UpdateNextMovementAction() {
 
 void Game_Event::SetMaxStopCountForRandom() {
 	auto st = GetMaxStopCountForStep(GetMoveFrequency());
-	st *= (Rand::GetRandomNumber(0, 3) + 3) / 5;
+	st *= Rand::GetRandomNumber(0, 3) + 3;
+	st /= 5;
 	SetMaxStopCount(st);
 }
 


src/game_event.cpp
```diff
@@ -465,7 +465,7 @@ void Game_Event::UpdateNextMovementAction() {
 
 void Game_Event::SetMaxStopCountForRandom() {
 	auto st = GetMaxStopCountForStep(GetMoveFrequency());
-	st *= (Rand::GetRandomNumber(0, 3) + 3) / 5;
+	st *= (Rand::GetRandomNumber(0, 3) + 3) / 5.0;
```
Ghabry, line 468: Can you test something for me?

Change this to:

```cpp
st *= Rand::GetRandomNumber(0, 3) + 3;
st /= 5;
```

Ghabry: Guess the precedence is just wrong because of the "*=". Everything is divided by 5, not just the random number.

rueter37: The change works for me. Should I update this PR this way?

Ghabry: Yes please as we try to be close to rpg_rt here

Ghabry (approved):

src/game_event.cpp
rueter37: Updated this PR.

fdelapena (approved):

fdelapena:
jenkins: test this please

PR #2430, opened 2020-12-23
Fix EXP table on actors with changed classes on savegame loading

rueter37:
This PR fixes a regression in relation to savegame loading: If a RPG Maker 2003 savegame gets loaded then any actors with changed classes used the default EXP table instead of the EXP table of the current class. Adding the `MakeExpList()` function fixes this.

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -106,6 +106,7 @@ void Game_Actor::SetSaveData(lcf::rpg::SaveActor save) {
 		data.super_guard = dbActor->super_guard;
 	}
 
+	MakeExpList();
 	Fixup();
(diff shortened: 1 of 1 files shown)


Ghabry:
@rueter37 
Thanks for finding this!

Could you move this before ``Fixup()`` in ``SetSaveData(lcf::rpg::SaveActor save)`` instead?
Otherwise the ExpList is calculated twice on startup because the constructor also calls Fixup.

rueter37:
Did the requested change.

fmatthew5876 (approved):

fdelapena (approved):

PR #2438, opened 2021-01-11
Fix lcf::Data linker errors for Make platforms

fmatthew5876:
Since this is going away soon, I opted to just move the file instead of patching all the makefiles and then needing to undo the patch later.


Ghabry:
oh, so this wasn't a dirty workspace? wups .

Ghabry (approved):

Issue #2436, opened 2021-01-10
Bug "Random Movement" with Frequency "nX"


Br4ssman:
**Name of the game**: Super Crystal Hunter (2k3 steam) - Minigame "Grenade Sheep"

**Player platform**: Windows x64

**Attach files (as a .zip archive or link them)**
[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/5792400/easyrpg_log.txt)
https://user-images.githubusercontent.com/1331889/104124145-56e06380-534f-11eb-9a96-49674fb43e58.mp4

Normal behavior: https://youtu.be/a-puNA6NO1Y

Config of the wolves:
![imagen](https://user-images.githubusercontent.com/18174790/104120141-c9901580-5334-11eb-9e4c-00647114a8d3.png)

**Describe the issue in detail and how to reproduce it**:

There is nothing special about the mini-game, simply when the wolves fall (spritesheet) you place the events of each wolf with the settings of the previous screenshot.

When the wolves die, another spritesheet is processed. But what the random movement of the wolves is is no mystery.

For some strange reason the wolves get stuck on the leftmost tiles and only move up and down

Thanks guys.

Ghabry:
Github allows embedding of videos now, did that for your video.

Br4ssman:
Thanks @Ghabry I didn't knew that ^^'
btw, I want to note that the wolves have his graphic as "Direction Fix/Inanimated", because that can be related with the bug, maybe.

rueter37:
@Br4ssman you are right with your assumption that the graphic animation setting set to "Direction Fix/Inanimated" is the culprit. I can reproduce the bug this way. As soon the face direction is locked, the random movement behavior for some reason doesn't allow moving towards the direction the event is facing. This explains why the wolves are never moving to the right, because they are facing right and have the face direction locked. Moreover this bug occurs independent of the move frequency.

PR #2447, opened 2021-01-30
Fix random movement behavior on locked face direction

rueter37:
The random movement behavior uses the current move direction instead of the current face direction now if the face direction is locked. This fixes the bug which disallows the event moving towards the direction he is facing if the face direction is locked and the
movement type is set random movement.

Fixes: #2436

### DIFF
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -497,15 +497,15 @@ bool Game_Character::Move(int dir) {
 }
 
 void Game_Character::Turn90DegreeLeft() {
-	SetDirection(GetDirection90DegreeLeft(GetFacing()));
+	SetDirection(GetDirection90DegreeLeft(GetDirection()));
 }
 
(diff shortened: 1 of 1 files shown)


fdelapena (approved):

fmatthew5876 (changes requested):
This is actually a larger bug. The Turn90 and Turn180 functions should *always* use direction and not facing. This affects random movement and also the turning move route commands.

Please remove the conditional and change it always use direction. Don't forget to run the unit tests after.

DynRPG Address
-----

| Addr | Name |
| -- | -- |
| 004C45C8 | Turn90DegLeft |
| 004C45F8 | Turn90DegRight |
| 004C4628 | Turn180Deg |

rueter37:
Did the requested change. The unit tests were successful.

fmatthew5876 (approved):

Ghabry:
Thanks!

Issue #2337, opened 2020-09-14
Title image disappears too early and system graphic changes on load


Ghabry:
1st bug:

Select "New Game". The title screen is black before the fadeout starts. In 0.6.2.2 this doesn't happen.

2nd bug:
Is a regression #2111 
Load a savegame that has an altered system graphic. The system graphic is already changed in the load menu before the fadeout happens. In 0.6.2.2 this doesn't happen.

rueter37:
The second bug mentioned here has been resolved by #2456.

PR #2451, opened 2021-02-06
Savegame window system graphic fix

rueter37:
The system graphic is no longer changed in the load menu before the fadeout happens. This fixes the second bug mentioned in #2337.


fmatthew5876:
Since it's related, lets find a better more general solution to this problem in #2452 and add this fix there as well using it.

PR #2452, opened 2021-02-07
Disallow system updates on closing windows

rueter37:
System changes are applied now after the window has been closed. This fixes a cosmetic bug in "Vampires Dawn" when you interact with the graves in the graveyard in "Asgars Schloss", the message window uses a different system graphic. But as soon as the window starts closing, the system graphic switches back to the default system graphic which looks a bit awkward. This PR fixes this by applying the change after the window has been closed. You can reproduce the bug and confirm the fix with this savegame: [Save01.zip](https://github.com/EasyRPG/Player/files/5939146/Save01.zip).

### DIFF
--- a/src/window_selectable.cpp
+++ b/src/window_selectable.cpp
@@ -123,8 +123,13 @@ void Window_Selectable::UpdateArrows() {
 }
 
 // Update
-void Window_Selectable::Update() {
-	Window_Base::Update();
+void Window_Selectable::Update(bool block_system_change) {
+	if (block_system_change) {
+		Window::Update();
+		Window_Base::UpdateMovement();
+	} else {
+		Window_Base::Update();
+	}
 	if (active && item_max > 0 && index >= 0) {
 		if (Input::IsRepeated(Input::DOWN) || Input::IsTriggered(Input::SCROLL_DOWN)) {
 			if (index < item_max - column_max || column_max == 1) {
--- a/src/window_message.cpp
+++ b/src/window_message.cpp
@@ -381,7 +381,7 @@ void Window_Message::Update() {
 
 	const bool was_closing = IsClosing();
 
-	Window_Selectable::Update();
+	Window_Selectable::Update(IsClosing());
 	number_input_window->Update();
 	gold_window->Update();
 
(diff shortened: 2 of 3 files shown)


src/window_selectable.cpp
```diff
@@ -123,8 +123,13 @@ void Window_Selectable::UpdateArrows() {
 }
 
 // Update
-void Window_Selectable::Update() {
-	Window_Base::Update();
+void Window_Selectable::Update(bool block_system_change) {
+	if (block_system_change) {
+		Window::Update();
+		Window_Base::UpdateMovement();
```
fmatthew5876, line 129: This is a gross hack, especially in #2451 where these 2 lines just appear and is totally not obvious why.

You should either:
1. Refactor the update routines so that all child classes must explicitly call an `UpdateSystemGraphic()` in their update routines.
2. Add some member variable flag or virtual function which says whether or not to update the system graphic and adjust it / call it as needed.

Basically lets make "should we update the system graphic?" a first class concept of the `Window` classes

fmatthew5876 (changes requested):
See inline comment

Ghabry:
> Refactor the update routines so that all child classes must explicitly call an UpdateSystemGraphic() in their update routines.

Imo just removing ``UpdateSystemGraphic`` would work. The only case where a live update is even possible is on map and battle.

The only way to update the system graphic is via ``CommandChangeSystemGraphics``.
Like e.g. with ```ChipsetUpdated`` there could be a ```SystemGraphicUpdated`` callback invoked on map and battle scene which invokes ``SetWindowskin(Cache::SystemOrBlack());`` on all windows.

rueter37:
I'm sorry to disappoint but I'm giving up here. This is way above my skill level (Honestly I don't know why I have tried that at all in the first place).

Ghabry:
Please don't take this personal. We appreciate all your contributions!

Even if they don't meet our "standards" they are usually a great base to work on (or give a hint where the problem is).

rueter37:
Thank you for the positive response despite my endless negativity, I really appreciate it! Of course I will continue contributing to the EasyRPG project, because as a "Vampires Dawn" fan (and RPG Maker 2k/2k3 games in general - for nostalgia reasons) I have a genuine personal interest that EasyRPG reaches the 1.0 release someday. As an amateur in programming I try to stick to the issues which are easier to fix, but I have underestimated this one.

PR #2453, opened 2021-02-07
Equip and shop window: Fix display errors on edge cases

rueter37:
This PR handles some edge cases in the equip value window and the shop window: In the equip value window, if you have equipped items which makes stats go above 999 (or below 1), then the new stat value is displayed wrong if you change equipment which affects stats which are "out of bounds". The shop window comparison indicators are affected as well. This one is hard to explain so I come up with an example in "Vampires Dawn II" with this savegame: [Save01.zip](https://github.com/EasyRPG/Player/files/5939178/Save01.zip). Take a look at Valnar's equipment: The defense stat sums up to 1090 with the equipment and the game clamps it to 999, which is correct. But if you want to equip the "Spiegelring", which gives a penalty of 30 to all stats, the preview shows a defense stat of 969. The expected value is 999 here because the stat will drop to 1060 which clamps to 999. In the shop window you have the following issue: Talk to the shopkeeper and buy things. Move the cursor to the armors or helmets and you will notice that the comparison indicator says that Valnar's stats will go down despite the defense stat remaining at 999. This PR intends to fix the mentioned issues.

### DIFF
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -489,6 +489,22 @@ static int AdjustParam(int base, int mod, int maxval, Span<const int16_t> states
 	return value;
 }
 
+int Game_Battler::CalcValueAfterAtkStates(int value) const {
+	return AdjustParam(value, 0, MaxStatBattleValue(), GetInflictedStates(), &lcf::rpg::State::affect_attack);
+}
+
+int Game_Battler::CalcValueAfterDefStates(int value) const {
+	return AdjustParam(value, 0, MaxStatBattleValue(), GetInflictedStates(), &lcf::rpg::State::affect_defense);
+}
+
+int Game_Battler::CalcValueAfterSpiStates(int value) const {
+	return AdjustParam(value, 0, MaxStatBattleValue(), GetInflictedStates(), &lcf::rpg::State::affect_spirit);
+}
+
+int Game_Battler::CalcValueAfterAgiStates(int value) const {
+	return AdjustParam(value, 0, MaxStatBattleValue(), GetInflictedStates(), &lcf::rpg::State::affect_agility);
+}
+
 int Game_Battler::GetAtk(Weapon weapon) const {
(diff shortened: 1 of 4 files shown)


Ghabry:
This PR here also went unreviewed for too long. Will do it ASAP.
I can already say that this cannot break games because it only alters the Equip and the Shop Scene, no interpreter related code 👍. So is low-risk change.

Ghabry (approved):

src/game_battler.cpp
```diff
@@ -489,6 +489,27 @@ static int AdjustParam(int base, int mod, int maxval, Span<const int16_t> states
 	return value;
 }
 
+int Game_Battler::CalcValueAfterStates(int value, bool lcf::rpg::State::*adj) const {
```
fmatthew5876, line 492: Does this logic need to be fully copy pasted like this? It's already existing above in `AdjustParams`.

Can you make this function just call that one?

Do be aware of the min/max clamping behavior

src/scene_equip.cpp
```diff
@@ -126,6 +131,11 @@ void Scene_Equip::UpdateStatusWindow() {
 		spi = Utils::Clamp(spi, 1, 999);
 		agi = Utils::Clamp(agi, 1, 999);
 
+		atk = actor.CalcValueAfterStates(atk, &lcf::rpg::State::affect_attack);
+		def = actor.CalcValueAfterStates(def, &lcf::rpg::State::affect_defense);
+		spi = actor.CalcValueAfterStates(spi, &lcf::rpg::State::affect_spirit);
+		agi = actor.CalcValueAfterStates(agi, &lcf::rpg::State::affect_agility);
```
fmatthew5876, line 137: We should add wrapper functions instead of requiring the caller to pass a pointer to member.

rueter37:
Please don't re-review this one now, I messed up the commits.

rueter37:
Fixed the commits and addressed the review comments now. This can be reviewed again now.

fdelapena (approved):

fmatthew5876 (approved):

Issue #2306, opened 2020-08-26
Optimize filefinder directory tree data structure


fmatthew5876:
This program, when run on hh3's project tree

```
int main(int argc, char** argv) {
    auto tree = FileFinder::CreateDirectoryTree(argv[1]);
}
```

Runtime is:
```
real    0m0.664s
user    0m0.542s
sys     0m0.120s
```

And memory usage, I'll let this picture speak for itself:

![ffmem](https://user-images.githubusercontent.com/1198466/91367123-a6c0f280-e7d3-11ea-8e2c-af5f7cdb750d.png)

Ghabry:
I started working on this.
Will split the DirectoryTree and the RTP handling in two files.
Then I can make the upcoming Filesystem class more slim because these two complecated problems (directory enumeration & caching and RTP file rewriting) are somewhere else (also better testable)

Ghabry:
Current WIP: https://github.com/Ghabry/easyrpg-player/tree/directorytree

Instead of ``std::unordered_map`` I use ``FlatUniqueMultiMap``. Also the enumeration is lazy: An folder is cached the first time it is accessed.

Still has some FIXME and lacks case-insensitive folder support -> RTP does not work.

PR #2455, opened 2021-02-09
Implement VFS and ZIP reading support (FS [3/4])

### LINKED ISSUE #1085: Packed games support
This feature would allow to read game data inside from zip files and other formats. This feature may be useful for systems with slow file copy, such Android via MTP.

Supporting several game installer or self extracting executable types, with the purpose of direct download and not requiring a Windows machine may be useful, too.

This would vastly improve the user experience for the Android and other non-Windows ports.

### LINKED ISSUE #2491: Unterwegs in Düsterburg: not loading (Unicode)
This very famous game doesn't open in Retroarch/EasyRPG Core and I get a red error screen :/

Name of the Game: Unterwegs in Düsterburg
Platform: Windows 10 64 bit, Retroarch 1.9, EasyRPG 0.6.2.3
Description: Red Screen with a simple "Error: EasyRPG will close now"

It works when I use the standalone EasyRPG player tough, just not with Retroarch.

However I prefer Retroarch because of the shaders . Would be nice if that function would be implemented into the EasyRPG-Player.

Ghabry:
Fix #1085 
Fix #2491

TODO/FIXME:

- [x] Platform Test: Emscripten
- [x] Platform Test: Linux
- [x] Tree has lots of duplications in the cache (path with \\, with /, with trailing /)
- [x] Improve the ZIP FS, the code is very old, still from ChrisBreiti from ~5 years ago
- [x] Creating Sub-Filesystems in the GameBrowser leaks memory (FS never deleted)
- [x] Unit tests fail with autotools (but work with CMake o_O)
- [x] Tests
- [x] Documentation
- [x] Refactor RTP
- [x] Crash when the game is directly in the root of a ZIP archive
- [x] ZIP parser does not notice intermediate folders
- ~~Restarting lacks some cleanup (e.g. unload core in RetroArch and loading again segfaults)~~ Cannot reproduce. RetroArch bug?

--------

Major change:
Before accessing a file you must decide if it is a game file (``FileFinder::Game()``), a save file (``FileFinder::Save()``) or something outside (``FileFinder::Root()``). These return a subview into the corresponding folders. Adding something like ``FileFinder::Config()`` to get the global config dir would make sense here.

--------

Loading games from ZIP files:
The Filesystem ```Create`` function will attempt to access the path directly and when this fails will split it in subcomponents and attempt to construct a VFS path.
Chaining is supported, this means a ZIP can contain a ZIP (though this becomes slow when you chain too much).

To boot a game from a zip simply specify the path as usual:

```
--project-path some_path/games.zip/TestGame-2000
```

---------

Saving inside zip files

ZIPs are obviously not writable (at least not in a sane way).
When the FS does not support the ``Write`` Feature (and ``--save-path`` is not set) then the FileFinder will construct a savepath next to the non-writable file and create the folders, e.g.:

- Game path is ``some_path/games.zip/TestGame-2000``
- Save path will be ``some_path/games.zip.save/TestGame-2000``

- Game path is ``some_path/tons_of_games.zip/TestGames.zip/TestGame-2000``
- Save path will be ``some_path/tons_of_games.zip.save/TestGames.zip/TestGame-2000``

---------

New classes:

- ``Filesystem`` - Base class of all filesystems`
- ``FilesystemView`` - Holds a reference to a filesystem and forwards all calls to it (this pattern makes storing references to file systems simpler because the child class does not matter)
- ``NativeFilesystem`` - Forwards all calls to the API of the operating system (used for the Root filesystem)
- ``ZipFilesystem - Provides the Filesystem Interface for ZIP files. The encoding is auto-detected.

-----------

I decided again to keep ``FileFinder::FindDefault`` and friends, at least until the gigantic PRs such as battle are in. Want to prevent rebase chaos.

-----------

As you can see I use ``DirectoryTree`` as part of the filesystem but the ``DirectoryTreeView`` I just invented is already deleted again. I used the knowledge gained here to invent ``FilesystemView`` and search replaced it everywhere. (The API is actually almost the same as ``DirectoryTreeView`` but it is layered over a VFS)


Ghabry:
Jenkins: Test this please

Ghabry:
I mark this now as ready for review.

Also have a solution for the "game inside APK" feature: Namespaces provided by a "root filesystem" that handles this. Idea is that the startup code will set the project-path to ``apk://game_dir`` and it will "just work". Though not implemented yet, this is just the base for this.

There are still some problems I have to solve but otherwise it is feature complete.

Open problems:

- [ ] Emscripten
- [ ] Unit tests fail with autotools (but work with CMake o_O)
- [ ] Creating Sub-Filesystems in the GameBrowser leaks memory (FS never deleted)

Ghabry:
I found the emscripten problem:
The Game VFS was not initialized when emscripten checked whether a index.json exists.

-----

Oh and btw there are already some namespaces available for testing (use as ``--project-path``)

- ``root://`` This shows a list of namespaces supported by the Root FS (mostly for debugging)
- ``file://some_path`` This uses a path as a game directory without any fancy processing (no zip support, uses the path as is)

Ghabry:
I noticed now a huge design issue with how the RTP is handled. This worked for DirectoryTree but is now broken for VFS:

FindImage etc. use a FilesystemView that is rooted at the Game-Directory. When an asset is not found the DirectoryTree checks the various RTP locations and returns a path into it. This worked before because this were just absolute paths in the real filesystem.
But now with the VFS this fails because the VFS at the Game()-Directory has no idea how to handle a global path.
Have to move the RTP lookup to a different location and return InputStreams directly... More work to do :/

Ghabry:
Done. In the end this was actually more complicated than expected. Everything Image and Audio related operates now on streams. This is needed because paths do not work here as the RTP is not reachable from the Game-VFS.
Everything that is not interfered by RTP can use normal Game-VFS paths.
But I think I got it now and no gigantic refactors are needed anymore.

Sorry for mostly breaking SDL_mixer. It is hard to get this right as the ownership of the streams when passing back and forth in SDL is a mess because of all our workarounds in there. So I decided to keep this broken and get #2302 ready soon. Then Mixer can be deleted.

Planned for [4/4]: Fixing Scene_Import and #2306

rueter37:
The changes work for me. But the `<memory>` header must be added to `directory_tree.h` and `filesystem.h` to make GCC 9.3 stop complaining.

Ghabry:
Thanks for testing. Added the missing headers. Guess this is also a PR that requires tests instead of reviews :/.

fdelapena (approved):

Tondorian:
I tested it on my switch. 
For some reason it is not able to create content to a log file, however, it creates an empty one and lists all my games zipped and unzipped. If I try to start it a red error "screen" appears:
"Error:
Error loading RPG_RT.ldb

EasyRPG player will close now.
Press [ENTER] key to exit ..."
this error happens on zipped and folder games.

Ghabry:
okay this is unfortunate. Are you somehow able to reproduce this on the PC? Otherwise this will hard to figure out for me. No Switch :(

Tondorian:
[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/6313038/easyrpg_log.txt)
here is the logfile on pc zip files are created by 7zip.
on switch it also creates the save folder for zipped games

Ghabry:
ah yeah  that is another bug reported by somebody else:
The game must not be in the root of the ZIP archive but in a subfolder of the ZIP. Otherwise the Player crashes. I have to fix this.

Also only the compression algorithm "Deflate" is supported. When you set it in 7zip to PPMD or Deflate64 it will not work. (but this should show a different warning)

Ghabry:
Fixed the "game in root" crash. This was a huge bug in the DirectoryTree code.

Also new quality of life feature: When a VFS contains in the root only one folder that folder is entered directly. This ensures that booting of zip archives feels more natural. They usually have a structure of "archive.zip/game_folder".

@Fishaphrodite  I cannot reproduce the cleanup problem with the core. Can you give me a step-by-step guide?

@Tondorian I requested a new switch build. Please try when build ready.

Ghabry:
I know this is a huge PR but I would be interested in reviews now :)

Except for some ZIP parser bugs everything seems to be "bugfree" by now...

Ghabry:
okay intermediate folders are now created. This fixes the final issue here.

rueter37:
I noticed some oddities in relation to the `--project-path` command line switch: Using absolute paths work fine unless the path has got a trailing slash. In that case the player exits with a red screen saying `Error loading RPG_RT.ldb`. Moreover if you use relative paths and the path begins with a dot (to start off from the current directory) the red screen appears as well with the same message. And more weird things happen if the path begins with two dots (to start off from the parent directory): Then the game browser opens listing the directories of the game instead of starting the game.

Ghabry:
The problem you mention is hard to solve. First I need system specific code to get the absolute path. Otherwise this ".." makes no sense when making the path canonical :/. So I would prefer to postpone this. 

Though the "." case should work.

Ghabry:
no fixes, just rebased the emscripten code

Ghabry:
This somehow completely breaks emscripten again... No files are found o_O.

Something funny: WHen you provide a game that does not exist the game browser will open and show the folder "Save" because the game browser learned how to browse xD. Not intentional ^^'

Ghabry:
The "." cases are fixed. ".." will not work, please don't use it (yet) ;).

Also finally found the reason why Frozen Triggers didn't work in the web. A bug in my cache flush code :/

rueter37:
Perfect, works for me!

Ghabry:
Still waiting for feedback here :)

rueter37:
Sadly I'm not able to review that huge amount of changes. But as I stated almost two weeks ago, the changes work for me (and still do). :+1:

fdelapena (approved):

Ghabry:
Thanks I will fix the nits and give a _7 days_ grace period for further reviews before merge. I requested reviews 30 days ago, so this is imo long enough by now.

fmatthew5876 (approved):

fdelapena:
Maybe there's a good chance to propose a last minute custom zip extension for games willing to hide the game data from accidental spoilers opening "compressed folders" on Windows. Something like `.ezrpg` or `.erdata` or whatever.

Ghabry:
Well one of my ideas for the editor is some kind of "release format", something like APK (this is just a renamed ZIP file with a metadata text file).

So ".easyrpg" including the game with Metadata. I can already reserve ".easyrpg" as an extension for ZIP files if you want. Doesn't hurt.

Ghabry:
Added now ".easyrpg". Is simply forwarded to the zip-handler

PR #2456, opened 2021-02-09
Disallow system updates on closing windows (retry)

rueter37:
This PR is a retry of #2452. I had overreacted and prematurely closed the old PR. After getting a clear mind I decided to try it again. This contains the fix intended by #2451 too.

### DIFF
--- a/src/window.h
+++ b/src/window.h
@@ -83,6 +83,8 @@ public:
 	bool IsOpeningOrClosing() const;
 
 protected:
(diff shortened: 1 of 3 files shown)


fmatthew5876 (approved):
I think Ghabry proposed solution of removing the update logic entirely and testing it is likely best solution, assuming it works.

Regardless, this is clearly better than the current code and more correct. So this PR gets my approval.

fdelapena (approved):

PR #2457, opened 2021-02-11
change the button input layout to match 3ds' better layout

kane159:


carstene1ns:
jenkins: ok to test.

src/platform/psvita/psp2_ui.cpp
```diff
@@ -209,7 +209,7 @@ void Psp2Ui::ProcessEvents() {
 	sceCtrlPeekBufferPositive(0, &input, 1);
 	keys[Input::Keys::Z] = (input.buttons & SCE_CTRL_CROSS);
 	keys[Input::Keys::X] = (input.buttons & SCE_CTRL_CIRCLE);
-	keys[Input::Keys::N2] = (input.buttons & SCE_CTRL_TRIANGLE);
+	keys[Input::Keys::F] = (input.buttons & SCE_CTRL_TRIANGLE);
```
carstene1ns, line 212: The decision was taken to support yume nikki here.
The 3DS has the advantage to support the "2" key through touch screen. The vita port might need some love.

carstene1ns:
I think for users it is better to have the fast forward than the fps counter, so we should change <kbd>F2</kbd>-><kbd>F</kbd> in line 220 instead.
https://github.com/EasyRPG/Player/blob/77fee9ca8f4bbadf0f3e72d2f0739e7f796a68e1/src/platform/psvita/psp2_ui.cpp#L220

Ghabry:
best way is probably to port the on-screen numpad we have on the Nintendo Switch over to PsVita. Then there are lots of free keys.
Currently the touch screen is only used by the Shader selector.

Will not work on PS TV though.

carstene1ns:
Okay, for PSTV I think some kind of circle menu by button combo like we talked about makes sense, will see if i come up with something.

carstene1ns (approved):

Ghabry:
with your other PR the mapping of the right analog stick becomes unnecessary. Replace this with movement?

carstene1ns:
We already have movement with left stick and dpad, I have intentionally left it unused for the mentioned "special button stuff"

carstene1ns:
@kane159: Are you around to test a build with touchscreen input?

PR #2459, opened 2021-02-17
RNG refactoring

DNKpp:
As already discussed on discord, the current workflow of locking and unlocking rng seems quite dirty. My solution is wrapping the rng in a simple polymorphic adapter class, which let's us exchange the behavior during runtime (e.g. for tests).
I also added a simple ScopedRNGWrapper class, which will exchange the current global rng with a newly created one and exchanges it back on destruction.

Finally I added some test cases.

### DIFF
--- a/src/algo.cpp
+++ b/src/algo.cpp
@@ -155,11 +155,11 @@ int CalcCriticalHitChance(const Game_Battler& source, const Game_Battler& target
 	return crit_chance;
 }
 
-int VarianceAdjustEffect(int base, int var) {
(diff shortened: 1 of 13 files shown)


Ghabry:
Jenkins: Test this please

DNKpp:
Well, I've mixed up the c++ versions. Most _v type_trait prefixes are from c++17. Questions is, when do you plan upgrading version? If it's planned for the near future I'll simply wait and instead upgrading the other parts to c++17 .

carstene1ns:
Since we do use c++17 now, this should be looked at again...

PR #2460, opened 2021-02-18
Filesystem [2.5/3]: Fix regressions

Ghabry:
@rueter37 can you confirm?

- RTP lookup: Reordered the logic, fixes RTP assets not found when game lacks folder
- Saving: The save scene constructed a path that lacked the root path. This is fixed by accident in #2455 because the stream opening is now handled by the VFS, not by FileFinder.

### DIFF
--- a/src/directory_tree.cpp
+++ b/src/directory_tree.cpp
@@ -190,25 +190,25 @@ std::string DirectoryTree::FindFile(const DirectoryTree::Args& args) const {
 
 	auto* entries = ListDirectory(dir);
 	if (entries) {
-        std::string dir_key = make_key(dir);
-        auto dir_it = dir_cache.find(dir_key);
-        assert(dir_it != dir_cache.end());
+		std::string dir_key = make_key(dir);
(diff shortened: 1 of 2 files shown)


rueter37:
The RTP lookup is working now, but savegame files are no longer being created.

rueter37:
Changing [line 264 of directory_tree.cpp](https://github.com/Ghabry/easyrpg-player/blob/directorytree/src/directory_tree.cpp#L264) to `return tree->GetRootPath();` fixes the savegame file issue for me. But I have a bad feeling that this will probably break something else...

Ghabry:
@rueter37  sorry for the slow response time. This is all fixed now.

For FS3 I have to recheck if all the bugs I just fixed are also in FilesystemView....

Ghabry:
When this is in will also submit a new APK to Android. That one was a blocker :/

rueter37:
Perfect, can confirm that the issues are fixed now! :+1:

fdelapena (approved):

Issue #2450, opened 2021-02-05
Vampires Dawn II: Using the item "Speicherstein" triggers endless loop


rueter37:
Using the item "Speicherstein" (item which allows to save the game once) in "Vampires Dawn II" triggers an endless loop which means the save file window is always opening regardless if you save the game or exit the save file window. This happens on current master, but not on stable version 0.6.2.3 so we have a regression here.

rueter37:
One observation I made regarding this bug: As soon the save file scene is called the remaining commands of the event are not executed including unsetting the switch which triggers the event so this is the reason for the endless loop.

Ghabry:
Because you know programming and this is a regression: Could you try to bisect this for us?

Start it with
```
git bisect start
git bisect good 0.6.2
git bisect bad HEAD
```
Then compile it and test the build.
If it is broken:
```
git bisect bad
```
If it works:
```
git bisect good
```
Repeat until the bad commit is found and tell us the commit (takes log(n) steps because it is a binary search [always takes the commit in the middle of the remaining]).

Problem is that you must have the right liblcf version for each revision because of incompatible changes. I recommend that you clone liblcf to ``lib/liblcf``, use CMake with ``-DPLAYER_BUILD_LIBLCF=ON`` and then reset liblcf to a revision that is close to the Player date.

rueter37:
After typing in `git bisect bad` I got the following output: `binäre Suche: danach noch 50 Commits zum Testen übrig (ungefähr 6 Schritte) [643b83b5038a51a1aebca92534a67ad743d007d0] Merge pull request #2301 from Ghabry/inputrecv2`.

I don't know if this is of any help, but the last `git bisect good` outputted the following: `binäre Suche: danach noch 101 Commits zum Testen übrig (ungefähr 7 Schritte) [7ec99d632c0a3399d01794c48faaac71c3049198] Ineluki: Documentation`

Ghabry:
Yeah this is kinda how it works.
The ``ungefähr X Schritte`` tells you how often you must retest and bisect good/bad until it reports something like:
``Der erste schlechte Commit ist``

rueter37:
The search has been narrowed down to 4 commits, here is the output:
`There are only 'skip'ped commits left to test.`
`The first bad commit could be any of:`
`e38d5951af54eeff43e15d6911674e5ce8dfe1c4`
`6fe4dfb4ed913ceb992a1b74a3d1044f40fbf166`
[...]

PR #2461, opened 2021-02-19
Fix save count incrementing on game save

rueter37:
The PR aims to fix the save count incrementing on game save. For some reason the [save count increment in liblcf](https://github.com/EasyRPG/liblcf/blob/master/src/lsd_reader.cpp#L37) is no longer working some commits after the 0.6.2 release.

Fixes: #2450

### DIFF
--- a/src/game_system.h
+++ b/src/game_system.h
@@ -276,6 +276,7 @@ public:
 	void SetAllowMenu(bool allow);
 
 	int GetSaveCount();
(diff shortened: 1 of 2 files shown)


fmatthew5876 (approved):

Ghabry (approved):
This likely broke because Game_* are not referencing LCF data directly anymore but are copies of it.

PR #2465, opened 2021-03-02
Fix translation reading on Windows

Ghabry:
+ReadLine API improvements

### DIFF
--- a/src/game_system.cpp
+++ b/src/game_system.cpp
@@ -513,7 +513,11 @@ void Game_System::OnBgmReady(FileRequestResult* result) {
 		}
 
 		// The first line contains the path to the actual audio file to play
-		std::string line = Utils::ReadLine(stream);
+		std::string line;
+		if (!Utils::ReadLine(stream, line)) {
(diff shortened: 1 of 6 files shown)


Ghabry:
@fdelapena  please give the one having the error the Player.exe for testing, thanks

fdelapena (approved):
Confirmed fixed, thanks!

src/game_system.cpp
```diff
@@ -513,7 +513,11 @@ void Game_System::OnBgmReady(FileRequestResult* result) {
 		}
 
 		// The first line contains the path to the actual audio file to play
-		std::string line = Utils::ReadLine(stream);
+		std::string line;
+		if (!Utils::ReadLine(stream, line)) {
+			Output::Warning("Ineluki MP3: Link file is empty: {}", path);
```
fdelapena, line 518: About this one, are games with broken links closing in RPG_RT.exe or just silently fail?

Ghabry: No idea. Never tested. This is only a new warning for empty link files. The one for file not found a bit below is there since years and nobody ever complained

PR #2468, opened 2021-03-04
Fix WildMidi becoming stuck

Ghabry:
Was a bug in the pitch calculation.

Note that the looping is not seemless at the end, I don't consider this a bug, is just really hard to get the timing right because the midi messages and Wildmidi are independent of each other (sample based as you know...). So this is a gigantic hack that kinda works.

For gapless playback use FluidLite or FmMidi.


fdelapena (approved):

carstene1ns (approved):
Looks good to me.

PR #2471, opened 2021-03-07
Filesystem [2.75/4]: Add sanity checks after OpenInput/OutputStream

Ghabry:
OpenInput/OutputStream are invalid when the stream cannot be opened.
This results in segfaults when they are passed by reference.

Most important is a regression fix: When RPG_RT.ini was missing the Player crashed :(

For all others: Good to have (most of them are unlikely to be hit because other code checks first if a file was found so they will only fail for stuff like "No permission" or "Out of handles")


src/input_source.cpp
```diff
@@ -145,7 +150,7 @@ bool Input::Source::InitRecording(const std::string& record_to_path) {
 		record_log = std::make_unique<Filesystem_Stream::OutputStream>(FileFinder::OpenOutputStream(path, std::ios::out | std::ios::trunc));
 
 		if (!record_log) {
-			Output::Warning("Failed to open file {} for input recording : {}", path, strerror(errno));
+			Output::Error("Failed to open file {} for input recording : {}", path, strerror(errno));
```
Ghabry, line 153: promoted this to an error because this is an incorrect command line argument

fdelapena (approved):

fmatthew5876 (approved):

PR #2475, opened 2021-03-11
Migrate Android build to gradle+CMake

### LINKED ISSUE #621: ndk-build is stupid and tries to compile any .cpp file in src
Maybe ndk-build should depend on a list like all other sane build systems and stop compiling my non-clean upd conflict files player.BASE.cpp because they end on .cpp -.-

Ghabry:
To build set env var EASYRPG_TOOLCHAIN_DIRS to the paths that contain all the Android toolchains.
Then run "gradle assembleRelease" and the magic happens.

Fix #621


fdelapena (approved):

Ghabry:
Dont merge this yet.
I want to invent a way to pass custom CMake flags to the gradle CMake invocation.
It is completely unclear to me who at Google thought it is a good idea to push a feature that has literally zero official customization points (also the documentation is really lacking). They had at least 3 years to fix this, they just dont care.

Ghabry:
CMake options are now possible with -PcmakeOptions="-DOPTION1=ON OPTION2=OFF" or env var ORG_GRADLE_PROJECT_cmakeOptions="-DOPTION1=ON OPTION2=OFF" (the env var is a built-in gradle feature)

fdelapena (approved):

builds/android/app/src/gamebrowser/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+cmake_minimum_required(VERSION 3.7)
```
carstene1ns, line 1: Was just thinking we could add this to the main CMakeLists as well.
However, it is currently really convoluted, so maybe we should split other things out like this.

Ghabry: Imo it would make sense to move all the system dependent stuff in CmakeLists.txt in the build/ folder. Currently the main file is quite messy.

When devkitXXX and VitaSDK are ever built through CMake we could hide all the custom stuff (build_vpk, make_dol, etc.) in subfolders.

carstene1ns: > When devkitXXX and VitaSDK are ever built through CMake we could hide all the custom stuff (build_vpk, make_dol, etc.) in subfolders.

of course!

builds/android/gradle.properties
```diff
@@ -1,3 +1,8 @@
 android.enableJetifier=true
 android.useAndroidX=true
-android.useDeprecatedNdk=true
+
+VERSION_NAME=1.0
```
carstene1ns, line 4: We should start to set these with each release from now on.
This way f-droid will have it easier at least.

carstene1ns (approved):
It works 👍🏼

carstene1ns:
jenkins: test this please.

PR #2478, opened 2021-03-14
Improve MIDI playback accuracy

Ghabry:
by processing messages in smaller chunks. This reduces the resolution from 4-8ms to 1ms.

[dk.zip](https://github.com/EasyRPG/Player/files/6137688/dk.zip)

- dkbad.ogg - Upstream
- dkgood.ogg - This patch
- dkorig.ogg - RPG_RT (different soundfont)

### DIFF
--- a/src/decoder_midigeneric.cpp
+++ b/src/decoder_midigeneric.cpp
@@ -25,6 +25,10 @@
 
 constexpr int GenericMidiDecoder::midi_default_tempo;
 
+// 1 ms of MIDI message resolution for a 44100 Hz samplerate
+constexpr int samples_per_play = 512;
+constexpr int bytes_per_sample = sizeof(int16_t) * 2;
+
 GenericMidiDecoder::GenericMidiDecoder(MidiDecoder* mididec)
 	: mididec(mididec) {
 	assert(mididec);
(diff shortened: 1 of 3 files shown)


fdelapena (approved):

fdelapena:
Though a previous hotfix for speed broke midi tempo in master (not just for wildmidi), https://easyrpg.org/play/pr2478/?game=smashbros note timings play properly now :+1:

Ghabry:
The playback speed does not change anymore?
Okay will check this

Ghabry:
My last Pitch patch had a bug: The calculation was wrong, must be 100/pitch, not pitch/100.

I also added a clamping for invalid BGM/SE values. I read in the Yume2kki Discord that a song in the latest version uses a pitch of 20. This is adjusted to 100 because it is out-of-range. Now it becomes 50. This is at least closer. Lifting the range would need testing first, so this fix was faster.

fdelapena:
The pitch/tempo works again, thanks!
About clamping values: [RPG_RT.exe supports modified pitch values below 50% / above 150%](https://www.youtube.com/watch?v=oxT0rsHAVjo&t=78s). It's unlikely somebody modified this by hand, but based on experience with some weird games tweaking some values... well, we can wait until some bug report appears :+1:.

src/decoder_midigeneric.cpp
```diff
@@ -25,6 +25,10 @@
 
 constexpr int GenericMidiDecoder::midi_default_tempo;
 
+// 1 ms of MIDI message resolution for a 44100 Hz samplerate
+constexpr int samples_per_play = 512;
+constexpr int bytes_per_sample = sizeof(int_least16_t) * 2;
```
fmatthew5876, line 30: Why `int_least16_t`? This could be larger than 16 bit.

Why not `int16_t`?

Ghabry: Because fmmidi used it, was a copy-paste from there :sweat_smile: . But you are right, this makes no sense.

fmatthew5876 (approved):

src/decoder_midigeneric.cpp
Ghabry: fixed

PR #2481, opened 2021-03-15
Fixes for NovemberJoy's remaining battle issues

rueter37:
This PR fixes the death animation loop by playing death animations only once (RPG_RT does this as well) and normal enemy attacks do not longer wait for the battle animation (you will see the animation and the damage number at the same time like in RPG_RT). This affects RPG Maker 2003 battles.

Fixes: #1681
Fixes: #2335

### DIFF
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -2214,10 +2214,15 @@ Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleAction
 			auto* enemy = static_cast<Game_Enemy*>(target);
 			enemy->SetBlinkTimer();
 		} else if (action->GetAffectedHp() < 0) {
-			target_sprite->SetAnimationState(Sprite_Actor::AnimationState_Damage, Sprite_Actor::LoopState_DefaultAnimationAfterFinish);
+			if (!target->IsDead()) {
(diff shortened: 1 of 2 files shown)


fdelapena (approved):

src/scene_battle_rpg2k3.cpp
```diff
@@ -2214,10 +2214,15 @@ Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleAction
...
 			enemy->SetBlinkTimer();
 		} else if (action->GetAffectedHp() < 0) {
-			target_sprite->SetAnimationState(Sprite_Actor::AnimationState_Damage, Sprite_Actor::LoopState_DefaultAnimationAfterFinish);
+			if (!target->IsDead()) {
+				target_sprite->SetAnimationState(Sprite_Actor::AnimationState_Damage, Sprite_Actor::LoopState_DefaultAnimationAfterFinish);
+			}
 		}
 	}
 
+	if (action->IsSuccess() && target->GetType() == Game_Battler::Type_Ally && target->IsDead()) {
```
fmatthew5876, line 2223: A better condition is `!was_dead && IsDead()`

Ghabry (approved):
The code looks fine and I trust rueter that they did lots of testing to ensure this works as expected.

I'm far less knowledgeable in how the battle system should behave.

So when fmatthews concern is addressed can be merged.

src/scene_battle_rpg2k3.cpp

fmatthew5876 (approved):

PR #2482, opened 2021-03-15
Battle 2k3: Battle animation sprites changes

rueter37:
This PR intends to do the following fixes:

- Placement of battlers: The battlers were placed wrong (too much distance between each other and partially and full off-screen placing starting with the third party member) if battle animation sprites were used. BattleCharset sprites are working fine.
- Displaying state afflictions: While BattleCharset sprites are showing its state afflictions at the beginning of the battle, the battle animation sprite battlers did so not until their first turn.
- Battle animation sprites are flipped now if the flip flag is set.
- BattleAnimation height: Battle animations are displayed on the right height if the target is a battle animation sprite.

### DIFF
--- a/src/battle_animation.cpp
+++ b/src/battle_animation.cpp
@@ -293,8 +293,12 @@ void BattleAnimationBattle::Draw(Bitmap& dst) {
 	for (auto* battler: battlers) {
 		const Sprite_Battler* sprite = battler->GetBattleSprite();
 		int offset = 0;
-		if (sprite && sprite->GetBitmap()) {
-			offset = CalculateOffset(animation.position, sprite->GetHeight());
+		if (sprite) {
+			if (sprite->GetBitmap()) {
+				offset = CalculateOffset(animation.position, sprite->GetHeight());
+			} else {
+				offset = CalculateOffset(animation.position, GetAnimationCellHeight() / 2);
+			}
 		}
 		DrawAt(dst, battler->GetBattlePosition().x, battler->GetBattlePosition().y + offset);
 	}
@@ -306,6 +310,37 @@ void BattleAnimationBattle::FlashTargets(int r, int g, int b, int p) {
 }
 
 void BattleAnimationBattle::ShakeTargets(int str, int spd, int time) {
+	for (auto& battler: battlers) {
+		battler->ShakeOnce(str, spd, time);
+	}
+}
+
+BattleAnimationBattler::BattleAnimationBattler(const lcf::rpg::Animation& anim, std::vector<Game_Battler*> battlers, bool only_sound, int cutoff_frame, bool set_invert) :
+	BattleAnimation(anim, only_sound, cutoff_frame), battlers(std::move(battlers))
+{
+	invert = set_invert;
+}
+
+void BattleAnimationBattler::Draw(Bitmap& dst) {
+	if (IsOnlySound())
+		return;
+	if (animation.scope == lcf::rpg::Animation::Scope_screen) {
+		DrawAt(dst, SCREEN_TARGET_WIDTH / 2, SCREEN_TARGET_HEIGHT / 3);
+		return;
+	}
+
+	for (auto* battler: battlers) {
+		DrawAt(dst, battler->GetBattlePosition().x, battler->GetBattlePosition().y);
+	}
+}
+
+void BattleAnimationBattler::FlashTargets(int r, int g, int b, int p) {
+	for (auto& battler: battlers) {
+		battler->Flash(r, g, b, p, 0);
+	}
+}
+
+void BattleAnimationBattler::ShakeTargets(int str, int spd, int time) {
 	for (auto& battler: battlers) {
 		battler->ShakeOnce(str, spd, time);
 	}
--- a/src/battle_animation.cpp
+++ b/src/battle_animation.cpp
@@ -326,3 +326,6 @@ void BattleAnimation::SetFrame(int frame) {
 	this->frame = frame;
 }
 
+void BattleAnimation::SetInvert(bool inverted) {
+	invert = inverted;
+}
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -342,9 +342,10 @@ void Scene_Battle_Rpg2k3::ResetAllBattlerZ() {
 	}
 
 	for (auto* actor: Main_Data::game_party->GetActors()) {
-		auto* sprite = actor->GetBattleSprite();
+		auto* sprite = actor->GetActorBattleSprite();
 		if (sprite) {
 			sprite->ResetZ();
+			sprite->DetectStateChange();
 		}
 	}
 }
--- a/src/sprite_actor.cpp
+++ b/src/sprite_actor.cpp
@@ -59,6 +59,7 @@ void Sprite_Actor::Update() {
 		if (Player::IsRPG2k3()) {
 			if (animation) {
 				// Is a battle animation
+				animation->SetInvert(battler->IsDirectionFlipped());
 				animation->Update();
 
 				if (animation->IsDone()) {
@@ -127,9 +128,6 @@ void Sprite_Actor::Update() {
 
 	const bool flip = battler->IsDirectionFlipped();
 	SetFlipX(flip);
-	if (animation) {
-		SetFlipX(flip);
-	}
 }
 
 void Sprite_Actor::SetAnimationState(int state, LoopState loop) {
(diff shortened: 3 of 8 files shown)


fdelapena:
jenkins: test this please

fdelapena (approved):

src/sprite_actor.cpp
```diff
@@ -203,14 +203,14 @@ bool Sprite_Actor::IsIdling() {
...
-		return animation->GetWidth();
+		return animation->GetAnimationCellWidth() / 2;
 	}
 	return Sprite::GetWidth();
 }
 
 int Sprite_Actor::GetHeight() const {
 	if (animation) {
-		return animation->GetHeight();
+		return animation->GetAnimationCellHeight() / 2;
```
Ghabry, line 212: RPG Maker 2003 has two types of battle animations (small and big), did you check this?

rueter37: Thanks for the pointer, will make some tests tomorrow with the big sprites. The small ones should work fine.

rueter37: Thanks again for checking on this one, the big sprites were indeed broken. `GetAnimationCellWidth()` and `GetAnimationCellHeight()` always returned 96 (which is correct for normal sprites) but must return 128 on big sprites.

src/battle_animation.cpp
```diff
@@ -293,8 +293,12 @@ void BattleAnimationBattle::Draw(Bitmap& dst) {
 	for (auto* battler: battlers) {
 		const Sprite_Battler* sprite = battler->GetBattleSprite();
 		int offset = 0;
-		if (sprite && sprite->GetBitmap()) {
-			offset = CalculateOffset(animation.position, sprite->GetHeight());
+		if (sprite) {
+			if (sprite->GetBitmap()) {
+				offset = CalculateOffset(animation.position, sprite->GetHeight());
+			} else {
+				offset = CalculateOffset(animation.position, GetAnimationCellHeight() / 2);
```
Ghabry, line 300: just a marker for me to test this in emscripten

src/scene_battle_rpg2k3.cpp
```diff
@@ -342,9 +342,11 @@ void Scene_Battle_Rpg2k3::ResetAllBattlerZ() {
 	}
 
 	for (auto* actor: Main_Data::game_party->GetActors()) {
-		auto* sprite = actor->GetBattleSprite();
+		auto* sprite = actor->GetActorBattleSprite();
 		if (sprite) {
 			sprite->ResetZ();
+			sprite->DetectStateChange();
+			sprite->RefreshZ();
```
Ghabry, line 349: can you put the code of ``RefreshZ`` in ``ResetZ`` or is there a reason why this doesnt work?

src/battle_animation.cpp
```diff
@@ -311,6 +315,92 @@ void BattleAnimationBattle::ShakeTargets(int str, int spd, int time) {
 	}
 }
 
+BattleAnimationBattler::BattleAnimationBattler(const lcf::rpg::Animation& anim, std::vector<Game_Battler*> battlers, bool only_sound, int cutoff_frame, bool set_invert) :
+	BattleAnimation(anim, only_sound, cutoff_frame), battlers(std::move(battlers))
+{
+	invert = set_invert;
+}
```
Ghabry, line 322: If I understand this correctly a BattleAnimationBattler is this Animation2 stuff from RPG2k3 (as alternative to CBA), right?

src/battle_animation.cpp
```diff
@@ -311,6 +315,92 @@ void BattleAnimationBattle::ShakeTargets(int str, int spd, int time) {
...
+
+	for (auto* battler: battlers) {
+		DrawAt(dst, battler->GetDisplayX(), battler->GetDisplayY());
+	}
+}
+
+void BattleAnimationBattler::UpdateScreenFlash() {
+	int r, g, b, p;
+	UpdateFlashGeneric(screen_flash_timing, r, g, b, p);
+	if (r > 0 || g > 0 || b > 0 || p > 0) Main_Data::game_screen->FlashOnce(r, g, b, p, 0);
```
Ghabry, line 340: This rgba stuff was controversial the last time we had this PR but now it is isolated into Animation2 only so it cannot break other code.
Can you indent this properly? Add {} after the if as in FlashTargets below.

src/battle_animation.cpp
```diff
@@ -311,6 +315,92 @@ void BattleAnimationBattle::ShakeTargets(int str, int spd, int time) {
...
+			cell.tone_blue * 128 / 100,
+			cell.tone_gray * 128 / 100));
+		SetOpacity(255 * (100 - cell.transparency) / 100);
+		SetZoomX(cell.zoom / 100.0);
+		SetZoomY(cell.zoom / 100.0);
+		SetFlipX(invert);
+		for (auto& battler: battlers) {
+			SetFlashEffect(battler->GetFlashColor());
+		}
+		Sprite::Draw(dst);
```
Ghabry, line 394: Is this copy-paste here really necessary?
You are already overloading ``FlashTargets`` can you tell us what is not working here making this necessary? Then we can work on a fix.

Ghabry (changes requested):
-

src/battle_animation.cpp
rueter37: You are correct. The BattleAnimationBattler class is meant to handle battle animation sprite battlers.

src/scene_battle_rpg2k3.cpp
rueter37: Unfortunately the code cannot be put into `ResetZ`, because `ResetZ` is in `sprite_battler` while the code has to be run inside of `sprite_actor` because of the `animation` variable.

src/battle_animation.cpp
rueter37: The copy-pasted code is unfortunately needed for making the flash effects on battle animation sprites work.

src/battle_animation.cpp
rueter37: Fixed the indentation.

src/scene_battle_rpg2k3.cpp
Ghabry: @fmatthew5876 can ResetZ be made virtual or do you have any performance worries.

Ghabry:
When I test with a battle animation that has a shake RPG_RT ignores them for both Target and Screen but in Player the shake always happens.

rueter37:
I have removed the flash and shake changes now, because the were broken anyway. Fixing the flash and shake stuff will be handled in another PR.

src/battle_animation.cpp
rueter37: Removed the flash and shake stuff now ([see below](https://github.com/EasyRPG/Player/pull/2482#issuecomment-810398696)).

src/scene_battle_rpg2k3.cpp

fmatthew5876: `ResetZ()` is called by the constructors of `Sprite_Actor()` and `Sprite_Enemy()`. This means that if someone even subclassed them and tried to override it, the child class function would not get called as the child class hasn't been constructed yet.

If we're going to do this, lets make `ResetZ()` `final` in both of these classes to prevent overriding it.

rueter37: Added a commit now which disallows overriding the `ResetZ()` function in `sprite_actor` and `sprite_enemy`.

Ghabry: My bad forgot to check for this annoying limitation :/

rueter37:
@Ghabry Despite being outside of the scope of this PR, I have to add something about the flashes:
> When Set to "Screen" the Actor flashes (wtf)

This is already broken in master. You cannot reproduce it using a battle animation sprite, because flashes are not working there at all, but on enemies. Create a battle animation which uses screen flashes only and watch the enemies upon usage of the battle animation. You will see the enemies flash once in the flash color set in the battle animation settings despite the scope being set to "Screen".

rueter37:
@Ghabry I forgot to answer to [this comment](https://github.com/EasyRPG/Player/pull/2482#issuecomment-809773844) from you: The problems you mentioned are now handled by #2505. And BattleCharset sprites are affected by flash and shake effects which is already working (this is meant with "BattleCharset sprites fully work here").

src/battle_animation.cpp
Ghabry: well yeah this works. As this depends on the battle sprite the sprite is usually already loaded before anyway, so no emscripten worries here

Ghabry (approved):

Ghabry:
Just a reminder that, when there is no further feedback, this will be merged Sunday (6 days, 2 weeks after last review).

fmatthew5876 (approved):

PR #2483, opened 2021-03-15
RPG Maker 2003 battle scene changes

rueter37:
This PR does the following changes:

- The status window components have been rearranged to match the RPG_RT style.
- The transparent windows use an alpha of 160 instead of 128 now (tested with truecolor RPG_RT).
- In traditional style battles two bugs have been fixed: The SP window got printed over when a skill which costs SP has been used and the HP value didn't change its color if the HP reached critical or zero.
- In alternative style battles the gauges are now shown transparent if the window is opaque.
- The ally cursor is always drawn 40 pixels above the Y-Position of the battler.
- The enemy cursor is always drawn on the same height as the Y-Position of the enemy.
- Flipped actors use now the correct walking animation on escaping.
- The battle status window gets refreshed on applying state effects. This stops the HP value "lagging" behind on traditional battle mode.
- The enemy target window gets refreshed as soon as an enemy dies.

### DIFF
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -302,7 +302,7 @@ void Scene_Battle_Rpg2k3::CreateUi() {
 	}
 
 	if (lcf::Data::battlecommands.battle_type != lcf::rpg::BattleCommands::BattleType_traditional) {
-		int transp = lcf::Data::battlecommands.transparency == lcf::rpg::BattleCommands::Transparency_transparent ? 128 : 255;
+		int transp = lcf::Data::battlecommands.transparency == lcf::rpg::BattleCommands::Transparency_transparent ? 160 : 255;
 		options_window->SetBackOpacity(transp);
 		item_window->SetBackOpacity(transp);
 		skill_window->SetBackOpacity(transp);
@@ -506,7 +506,7 @@ void Scene_Battle_Rpg2k3::CreateBattleTargetWindow() {
 	target_window->SetZ(Priority_Window + 10);
 
 	if (lcf::Data::battlecommands.battle_type != lcf::rpg::BattleCommands::BattleType_traditional) {
-		int transp = lcf::Data::battlecommands.transparency == lcf::rpg::BattleCommands::Transparency_transparent ? 128 : 255;
+		int transp = lcf::Data::battlecommands.transparency == lcf::rpg::BattleCommands::Transparency_transparent ? 160 : 255;
 		target_window->SetBackOpacity(transp);
 	}
 }
@@ -594,7 +594,7 @@ void Scene_Battle_Rpg2k3::CreateBattleCommandWindow() {
 	command_window->SetZ(Priority_Window + 20);
 
 	if (lcf::Data::battlecommands.battle_type != lcf::rpg::BattleCommands::BattleType_traditional) {
-		int transp = lcf::Data::battlecommands.transparency == lcf::rpg::BattleCommands::Transparency_transparent ? 128 : 255;
+		int transp = lcf::Data::battlecommands.transparency == lcf::rpg::BattleCommands::Transparency_transparent ? 160 : 255;
 		command_window->SetBackOpacity(transp);
 	}
 }
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -397,7 +397,7 @@ void Scene_Battle_Rpg2k3::UpdateAnimations() {
 
 				ally_cursor->SetVisible(true);
 				ally_cursor->SetX(actor->GetBattlePosition().x);
-				ally_cursor->SetY(actor->GetBattlePosition().y - sprite->GetHeight() / 2);
+				ally_cursor->SetY(actor->GetBattlePosition().y - 40);
 
 				if (frame_counter % 30 == 0) {
 					SelectionFlash(actor);
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -515,6 +515,9 @@ void Scene_Battle_Rpg2k3::RefreshTargetWindow() {
 	// FIXME: Handle live refresh in traditional when the window is always visible
 	auto commands = GetEnemyTargetNames();
 	target_window->ReplaceCommands(std::move(commands));
+	if (!target_window->GetActive()) {
+		target_window->SetIndex(-1);
+	}
 }
 
 void Scene_Battle_Rpg2k3::CreateBattleStatusWindow() {
@@ -942,6 +945,7 @@ bool Scene_Battle_Rpg2k3::CheckBattleEndAndScheduleEvents(EventTriggerType tt) {
 #else
 	(void)page;
 #endif
+	RefreshTargetWindow();
 
 	return !interp.IsRunning();
 }
@@ -2237,6 +2241,7 @@ Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleAction
 		if (!was_dead && enemy->IsDead()) {
 			Main_Data::game_system->SePlay(Main_Data::game_system->GetSystemSE(Main_Data::game_system->SFX_EnemyKill));
 			enemy->SetDeathTimer();
+			RefreshTargetWindow();
 		}
 	}
 
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -302,7 +302,7 @@ void Scene_Battle_Rpg2k3::CreateUi() {
 	}
 
 	if (lcf::Data::battlecommands.battle_type != lcf::rpg::BattleCommands::BattleType_traditional) {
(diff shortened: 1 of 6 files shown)


fdelapena (approved):

src/window_battlestatus.cpp
```diff
@@ -185,7 +186,12 @@ void Window_BattleStatus::DrawGaugeSystem2(int x, int y, int cur_value, int max_
 		gauge_width = 25 * cur_value / max_value;
 	}
 
-	contents->StretchBlit(Rect(x, y, gauge_width, 16), *system2, Rect(48 + gauge_x, 32 + 16 * which, 16, 16), Opacity::Opaque());
+	if (gauge_width >= 13) {
+		contents->StretchBlit(Rect(x, y, 13, 16), *system2, Rect(48 + gauge_x, 32 + 16 * which, 8, 16), Opacity::Opaque());
+		contents->StretchBlit(Rect(x + 13, y, gauge_width - 13, 16), *system2, Rect(56 + gauge_x, 32 + 16 * which, 8, 16), Opacity::Opaque());
+	} else {
+		contents->StretchBlit(Rect(x, y, gauge_width, 16), *system2, Rect(48 + gauge_x, 32 + 16 * which, 8, 16), Opacity::Opaque());
```
Ghabry, line 201: > Battle 2k3: Change gauge display 

Can you explain what the magic number 13 means here?
Also for what is the pixman expert?

src/window_battlestatus.cpp
```diff
@@ -155,7 +159,9 @@ void Window_BattleStatus::RefreshGauge() {
 				int y = 2 + i * 16;
 
 				if (lcf::Data::battlecommands.battle_type == lcf::rpg::BattleCommands::BattleType_alternative) {
-					DrawGauge(*actor, 202 - 10, y - 2);
+					if (lcf::Data::battlecommands.transparency == lcf::rpg::BattleCommands::Transparency_opaque || (2 + index * 16 != y)) {
```
Ghabry, line 162: What is the purpose of the ``2 + index * 16 != y`` check?

src/scene_battle_rpg2k3.cpp
```diff
@@ -508,6 +508,7 @@ void Scene_Battle_Rpg2k3::RefreshTargetWindow() {
 	// FIXME: Handle live refresh in traditional when the window is always visible
 	auto commands = GetEnemyTargetNames();
 	target_window->ReplaceCommands(std::move(commands));
+	if (!target_window->GetActive()) target_window->SetIndex(-1);
```
Ghabry, line 511: Missing {} and indent.

Can your change cause any race condition? E.g. the target window is open and while open an enemy dies (if this is even possible).
Does this fix any known bug?

src/window_battlestatus.cpp
```diff
@@ -88,7 +88,7 @@ void Window_BattleStatus::Refresh() {
 			} else {
 				if (lcf::Data::battlecommands.battle_type == lcf::rpg::BattleCommands::BattleType_traditional) {
 					DrawActorState(*actor, 84, y);
-					contents->TextDraw(136 + 4 * 6, y, Font::ColorDefault, std::to_string(actor->GetHp()), Text::AlignRight);
+					contents->TextDraw(136 + 4 * 6, y, actor->GetHp() == 0 ? Font::ColorKnockout : actor->GetHp() <= actor->GetMaxHp() / 4 ? Font::ColorCritical : Font::ColorDefault, std::to_string(actor->GetHp()), Text::AlignRight);
```
Ghabry, line 91: I agree with c1 that this ternary is hard to read.
https://github.com/EasyRPG/Player/blob/master/src/window_base.cpp#L177-L181

carstene1ns: We have this in 3 places now at least. Maybe is time to make this a real function.

src/scene_battle_rpg2k3.cpp
```diff
@@ -301,7 +301,7 @@ void Scene_Battle_Rpg2k3::CreateUi() {
 	}
 
 	if (lcf::Data::battlecommands.battle_type != lcf::rpg::BattleCommands::BattleType_traditional) {
-		int transp = lcf::Data::battlecommands.transparency == lcf::rpg::BattleCommands::Transparency_transparent ? 128 : 255;
+		int transp = lcf::Data::battlecommands.transparency == lcf::rpg::BattleCommands::Transparency_transparent ? 160 : 255;
```
carstene1ns, line 304: btw. this is also referenced in multiple places, so maybe make this a function as well? `…IsTransparent()`
https://github.com/EasyRPG/Player/search?q=battlecommands.transparency

src/window_battlestatus.cpp
Ghabry: yeah.
Maybe add two new functions to window_base.h which simply draw the currect HP/MP value at cx/cy. There is already DrawActorHp but this one is not generic enough for this.
```
void DrawActorHpValue(const Game_Battler& actor, int cx, int cy) const;
void DrawActorSpValue(const Game_Battler& actor, int cx, int cy) const;
```

src/scene_battle_rpg2k3.cpp
Ghabry: good idea, as this is 2k3 only feature the  `IsTransparent`` func can be in the 2k3 battle scene directly.

src/window_battlestatus.cpp
rueter37: I admit this one is more of a hack. I take the left 8 pixels of the gauge and stretch them to 13 pixels, and the right 8 pixels of the gauge gets stretched to 12 pixels, which sums to 25 pixels like in RPG_RT. By putting the two halves together you get a more accurate output than directly stretch the entire gauge from 16 to 25 pixels. The so-called "pixman expert" is somebody we need to refactor the StretchBlit and similar functions to match RPG_RT behavior (see issue #446). As far as I know we are using pixman for the pictures, so I went with the term "pixman expert" for people who are experienced in using this library. (This entire stuff here is kinda hard to explain...)

src/window_battlestatus.cpp
rueter37: This check stops the gauge (if the gauge is opaque) being drawn if the menu cursor is on the same position.

src/window_battlestatus.cpp
Ghabry: Can I suggest something?
Could you try to replicate a RPG2k3 gauge using operations (stretch) in a normal image editor like Gimp or Paint.net?
When the exact operations are known this is easier to replicate in the Player instead of experimenting.

src/window_battlestatus.cpp
rueter37: Added the functions and replaced the ternary.

src/scene_battle_rpg2k3.cpp
rueter37: Fixed the missing {} and indent. This change fixes the cursor being shown even if the target window is not active. In RPG_RT the cursor is only shown in the target window if you are selecting a target. If the target window is open the active flag should always be set which means that the case you mentioned should never occur.

src/window_battlestatus.cpp
carstene1ns: The color logic is still duplicated though.

src/window_battlestatus.cpp
Ghabry: But with gimp you get the result you want?

rueter37: No.

rueter37: Alternatively we can remove the gauge stuff from this PR and wait until #446 is fixed (this should automatically fix the gauge display).

Ghabry: Okay. Please remove this commit then. As this is not a 100% fix I do not think that making the code harder to read is worth it here. Sorry

src/window_battlestatus.cpp
Ghabry: yes looks good

src/window_battlestatus.cpp
Ghabry: I thought about this now for 15 minutes but now I fully understand why they did this (typical 2k3 half assed job):

The gauge is drawn transparent but above the selection box. This means that the color of the gauge would be wrong because of the selection box as the draw order is "Background, Selection Box, Gauge, Text".

Instead of thinking about a solution they simply hid the bug. Genious move -_-.

Could you add a comment there:
```
RPG_RT Bug (?): Gauge hidden when selected due to transparency (wrong color)
```
Or something similiar.

Ghabry:
Have to check the following if the pixels match:
- Battle 2k3: Rearrange components in the battler status window (1/2) 
- Battle 2k3: Rearrange components in the battler status window (2/2)
But rest looks already good :+1:

src/window_battlestatus.cpp
rueter37: Removed the gauge commit now.

src/window_battlestatus.cpp
rueter37: Added.

Ghabry (approved):
Because this is pure visual I give a 7 day grace period for reviews.

fmatthew5876 (approved):

Issue #1708, opened 2019-03-17
DBZ Legend of Z: Blank pose/CBA for skill animation


OmegaMagnusDA:
Name of the game: **Dragon Ball Z: Legend of Z RPG**

Player platform: **Android**, **Windows**, and possibly others

-Download **DBZ LOZ (No music version)** for quickest download (third option) https://gamejolt.com/games/dragon-ball-z-legend-of-z-rpg/145494

-**Save(s)**: https://drive.google.com/open?id=1nnymRrmyPsXMfICy5NARPjVZQsACRH-o (roam around current map for battle. Use skills at bottom of skill lists to demonstrate)

-**easyrpg_log.txt log file**: https://drive.google.com/open?id=1O7Hr7P-puoby0-YWZQHLB1pKCR2AKyHJ

**Issue in detail**:
Skills that use blank frames from battle character sets to have character disappear during skill animation, do not execute correctly. (CBA = character battle animation)

Bug Example:
![image](https://user-images.githubusercontent.com/48193819/54484469-ecc70800-4824-11e9-97c7-0fa8e350d20b.png)

Use of Blank CBA (character battle animation) during skill:
![image](https://user-images.githubusercontent.com/48193819/54484436-1895be00-4824-11e9-8306-cd8cab07758a.png)

How it regularly executes (Regular RM2K3):
![image](https://user-images.githubusercontent.com/48193819/54484493-52b38f80-4825-11e9-8ea6-1385a9bd35cf.png)

fdelapena:
Thank you for the detailed report. This is because CBA is currently not implemented (it's the major 2k3 battle feature missing in Player). See #1269 for details and implementation progress (and consider subscribing to the issue for updates).

PR #2484, opened 2021-03-15
Partial CBA implementation

rueter37:
Depends on #2399

This PR adds an about 80% accurate (and about 50% complete) implementation of CBA in RPG Maker 2003 battles. It adds the CBA movement and the melee weapons. Not covered by this are the ranged weapons and the afterimage. Maybe this can be used as a base for a 100% accurate and complete CBA implementation.

PS: Enjoy Captain Novolin jumping on the enemies' heads with this one :smile:


Ghabry:
You probably forgot to add a new file to Makefile.am

CMakeLists.txt
```diff
@@ -311,6 +311,8 @@ add_library(${PROJECT_NAME} STATIC
 	src/sprite_character.h
 	src/sprite_picture.h
 	src/sprite_picture.cpp
+	src/sprite_weapon.cpp
+	src/sprite_weapon.h
```
fdelapena, line 315: Missing from Makefile.am, causing Jenkins build (uses Autotools) to fail.

rueter37: Fixed.

fdelapena:
jenkins: test this please

src/game_battlealgorithm.cpp
```diff
@@ -713,6 +717,25 @@ int Game_BattleAlgorithm::Normal::GetSourcePose() const {
 		: lcf::rpg::BattlerAnimation::Pose_AttackRight;
 }
 
+int Game_BattleAlgorithm::Normal::GetCBAMovement() const {
```
fmatthew5876, line 724: How does this work when dual wielding and the second weapon is used?

What if the the first weapon causes you to walk to the enemy and the second one doesn't? How about the opposite case?

src/scene_battle_rpg2k3.cpp
```diff
@@ -2083,6 +2089,65 @@ Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleAction
...
+	auto* source = action->GetSource();
+
+	if (source->GetType() == Game_Battler::Type_Ally) {
+		auto* target = action->GetTarget();
+
+		if (cba_move_frame == 0) {
+			auto* actor = static_cast<Game_Actor*>(source);
+			auto* sprite = actor->GetActorBattleSprite();
+			if (sprite) {
+				sprite->SetAnimationState(Sprite_Actor::AnimationState_WalkingLeft);
```
fmatthew5876, line 2116: Does this work correctly if the target is to the right of the attacker? (Back attack etc..)

src/scene_battle_rpg2k3.cpp
```diff
@@ -2083,6 +2089,65 @@ Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleAction
...
+					auto* enemy = static_cast<Game_Enemy*>(target);
+					auto* enemysprite = enemy->GetEnemyBattleSprite();
+					if (enemysprite) {
+						cba_end_pos = Point(target->GetBattlePosition().x + (source->IsDirectionFlipped() ? -(enemysprite->GetWidth() / 2) : enemysprite->GetWidth() / 2), target->GetBattlePosition().y);
+					}
+				}
+			}
+		}
+
+		if (cba_move_frame < 15) {
```
fmatthew5876, line 2130: 15 frames needs confirmation in RE tools. Leaving a note

src/scene_battle_rpg2k3.cpp
```diff
@@ -2083,6 +2089,65 @@ Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleAction
...
+	if (action->GetCBAMovement() != lcf::rpg::BattlerAnimationItemSkill::Movement_none) {
+		SetBattleActionState(BattleActionState_MoveForward);
+	} else {
+		SetBattleActionState(BattleActionState_StartAlgo2);
+	}
+	return BattleActionReturn::eContinue;
+}
+
+Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleActionMoveForward(Game_BattleAlgorithm::AlgorithmBase* action) {
+	auto* source = action->GetSource();
```
fmatthew5876, line 2107: To make this more robust. We can use battle action substates. Make an initialization substate which resets the `cba_move_frame` variable. Then immediately go into the `Moving` substate which animates the move.

Basically, nothing outside of this function should be touching the `cba_move_...` variables. The function itself should be completely self contained.

src/scene_battle_rpg2k3.cpp
```diff
@@ -2096,6 +2161,10 @@ Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleAction
 						Sprite_Actor::LoopState_WaitAfterFinish);
 			}
 		}
+
+		if (action->GetCBAMovement() != lcf::rpg::BattlerAnimationItemSkill::Movement_none) {
+			cba_move_frame = 0;
+		}
```
fmatthew5876, line 2197: See comment above about substates, and remove this

src/scene_battle_rpg2k3.cpp
```diff
@@ -2133,6 +2206,58 @@ Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleAction
...
+			return BattleActionReturn::eWait;
+		} else {
+			auto* actor = static_cast<Game_Actor*>(source);
+			auto* sprite = actor->GetActorBattleSprite();
+			if (sprite) {
+				sprite->DoIdleAnimation();
+			}
+			cba_move_frame = 0;
+		}
+	}
```
fmatthew5876, line 2302: This whole block looks like copy and paste. Maybe we can make an `UpdateMovement()` function which both state handlers call and add some flag for the direction. Or you could add a multiplier, which is normally + or - 2.

src/sprite_actor.cpp
```diff
@@ -267,7 +264,7 @@ void Sprite_Actor::OnBattlercharsetReady(FileRequestResult* result, int32_t batt
 
 void Sprite_Actor::Draw(Bitmap& dst) {
 	auto* battler = GetBattler();
-	if (battler->IsHidden()) {
+	if (battler->IsHidden() || do_not_draw) {
```
fmatthew5876, line 268: Is this fixing some other bug? It would be good to describe all the things implemented and bugs fixed in the commit message.

src/game_battlealgorithm.cpp
```diff
@@ -736,6 +740,25 @@ int Game_BattleAlgorithm::Normal::GetCBAMovement() const {
 	return lcf::rpg::BattlerAnimationItemSkill::Movement_none;
 }
 
+const lcf::rpg::BattlerAnimationItemSkill* Game_BattleAlgorithm::Normal::GetWeaponAnimationData() const {
+	const auto weapon = GetWeapon();
+	auto* source = GetSource();
+	if (source->GetType() == Game_Battler::Type_Ally) {
+		auto* ally = static_cast<Game_Actor*>(source);
+		auto weapons = ally->GetWeapons(weapon);
+		auto* item = weapons[0];
```
fmatthew5876, line 749: Again, need to support dual wield here

src/scene_battle_rpg2k3.cpp
rueter37: This one works because the sprites are mirrored on back attack. For setting the `cba_end_pos` variable there is even a check if the sprite is mirrored.

src/sprite_actor.cpp
rueter37: Yes, this one is the fix for #1708. The `do_not_draw` variable is set to true if the CBA battler name is empty. I'm going to update the commit message accordingly when the commit gets changed.

src/game_battlealgorithm.cpp
rueter37: I must admit here that what I "built" here doesn't exactly match RPG_RT (hence the accuracy of only about 80%). What is different here is that we have separate battle algorithms for every attack. This means that on every attack a move is started (depending on the weapon's settings), then the battle animation is played on the target and then the move back occurs. The effects of the attack are applied afterwards. Sadly I had trouble to get it exactly the way RPG_RT does. In RPG_RT it works this way: First the actor starts the move, then make one or more attacks on the targets where the effects are applied after the attack animation and then moves back. In the case of moving to the target the actor remains at the first target even if it gets killed and the actor attacks the next target. The problem I had with the RPG_RT way is when a victory condition was triggered, then the attacking actor didn't move back to his starting point and trying to work around the victory condition led to even more bad code in my case. The version I built has the advantage that the actor is always in the right position when attacking a target, but the disadvantage is it looks awkward when attacking the same target two or more times in a row (always moving back and forth).

src/game_battlealgorithm.cpp
rueter37: Please see my answer to your first comment. Sadly I need guidance to make this work the RPG_RT way.

src/scene_battle_rpg2k3.cpp
rueter37: I'm probably missing something, but do battle action substates even exist? I only see scene action substates.

src/game_battlealgorithm.cpp
fmatthew5876: My bad, I forgot how my own code works. I believe this should work fine. `weapons[0]` should be the primary or second weapon based on which weapon is currently active from `GetWeapon()`

src/scene_battle_rpg2k3.cpp
```diff
@@ -2156,9 +2162,33 @@ Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleAction
 			if (pose != lcf::rpg::BattlerAnimation::Pose_Idle) {
 				// FIXME: This gets cleaned up when CBA is implemented
 				auto action_state = static_cast<Sprite_Actor::AnimationState>(pose + 1);
-				sprite->SetAnimationState(
-						action_state,
-						Sprite_Actor::LoopState_WaitAfterFinish);
+
+				if (action->GetType() == Game_BattleAlgorithm::Type::Normal) {
+					sprite->SetNormalAttacking(true);
+					auto* weapon = actor->GetWeaponSprite();
```
fmatthew5876, line 2168: You'll need 2 weapon sprites for dual wield. You could use `Game_Battler::Weapon` here to select which one to return like all the other weapon related functions do.

fmatthew5876:
It might be easier to break this into 2 separate PRs?

I feel like the weapon sprites are much simpler to implement and perfect than the movement.

src/scene_battle_rpg2k3.cpp
fmatthew5876: Ahh you're right, battle action substates only exist in the 2k battle system.

In this case you could just create 2 battle action states. One which sets everything up and then immediately dispatches to the next which animates frame by frame.

src/game_battlealgorithm.cpp
fmatthew5876: I'm all for an 80% solution being better than the 0% we have now, however I do think having the actor go to and back for every hit will greatly slow the pace of battles. We should try to fix that.

I would be happy with fixing it correctly, or failing that have the actor only walk to the first target and stay there until the battle sequence is done.

Why can't you do the "move back" before `PostAction` and after the repeated algorithm executions. Only one battle algo is used, its' just replayed multiple times for each hit.

Basically here in `ProcessBattleActionApply` you would inject your MoveBack before PostAction

```
// If action does multiple attacks, repeat again.
if (action->RepeatNext(false)) {
    SetBattleActionState(BattleActionState_StartAlgo);
    return BattleActionReturn::eContinue;
}

SetBattleActionState(BattleActionState_PostAction);
return BattleActionReturn::eContinue;
```

The way I think this would work:

* Start - setup the algo
* MoveToTarget - move to the target from whatever your current position is, or don't move if you're already at the target. This will work seamlessly for multiple hits against the same target
* Animations, Reflect, execute, blah blah..
* Apply - apply effects. Check if we need to do another hit, if so, go back to Start state
* MoveBackToPosition
* PostAction ... Finished

rueter37:
The code has been updated according to the review comments. Movement and weapons behave now like in RPG_RT with about 95% accuracy. But I need help with one outstanding problem: How to make the actors move back to their starting point if a victory/defeat condition is triggered? The victory condition is triggered as soon the last enemy gets killed, so the attacking actor remains on its spot on victory. And defeat is possible if the player attacks and gets his entire party killed due to a reflected spell. Even in this case the actor moves back (despite being dead) to his starting point in RPG_RT.

src/game_battlealgorithm.cpp
rueter37: To answer your first comment: In case of dual wielding the movement information of the first weapon is used.

fmatthew5876: You mean that's how RPG_RT does it? Always the first weapon?

We can double check. Equip 2 weapons. First one moves, second one does not. First attack kills enemy. Does the second attack move to the second enemy or stay put?

If that's the case we should change our code to reflect that + a comment describing why.

fmatthew5876:
> The code has been updated according to the review comments. Movement and weapons behave now like in RPG_RT with about 95% accuracy. 

> But I need help with one outstanding problem: How to make the actors move back to their starting point if a victory/defeat condition is triggered? The victory condition is triggered as soon the last enemy gets killed, so the attacking actor remains on its spot on victory. 

At least this is still better than what we have now.

What you're saying is that the enemy dies and the battle state machine gets aborted, and so it has no time to animate the actor back into position.

So lets investigate exactly what the interaction is now because I don't remember. You can set a breakpoint on `SetState()` when the `state` is `Victory` and then see from where it gets called. Specificially I'd like to know the code path from the time the Apply hp effect kills the enemy until the time the state changes to victory/defeat. 

Once we know exactly how that is occuring, we can brainstorm a solution.



> Even in this case the actor moves back (despite being dead) to his starting point in RPG_RT.

Really? Does he move back while using the dead pose? :)

> And defeat is possible if the player attacks and gets his entire party killed due to a reflected spell. 

This could also be possible with the various times the event interpreter is called in between actions. I'd be curious to know how RPG_RT behaves with positioning.

rueter37:
Well, he first moves back (with 0 HP :smile:) and then goes into the dead pose.

src/game_battlealgorithm.cpp
rueter37: I tested this one. When the first enemy gets killed the actor remains at the place from the first enemy when attacking the second enemy. After that he returns to his starting point. So the information of the first weapon is used. The code already works.

rueter37:
`SetState(State_Victory)` can be called by the function `CheckBattleEndConditions()`. `CheckBattleEndConditions()` can be called by the functions `UpdateBattleState()`, `CheckBattleEndAndScheduleEvents(EventTriggerType tt)` and `ProcessSceneActionBattle()`.

fmatthew5876:
I see.

What we really need is some kind of unconditional "walk back" anytime the battle state machine is exited. Adding that and allowing it animate the walk frame by frame is going to be tricky. It's not clear to me the best way to do it given the current state of the code.

What you could do as a temporary hack is in `SetState()` reset any CBA movement. In other words, just snap the actor back to their original position. That's not complete but it's better than leaving them there and better than us not having any CBA movement at all. Then we can think about how to implement this unconditional walk back properly. The real solution will need some refactor. You'll also want to do the same position reset if the actor is killed.


I think this relates to Ghabry's comments about ATB. What we probably want here are 2 parallel state machines. The main battle scene state machine and the battle action state machine. With both of them updating every frame. That would allow battle actions to run while the user navigates the menus.

Once that's achieved, the battle state machine can always do a "walk back" when actions finish, because it's still running. In the case of non-ATB or transition to victory/defeat, the main state machine can be programmed to wait for the walk back to complete.

rueter37:
Added the temporary hack now which snaps the actor back on victory/defeat condition.

Ghabry:
Please start to use small commits, especially when we ask you for changes during reviews. Do not simply squash everything in one single commit. You can still do this at the end when we ask you for it.

This is just not helpful. Makes it hard for others later to bisect, hard to understand what you "thought" while implementing it, etc. 

Look at the many changes requested by matthew. For others (me) it is impossible now to figure out what you did to address them.
Also where was this "temporary hack" placed? This needs a FIXME comment in the codebase. Also it should be an extra commit.

Ghabry:
Also maybe add comments at places that look "wrong" but are actually correct, so the next one looking at it doesn't fix it on accident (there were some behaviour discussions in the review comments that apply for this)

src/sprite_actor.cpp
Ghabry: can you add a comment in the code?

src/scene_battle_rpg2k3.cpp
Ghabry: dito (comment)

src/scene_battle_rpg2k3.cpp
```diff
@@ -2293,15 +2438,38 @@ Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleAction
...
+			auto* actor = static_cast<Game_Actor*>(source);
+			auto* source_sprite = actor->GetActorBattleSprite();
+			if (source_sprite) {
+				source_sprite->SetNormalAttacking(false);
+				auto* weapon = actor->GetWeaponSprite();
+				int weapon_animation_id = 0;
+				if (weapon) {
+					auto* weapon_animation_data = action->GetWeaponAnimationData();
+					if (weapon_animation_data) {
+						if (weapon_animation_data->type == 0) {
```
Ghabry, line 2458: can you check if there is a enum for this 0 somewhere in liblcf?

src/sprite_weapon.cpp
```diff
@@ -0,0 +1,169 @@
...
+	int z = battler->GetType();
+	z *= SCREEN_TARGET_HEIGHT * 2;
+	z += y;
+	z *= id_limit;
+	z += id_limit - battle_index;
+	z += Priority_Battler;
+
+	default_z = z;
+
+	SetZ(z);
```
Ghabry, line 168: Is there dead code in the Sprite_Weapon class? This looks like a copy-paste of the old Sprite_Battler class.
There are comments and checks concerning enemies in here. Enemies cannot use CBA, so I'm confused.

src/sprite_weapon.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+	if (battler->GetBattleAnimationId() > 0) {
+		const lcf::rpg::BattlerAnimation* anim = lcf::ReaderUtil::GetElement(lcf::Data::battleranimations, battler->GetBattleAnimationId());
+		if (!anim) {
+			Output::Warning("Invalid battler animation ID {}", battler->GetBattleAnimationId());
+			return;
+		}
+
+		const auto* battler_animation_weapon = lcf::ReaderUtil::GetElement(anim->weapons, weapon_animation_id);
+		if (!battler_animation_weapon) {
```
Ghabry, line 97: Is no valid battler_anim_weapon a game bug? If yes add a Output::Warning

src/scene_battle_rpg2k3.cpp
rueter37: Added FIXME comment.

src/scene_battle_rpg2k3.cpp
rueter37: Fixed. A enum exists for this one.

src/sprite_weapon.cpp
rueter37: Fixed. The enemy code has been removed.

src/sprite_weapon.cpp
rueter37: Added warning message.

src/sprite_actor.cpp
rueter37: Added the comment now.

src/scene_battle_rpg2k3.cpp
```diff
@@ -671,6 +677,15 @@ void Scene_Battle_Rpg2k3::SetState(Scene_Battle::State new_state) {
 		auto_battle = true;
 	}
 
+	// FIXME: Remove this block when a proper solution has been found
+        // for moving the attacking actor back on victory/defeat condition
+        // Moreover remove all occurences of "cba_actor" in the code then
```
fdelapena, line 682: Spaces vs tabs.

PR #2485, opened 2021-03-15
Help window space width changes

rueter37:
Depends on #2399 because of the help window changes made there.

The target status window and end game window and the RPG Maker 2003 battle notifications use full width spaces instead of half width spaces in RPG_RT. This PR does these changes.

Fixes the bug reported of the end screen in #2264.

### DIFF
--- a/src/window_help.h
+++ b/src/window_help.h
@@ -40,8 +40,9 @@ public:
 	 *
 	 * @param text text to show.
 	 * @param align text alignment.
+	 * @param halfwidthspace if half width spaces should be used.
 	 */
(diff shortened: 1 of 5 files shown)


src/window_help.h
```diff
@@ -41,7 +41,7 @@ class Window_Help : public Window_Base {
 	 * @param text text to show.
 	 * @param align text alignment.
```
Ghabry, line 42: you forgot to add the new param to the doc comment.
Otherwise I think this patch is fine.

rueter37: Fixed.

Ghabry (approved):

fdelapena (approved):

fmatthew5876 (approved):

PR #2486, opened 2021-03-16
Stub Maniac Patch commands, add --patch cli arg

Ghabry:
This stubs Maniac Patch commands in the interpreter and emits warnings.
Can help in finding problematic games.

Also added a ``--patch`` cli arg because there is currently no way to configure this stuff in emscripten.


fdelapena:
Because this adds commands to battle interpreter, better slip until battle rewrite gets merged to prevent rebase hell.

fdelapena (approved):

fmatthew5876 (approved):

Issue #2487, opened 2021-03-16
Ara Fell: Not all states shown in Info box


Ghabry:
#### Name of the game: Ara Fell Remastered

This is a commercial game so I cannot link it.

Basicly RPG Maker 2003 shows states in a info box when you hover a target:

![Screenshot_20210316_204453-fs8](https://user-images.githubusercontent.com/1331889/111370853-019b3800-8699-11eb-99f5-dbc8fdea52e8.png)

Ara Fell uses this in a quite creative way: To show the enemy name and some metadata.

Player only shows IMMUNE.

![3-fs8](https://user-images.githubusercontent.com/1331889/111371417-a7e73d80-8699-11eb-9d83-81eea797e035.png)
![2-fs8](https://user-images.githubusercontent.com/1331889/111371418-a87fd400-8699-11eb-8178-6e0632246b64.png)
![1-fs8](https://user-images.githubusercontent.com/1331889/111371420-a87fd400-8699-11eb-9b80-24e755e66a97.png)

rueter37:
As far as I know in RPG_RT the info box shows up to 5 states ordered by state priority. If two states have the same priority then that state with the higher editor ID "wins". 9 full space characters are reserved per state because the state name is allowed to have up to 8 characters.

rueter37:
Dropping out from this one - it is too likely that I break the help window on trying to fix this issue.

rueter37:
I have decided to take on this issue again. I have added a PR (#2492) now which should fix this.

PR #2492, opened 2021-03-19
Show more than one state in help window

rueter37:
This PR makes the help window in RPG Maker 2003 battles show more than one state if you target an enemy. Up to 5 states are shown ordered by priority.

Fixes: #2487

### DIFF
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -678,6 +678,11 @@ void Game_Battler::Flash(int r, int g, int b, int power, int frames) {
 	flash.time_left = frames;
 }
 
+const std::vector<lcf::rpg::State*> Game_Battler::GetInflictedStatesOrderedByPriority() const {
+	std::vector<lcf::rpg::State*> state_list = State::GetObjects(GetStates());
+	return State::SortedByPriority(state_list);
+}
+
 int Game_Battler::CanChangeAtkModifier(int modifier) const {
 	const auto prev = atk_modifier;
 	const auto base = GetBaseAtk();
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -44,6 +44,7 @@
 #include "output.h"
 #include "autobattle.h"
 #include "enemyai.h"
+#include <algorithm>
 
 Scene_Battle_Rpg2k3::Scene_Battle_Rpg2k3(const BattleArgs& args) :
 	Scene_Battle(args),
@@ -427,9 +428,15 @@ void Scene_Battle_Rpg2k3::UpdateAnimations() {
 					enemy_cursor->SetX(enemy->GetBattlePosition().x + sprite->GetWidth() / 2 + 2);
 					enemy_cursor->SetY(enemy->GetBattlePosition().y - enemy_cursor->GetHeight() / 2);
 
-					auto* state = enemy->GetSignificantState();
-					if (state) {
-						help_window->SetText(ToString(state->name), state->color);
+					std::vector<lcf::rpg::State*> ordered_states = enemy->GetInflictedStatesOrderedByPriority();
+					if (ordered_states.size() > 0) {
+						help_window->Clear();
+						int state_counter = 0;
+						for (lcf::rpg::State* state : ordered_states) {
+							std::string state_name = fmt::format("{:9s}", state->name);
+							help_window->AddText(state_name, state->color, Text::AlignLeft, false);
+							if (++state_counter >= 5) break;
+						}
 						help_window->SetVisible(true);
 					} else {
 						help_window->SetVisible(false);
--- a/src/state.cpp
+++ b/src/state.cpp
@@ -203,4 +203,23 @@ int GetStateRate(int state_id, int rate) {
 
 }
 
+const std::vector<lcf::rpg::State*> GetObjects(const StateVec& states) {
+	std::vector<lcf::rpg::State*> state_list;
+	for (size_t i = 0; i < states.size(); ++i) {
+		if (states[i] > 0) {
+			lcf::rpg::State* state = lcf::ReaderUtil::GetElement(lcf::Data::states, i + 1);
+			state_list.push_back(state);
+		}
+	}
+	return state_list;
+}
+
+const std::vector<lcf::rpg::State*> SortedByPriority(const std::vector<lcf::rpg::State*>& states) {
+	std::vector<lcf::rpg::State*> state_list = states;
+	std::sort(state_list.begin(), state_list.end(), [](lcf::rpg::State* lhs, lcf::rpg::State* rhs) {
+		return std::tie(lhs->priority, lhs->ID) > std::tie(rhs->priority, rhs->ID);
+	});
+	return state_list;
+}
+
 } //namspace State
--- a/src/window_help.cpp
+++ b/src/window_help.cpp
@@ -33,24 +33,8 @@ void Window_Help::SetText(std::string text, int color, Text::Alignment align, bo
 	if (this->text != text || this->color != color || this->align != align) {
 		contents->Clear();
 
-		int x = 0;
-		std::string::size_type pos = 0;
-		std::string::size_type nextpos = 0;
-		while (nextpos != std::string::npos) {
-			nextpos = text.find(' ', pos);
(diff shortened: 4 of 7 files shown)


fdelapena (approved):

src/scene_battle_rpg2k3.cpp
```diff
@@ -427,9 +428,33 @@ void Scene_Battle_Rpg2k3::UpdateAnimations() {
...
+							}
+						});
+
+						help_window->Clear();
+						int state_counter = 0;
+						int x = 0;
+						for (lcf::rpg::State* state : ordered_states) {
+							std::string state_name = fmt::format("{:9s}", state->name);
+							help_window->AddText(x, state_name, state->color, Text::AlignLeft, false);
+							x += Font::Default()->GetSize(state_name).width;
```
fmatthew5876, line 455: Instead of recomputing the size here, make `AddText()` return the size and reuse it from there.

src/scene_battle_rpg2k3.cpp
```diff
@@ -427,9 +428,33 @@ void Scene_Battle_Rpg2k3::UpdateAnimations() {
...
+					for (size_t i = 0; i < states.size(); ++i) {
+						if (states[i] > 0) {
+							lcf::rpg::State* state = lcf::ReaderUtil::GetElement(lcf::Data::states, i + 1);
+							ordered_states.push_back(state);
+						}
+					}
+
+					if (ordered_states.size() > 0) {
+						std::sort(ordered_states.begin(), ordered_states.end(), [](lcf::rpg::State* lhs, lcf::rpg::State* rhs) {
+							if (lhs->priority == rhs->priority) {
```
fmatthew5876, line 442: I think this sort should be a utility function in the `State` library

src/scene_battle_rpg2k3.cpp
```diff
@@ -427,9 +428,33 @@ void Scene_Battle_Rpg2k3::UpdateAnimations() {
...
-					if (state) {
-						help_window->SetText(ToString(state->name), state->color);
+					std::vector<lcf::rpg::State*> ordered_states;
+					const auto& states = enemy->GetStates();
+					for (size_t i = 0; i < states.size(); ++i) {
+						if (states[i] > 0) {
+							lcf::rpg::State* state = lcf::ReaderUtil::GetElement(lcf::Data::states, i + 1);
+							ordered_states.push_back(state);
+						}
+					}
```
Ghabry, line 438: This for loop here (which basicly fetches the objects for state ids) should be also a helper function in the state class.

Something like ``std::vector<State*> GetObjects()``

src/window_help.cpp
```diff
@@ -60,3 +60,23 @@ void Window_Help::Clear() {
...
+		x += Font::Default()->GetSize(segment).width;
+
+		if (nextpos != decltype(text)::npos) {
+			if (halfwidthspace) {
+				x += Font::Default()->GetSize(" ").width / 2;
+			} else {
+				x += Font::Default()->GetSize(" ").width;
+			}
+			pos = nextpos + 1;
+		}
```
Ghabry, line 80: Because this is a copy-paste from "SetText" you can make SetText call AddText, then the code is not duplciated anymore.

src/scene_battle_rpg2k3.cpp
rueter37: Fixed.

src/window_help.cpp
rueter37: Fixed.

src/window_help.h
```diff
@@ -49,13 +49,25 @@ class Window_Help : public Window_Base {
...
+
 private:
 	/** Text to draw. */
 	std::string text;
 	/** Color of Text to draw. */
 	int color = Font::ColorDefault;
 	/** Alignment of text. */
 	Text::Alignment align;
+	/** Current text position */
+	int x = 0;
```
Ghabry, line 70: this shadows (same name) the x from the Window class which indicates the x position of the window on screen.
Can you rename it to something else like ``text_x_offset`` ?

src/scene_battle_rpg2k3.cpp
rueter37: Added.

src/scene_battle_rpg2k3.cpp
rueter37: Added.

src/window_help.h
rueter37: Fixed.

src/state.cpp
```diff
@@ -203,4 +203,30 @@ int GetStateRate(int state_id, int rate) {
...
+	std::vector<lcf::rpg::State*> state_list;
+	for (size_t i = 0; i < states.size(); ++i) {
+		state_list.push_back(states[i]);
+	}
+	std::sort(state_list.begin(), state_list.end(), [](lcf::rpg::State* lhs, lcf::rpg::State* rhs) {
+		if (lhs->priority == rhs->priority) {
+			return lhs->ID > rhs->ID;
+		} else {
+			return lhs->priority > rhs->priority;
+		}
```
Ghabry, line 227: look in Tone and Color headers.
This compare stuff can be easier written and less error prone via ``std::tie()``

src/state.cpp
```diff
@@ -203,4 +203,30 @@ int GetStateRate(int state_id, int rate) {
...
+	}
+	return state_list;
+}
+
+const std::vector<lcf::rpg::State*> SortedByPriority(const std::vector<lcf::rpg::State*>& states) {
+	std::vector<lcf::rpg::State*> state_list;
+	for (size_t i = 0; i < states.size(); ++i) {
+		state_list.push_back(states[i]);
+	}
```
Ghabry, line 221: This is a copy: ``state_list = states``.

src/game_battler.cpp
```diff
@@ -678,6 +678,11 @@ void Game_Battler::Flash(int r, int g, int b, int power, int frames) {
 	flash.time_left = frames;
 }
 
+const std::vector<lcf::rpg::State*> Game_Battler::GetInflictedStatesOrderedByPriority() const {
+	std::vector<lcf::rpg::State*> state_list = State::GetObjects(GetStates());
+	return State::SortedByPriority(state_list);
+}
```
Ghabry, line 684: This is more an opinion:
I would not use a new method and simply chain it in the battle class:
``std::vector<lcf::rpg::State*> ordered_states = State::Sorted(State::GetObject(enemy->GetStates()))``

Ghabry (commented):
-

src/state.cpp
rueter37: Fixed.

src/state.cpp
rueter37: Fixed.

src/game_battler.cpp
rueter37: Well, I thought that would be a good place to use the new `GetObjects()` methods...

src/scene_battle_rpg2k3.cpp
```diff
@@ -427,9 +428,15 @@ void Scene_Battle_Rpg2k3::UpdateAnimations() {
...
-					auto* state = enemy->GetSignificantState();
-					if (state) {
-						help_window->SetText(ToString(state->name), state->color);
+					std::vector<lcf::rpg::State*> ordered_states = enemy->GetInflictedStatesOrderedByPriority();
+					if (ordered_states.size() > 0) {
+						help_window->Clear();
+						int state_counter = 0;
+						for (lcf::rpg::State* state : ordered_states) {
+							std::string state_name = fmt::format("{:9s}", state->name);
+							help_window->AddText(state_name, state->color, Text::AlignLeft, false);
```
fmatthew5876, line 437: What does RPG_RT do when thee state text runs off the edge of the window? Does it display partial text cutoff on the end like "poi" instead of "poison"? Or is it smart enough to not display the text if the entire word doesn't fit?

rueter37: Because RPG_RT shows at most 5 states (tested it myself), the state text will never run off the edge of the window. A state name can have at most 8 characters, and RPG_RT reserves 9 characters per state, so we have a maximum of 45 characters. The character limit of the help window is 50.

fmatthew5876 (approved):

Issue #2494, opened 2021-03-20
Game fails to read assets at random times.


axzxc1236:
Update 1: I saw the solution and I build the patched PKGBUILD for my self but I only have the time to play the game on next weekend.

---

Sorry, I didn't use continuous build, but I built from latest source (239686f) using easyrpg-player-git from AUR.

---
Please fill in the following fields before submitting an issue:
#### Name of the game:
(This is a Chinese RPG game) [勇者特訓班](https://forum.gamer.com.tw/C.php?bsn=4918&snA=29683)

#### Player platform:
Linux (Arch Linux based distro)

#### Attach files (as a .zip archive or link them)
[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/6175414/easyrpg_log.txt)

#### Describe the issue in detail and how to reproduce it:
I am now playing this RPG game with long play hours and at random times easyrpg will fail to load game assets... and then it's too late to save the game, if I tried to do so it will make easyrpg force close the game and end up losing progress.
![force quit screenshot](https://user-images.githubusercontent.com/3145859/111868040-ad52c980-89b2-11eb-88c6-54d338d5f3ee.png)
First it will show the message `Warning: Music not readable: xxxxx.mid` then I tries to get to menu by pressing X then I get the red screen that says `couldn't open image file.....  EasyRPG Player will close now.`
I couldn't figure out what is triggering this odd behavior (I mean I know it will get me red screen if I tried to open menu by pressing X, but I don't know why it fails to load game assets).... seconds before game was fine and the next second it isn't, if there is something I can do to provide more debug information I can do it.

carstene1ns:
Could be memory related. Can you try getting the used ram?
```sh
free -hm
sudo pmap `pidof easyrpg-player`
```

Ghabry:
Or we maybe leak file handles.
This is an interesting problem.

Ghabry:
My educated guess by now are the error paths.
It could be that some libraries we use do not invoke the close callback on error.

Ghabry:
@axzxc1236 
Two workarounds for now:
Apply this commit to your build and recompile: https://github.com/Ghabry/easyrpg-player/commit/73cbc9f888cc0f888b67b8cdd560f5b2f78eab1c

Or:
Install wildmidi and freepats-legacy from AUR.
Copy /usr/share/freepats/wildmidi-freepats.cfg to /etc/wildmidi/wildmidi.cfg
[...]

PR #2496, opened 2021-03-20
SDL mixer: Do not leak handles

Ghabry:
This one is interesting and very old :(.

- [x] Test Native MIDI on Windows

The good news: This only affects SDL Mixer (not SDL) and in June 2016 we added the Audio Decoder classes which resulted in this codepath being almost never used anymore (except for Native MIDI playback).

First: ``Load_MUS_RW`` of SDL Mixer 1.2 _never_ closes handles. So for SDL1.2 we leak them since June 2014:
https://github.com/EasyRPG/Player/commit/a441cbc16e7ca9fee42a9527536d824e657deaef
(I will not fix this, we do not use Mixer 1.2 anywhere anymore)

For SDL2 the bug was introduced here in Feb 2016:
https://github.com/EasyRPG/Player/commit/4f2b8c9cb5beb109c65e7838f71b986c3e6a2892
The commit messages links to #710 which discusses a fix for a double free in the SDL ADPCM loader.

I also think that this comment here is wrong "The crash can be worked around by passing 0 instead of 1 to SDL_LoadMUS_RW() and the free call is already done from the shared_ptr reset deleter parameter.". The shared_ptr deleter only applied to SE, not to BGM.

_I simply set the "free_src" param to 1, I do not really care that this breaks ADPCM, we use libsndfile for this. SDL mixer can die once the native Midi PR is in_

Fix: #2494

--------

I also noticed that BGS_Play does use filenames directly, so this is also broken with my VFS-PR... Not going to fix this.

### DIFF
--- a/src/audio_sdl_mixer.cpp
+++ b/src/audio_sdl_mixer.cpp
@@ -318,12 +318,7 @@ void SdlMixerAudio::BGM_Play(std::string const& file, int volume, int pitch, int
 		return;
 	}
 
-	filestream = FileFinder::OpenInputStream(file);
-	if (!filestream) {
-		// Unlikely, just in case
-		Output::Warning("Music not readable: {}", FileFinder::GetPathInsideGamePath(file));
-		return;
-	}
+	filestream.seekg(0, std::ios_base::beg);
 	SDL_RWops* rw = create_StreamRWOps(std::move(filestream));
 
 	bgm_stop = false;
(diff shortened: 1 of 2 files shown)


fdelapena (approved):

fmatthew5876 (approved):

carstene1ns (approved):

Ghabry:
Running this on Windows gives "vector subscript out of range". Have to attach a debugger first before this is mergeable...

-----

tenor

Ghabry:
The assertion was unrelated, but I fixed it.

carstene1ns:
> The assertion was unrelated, but I fixed it.

Good thing this only was in for 4 days :P
(since #2478)

Ghabry:
I wonder why ASAN didnt catch it.

Issue #2497, opened 2021-03-22
Maniac: Support of custom chunks


Ghabry:
The Maniac patch adds certain chunks. Some of them are useful in general, e.g. for debugging purposes.

see also #1818 

Here is the list:

Enemy:
- [x] maniac_unarmed_animation. Animation for normal enemy attacks. This is a cool addition for the battle system imo :+1: (also doesnt sound hard to implement. CC @rueter37 #2498  since 0.7

ExecFrame:
- [ ] maniac_event_info. Information about the current event. Useful for debugging.
- [ ] maniac_event_id. The event ID (contrary to the other event ID field this one is not reset on map change, Useful for debugging)
- [ ] maniac_event_page_id. The current event page. Useful for debugging
- [x] maniac_loop_info_size. Here we must only write the correct value (it is possible to let liblcf do this would have required a custom type, which was not worth it for a patch imo) #2649
- [x] maniac_loop_info: Needed for the loop event command #2649

System:
- [x] maniac_frameskip - Imo this one is not interesting. WONT FIX.
- [ ] maniac_picture_limit - Not really useful, we could just copy our Picture limit over (but we are dynamic so this doesn't really matter)
- [ ] maniac_options: Imo this stuff should be global options, not game options. Except for TestPlay toggle feature.
- [ ] maniac_joypad_bindings: Maybe interesting?

PR #2498, opened 2021-03-22
Implement customizable enemy unarmed battle animation

rueter37:
This PR allows to set the unarmed battle animation of an enemy via the "maniac_unarmed_animation" setting in the LCF database. This fixes the first bullet point in #2497.

### DIFF
--- a/src/game_enemy.h
+++ b/src/game_enemy.h
@@ -21,6 +21,7 @@
 // Headers
 #include "game_battler.h"
 #include "sprite_enemy.h"
+#include "player.h"
 #include <lcf/rpg/enemy.h>
 #include <lcf/rpg/enemyaction.h>
 #include <lcf/rpg/troopmember.h>
@@ -224,6 +225,9 @@ public:
 	/** @return true if enemy is flying */
 	bool IsFlying() const;
 
+	/** @return the id of the enemy's unarmed battle animation */
+	int GetUnarmedBattleAnimationId() const;
+
 	Sprite_Enemy* GetEnemyBattleSprite() const;
 
 protected:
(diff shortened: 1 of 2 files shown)


src/game_enemy.h
```diff
@@ -362,6 +365,10 @@ inline bool Game_Enemy::IsFlying() const {
 	return enemy->levitate;
 }
 
+inline int Game_Enemy::GetUnarmedBattleAnimationId() const {
+        return enemy->maniac_unarmed_animation;
```
Ghabry, line 369: Please guard this with a Player::IsPatchManiac and return 1 when the patch is unavailable.

Ghabry:
This was fast. Did you also test it in RPG_RT with Maniac Patch?

src/game_enemy.h
rueter37: Fixed.

rueter37:
Unfortunately not.

src/game_enemy.h
fmatthew5876: This should also be an easyrpg extension outside of just maniac patch, but we can worry about that later. We don't have any framework for this yet.

fmatthew5876: Also @rueter37 Please put the conditional in `Game_Enemy` not at the call site.

rueter37: I'm probably lacking intelligence but I don't understand this one.

Ghabry: He meant that you should put the Maniac-Check and the return 1 here where I marked it with my comment and in game_battlealgorithm just call enemy->GetUnarmedBattleAnimationId().
This has the advantage that any caller of this function will get the expected value.

rueter37: Thanks for the explanation! I hope I got this right now.

Ghabry: yeah this way

fmatthew5876 (approved):

fdelapena (approved):

Ghabry (changes requested):
Needs test before merge (my job^^)

carstene1ns (approved):
This should work even if the maniac patch is unused, since we default on 1 in liblcf.

Ghabry:
Yeah, but this is less sloppy. For RPG2k3E we are quite sloppy with the chunk handling everywhere, e.g. "Skip title" works in all engines but should only work in 2k3E :/

Ghabry (approved):

Issue #2362, opened 2020-09-26
List of custom EasyRPG terms


fmatthew5876:
This issue is to create a list of all the hard coded strings in RPG_RT that we would like to add as customizable Terms as an easyrpg extension. The rule can follow `SaveActor::name`, in that the string defaults to `\x1` which means use the RPG_RT compatible default value.

We could also extend the RPG_RT defaults to use "advocate" or "2k3e" translations, depending on the engine version.

Menu Scene (#2473)
-------------

- [x] Item number separator
- [x] Skill cost separator
- [x] Equipment arrow
- [x] Status Screen Name
- [x] Status Screen Class
- [x] Status Screen Title
- [x] Status Screen Condition
- [x] Status Screen Frnt
- [x] Status Screen Back

Battle Scene (#2502)
-------------

- [x] 2k3 battle monster double attack notification
- [x] 2k3 battle monster defend notification
- [x] 2k3 battle monster observe notification
- [x] 2k3 battle monster charge up notification
- [x] 2k3 battle monster self-destruct notification
- [x] 2k3 message for back and pincer attack (default is no message)

Ghabry:
The 2k3 battle notification for skill (currently is the skill name) should be also customizable (with placeholder support as in 2k)

Tayruu:
I've edited things like these myself in my RPG_RT, so the ability to set them somehow without compiling my own Player would be neat. (Though I have some functions I'll need to custom compile anyway.)

Just missing the "-----/-----" of max EXP.

Ghabry:
This --------/------- vs EXP/-------- could be also a configuration. But more a flag than a term.

--------------------

Or maybe we should just add the "Scene Push" event command I suggested, reimplement the entire Menu scene in RPG Maker Event script and then you can alter stuff by modifying event code ;))).

rueter37:
The "Confirm" and "Redo" commands in the party order scene are hard coded as well.

Ghabry:
@rueter37
I just saw this commit in liblcf: https://github.com/rueter37/liblcf/commit/cf01c364b07c974aa0fb3a0a13cbd69345202386
Can you PR this commit?
Looks like a good start for this feature.

rueter37:
@Ghabry I have opened a [PR with the commit you mentioned](https://github.com/EasyRPG/liblcf/pull/414) now in liblcf.

rueter37:
[...]

PR #2502, opened 2021-03-31
Make RPG Maker 2003 battle hardcoded terms customizable

rueter37:
This PR makes the hardcoded terms in RPG Maker 2003 battle customizable and adds skill and item start messages in RPG Maker 2003 battles.

Fixes: #2362

### DIFF
--- a/src/battle_message.cpp
+++ b/src/battle_message.cpp
@@ -432,8 +432,15 @@ std::string GetItemStartMessage2k(const Game_Battler& source, const lcf::rpg::It
 	return ToString(source.GetName()) + particle + ToString(item.name) + ToString(lcf::Data::terms.use_item);
 }
 
-std::string GetDoubleAttackStartMessage2k3() {
-	return "Double Attack";
+std::string GetDoubleAttackStartMessage2k3(const Game_Battler& source) {
+	if (lcf::Data::terms.easyrpg_battle2k3_double_attack == lcf::Data::terms.kDefaultTerm) {
+		return "Double Attack";
+	}
+	return Utils::ReplacePlaceholders(
+		lcf::Data::terms.easyrpg_battle2k3_double_attack,
+		Utils::MakeArray('S'),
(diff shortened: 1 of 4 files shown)


src/battle_message.cpp
```diff
@@ -432,36 +440,110 @@ std::string GetItemStartMessage2k(const Game_Battler& source, const lcf::rpg::It
...
+		lcf::Data::terms.easyrpg_battle2k3_double_attack,
+		Utils::MakeArray('S'),
+		Utils::MakeSvArray(source.GetName())
+	);
 }
 
-std::string GetSkillStartMessage2k3(const lcf::rpg::Skill& skill) {
-	return ToString(skill.name);
+std::string GetSkillStartMessage2k3(const Game_Battler& source, const Game_Battler* target, const lcf::rpg::Skill& skill) {
+	StringView target_name = "???";
```
Ghabry, line 455: As these placeholders are from 2kE: When you use a skill that targets the entire party is the %O also returning ??? there?

Ghabry (approved):
Feels a bit strange to see placeholders which is a 2kE feature in 2k3 code but considering these are EasyRPG extensions it makes sense to use them. :)
Code-wise I see no problems here and this all implements stuff I consider useful.

fdelapena (approved):

fmatthew5876 (approved):

src/battle_message.cpp
rueter37: Tested with RPG_RT 2kE and confirmed that %O returns ??? if a skill targets the entire party.

carstene1ns (approved):

carstene1ns:
While we are at it, these messages ("Double Attack", "Observe",...) should be added somehow to the translation feature without using custom terms.

Ghabry:
At least lcftrans and translation already supports this because it is handled by ForEachString.
For the database it needs our editor 😁

Issue #1269, opened 2017-09-27
Battle 2k3: Item animations


Ghabry:
![sin titulo](https://user-images.githubusercontent.com/1331889/30932985-7f425d80-a3c9-11e7-8200-a87cd61ecf90.png)

~~wtf is CBA~~

Weapon CBA:

- [x] Show the right handed battle weapon in battle (#2510)
- [x] Show left handed (TODO: Both handed?) (#2510)
- [x] Movement (#2503):
  - [x] No Movement: Use the skill without moving.
  - [x] Step Forward: Step forward, then use the skill.
  - [x] Jump Forward: Jump forward, then use the skill.
  - [x] Move to Enemy: Move in front of the enemy, then use the skill.
- [x] Attack Count: How often to swing the weapon, just optical. Not for animations (#2399)
- [x] CBA Weapon (#2510)
- [x] Afterimage yes/no "Choose from the dropdown menu whether an afterimage should be shown when the actor is moving during skill usage. If No movement was selected, there is no need to specify this. (#2538)

Battle Animation CBA:
- [x] Battle animation (#2510)

Ranged CBA:
- [x] "Use" checkbox (#2515)
- [ ] Ranged Path
  - [x] Dagger "fly in a straight line like a knife" (#2515)
  - [ ] Boomerang "Follow an elliptic path"
- [ ] Ranged Attack range
  -  [x] Single Enemy (#2515)
  - [x] Center (#2537)
  - [X] All Enemies (#2537)
  - [ ] Attack in Sequence
- [x] Flying speed (#2515):
   - [x] Slow
   - [x] Normal
   - [x] Fast

32 different battle weapon types can be configured (Animation 2 tab). Each item has one of these types with corresponding animation. The graphic used is per battler.

~~Open questions:
Can this be used together with "Battle animation" battlers? Based on the GUI: No~~

CherryDT:
CBA = Character Battle Animation

fmatthew5876:
CBA Movement details
=======================

Here some details about how CBA movement works. More investigation and testing is needed.

Actor::UpdateBattle 004BBD7C
```
if (hidden) return;
if (battle event is waiting for wait command or waiting for "wait for key" (from wait command, not keyinputproc)) return;

damagePopupTimer -= 4
framecounter++
framecounter++ //Not a typo, this really happens twice?!

// Handle victory conditions
// Handle row command placement

// Handle weapon attacks

```


[...]

PR #2503, opened 2021-04-01
Battle 2k3: Add CBA movement

rueter37:
This PR implements the CBA movement in RPG Maker 2003 battles.

Fixes: #1277
Fixes: #1708

All movement related checkboxes in #1269 are handled by this PR.

Addressing review comments will be handled now in subsequent commits unless stated otherwise by the reviewers in order to avoid confusion.

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -726,10 +726,8 @@ int Game_BattleAlgorithm::Normal::GetCBAMovement() const {
 		auto weapons = ally->GetWeapons(weapon);
 		auto* item = weapons[0];
 		if (item) {
-			if (item->animation_id > 0) {
-				if (static_cast<int>(item->animation_data.size()) > source->GetId() - 1) {
-					return item->animation_data[source->GetId() - 1].movement;
-				}
+			if (static_cast<int>(item->animation_data.size()) > source->GetId() - 1) {
+				return item->animation_data[source->GetId() - 1].movement;
 			}
 		}
 	}
@@ -1041,7 +1039,7 @@ std::string Game_BattleAlgorithm::Skill::GetStartMessage(int line) const {
 
 int Game_BattleAlgorithm::Skill::GetSourcePose() const {
 	auto* source = GetSource();
-	if (source->GetType() == Game_Battler::Type_Ally && skill.animation_id > 0) {
+	if (source->GetType() == Game_Battler::Type_Ally) {
 		if (static_cast<int>(skill.battler_animation_data.size()) > source->GetId() - 1) {
 			return skill.battler_animation_data[source->GetId() - 1].pose;
 		}
@@ -1052,7 +1050,7 @@ int Game_BattleAlgorithm::Skill::GetSourcePose() const {
 
 int Game_BattleAlgorithm::Skill::GetCBAMovement() const {
 	auto* source = GetSource();
-	if (source->GetType() == Game_Battler::Type_Ally && skill.animation_id > 0) {
+	if (source->GetType() == Game_Battler::Type_Ally) {
 		if (static_cast<int>(skill.battler_animation_data.size()) > source->GetId() - 1) {
 			return skill.battler_animation_data[source->GetId() - 1].movement;
 		}
@@ -1193,7 +1191,7 @@ int Game_BattleAlgorithm::Item::GetSourcePose() const {
 
 int Game_BattleAlgorithm::Item::GetCBAMovement() const {
 	auto* source = GetSource();
-	if (source->GetType() == Game_Battler::Type_Ally && item.animation_id > 0) {
+	if (source->GetType() == Game_Battler::Type_Ally) {
 		if (static_cast<int>(item.animation_data.size()) > source->GetId() - 1) {
 			return item.animation_data[source->GetId() - 1].movement;
 		}
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -322,6 +322,10 @@ int Game_BattleAlgorithm::AlgorithmBase::GetSourcePose() const {
 	return lcf::rpg::BattlerAnimation::Pose_Idle;
 }
 
+int Game_BattleAlgorithm::AlgorithmBase::GetCBAMovement() const {
+	return lcf::rpg::BattlerAnimationItemSkill::Movement_none;
+}
+
 void Game_BattleAlgorithm::AlgorithmBase::Start() {
 	reflect_target = nullptr;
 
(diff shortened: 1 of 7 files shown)


src/scene_battle_rpg2k3.h
```diff
@@ -18,6 +18,11 @@
 #ifndef EP_SCENE_BATTLE_RPG2K3_H
 #define EP_SCENE_BATTLE_RPG2K3_H
 
+// FIXME: How many CBA move frames does RPG_RT actually use?
+#define CBA_NUM_MOVE_FRAMES_STEP 15
```
Ghabry, line 22: Please use constexpr and follow the naming convention . See https://github.com/EasyRPG/Player/blob/831c1a42141252eaf2f84aa01634de7318d9160e/src/scene_battle.h#L65

(you can squash this commit)

rueter37: Fixed.

rueter37:
Changed the CBA move frame count according to @fmatthew5876's [findings about the CBA movement](https://github.com/EasyRPG/Player/issues/1269#issuecomment-811899764).

Ghabry:
Not fully reviewed and tested yet but on first look this looks clean :+1:

fdelapena (approved):

src/game_battlealgorithm.cpp
```diff
@@ -1157,6 +1191,17 @@ int Game_BattleAlgorithm::Item::GetSourcePose() const {
 	return lcf::rpg::BattlerAnimation::Pose_Item;
 }
 
+int Game_BattleAlgorithm::Item::GetCBAMovement() const {
+	auto* source = GetSource();
+	if (source->GetType() == Game_Battler::Type_Ally && item.animation_id > 0) {
```
fmatthew5876, line 1196: Is the `animation_id > 0` check actually needed for Item, Skill and Normal?

fmatthew5876:
Looks ok to me as well from a code review perspective. This needs a good deal of testing from someone other than the PR author.

fmatthew5876:
FYI for testing pixel accuracy of movment, I find the `bath` RTP battle background to be pretty good.

src/game_battlealgorithm.cpp
rueter37: Did some tests myself and it seems to work without these "safeguards". Should I remove them then?

fmatthew5876: I think so, since the `animation_id` is not used in the CBA data lookup. Also I don't see such checks in RE, but I could have missed it somewhere. The RPG_RT code for this stuff is horrible.

rueter37: Removed the `animation_id > 0` checks now.

Ghabry:
Will take a look on Friday or early next week, depending on my time.

Besides OFF, DBZ: Legend of Z and VD2: Are there any other popular games that use CBA?

rueter37:
Unfortunately I'm not well informed on this one, personally I only know VD2 as a game which uses CBA movement. I mostly played RPG Maker 2000 games back then (and that was more than 16 years ago, time flies really fast :smile:)

Ghabry:
Well I used rmarchiv again here. Unfortunately searching for "Jump Forward" is kinda impossible without adding lots of other filtering code because when you add more Actors they get "Jump Forward" by default.

Some games that use "Step". The value in front is how open it is used in the _entire game_.

- 10 06 Alte Macht X
-     12 DSdLHdTGdE
-     12 Legion Saga R
-     17 Final Fantasy Unknown (Demo)
-     17 Holy Trinity - Die Apokalypse
-     21 Legends of System Gaia
-     25 Der Brief für den König - Demo II (Ver1.5)
-     27 Echoes Of The Past
-     31 The Legend of Phylandia (Demo)
-     52 Oracle of Tao
-     66 Blackmoon Prophecy

Some games that use "Move to enemy":

-    102 06 Alte Macht X
-    125 Hero of Diamantia (Demo)
-    127 Balmung Cycle
-    127 Die letzte Schlacht der Elfen VIII
-    181 The Fadestrider (Demo v1.25)
-    185 Vampire Chronicles 3 (Demo 3)
-    192 Sylfia - Legacy of the Moonmirror
-    332 Oracle of Tao
-    849 Von Händlern und Helden (Beta 0.1)
-   1186 THEIA_Orihalcon_Edition_ENG_V1-02

1186 times? That sounds like some bloat project :D

Ghabry:
Well there are some timing issues (off by some frames) all over the place. Not just in CBA but imo this can be adressed when everything is implemented. Otherwise this is a too time-consuming task.

@rueter37 
In Blackmoon Prophecy (you find it on rmarchiv). Just fight the first monster group. After the attack your hero moves back.
In your PR the movement back is not flipped (faces left, not right)

Ghabry:
@CherryDT
Do you know about a bug in the editor that appears to write partially uninitialized memory into the battle test equipment data?

When I alter the battletest equipment the values saved are sometimes really high. The solution seems to be to strip of the upper 16 bit with "0xFFFF".

rueter37:
@Ghabry About the "Step Forward" issue: I found out it depends if you are doing a normal attack or using a skill. The move back is flipped in case of a normal attack, but not flipped in case of using a skill. "Jump Forward" is affected by this as well.

CherryDT:
> @CherryDT
> Do you know about a bug in the editor that appears to write partially uninitialized memory into the battle test equipment data?
> 
> When I alter the battletest equipment the values saved are sometimes really high. The solution seems to be to strip of the upper 16 bit with "0xFFFF".

No, sorry... But if you discover the cause/the code address where it happens, I'd be happy to hear about it

Ghabry:
@rueter37 
Could you apply this patch here. Is unrelated to CBA but without it most battle tests will launch with the weapons removed...

```
From 9e362907375d746fd309e76e2a28bc8ea9216c3d Mon Sep 17 00:00:00 2001
From: Ghabry <gabriel+github@mastergk.de>
Date: Sat, 17 Apr 2021 17:01:19 +0200
Subject: [PATCH 1/1] Remove garbage data in BattleTest, the upper 16bit look
 uninitialized

This garbage data was already noticed in ad1c5be but the fix simply removed the entire equipment, which is incorrect.
---
 src/game_party.cpp | 8 +++++++-
 1 file changed, 7 insertions(+), 1 deletion(-)

diff --git a/src/game_party.cpp b/src/game_party.cpp
index 6e02ad2e7..9640c1e92 100644
--- a/src/game_party.cpp
+++ b/src/game_party.cpp
@@ -108,7 +108,13 @@ void Game_Party::SetupBattleTest() {
 		Game_Actor* actor = Main_Data::game_actors->GetActor(btdata.actor_id);
 
 		// Filter garbage btdata inserted by the editor
-		std::array<int, 5> ids = {{ btdata.weapon_id, btdata.shield_id, btdata.armor_id, btdata.helmet_id, btdata.accessory_id }};
+		// The upper 16 bit look like uninitialized data
+		std::array<int, 5> ids = {{
+			btdata.weapon_id & 0xFFFF,
+			btdata.shield_id & 0xFFFF,
+			btdata.armor_id & 0xFFFF,
+			btdata.helmet_id & 0xFFFF,
+			btdata.accessory_id & 0xFFFF }};
 		std::replace_if(ids.begin(), ids.end(), [] (const int& item_id) {
 			return lcf::ReaderUtil::GetElement(lcf::Data::items, item_id) == nullptr;
 		}, 0);
-- 
2.30.1

```

Ghabry:
> The move back is flipped in case of a normal attack, but not flipped in case of using a skill.

The question is now if this is a bug or a feature xD

rueter37:
@Ghabry Added your patch now.

Ghabry:
Yeah besides the sometimes incorrect facing after step or jump forward I cannot find any issues :+1: 

When this is fixed can be approved.

rueter37:
Fixed the facing direction on "step forward" and "jump forward" CBA movement types when moving back now.

Ghabry:
Found another bug in the editor: When you edit the CBA animation of an item in that modal dialog and click Cancel the change is still applied. Just not shown until you change the item selection once :facepalm:

Ghabry (approved):
I tested this directly together with the upcoming Weapon CBA PR.
The facing looks now correct in all cases :+1: . Also tested pincer, surprise etc.

fmatthew5876 (approved):
Code looks good to me, and this has been tested. I approve

PR #2510, opened 2021-04-17
Battle 2k3: Add CBA melee weapons

rueter37:
This PR implements the CBA melee weapons in RPG Maker 2003 battles.

### DIFF
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -28,6 +28,7 @@
 #include "sprite.h"
 #include "sprite_enemy.h"
 #include "sprite_actor.h"
+#include "sprite_weapon.h"
 #include "cache.h"
 #include "game_actors.h"
 #include "game_system.h"
@@ -333,6 +334,7 @@ void Scene_Battle_Rpg2k3::CreateEnemySprites() {
 void Scene_Battle_Rpg2k3::CreateActorSprites() {
 	for (auto* actor: Main_Data::game_party->GetActors()) {
 		actor->SetBattleSprite(std::make_unique<Sprite_Actor>(actor));
+		actor->SetWeaponSprite(std::make_unique<Sprite_Weapon>(actor));
 	}
 }
 
@@ -349,6 +351,10 @@ void Scene_Battle_Rpg2k3::ResetAllBattlerZ() {
 		if (sprite) {
 			sprite->ResetZ();
 			sprite->DetectStateChange();
+		}
+		auto* weapon = actor->GetWeaponSprite();
+		if (weapon) {
+			weapon->ResetZ();
 		}
 	}
 }
@@ -1631,6 +1637,17 @@ Scene_Battle_Rpg2k3::SceneActionReturn Scene_Battle_Rpg2k3::ProcessSceneActionVi
 	};
 
 	if (scene_action_substate == eCBAInit) {
+		for (auto* actor: Main_Data::game_party->GetActors()) {
+			auto* sprite = actor->GetActorBattleSprite();
+			if (actor->Exists() && sprite) {
+				sprite->SetNormalAttacking(false);
+				auto* weapon = actor->GetWeaponSprite();
+				if (weapon) {
+					weapon->StopAttack();
+				}
+			}
+		}
+
 		if (cba_action != nullptr && cba_direction_back) {
 			CBAInit();
 			SetSceneActionSubState(eCBAMove);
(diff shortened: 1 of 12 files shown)


Ghabry:
According to #1269 this implements "[x] Animation" is this correct?

trail3

rueter37:
Yes, the battle animation weapons are implemented by this PR as well. I have tested it with "Vampires Dawn 2".

Ghabry:
Also confirmed that this works flawless in Mario: The Seven Sages :+1:

Ghabry (commented):
Ingame this works perfect. I cannot find any issues with it.
But this still needs a code review :)

src/sprite_weapon.cpp
```diff
@@ -0,0 +1,161 @@
...
+	int z = battler->GetType();
+	z *= SCREEN_TARGET_HEIGHT * 2;
+	z += y;
+	z *= id_limit;
+	z += id_limit - battle_index;
+	z += Priority_Battler;
+
+	default_z = z;
+
+	SetZ(z);
```
Ghabry, line 167: One suggestion to reduce a bit code duplication:

Instead of ``Game_Actor*`` pass a ``Sprite_Battler*`` (the one that draws the Sprite of the Actor in battle) and store this ``Sprite_Battler`` in a variable.
You can still make the base class constructor happy with ``Sprite_Battler(sprite_actor->GetBattler(), sprite_actor->GetBattler()->GetId())``.

ResetZ can be now replaced with a call to ``sprite_actor->GetZ()`` and the stuff needed for ``CreateSprite`` and ``Draw`` can be also directly read from the ``sprite_actor``

src/sprite_actor.cpp
```diff
@@ -160,9 +160,10 @@ void Sprite_Actor::SetAnimationState(int state, LoopState loop) {
 		if (ext->animation_type == lcf::rpg::BattlerAnimationPose::AnimType_battle) {
 			do_not_draw = false;
 			SetBitmap(BitmapRef());
-			lcf::rpg::Animation* battle_anim = lcf::ReaderUtil::GetElement(lcf::Data::animations, ext->battle_animation_id);
+			if (animation_id == 0) animation_id = ext->battle_animation_id;
```
Ghabry, line 168: add {} and linebreaks/indent

src/sprite_actor.cpp
```diff
@@ -80,7 +80,7 @@ void Sprite_Actor::Update() {
 			// Is a battle charset animation
 
 			static const int frames[] = {0,1,2,1,0};
-			int frame = (battler->IsDefending() ? 0 : frames[cycle / 10]);
+			int frame = (battler->IsDefending() ? 0 : (normal_attacking ? std::min(2, cycle / 10) : frames[cycle / 10]));
```
Ghabry, line 83: Can you explain what normal_attacking is?

Ghabry (commented):
Imo this looks already pretty solid. Much better than the last time you PR'ed this :+1:

src/sprite_actor.cpp
rueter37: Fixed.

src/sprite_actor.cpp
rueter37: normal_attacking is set to true if the actor starts a normal attack. If this flag is set, the frame sequence is only played once and stays at the last frame until the actor finishes his attack.

src/sprite_weapon.cpp
rueter37: I'm probably missing something obvious, but the suggestion does not work for me. Moreover the sprite_battler class has not got the DisplayX, DisplayY and GetFlashColor functions which are used in the Draw function from sprite_weapon.

Ghabry: actually you can completely remove ``ResetZ`` because the entire implementation is in ``Sprite_Battler``.

Yes it lacks DisplayX, DisplayY and GetFlashColor but you can use sprite_actor->GetX, sprite_actor->GetY and sprite_actor->GetFlashEffect.

rueter37: The `ResetZ` function has been removed now from `Sprite_Weapon`.

src/sprite_weapon.cpp
```diff
@@ -80,13 +80,13 @@ void Sprite_Weapon::SetWeaponAnimation(int nweapon_animation_id) {
 }
 
 void Sprite_Weapon::StartAttack(bool secondary_weapon) {
-	cycle = 0;
+	auto* battler = GetBattler();
 
-	attacking = true;
+	SetZ(battler->GetBattleSprite()->GetZ() + ((secondary_weapon ^ battler->IsDirectionFlipped()) ? 1 : -1));
```
Ghabry, line 85: New question. I missed this during the review:
Can you replace this xor with a if-else? I always have a really hard time to understand what is happening when XOR is used.
But please do not use multiple layers of "ternary ifs" to solve this :)

src/scene_battle_rpg2k3.cpp
```diff
@@ -352,10 +352,6 @@ void Scene_Battle_Rpg2k3::ResetAllBattlerZ() {
 			sprite->ResetZ();
 			sprite->DetectStateChange();
 		}
-		auto* weapon = actor->GetWeaponSprite();
-		if (weapon) {
-			weapon->ResetZ();
-		}
```
Ghabry: uhm these 4 lines can stay. I do not think they are wrong and there is a ResetZ in the sprite class, so it will "just work".

rueter37: Well, these lines are no longer needed because `ResetZ` has been removed from `Sprite_Weapon`.

src/sprite_weapon.cpp
rueter37: Fixed (I hope the correct way).

src/scene_battle_rpg2k3.cpp
Ghabry: Yes but ResetZ is still on Sprite_Battler and weapon inherits from it so you can still call it. But if this does not cause any issues it can stay removed.

src/sprite_weapon.cpp
Ghabry: Looks okay

src/scene_battle_rpg2k3.cpp
rueter37: For me it works, because the Z-value gets set explicitly when `StartAttack` is called which makes the weapon visible until `StopAttack` is called. Outside from normal attacks the weapon is invisible so the Z-value does not matter then.

Ghabry (approved):
looks good from my side. Can be used as a base for ranged stuff

Ghabry:
No further feedback since 9 days. Will do a merge on Tuesday.

src/sprite_weapon.h
```diff
@@ -0,0 +1,71 @@
...
+#include "sprite_battler.h"
+#include "async_handler.h"
+
+class Game_Battler;
+class Game_Actor;
+
+/**
+ * Sprite_Weapon class, used for weapon battle sprites
+ */
+class Sprite_Weapon : public Sprite_Battler {
```
fmatthew5876, line 31: Why does this inherit from `Sprite_Battler` and not `Sprite`?

rueter37: Well, when I started implementing the CBA weapons I took `Sprite_Actor` as a base which inherits from `Sprite_Battler`.

Ghabry: Considering that you do not use ``ResetZ`` anymore this is a legitimate request.
Just change it to ``Sprite``. It should work. Can be squashed into commit "Remove ResetZ function from sprite_weapon "

rueter37: Did the requested changed now.

fmatthew5876 (approved):

Issue #2506, opened 2021-04-03
Wadanohara (vgperson) displaying error "can't set equipment" during the title screen


Fishaphrodite:
Please fill in the following fields before submitting an issue:
#### Player platform:
Linux (built from master 1 day ago I believe, maybe 2?)
#### Attach files (as a .zip archive or link them)
[Wadanohara_Set_equip_title_creen_issue.zip](https://github.com/EasyRPG/Player/files/6253106/Wadanohara_Set_equip_title_creen_issue.zip)

> Contains a screenshot of the issue, as well as the log (it seems to happen everytime the title sceen is accessed, and also when the game is closing, though that one won't be an issue).

> Log mentions a save, but it isn't required to reproduce (actually, the save was added afterwards just to try to go back from a currently running game back to the main menu).

#### Describe the issue in detail and how to reproduce it:
Open the translation by VGPerson of the game "Wadanohara and the great blue sea" (the Data folder is the game itself), and those error messages will pop up as the game tries to go to the title screen.

The issue isn't game breaking it seems so far, but it look wrong. (I didn't try to replay the game to see if those could be equiped or not during a regular playthrough).

Ghabry:
This looks like a game bug.
Needs a check in RPG_RT maybe it simply ignores it than the warning can be changed from warning to debug.

Ghabry:
This is caused by these lines here:

https://github.com/EasyRPG/Player/blob/c7b5a323bfb3f610fd1ab9cf01fecb22e365f016/src/game_actor.cpp#L89-L93

Wadanohara has invalid items in the actor equipment (out of range). ``SetEquipment`` shows a warning when the item is out of range. The Actor constructor needs a check and should debug log this.

rueter37:
In RPG_RT the equipment slot is left empty (set to 0) if an out of range item is set in the initial equipment, so we have nothing game breaking here. We can change [line 300 in game_actor.cpp](https://github.com/EasyRPG/Player/blob/c7b5a323bfb3f610fd1ab9cf01fecb22e365f016/src/game_actor.cpp#L300) to `Output::Debug("SetEquipment: Can't equip item with invalid ID {}", new_item_id);` to make the message show only in the console.

Ghabry:
What does RPG_RT do when you use set equipment event command with something invalid? They all use the same function

rueter37:
[...]

PR #2511, opened 2021-04-18
Remove out of range items from equipment on actor initialization

rueter37:
This PR removes out of range items from actor equipment upon initialization of the actor.

Fixes: #2506

### DIFF
--- a/src/game_actor.cpp
+++ b/src/game_actor.cpp
@@ -86,11 +86,20 @@ Game_Actor::Game_Actor(int actor_id) {
 	SetHp(GetMaxHp());
 	SetSp(GetMaxSp());
 
-	SetEquipment(1, dbActor->initial_equipment.weapon_id);
-	SetEquipment(2, dbActor->initial_equipment.shield_id);
-	SetEquipment(3, dbActor->initial_equipment.armor_id);
-	SetEquipment(4, dbActor->initial_equipment.helmet_id);
-	SetEquipment(5, dbActor->initial_equipment.accessory_id);
+	// Remove items that do not exist in the database anymore
+	std::array<int, 5> ids = {{
+		dbActor->initial_equipment.weapon_id,
+		dbActor->initial_equipment.shield_id,
+		dbActor->initial_equipment.armor_id,
+		dbActor->initial_equipment.helmet_id,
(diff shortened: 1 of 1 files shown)


src/game_actor.cpp
```diff
@@ -86,11 +86,21 @@ Game_Actor::Game_Actor(int actor_id) {
...
-	SetEquipment(5, dbActor->initial_equipment.accessory_id);
+	std::array<int, 5> ids = {{
+		dbActor->initial_equipment.weapon_id & 0xFFFF,
+		dbActor->initial_equipment.shield_id & 0xFFFF,
+		dbActor->initial_equipment.armor_id & 0xFFFF,
+		dbActor->initial_equipment.helmet_id & 0xFFFF,
+		dbActor->initial_equipment.accessory_id & 0xFFFF }};
+	std::replace_if(ids.begin(), ids.end(), [] (const int& item_id) {
+		return lcf::ReaderUtil::GetElement(lcf::Data::items, item_id) == nullptr;
+	}, 0);
```
Ghabry, line 97: This & 0xFFFF is not needed here. This was just because of the garbage data in battle test.

Also add a comment like "// Remove items that do not exist in the database anymore"

src/game_actor.cpp
```diff
@@ -86,11 +86,21 @@ Game_Actor::Game_Actor(int actor_id) {
...
+
+	SetEquipment(1, ids[0]);
+	SetEquipment(2, ids[1]);
+	SetEquipment(3, ids[2]);
+	SetEquipment(4, ids[3]);
+	SetEquipment(5, ids[4]);
```
Ghabry, line 103: And these 5 SetEquipment calls can be changed to a for-loop.

Ghabry:
This is exactly what I had in mind. thanks for the patch.

src/game_actor.cpp
rueter37: Fixed.

src/game_actor.cpp
rueter37: Changed.

Ghabry (approved):

fmatthew5876 (approved):

PR #2513, opened 2021-04-18
Fix #2509

### LINKED ISSUE #2509: Performance issues in gauge battle system type
There are numerous reports about Battle Animations, especially Battle Animations used as Battler sprites like VD2 does. They cause a significant drop in FPS or high CPU usage.

Should be investigated.

Ghabry:
When the cache is exhausted do not delete unreferenced assets that were used during the last 50ms.

The current behaviour (flush everything unreferenced when cache is full) was added to workaround out-of-memory issues on systems with limited RAM but this direct unload made some use cases like face rendering in the Battle scene with Gauge style really slow (face was deleted and reloaded from disk each frame).

I hope that 50ms is a good compromise here.

Fix #2509

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -101,8 +101,15 @@ namespace {
 				continue;
 			}
 
-			if (cache_size <= cache_limit && cur_ticks - it->second.last_access < 3s) {
-				// Below memory limit and last access < 3s
+			auto last_access = cur_ticks - it->second.last_access;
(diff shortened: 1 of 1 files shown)


rueter37:
The fix works for me :+1:

fmatthew5876 (approved):

fdelapena (approved):

PR #2435, opened 2021-01-09
Emscripten bugfixes

Ghabry:
- Regression fix: Path was not made canonical anymore
- Replace regex with a replace function

PR #2514, opened 2021-04-27
URL encode +

krmbn0576:
Some servers (e.g. CloudFront) will not respond properly if the file name contains `+`, such as `a+b.png`.

Therefore, URL encoding is helpful. (See also #894)

### DIFF
--- a/src/async_handler.cpp
+++ b/src/async_handler.cpp
@@ -253,9 +253,10 @@ void FileRequestAsync::Start() {
 		request_path += path;
 	}
 
-	// URL encode % and #
+	// URL encode % and # and +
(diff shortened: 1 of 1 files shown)


fdelapena (approved):

Ghabry:
Jenkins: Test this please

Ghabry (approved):
not seeing a problem with this. Thanks

Ghabry:
uhm yeah this completely breaks emscripten because the "+" is a regex that matches the entire string...

I will add a string.replaceall function to ensure that this never happens again -_-

krmbn0576:
sorry for my wrong pull request, and thanks for fixing it correctly in #2435 👍 👍 👍

PR #2515, opened 2021-04-27
Battle 2k3: Add CBA single target ranged weapons

rueter37:
This PR adds CBA ranged weapons which target a single enemy and travel straight to it. The following checkboxes from #1269 are handled by it: "Use" checkbox, Dagger "fly in a straight line like a knife", Single Enemy and all Flying speed checkboxes.

### DIFF
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -2268,7 +2268,15 @@ Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleAction
 
 Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleActionCBARangedWeaponMove(Game_BattleAlgorithm::AlgorithmBase* action) {
 	auto* source = action->GetSource();
-	auto* target = action->GetOriginalSingleTarget();
+	Game_Battler* target = nullptr;
+	// The ranged weapon animation targets the original single target
+	// if the weapon has the "Attack All" flag set and the ranged attack
+	// range is set to "Single Enemy"
+	if (action->GetWeaponData()->attack_all) {
+		target = action->GetOriginalSingleTarget();
+	} else {
+		target = action->GetTarget();
+	}
 
 	if (cba_ranged_weapon_move_frame < cba_num_ranged_weapon_move_frames) {
 		cba_ranged_weapon_move_frame++;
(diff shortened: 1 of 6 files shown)


fdelapena (approved):

src/scene_battle_rpg2k3.cpp
```diff
@@ -1901,7 +1901,7 @@ Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleAction
 
 	if (source->GetType() == Game_Battler::Type_Ally) {
 		auto* sprite = static_cast<Game_Actor*>(source)->GetActorBattleSprite();
-		if (sprite && !sprite->IsIdling() && battle_action_state != BattleActionState_CBAMove && battle_action_state != BattleActionState_Animation) {
+		if (sprite && !sprite->IsIdling() && battle_action_state != BattleActionState_CBAMove && battle_action_state != BattleActionState_StartAnimation && battle_action_state != BattleActionState_CBARangedWeaponInit && battle_action_state != BattleActionState_CBARangedWeaponMove && battle_action_state != BattleActionState_Animation) {
```
Ghabry, line 1904: This very long ``battle_action_state`` condition could be changed into a switch-case:

```
switch (battle_action_state) {
  case BattleActionState_CBAMove:
  case BattleActionState_StartAnimation:
   etc etc etc...
    return BattleActionReturn::eWait;
  default:;
```

Ghabry:
I would assume that the "bow" option is just some kind of sinus curve? Or is there also a behaviour change?

rueter37:
If you mean the boomerang path type with "bow" option, then unfortunately it is very difficult (see https://github.com/easyrpg/player/issues/1269#issuecomment-826091681).

src/scene_battle_rpg2k3.cpp
rueter37: Did the change now.

fmatthew5876 (approved):
Looks reasonable to me. Needs to be tested by someone other than code author.

Ghabry (commented):
I found one minor problem:

When the attack count is set to x2 or x3 and the first enemy dies than the arrow is still targetting the dead enemy. in RPG_RT the arrow targets the next enemy.

rueter37:
Fixed the targeting of the ranged weapon animation now. The weapon needs the "Attack All" flag set to make the ranged weapon animation target the original single target. Otherwise the current target is targeted by the ranged weapon animation. This applies to ranged weapons with the attack range set to "Single Enemy".

Ghabry (approved):
Works perfect now. Even better than in RPG_RT ;).

PR #2516, opened 2021-05-02
Fix make distcheck

### LINKED ISSUE #2489: Reminder: Add missing tests to Makefile.am and fix make dist
Many tests are missing.

(after battle is merged)

Ghabry:
Add all tests and missing files to Makefile.am

Fix make distcheck

The folders in tests/ are moved into a subfolder "assets" because the Makefile gets confused when a folder and a filename are the same (e.g. platform). So this solution was the easiest ^^. Because I updated ``EP_TEST_PATH`` to reflect this there are 0 source code changes needed for this.

Fix #2489


Ghabry:
Tested also CMake build of Linux and libretro with this and it works.

An open problem I do not really want to solve is about libretro:
This will only build when the submodule was cloned beforehand. Otherwise the files are missing :/.

carstene1ns (approved):
Looks good to me.

carstene1ns:
> An open problem I do not really want to solve is about libretro:
> This will only build when the submodule was cloned beforehand. Otherwise the files are missing :/.

This is really a case fixed by RTFM I guess. The readme and build output mention this.

fdelapena (approved):

Issue #2519, opened 2021-05-04
Yume2kki: Wallpaper selection looks different


purupurupuruin:
#### Name of the game:
Yume2kki (Version 0.116)
(Download is at https://yume2kki.fandom.com/wiki/Downloads)

#### Player platform:
Linux

#### Attach files (as a .zip archive or link them)

Wallpaper selection in RPG_RT:
![Screenshot_2021-05-03_18-22-12](https://user-images.githubusercontent.com/64451846/116953064-25aeda80-ac41-11eb-816a-dc6a806f12b8.png)

<details>
  <summary>Wallpaper selection in EasyRPG Player (0.6.2.3):</summary>
  
![Screenshot_2021-05-03_18-47-36](https://user-images.githubusercontent.com/64451846/116953100-3fe8b880-ac41-11eb-9c2d-52bac4912ae5.png)
![Screenshot_2021-05-03_18-49-38](https://user-images.githubusercontent.com/64451846/116953104-41b27c00-ac41-11eb-85a6-143588b2f7ba.png)
</details>

Wallpaper selection in EasyRPG Player (Continuous build downloaded on May 3rd 2021):
![Screenshot_2021-05-03_19-39-39](https://user-images.githubusercontent.com/64451846/116955557-c30d0d00-ac47-11eb-8f0f-f4c4c4ccaab3.png)

#### Describe the issue in detail and how to reproduce it:
~~In EasyRPG Player the wallpaper selection screen of Yume2kki doesn't show the image preview in the background, in RPG_RT you can see a preview of the wallpaper before setting it.~~ (Only applies to Player version 0.6.2.3 and earlier)

In Yume2kki 0.116 the wallpaper selection screen is supposed to have a faded preview of the wallpaper you're about to select in the background, in EasyRPG Player the fade effect doesn't show.

For whatever reason the fade effect _does work_ in earlier versions of 2kki in Player. (The wallpaper selection UI is _slightly_ different in 0.116 compared to earlier versions so 2kki's developers must of made some change that isn't compatible with Player yet)

Also, the wallpapers have a white outline arround their thumbnail that doesn't show up in Player. (Regardless of 2kki's version)

To reproduce: Start a new game or load a save, head over to the computer, select tools, and just keep pressing Z until you reach the wallpaper selection screen. 

(Very sorry if this turns out to already be fixed in some branch! Or if it turns out this is a duplicate...)

[...]

PR #2523, opened 2021-05-05
ShowPicture: Support empty names

Ghabry:
Yume2kki 0.116 started to show a picture with no filename on the PC-wallpaper map.

No idea why they are doing this but when this is not filtered it unloads a black picture that is used for a background effect.

See #2519

a

@purupurupuruin

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -30,11 +30,8 @@
 #include "output.h"
 
 #ifdef GEKKO
-#  include <unistd.h>
 #  include <gccore.h>
 #  include <sys/iosupport.h>
-#elif defined(__SWITCH__)
-#  include <unistd.h>
 #elif defined(__ANDROID__)
 #  include <android/log.h>
 #elif defined(EMSCRIPTEN)
@@ -43,15 +40,13 @@
 
 #include "external/rang.hpp"
 
+#include "output.h"
 #include "filefinder.h"
 #include "input.h"
 #include "options.h"
-#include "output.h"
 #include "player.h"
 #include "bitmap.h"
-#include "main_data.h"
 #include "message_overlay.h"
-#include "utils.h"
 #include "font.h"
 #include "baseui.h"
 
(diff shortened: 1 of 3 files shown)


purupurupuruin:
I tried it but something weird happens, when I select a wallaper to preview it and cancel out it the background effect doesn't come back.

Ghabry:
You are right :(
So this is not the only issue.

fdelapena:
Not sure if related to the bug mentioned below and RPG_RT version used:

https://github.com/EasyRPG/Player/blob/602b229d311d31c32c897ddaed7482d3ff7bc0c1/src/game_pictures.cpp#L265-L269

Ghabry:
Looked into RPG_RT to figure this out:

It will apply all the Picture command data as usual but when there is no string the image buffer is not overwritten.

So this reuses the image buffer of whatever was in that slot before.

Ghabry:
There were multiple minor differences to RPG_RT that made this incompatible:

1. ShowPicture in RPG_RT will apply all the parameters but ignore the image buffer when the filename is empty
- Player hid the image and then did nothing
- I still hide the image for the "file has name case" due to Async loading (is made visible right after Async loading as before)

2. RPG_RT will render images with empty names
- Player aborted the rendering here

3. ErasePicture in RPG_RT will unload the image (and allocate a new image with 0 bytes, but this doesn't matter)
- Player only hid the sprite. This is actually bad because this means the reference is never lost until the picture is replaced -> waste of memory

With all these behaviours matching now the effect in the wallpaper select works.

src/output.cpp
```diff
@@ -341,7 +341,7 @@ void Output::ErrorStr(std::string const& err) {
 		std::cout << "EasyRPG Player will close now.";
 #if defined (GEKKO) || defined(__SWITCH__) || defined(_3DS)
 		// stdin is non-blocking
-		sleep(5);
+		Game_Clock::SleepFor(5s);
```
carstene1ns, line 344: I guess <unistd.h> might be removed for switch at least, wii likely.

Ghabry: ok, also tidyed the file a bit

purupurupuruin:
Tested it, works perfectly as far as I can tell!

carstene1ns (approved):

fdelapena (approved):

PR #2525, opened 2021-05-09
Enable sound on standalone game

carlsonsantana:
This change load settings on standalone games and consequently enable the sound of these apps.

### DIFF
--- a/builds/android/app/src/main/java/org/easyrpg/player/MainActivity.java
+++ b/builds/android/app/src/main/java/org/easyrpg/player/MainActivity.java
@@ -85,6 +85,7 @@ public class MainActivity extends Activity {
 
         if (standaloneMode) {
             // Launch the game
+            retrieveUserPreferences();
             GameInformation project = new GameInformation(dataDir + "/game");
             GameBrowserHelper.launchGame(this, project);
             finish();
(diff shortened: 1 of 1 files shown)


fdelapena:
jenkins CI: ok to test

fdelapena:
jenkins: ok to test

fdelapena (approved):
Thank you for the patch! (and for your [convert_rpggame_to_android](https://github.com/carlsonsantana/convert_rpggame_to_android) script! Standalone game improvements for reading from apk/obb without initial extraction are coming soon)

fmatthew5876 (approved):

carstene1ns (approved):

carstene1ns:
The actual problem lies here, because SettingsManager is not initialized:
https://github.com/EasyRPG/Player/blob/1fd6e7412cabdfa497574c56b6a075b0e04bf01c/builds/android/app/src/main/java/org/easyrpg/player/game_browser/GameBrowserHelper.java#L255-L257

Maybe we should revert the logic anyway.

Ghabry:
the entire SettingManager is a code-smell anyway. Should be a class instead than this would simply crash with a NullPointerException instead of using default data... (also no idea why there are no sane defaults).

PR #2528, opened 2021-05-13
Emscripten: Solution for savegame collisions

Ghabry:
Same solution as proposed in https://community.easyrpg.org/t/browser-version-saves-not-being-kept-when-leaving-and-returning-to-game-page/927  just a bit cleaner.

Usage (before the shell js file)

```js
<script>
Module = {
  EASYRPG_GAME = "UniqueGameId"
};
</script>
```

-------

Btw atsumaru loads the shell file and an "override script" with ``script defer``. Defer is nice because it enforces an execution order. Normal emscripten generator uses ``async`` so we cannot use this behaviour. :(

------

Also I got errors now because of missing dynCall. Maybe SDL update.


Ghabry:
Jenkins: test this please

fdelapena (approved):
Looks good to me, however it requires further setup apart of gencache. Though I'd prefer an automatic way to create an unique name (e.g.) based on server own game directory if possible, or set this from json cache file, and trying to detect existing savegame data in FS for backwards compatibility would be cool.

Ghabry:
server path would be obviously the easiest solution but this would break the saves for our own hosted web player (all versions share the same savegames, sometimes useful).

Will think about this again. Converting to a draft...

Ghabry:
This is possible now through the ``game`` arg in the easyrpg module

PR #2529, opened 2021-05-14
When Death "Ends after Battle" set the HP to 1 in case of dead.

### LINKED ISSUE #2521: Death with "Ends after Battle" is removed but HP stays at 0
#### Name of the game:
さいはてHOSPITAL

After a certain battle the game gives a game over. The battle is scripted, you can only lose. There is a "On Lose" handler, so the game continues on the map.

The problem is here that the game has "Ends after Battle" configured for the "Death" state. The state itself is removed but the HP stay at 0. They must be increased to 1.

Not sure where to put the fix for this, otherwise would have opened a PR already.

----

Reported via Mail. Thanks :)

Ghabry:
Fix #2521

### DIFF
--- a/src/game_battler.cpp
+++ b/src/game_battler.cpp
@@ -361,31 +361,56 @@ bool Game_Battler::AddState(int state_id, bool allow_battle_states) {
 	return was_added;
 }
 
-bool Game_Battler::RemoveState(int state_id, bool always_remove_battle_states) {
-	PermanentStates ps;
+template<typename F>
+bool RemoveStates(Game_Battler& battler, F&& f) {
+	auto prev_restriction = battler.GetSignificantRestriction();
 
-	auto prev_restriction = GetSignificantRestriction();
-	auto* state = lcf::ReaderUtil::GetElement(lcf::Data::states, state_id);
+	auto check_dead = [&]() {
+		// Cannot use IsDead here, as it checks for HP == 0
+		return State::Has(lcf::rpg::State::kDeathID, battler.GetStates());
+	};
 
-	if (!(always_remove_battle_states && state && state->type == lcf::rpg::State::Persistence_ends)) {
-		ps = GetPermanentStates();
-	}
-
-	auto was_removed = State::Remove(state_id, GetStates(), ps);
+	bool is_dead = check_dead();
+	bool was_removed = f();
 	if (was_removed) {
-		if (state_id == lcf::rpg::State::kDeathID) {
-			SetHp(1);
+		if (is_dead != check_dead()) {
+			// Was revived
+			battler.SetHp(1);
 		}
 
-		auto cur_restriction = GetSignificantRestriction();
-		if (GetBattleAlgorithm() != nullptr
-				&& GetBattleAlgorithm()->GetType() != Game_BattleAlgorithm::Type::None
-				&& cur_restriction != prev_restriction) {
-			SetBattleAlgorithm(std::make_shared<Game_BattleAlgorithm::None>(this));
+		auto cur_restriction = battler.GetSignificantRestriction();
+		if (battler.GetBattleAlgorithm() != nullptr
+			&& battler.GetBattleAlgorithm()->GetType() != Game_BattleAlgorithm::Type::None
+			&& cur_restriction != prev_restriction) {
+				battler.SetBattleAlgorithm(std::make_shared<Game_BattleAlgorithm::None>(&battler));
 		}
 	}
-
 	return was_removed;
+}
+
+void Game_Battler::RemoveBattleStates() {
+	RemoveStates(*this, [&]() {
+		return State::RemoveAllBattle(GetStates(), GetPermanentStates());
+	});
+}
+
+void Game_Battler::RemoveAllStates() {
+	RemoveStates(*this, [&]() {
+		return State::RemoveAll(GetStates(), GetPermanentStates());
+	});
+}
+
+bool Game_Battler::RemoveState(int state_id, bool always_remove_battle_states) {
+	return RemoveStates(*this, [&]() {
+		PermanentStates ps;
+
+		auto* state = lcf::ReaderUtil::GetElement(lcf::Data::states, state_id);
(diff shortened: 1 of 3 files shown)


src/game_battler.cpp
```diff
@@ -437,7 +437,12 @@ int Game_Battler::ApplyConditions() {
...
-	State::RemoveAllBattle(GetStates(), GetPermanentStates());
+	std::vector<bool> removed_states = State::RemoveAllBattle(GetStates(), GetPermanentStates());
+
+	if (!removed_states.empty() && removed_states[0]) {
+		// Death was removed
+		SetHp(1);
+	}
 }
 
 void Game_Battler::RemoveAllStates() {
```
fmatthew5876, line 448: Does this also need to set hp to 1 when death is removed?

src/game_battler.cpp
```diff
@@ -437,7 +437,12 @@ int Game_Battler::ApplyConditions() {
 }
 
 void Game_Battler::RemoveBattleStates() {
-	State::RemoveAllBattle(GetStates(), GetPermanentStates());
+	std::vector<bool> removed_states = State::RemoveAllBattle(GetStates(), GetPermanentStates());
+
+	if (!removed_states.empty() && removed_states[0]) {
+		// Death was removed
+		SetHp(1);
+	}
```
fmatthew5876, line 445: Instead of returning `vector<bool>` and allocating memory, you could just do:

1. was_dead = isDead()
2. RemoveAllBattle()
3. if (was_dead && !isDead()) { ... }

src/game_battler.cpp
Ghabry: This function is only called as part of FullHeal so doesn't really matter.

But will change it for consistency reasons.

fmatthew5876:
Can we try to clean up and streamline this a bit? There is state removal cleanup logic in `Game_Battler::Remove()` which needs to be duplicated in all 3 functions:
* Remove
* RemoveAll
* RemoveAllBattle

In addition to setting hp to 1, there is also logic to reset the battle algo for restrictions

You could do something like write a template private function which takes a lambda. The template function will do everything `RemoveState` does before and after actually removing the state or state(s), then do the cleanup logic after.


Basically take this funciton

```bool Game_Battler::RemoveState(int state_id, bool always_remove_battle_states) {
	PermanentStates ps;

	auto prev_restriction = GetSignificantRestriction();
	auto* state = lcf::ReaderUtil::GetElement(lcf::Data::states, state_id);

	if (!(always_remove_battle_states && state && state->type == lcf::rpg::State::Persistence_ends)) {
		ps = GetPermanentStates();
	}

	auto was_removed = State::Remove(state_id, GetStates(), ps);
	if (was_removed) {
		if (state_id == lcf::rpg::State::kDeathID) {
			SetHp(1);
		}

		auto cur_restriction = GetSignificantRestriction();
		if (GetBattleAlgorithm() != nullptr
				&& GetBattleAlgorithm()->GetType() != Game_BattleAlgorithm::Type::None
				&& cur_restriction != prev_restriction) {
			SetBattleAlgorithm(std::make_shared<Game_BattleAlgorithm::None>(this));
		}
	}

	return was_removed;
}```

And template out the `State::Remove()` piece. Then you can reuse the same code in all 3 cases.

**Comment by fmatthew5876** [comment]
A good example of this idiom is here: https://github.com/EasyRPG/Player/blob/master/src/game_actor.cpp#L425

Notice how the logic is reused many times in all the equipment functions

**Comment by Ghabry** [comment]
Ooooh right. You just spotted another bug: "Full Heal" does not invoke the action restriction code.

**Comment by Ghabry** [comment]
The streamlining opens the equipment state can of worms again :/

File: src/game_battler.cpp [hunk 3]
```diff
@@ -361,10 +361,36 @@ bool Game_Battler::AddState(int state_id, bool allow_battle_states) {
...
 
+void ForEachStateRemoved(Game_Battler& battler, Span<int16_t> removed_states) {
+	for (const auto& state_id: removed_states) {
+		auto prev_restriction = battler.GetSignificantRestriction();
+
+		if (state_id == lcf::rpg::State::kDeathID) {
+			battler.SetHp(1);
+		}
+
+		auto cur_restriction = battler.GetSignificantRestriction();
```
rueter37, line 372: I'm probably missing something but doesn't this always make `prev_restriction` and `cur_restriction` both equal?

src/game_battler.cpp
```diff
@@ -361,10 +361,36 @@ bool Game_Battler::AddState(int state_id, bool allow_battle_states) {
 	return was_added;
 }
 
+void ForEachStateRemoved(Game_Battler& battler, Span<int16_t> removed_states) {
+	for (const auto& state_id: removed_states) {
```
fmatthew5876, line 365: It's not efficient to loop over every state like this, and you don't have to.

Just do:
1. get prev restriction
2. get was_dead
3. apply all state removals (templated lambda which each main function passes in)
4. If was_dead and not is_dead -> Set hp to 1
5. get cur restriction
6. clear battle algo if restriciton changed

Ghabry: ah sorry. Now I get it. Also will not use ``IsDead`` but do a state check instead because IsDead checks for HP due to RPG_RT compat.

fmatthew5876: Good catch, yes state check is the right way to go. I would put a comment explaining why in case some developer tries to "fix" it later.

src/game_battler.cpp
fmatthew5876: Yes this is a bug

src/game_battler.cpp
Ghabry: Should be correct now.

fmatthew5876 (approved):
Looks good! Did anyone test the new code?

Ghabry:
the in-battle test is missing. Only tested the reported game.

Interesting would be how "Full Heal" behaves. I'm afraid that RPG_RT could have some more bugs hidden here.

rueter37:
The new code works for me. :+1:

fdelapena (approved):

PR #2530, opened 2021-05-14
Port some audio code from SDL2 Mixer to SDL2

Ghabry:
The frequency change code was missing.

Also changed the emscripten format to F32, this is the native format and maybe a bit faster.

### DIFF
--- a/src/audio_sdl.cpp
+++ b/src/audio_sdl.cpp
@@ -25,6 +25,10 @@
 #include <SDL_audio.h>
 #include <SDL_version.h>
 
+#ifdef EMSCRIPTEN
+#  include <emscripten.h>
+#endif
+
 #include "audio_sdl.h"
 #include "output.h"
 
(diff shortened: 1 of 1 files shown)


Ghabry:
okay this needs more testing, sounds a bit broken xD

src/audio_sdl.cpp
```diff
@@ -70,9 +74,27 @@ SdlAudio::SdlAudio() :
...
+#elif defined(EMSCRIPTEN)
+	// Get preferred sample rate from Browser (-> OS)
+	const int frequency = EM_ASM_INT_V({
+		var context;
+		try {
+			context = new AudioContext();
+		} catch (e) {
+			context = new webkitAudioContext();
+		}
+		return context.sampleRate;
```
fmatthew5876, line 89: Isn't this a memory leak?

Ghabry: This is JavaScript code so will be garbage collected.

carstene1ns:
I guess this needs a test on Wii. In audio playback this should not be really noticeable, however general responsiveness should be better.

Ghabry:
The F32 audio is obviously broken when using SDL (#2619) so I dropped this for now and only added the samplerate part

fdelapena (approved):

carstene1ns (approved):

PR #2533, opened 2021-05-16
Equipment scene: Use 4 digits for stat values in RPG Maker 2003

rueter37:
This PR reserves 4 digits for the stat values in the equipment scene in RPG Maker 2003 games. RPG Maker 2000 games reserves 3 digits for the stat values. Tested and confirmed with RPG_RT, this is probably meant to handle the display of the stat values if they go beyond 999 because of positive states.

### DIFF
--- a/src/window_equipstatus.cpp
+++ b/src/window_equipstatus.cpp
@@ -117,7 +117,7 @@ void Window_EquipStatus::DrawParameter(int cx, int cy, int type) {
 	contents->TextDraw(cx, cy, 1, name);
 
 	// Draw Value
-	cx += 10 * 6 + 6 * 3;
+	cx += (Player::IsRPG2k3() ? (8 * 6 + 6 * 4) : (10 * 6 + 6 * 3));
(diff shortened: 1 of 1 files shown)


fdelapena (approved):

Ghabry:
I just noticed another difference: The ">" are always rendered in RPG_RT. In Player they appear when you start an equipment selection.

rueter37:
As far as I know this was a design decision that the arrows only appear if you are changing equipment. That was years before I joined the EasyRPG project.

Ghabry:
Really? Uhm okay and I bet it was me. But yeah, it really makes sense that it is not rendered before the selection starts.

There are similiar UX changes in other parts. E.g. the "End Scene" (Do you want to quit) has "No" preselected instead of "Yes" because most UI guidelines agree that the non-destructive operation should be preselected.

Ghabry (approved):

fmatthew5876 (approved):

PR #2537, opened 2021-05-17
Battle 2k3: Add CBA multi target ranged weapons

rueter37:
This PR adds CBA ranged weapons which targets multiple enemies and travel straight to them. The following checkboxes from #1269 are handled by it: "Center" and "All Enemies" from the "Ranged Attack range" section. Not handled by this PR is the "Attack in Sequence" ranged attack range, because this one is plain insane.

### DIFF
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -2245,6 +2245,7 @@ Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleAction
 Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleActionCBARangedWeaponInit(Game_BattleAlgorithm::AlgorithmBase* action) {
 	auto* source = action->GetSource();
 	cba_ranged_weapon_move_frame = 0;
+	cba_ranged.clear();
 
 	if (source->GetType() == Game_Battler::Type_Ally) {
 		auto* actor = static_cast<Game_Actor*>(source);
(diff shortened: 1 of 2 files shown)


src/scene_battle_rpg2k3.cpp
```diff
@@ -2253,11 +2255,50 @@ Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleAction
...
+							if (action->GetWeaponData()->ranged_target == lcf::rpg::Item::Target_simultaneous) {
+								Main_Data::game_enemyparty->GetActiveBattlers(cba_ranged_targets);
+							}
+						}
+					}
+				} else {
+					cba_ranged_targets.push_back(action->GetTarget());
+				}
+
+				for (int i = 0; i < cba_ranged_targets.size(); i++) {
```
Ghabry, line 2294: You can use a for-each loop here.

fdelapena (approved):

Ghabry (changes requested):
Don't want to be rude, but I quickly made a patch instead of explaining it:
https://gist.github.com/Ghabry/8455a646b8bc7ac786b2b4d0aa7fb198

No need to credit me, just apply it with "patch -Np1 < name.patch" and squash it in your commit.

What I changed: Unified ``cba_ranged_weapons`` and ``cba_ranged_targets`` into a pair. This way they cannot go out-of-sync. Then converted everything to range-based for. Logic stayed the same.

I made some assumptions. It is possible that the assumptions are wrong:
When you use a CBA Ranged Weapon there is _always_ a target. I marked this _always_ with an assert.
Also there is _always_ a weapon, marked this with an assert.

rueter37:
@Ghabry thanks for the patch, it works for me and I have included it now. And no worries, I didn't see that as rude. I'm thankful for any help, even if it is 100% in code form.

Ghabry:
The only case broken (well, not broken like in RPG_RT :rofl:) for me is "Center" with "Attack all enemies" disabled.

This should result in a normal single attack. Instead it kinda does a single attack BUT the spear flies into the top-left corner instead of hitting enemies (also test with x2 and x3 attacks, still goes top-left)

For "Attack all" with "Attack all enemies" disabled it already does a normal single attack.

src/scene_battle_rpg2k3.cpp
```diff
@@ -2253,11 +2254,55 @@ Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleAction
...
+							// even if the real target of the animation is the center
+							cba_ranged.emplace_back(*action->GetTarget(), nullptr);
+						} else {
+							if (action->GetWeaponData()->ranged_target == lcf::rpg::Item::Target_simultaneous) {
+								std::vector<Game_Battler*> battlers;
+								Main_Data::game_enemyparty->GetActiveBattlers(battlers);
+								for (auto& b: battlers) {
+									cba_ranged.emplace_back(*b, nullptr);
+								}
+							}
```
Ghabry, line 2290: Just saw one more code problem:
You have a
```
if () {
} else {
  if () {
  } else {
}
```
here. This can be reduced to multiple ``else if`` instead.

src/scene_battle_rpg2k3.cpp
```diff
@@ -2268,32 +2313,34 @@ Scene_Battle_Rpg2k3::BattleActionReturn Scene_Battle_Rpg2k3::ProcessBattleAction
...
 		cba_ranged_weapon_move_frame++;
-		int offset_x = 0;
-		int offset_y = 0;
-		if (target) {
-			offset_x = target->GetBattlePosition().x - source->GetBattlePosition().x;
-			offset_y = target->GetBattlePosition().y - source->GetBattlePosition().y;
+		for (auto& it: cba_ranged) {
+			int offset_x = 0;
+			int offset_y = 0;
+			if (action->GetWeaponData()->ranged_target == lcf::rpg::Item::Target_center) {
```
Ghabry, line 2322: About the bug: My guess is that the problem is here. This must also check if you attack all enemies.

rueter37: Yes, this is true. I'm 99% sure that this is the cause.

src/scene_battle_rpg2k3.cpp

src/scene_battle_rpg2k3.cpp
rueter37: That was indeed the cause. It is fixed now.

Ghabry (approved):
Works and looks fine to me.
Imo you can squash the 2nd into the 1st commit here

rueter37:
Did the requested commit change now.

PR #2538, opened 2021-05-17
Battle 2k3: Implement CBA Afterimage

Ghabry:
~TODO: Figure out the transparency values.~

I made the API a bit more flexible: You can set how many afterimages you want :).

See #1269

Slightly exaggerated picture (manually set it to 10 images ;))

a-fs8

### DIFF
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -350,6 +350,7 @@ void Scene_Battle_Rpg2k3::ResetAllBattlerZ() {
 		auto* sprite = actor->GetActorBattleSprite();
 		if (sprite) {
 			sprite->ResetZ();
+			sprite->UpdatePosition();
 			sprite->DetectStateChange();
 		}
 	}
@@ -2765,6 +2766,10 @@ void Scene_Battle_Rpg2k3::CBAInit() {
 			}
 		}
 	}
+
+	if (cba_action->GetCBAMovement() != lcf::rpg::BattlerAnimationItemSkill::Movement_none) {
+		sprite->SetAfterimageAmount(cba_action->GetWeaponAnimationData()->after_image == lcf::rpg::BattlerAnimationItemSkill::AfterImage_add ? 3 : 0);
+	}
 }
 
 void Scene_Battle_Rpg2k3::CBAMove() {
@@ -2791,12 +2796,17 @@ void Scene_Battle_Rpg2k3::CBAMove() {
 		source->SetBattlePosition(Point(cba_start_pos.x + offset_x, cba_start_pos.y + offset_y));
 	}
 
-	if (cba_move_frame >= cba_num_move_frames && cba_direction_back) {
+	if (cba_move_frame >= cba_num_move_frames) {
 		auto* actor = static_cast<Game_Actor*>(source);
 		auto* sprite = actor->GetActorBattleSprite();
 		if (sprite) {
-			sprite->DoIdleAnimation();
+			sprite->DoAfterimageFade();
 		}
-		cba_action = nullptr;
+		if (cba_direction_back) {
+			if (sprite) {
+				sprite->DoIdleAnimation();
+			}
+			cba_action = nullptr;
+		}
 	}
 }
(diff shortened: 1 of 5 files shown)


Ghabry:
Figured out the transparency with a black-white-image and the 32 bit patch: Goes in steps of 64.

This is ready :)

fdelapena (approved):

src/scene_battle_rpg2k3.cpp
```diff
@@ -2765,6 +2766,10 @@ void Scene_Battle_Rpg2k3::CBAInit() {
 			}
 		}
 	}
+
+	if (cba_action->GetCBAMovement() != lcf::rpg::BattlerAnimationItemSkill::Movement_none) {
+		sprite->SetAfterimageAmount(cba_action->GetWeaponAnimationData()->after_image == lcf::rpg::BattlerAnimationItemSkill::AfterImage_add ? 3 : 0);
+	}
```
fmatthew5876, line 2772: When does this get reset to 0 after the movement is done?

What if you attack with after image and then..
* Escape?
* Row command?
* Change facing direction (but not movement) to attack an enemy on the other side?

Will those accidentally after-image as well?

Ghabry: The afterimage is removed when the battler is idling and all afterimage-frames have the same position.

The idling state is reached directly after the CBA movement is done, so 3 frames after the afterimages are gone. It is not possible to do any actions in this 3-frame window, as they are still part of the attack handling.

Though have to check if any event command can trigger it early. Could also set the amount of afterimages to "1" when the battle action starts. This way it is always reset.

rueter37:
Works for me. :+1:

fmatthew5876 (approved):

src/scene_battle_rpg2k3.cpp
Ghabry: Replaced the logic now with a fade-counter just for afterimage. This way it is ensured that the afterimage will always fade at the end of a CBA.

rueter37:
Did some more testing and unfortunately found a bug: As soon as an actor uses a skill which involves CBA movement, the game crashes. Normal attacks are not affected by this bug. The crash does not happen in current master.

rueter37:
@Ghabry: I have added a [commit](https://github.com/rueter37/player/tree/cba-afterimage) now which you can include in your PR now which should fix the afterimage issue with skills. There were more changes involved, because on normal attacks the afterimage setting is taken from the weapon animation data and on skills the afterimage setting is taken from the skill animation data.

Ghabry:
Oh, thanks!
I totally forgot that Skills can also have CBA. This also means I never tested any of your CBA code with skills. I hope it works ^^

src/scene_battle_rpg2k3.cpp
```diff
@@ -2768,7 +2768,7 @@ void Scene_Battle_Rpg2k3::CBAInit() {
 	}
 
 	if (cba_action->GetCBAMovement() != lcf::rpg::BattlerAnimationItemSkill::Movement_none) {
-		sprite->SetAfterimageAmount(cba_action->GetWeaponAnimationData()->after_image == lcf::rpg::BattlerAnimationItemSkill::AfterImage_add ? 3 : 0);
+		sprite->SetAfterimageAmount(cba_action->GetCBAAfterImage() == lcf::rpg::BattlerAnimationItemSkill::AfterImage_add ? 3 : 0);
```
Ghabry, line 2771: I just noticed: This is incorrect in the liblcf enum. It is an Afterimage, not an AfterImage.
Will submit a patch to liblcf to fix this. Good moment right now, will not break anythign except the PR :)

rueter37:
> I totally forgot that Skills can also have CBA. This also means I never tested any of your CBA code with skills. I hope it works ^^

CBA in skills are restricted to the options "Pose", "Movement" and "Afterimage". The first two are already implemented since #2503 so the skill CBA code works. Normal attacks have more CBA options available.

rueter37:
@Ghabry: I have updated my [commit](https://github.com/rueter37/player/tree/cba-afterimage) now according to the recent liblcf changes.

Ghabry:
@rueter37 thanks! I replaced yoru commit now.

PR #2544, opened 2021-05-23
Input: Rebind key used for fast forward x10 on desktop versions

rueter37:
This PR rebinds the fast forward x10 function from the `F` and `NumPlus` keys to the `G` key on desktop versions. This feature was suggested by rpgguest4188 and supported by Fran[e]1 on March 8, 2021 according to the [IRC chatlog](https://easyrpg.org/irc/log/2021/2021-03-08.html#msg-19:22:33). I'm supporting this feature request myself, because it is much more convenient not having to use two hands for the fast forward x10 function. One more advantage is that the change stops interferences with some RPG Maker 2003 games which use the `NumPlus` key for certain events. This change doesn't interfere with the current input scheme as the `G` key is currently unused.

### DIFF
--- a/src/input_buttons.h
+++ b/src/input_buttons.h
@@ -70,6 +70,7 @@ namespace Input {
 		SCROLL_UP,
 		SCROLL_DOWN,
 		FAST_FORWARD,
+		FAST_FORWARD_PLUS,
 		TOGGLE_FULLSCREEN,
 		TOGGLE_ZOOM,
 		BUTTON_COUNT
@@ -110,6 +111,7 @@ namespace Input {
 		"SCROLL_UP",
 		"SCROLL_DOWN",
 		"FAST_FORWARD",
+		"FAST_FORWARD_PLUS",
 		"TOGGLE_FULLSCREEN",
 		"TOGGLE_ZOOM",
 		"BUTTON_COUNT");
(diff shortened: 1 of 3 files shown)


fdelapena (approved):

Ghabry:
I'm not sure if adding another key is really an option for this. I would directly delete it again when the Scene_Config is ready (and put a fast forward multiplier config in there, just like Android).

How about: Instead of using PLUS for x10 you could use "F-Held + PLUS-Toggled" and "F-Held + MINUS-Toggled" for "Increase/Decrease multiplier by 1 in range [2, 10]". (toggled as in "pressed once", then you can configure this once on game start)

rueter37:
Well, I have use cases where I want fast forward x3 (RPG Maker 2003 battles) and where I want fast forward x10 (Traversing large empty or bland maps, skipping long cutscenes). Having two separate keys is more convenient for me. I personally don't like the idea to hit "F + Plus" and "F + Minus" seven times everytime to switch between fast forward x3 and x10.

Ghabry (approved):
This is okay for now. Can be revisited when we have a config scene.

carstene1ns (approved):
Not using this… *shrug*

Ghabry:
Not using this? That's the only way to make random encounters and walking simulators like Yume Nikki bearable xD.

Issue #2448, opened 2021-01-31
Add an --rtp-path flag override


fmatthew5876:
On test systems like WSL rtp detection fails. On other custom setups, one would want to just tell Player to use a specific RTP in a directory.

Add an `--rtp-path` command line flag to force the Player to use that path for it's RTP.

carstene1ns:
We have environment variables for this. Since this is linux-ish, should be easy to use in WSL.

fmatthew5876:
It would be good to add all the env vars  to `--help`

Ghabry:
The documentation is in the manpage.

Though some common env cars could be documented in help

fmatthew5876:
I get it, but if you just build a binary it would be convenient to have all docs built in. Another case is if a binary is shipped with a game with manages.

I see this as a convenience.

PR #2545, opened 2021-05-23
Add --rtp-path command line option

rueter37:
This PR adds a command line option which allows the user to add a custom RTP path. This path will be used with highest precedence.

Fixes: #2448

### DIFF
--- a/src/filefinder_rtp.cpp
+++ b/src/filefinder_rtp.cpp
@@ -31,7 +31,7 @@
 #   include <SDL_system.h>
 #endif
 
-FileFinder_RTP::FileFinder_RTP(bool no_rtp, bool no_rtp_warnings) {
+FileFinder_RTP::FileFinder_RTP(bool no_rtp, bool no_rtp_warnings, std::string rtp_path) {
 #ifdef EMSCRIPTEN
 	// No RTP support for emscripten at the moment.
 	disable_rtp = true;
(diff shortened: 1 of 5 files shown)


fdelapena (approved):

src/filefinder_rtp.cpp
```diff
@@ -131,6 +131,11 @@ FileFinder_RTP::FileFinder_RTP(bool no_rtp, bool no_rtp_warnings) {
 		env_paths.insert(env_paths.end(), tmp.begin(), tmp.end());
 	}
 
+	// If a custom RTP path is set, use this with highest precedence
+	if (!rtp_path.empty()) {
+		env_paths.insert(env_paths.begin(), rtp_path);
```
Ghabry, line 136: imo this should support Tokenizing (multiple paths split by ; or :) like the other env_paths. See the other cases a few lines above.

rueter37: Added tokenizing support.

Ghabry (approved):

Ghabry (approved):

Ghabry:
btw, the RTP loading also goes through the VFS handler, so with "--rtp-path 2000.zip" one can load a RTP from a ZIP archive. Tested and works :+1:

carstene1ns (approved):

PR #2549, opened 2021-05-31
Update IRC network

fdelapena:


.gitignore
```diff
@@ -110,3 +110,6 @@ local.properties
 
 # libretro
 easyrpg_libretro.*
+
+# GitHub
+/.github/
```
carstene1ns, line 115: I am against this change, I think this does not solve the problem since the file is pushed anyways on changes when you fast-forward.

fdelapena: Oops, thanks, this commit is from another PR, but needed to commit because it was forbidding me to open the PR. I'll edit the preferences on github instead and drop that commit.

carstene1ns (approved):

Issue #2236, opened 2020-05-30
Sound volume


Soeufans:
#### Describe the issue in detail and how to reproduce it:

Hello!

I was recently creating my menu for the game, and i notice, that the "Volume" for the sounds, works different from RM2k3

I creat an event with fading a sound to volume 100 to 0 and with RPG maker 2003 the sounds get offf gradually  like this:

Volume:
100 --- > 100
90   ---> 90
80   ---> 80
70   ---> 70
60   ---> 60
50   ---> 50
40   ---> 40
30   ---> 30
20   ---> 20
10   ---> 10
0     --->   1
In the volume lower than 60, theres a "noise" on them

With Player Easy RPG, the sounds get off, but it looks like it goes like this (or is my ear)
Editor   ---> EasyRPG
100      --->    100
90        --->      95
80        --->      85
70        --->      75
60        --->      65
50        --->       55
40        --->       45
30        --->       35
20        --->       25
10        --->       15
0          --->         0

not exactly as these numbers i expose, maybe its my ears, but theres some difference betwen them

Heres a Video about the sound
[Easy Video sounds.zip](https://github.com/EasyRPG/Player/files/4706444/Easy.Video.sounds.zip)

elsemieni:
I used to think there's a similar issue that I talked about different volume scales, but I couldn't find it. 
I will document it again.

Seems RPG_RT volume scale is more sort-of-exponential one, meanwhile in Player volume scale is linear.
There's a visual example of the attached video example for seeing more clearly.
![image](https://user-images.githubusercontent.com/1494687/83339937-a4b4c980-a2a0-11ea-8e88-35599efc1ccd.png)

As workaround hotfix for SuperCrystalHunter's Player repo I implemented some sort of lookup table for correcting volume values closer to RPG_RT ones in https://github.com/elsemieni/easyrpg-player-ne-sch/blob/master/src/game_system.h#L79 and https://github.com/elsemieni/easyrpg-player-ne-sch/blob/master/src/game_system.cpp#L194 . However, didn't PR this one beacuse the hotfix is ugly code-speaking.

rueter37:
The diagram looks like a cubic (power of 3) curve to me, so we could try
`volume = (int)std::ceil(100.0 * std::pow(volume / 100.0, 3.0));`

elsemieni:
[...]

PR #2551, opened 2021-06-01
Change sound volume scaling

rueter37:
This PR changes the volume scaling method of the sound and music effects from linear to a [logarithmic scale](https://github.com/easyrpg/player/issues/2236#issuecomment-803439029). This aims to fix some sound volume issues reported in #2236.

Related: #2236

### DIFF
--- a/builds/cmake/Modules/FindSDL2.cmake
+++ b/builds/cmake/Modules/FindSDL2.cmake
@@ -250,12 +250,13 @@ if(SDL2_FOUND)
 			find_library(AUDIOTOOLBOX AudioToolbox)
 			find_library(AUDIOUNIT AudioUnit)
 			find_library(METAL Metal)
+			find_library(GAMECONTROLLER GameController)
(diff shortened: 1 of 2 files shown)


fdelapena (approved):

Ghabry (approved):

carstene1ns (approved):
I think we should make this an Utils::AdjustVolume or something (DRY).
Maybe even local to game_system.

Ghabry:
in that case would prefer a function (in an anonymous namespace in the cpp) in Game_System. Could take the volume by reference and alter it.

rueter37:
Sorry, I didn't know that I was supposed to make further changes. I hope I added the function the right way.

Ghabry:
hmpf and of course this fails to compile on the 3DS because int32_t is not int there 😅

Ghabry:
oh I see now another flaw: This adjusted volume is also written in certain chunks. This shouldn't happen. Will also fix this

PR #2557, opened 2021-06-10
Fix 3ds and Vita build. Fix emscripten regression. Do not overwrite volume chunks.

Ghabry:
totally forgot that the volume is also part of the save data, making our saves incompatible with RPG_RT.

That PR fixes this.


rueter37:
The volume chunk fix works for me. :+1:

PR #2560, opened 2021-06-11
Android: Support launching of game.zip directly from the APK

Ghabry:
Save folder is now inside the data dir.

~~TODO: Also load Timidity stuff directly from the APK.~~

Usage: Put "game.zip" in the assets directory and it will launch on startup.

------

CC @carlsonsantana because you manage the scripts. FYI


Ghabry:
Video demo:

https://user-images.githubusercontent.com/1331889/121710674-21949900-cada-11eb-98ed-164099a6a890.mp4

Ghabry:
I will keep this PR on hold until the Storage Access Framework is in. Otherwise too much chaos with rebasing later and will likely need the SAF for something still...

Ghabry:
Standalone mode is now fixed and rebased.

The pending soundfont issue is also resolved automatically because it uses easyrpg.soundfont from the game directory.

Only problem that doesn't look solvable in an easy way (so I postbone this for now) is the bug report feature: I cannot use SAF here because standalone does not need SAF-stuff but I cannot attach savegames inside /data/data because that folder is not readable :/.


Also reenabled RTP by default as this is fast now and fixed that input crash.

Ghabry:
Added "asset delivery" support as this is needed for apps > 150 MB.

There are now two ways to do standalone mode:

If the game is part of the APK it must be in ``android/app/src/main/assets/game.zip`` and for the "Asset delivery" (only usable for Play Store!) it must be in ``android/assets/src/main/assets/game.zip``.

This means for publishing outside of the store there are two steps needed because app bundles are useless outside of it.

Ghabry:
Play Asset Delivery confirmed working by user mokodeoz, thanks

fdelapena (approved):

builds/android/app/src/main/java/org/easyrpg/player/InitActivity.java
```diff
@@ -95,36 +95,22 @@ public void prepareData() {
...
-            }
+            gameDir = "apk://game.zip";
         }
 
         if (standaloneMode) {
-            // Launch the game
-            DocumentFile gameFolder = DocumentFile.fromFile(new File(dataDir + "/game"));
-            Game project = new Game(gameFolder);
+            // Launch the game: FIXME
+            String saveDir = getApplication().getApplicationInfo().dataDir + "/Save";
```
Ghabry, line 110: To be tested for a follow up PR: To make the savegames available putting them in ``/sdcard/Android/data/[APP]`` (``getExternalFilesDirs``) would be better.

Then you can access them via ADB. At least for now... until Google fully restricts this m(

carstene1ns (approved):
Looks good to me.

PR #2564, opened 2021-06-24
Add custom game settings

rueter37:
Depends on [liblcf#421](https://github.com/easyrpg/liblcf/pull/421)

This PR adds the following game settings which can be set in the game database:

- BattleCommands: Default ATB mode
- BattleCommands: Enable in-battle row command
- BattleCommands: Sequential order battling
- Skills: Ignore reflect state
- Skills: State infliction hit rate
- Skills: Attribute shift hit rate
- Skills: Ignore physical rate restriction by states
- Skills: Ignore magical rate restriction by states
- Items: Maximum item count in inventory
- System: Alternative EXP curve
- System: Battle options
- System: Maximum actor HP value
- System: Maximum enemy HP value
- System: Maximum damage value
- System: Maximum EXP value
- System: Maximum level
- System: Enable stat absorbing
- System: Maximum number of savegame slots
- System: Default maximum item count in inventory
- System: Minimum variable value
- System: Maximum variable value
- System: Maximum actor SP value
- System: Maximum enemy SP value
- System: Maximum base stat value
- System: Maximum battle stat value

### DIFF
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -1149,8 +1149,9 @@ std::string Game_BattleAlgorithm::Skill::GetFailureMessage() const {
 }
 
 bool Game_BattleAlgorithm::Skill::IsReflected(const Game_Battler& target) const {
-	// Skills invoked by items ignore reflect
-	if (item) {
+	// Skills invoked by items and skills with the attribute
+        // "easyrpg_ignore_reflect" set ignore reflect
+	if (item || skill.easyrpg_ignore_reflect) {
 		return false;
 	}
 	return IsTargetValid(target) && target.HasReflectState() && target.GetType() != GetSource()->GetType();
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -51,7 +51,7 @@
 #include "spriteset_battle.h"
 
 static inline int MaxDamageValue() {
-	return Player::IsRPG2k() ? 999 : 9999;
+	return lcf::Data::system.easyrpg_max_damage == -1 ? (Player::IsRPG2k() ? 999 : 9999) : lcf::Data::system.easyrpg_max_damage;
 }
 
 Game_BattleAlgorithm::AlgorithmBase::AlgorithmBase(Type ty, Game_Battler* source, Game_Battler* target) :
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -1004,6 +1004,7 @@ bool Game_BattleAlgorithm::Skill::vExecute() {
 
 	bool heals_states = IsPositive() ^ (Player::IsRPG2k3() && skill.reverse_state_effect);
 	bool affected_death = false;
+	int to_hit_states = (skill.easyrpg_state_hit != -1 ? skill.easyrpg_state_hit : to_hit);
 	for (int i = 0; i < static_cast<int>(skill.state_effects.size()); i++) {
 		if (!skill.state_effects[i])
 			continue;
@@ -1017,7 +1018,7 @@ bool Game_BattleAlgorithm::Skill::vExecute() {
 			continue;
 		}
 
-		if (!Rand::PercentChance(to_hit)) {
+		if (!Rand::PercentChance(to_hit_states)) {
 			continue;
 		}
 
--- a/src/game_battlealgorithm.cpp
+++ b/src/game_battlealgorithm.cpp
@@ -973,6 +973,7 @@ bool Game_BattleAlgorithm::Skill::vExecute() {
 		const auto atk = target->CanChangeAtkModifier(effect);
 		if (atk != 0) {
 			SetAffectedAtk(atk);
+			if (lcf::Data::system.easyrpg_enable_stat_absorbing) SetIsAbsorbAtk(absorb);
 			SetIsSuccess();
 		}
 	}
@@ -980,6 +981,7 @@ bool Game_BattleAlgorithm::Skill::vExecute() {
 		const auto def = target->CanChangeDefModifier(effect);
 		if (def != 0) {
 			SetAffectedDef(def);
+			if (lcf::Data::system.easyrpg_enable_stat_absorbing) SetIsAbsorbDef(absorb);
 			SetIsSuccess();
 		}
 	}
@@ -987,6 +989,7 @@ bool Game_BattleAlgorithm::Skill::vExecute() {
 		const auto spi = target->CanChangeSpiModifier(effect);
 		if (spi != 0) {
 			SetAffectedSpi(spi);
+			if (lcf::Data::system.easyrpg_enable_stat_absorbing) SetIsAbsorbSpi(absorb);
 			SetIsSuccess();
 		}
 	}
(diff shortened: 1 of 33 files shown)


src/game_battlealgorithm.cpp
```diff
@@ -1149,8 +1149,9 @@ std::string Game_BattleAlgorithm::Skill::GetFailureMessage() const {
 }
 
 bool Game_BattleAlgorithm::Skill::IsReflected(const Game_Battler& target) const {
-	// Skills invoked by items ignore reflect
-	if (item) {
+	// Skills invoked by items and skills with the attribute
+        // "easyrpg_ignore_reflect" set ignore reflect
```
Ghabry, line 1159: > Add "easyrpg_ignore_reflect" attribute for skills 

Wrong spacing. Imo that entire comment is useless because it simply states what you can already see in the if. So delete it, thx.

src/scene_battle_rpg2k3.cpp
```diff
@@ -563,7 +563,7 @@ std::vector<std::string> Scene_Battle_Rpg2k3::GetBattleCommandNames(const Game_A
 			commands.push_back(ToString(cmd->name));
 		}
 	}
-	commands.push_back(ToString(lcf::Data::terms.row));
+	if (lcf::Data::battlecommands.easyrpg_enable_battle_row_command) commands.push_back(ToString(lcf::Data::terms.row));
```
Ghabry, line 567: Style: Wrap in ``{}``

src/scene_battle_rpg2k3.cpp
```diff
@@ -1226,6 +1231,15 @@ Scene_Battle_Rpg2k3::SceneActionReturn Scene_Battle_Rpg2k3::ProcessSceneActionAc
 			return SceneActionReturn::eWaitTillNextFrame;
 		}
 
+		if (lcf::Data::battlecommands.battle_type != lcf::rpg::BattleCommands::BattleType_traditional) {
+			if (Input::IsTriggered(Input::CANCEL)) {
+				SetActiveActor(-1);
+				Main_Data::game_system->SePlay(Main_Data::game_system->GetSystemSE(Main_Data::game_system->SFX_Cancel));
+				SetState(State_SelectOption);
+				return SceneActionReturn::eWaitTillNextFrame;
+			}
```
Ghabry, line 1252: > Add sequential order option in alternative and gauge battle types 

Is this like before where the Actor who is the first with a full gauge gets the turn?

rueter37: Yes.

Ghabry:
A commit to address all review comments is a good idea, otherwise you have too much to rebase & edit, is not worth it.

src/window_numberinput.cpp
```diff
@@ -84,8 +85,9 @@ int Window_NumberInput::GetMaxDigits() {
 }
 
 void Window_NumberInput::SetMaxDigits(int idigits_max) {
-	// Only accepts values between 1 and 6 (or 7) as RPG2K (or RPG2k3)
-	int top = Player::IsRPG2k() ? 6 : 7;
+	// At least 7 digits because of gold input in debug scene
+	// (free space and 6 digits for the gold value)
+	int top = std::max(7, digits_max);
```
Ghabry, line 90: >  Add settings for minimum and maximum variable values 
this looks wrong because digits_max is updated by the param idigits_max afterwards

Ghabry:
About these max stat base value things:
Imo it would make sense to have new function in Game_System for them but still delegate them through Game_Actor/Enemy for more flexibility (when we add new chunks to have per-character max values).
_I will append my own commit for this later as this is faster than adding 10 review comments_ :)

Ghabry:
> Add option to disable auto battling

Maybe this should be "overengineered" a bit: Instead of a bool-flag make it a number-vector like the BattleCommands or the Menu Items:
https://github.com/EasyRPG/Player/blob/master/src/scene_menu.cpp#L97

By default it is "0, 1, 2", representing Fight, Auto Battle, Escape and then allow removal of options. This makes it also easier to invent new options here :)

src/game_battlealgorithm.cpp
rueter37: Fixed.

src/scene_battle_rpg2k3.cpp

src/window_numberinput.cpp
rueter37: Fixed.

src/game_battlealgorithm.cpp
```diff
@@ -973,27 +973,31 @@ bool Game_BattleAlgorithm::Skill::vExecute() {
...
 			SetAffectedSpi(spi);
+			if (lcf::Data::system.easyrpg_enable_stat_absorbing) SetIsAbsorbSpi(absorb);
 			SetIsSuccess();
 		}
 	}
 	if (skill.affect_agility && Rand::PercentChance(to_hit)) {
 		const auto agi = target->CanChangeAgiModifier(effect);
 		if (agi != 0) {
 			SetAffectedAgi(agi);
+			if (lcf::Data::system.easyrpg_enable_stat_absorbing) SetIsAbsorbAgi(absorb);
```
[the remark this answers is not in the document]
rueter37: Fixed.

Ghabry:
btw, I'm sorry that this takes so long but be assured that this will land in 0.7.0

I'm 99% in favor of these improvements 👍

rueter37:
The idea with the number-vector for the battle options sounds interesting but I'm not able to implement it (tried it once). Too high a risk to break something if I implement this.

rueter37:
The number-vector for the battle options has been implemented now. System option 201 has been replaced now from "Enable auto battle" to "Battle options". The default setting is 0, 1, 2 (Battle, Auto battle and Escape) ([see this comment](https://github.com/EasyRPG/Player/pull/2564#issuecomment-879027751)).

Ghabry:
@rueter37  I rewrote parts of the code to reduce code duplication, please check if I broke something.

Also check the Sp Window in Battle 2k3, this needed a bit more changes because this depends now on the actor, before it was fixed...

rueter37:
@Ghabry Your changes work for me, I only had to fix the SP window, the correct window size wasn't applied if the maximum SP value was greater than 999. Moreover the "Enable stat absorbing" feature is a per skill option now.

Ghabry:
jenkins: test this please

src/scene_battle.cpp
```diff
@@ -159,9 +159,28 @@ void Scene_Battle::DrawBackground(Bitmap& dst) {
 
 void Scene_Battle::CreateUi() {
 	std::vector<std::string> commands;
-	commands.push_back(ToString(lcf::Data::terms.battle_fight));
-	commands.push_back(ToString(lcf::Data::terms.battle_auto));
-	commands.push_back(ToString(lcf::Data::terms.battle_escape));
+
+	for (std::vector<int16_t>::iterator it = lcf::Data::system.easyrpg_battle_options.begin();
```
Ghabry, line 163: this can be replaced with "for (auto option: easyrpg_battle_options)"

Same for "All all menu items" below

rueter37: Fixed.

Ghabry (approved):

fdelapena (approved):

PR #2571, opened 2021-07-11
Replace built-in "Fast WAV" decoder with dr_wav

Ghabry:
Also enabled it for all platforms. Should make WAV-loading everywhere faster because libsndfile scans the entire file once...

This also removes FastWAV by @Rinnegatamante . That decoder served us well but dr_wav is superior. :)

dr_wav states that it supports and has tests for:

  - Unsigned 8-bit PCM
  - Signed 12-bit PCM
  - Signed 16-bit PCM
  - Signed 24-bit PCM
  - Signed 32-bit PCM
  - IEEE 32-bit floating point
  - IEEE 64-bit floating point
  - A-law and u-law
  - Microsoft ADPCM
  - IMA ADPCM (DVI, format code 0x11)

This should be kinda everything we need?


Ghabry:
I must summon a autotools expert here @carstene1ns 

``WANT_DRWAV`` is not forwarded to Makefile.am for unknown reasons.

DrWav shall be enabled by default

I do:

```
AC_ARG_ENABLE([drwav],
	AS_HELP_STRING([--enable-drwav],[use internal WAV library dr_wav @<:@default=yes@:>@]), [], [enable_drwav="yes"])
AS_IF([test "x$enable_drwav" = "xyes"], <- set to "yes" if not provided
	AC_DEFINE([WANT_DRWAV],[1],[use internal WAV library dr_wav])]) <-- THIS WORKS PERFECT
AM_CONDITIONAL([WANT_DRWAV],[test "x$enable_drwav" = "xyes"]) <- THIS DOES NOT WORK (no WANT_DRWAV in Makefile.am)
```

Any idea why?

fdelapena:
> Any idea why?

Missing AC_SUBST maybe? For those with PKG_CHECK_MODULES is not required since a while because it does it implicitly.

carstene1ns:
configure complained `./configure: line 5845: ]: command not found`
And indeed there was a stray quotation mark, serving as shell brace.

src/decoder_drwav.cpp
```diff
@@ -0,0 +1,98 @@
...
+ * EasyRPG Player is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifdef WANT_DRWAV
```
carstene1ns, line 18: This does not work, as our autotools do not use CPPFLAGS, but config.h.

carstene1ns (approved):

fdelapena (approved):

PR #2572, opened 2021-07-12
Improve ZIP handling

### LINKED ISSUE #2562: ZIP: Improve encoding handling
Based on the interop discussion by apache commons below the encoding 437 should be always probed when there is no file with local encoding (used by Windows Zip function).

There should be also support for "general purpose bit 11", which indicates that the file is encoded in UTF-8.

------

Source: https://commons.apache.org/proper/commons-compress/zip.html

Recommendations for Interoperability

The optimal setting of flags depends on the archivers you expect as consumers/producers of the ZIP archives. Below are some test results which may be superseded with later versions of each tool.

- The java.util.zip package used by the jar executable or to read jars from your CLASSPATH reads and writes UTF-8 names, it doesn't set or recognize any flags or Unicode extra fields.
- Starting with Java7 java.util.zip writes UTF-8 by default and uses the language encoding flag. It is possible to specify a different encoding when reading/writing ZIPs via new constructors. The package now recognizes the language encoding flag when reading and ignores the Unicode extra fields.
- 7Zip writes CodePage 437 by default but uses UTF-8 and the language encoding flag when writing entries that cannot be encoded as CodePage 437 (similar to the zip task with fallbacktoUTF8 set to true). It recognizes the language encoding flag when reading and ignores the Unicode extra fields.
- WinZIP writes CodePage 437 and uses Unicode extra fields by default. It recognizes the Unicode extra field and the language encoding flag when reading.
- Windows' "compressed folder" feature doesn't recognize any flag or extra field and creates archives using the platforms default encoding - and expects archives to be in that encoding when reading them.
- InfoZIP based tools can recognize and write both, it is a compile time option and depends on the platform so your mileage may vary.
- PKWARE zip tools recognize both and prefer the language encoding flag. They create archives using CodePage 437 if possible and UTF-8 plus the language encoding flag for file names that cannot be encoded as CodePage 437.

Ghabry:
1. The language flag (Bit 11) is now respected for UTF-8
This is used by all zip programs except the Windows "compressed folder"

2. All files that are not ASCII and do not have Bit 11 are considered for encoding detection
The encoding detection will be only needed for Windows "compressed folder"

3. When the file is considered for encoding it is also stored in Codepage 437
7zip uses CP437 for Western European filenames

Fix #2562

----------

Umlauts in folder names won't work properly when 7zip created the archive. But this doesn't really matter as game directories do not contain umlauts.

When an upper folder contains umlauts (e.g. the Game directory itself "Unterwegs in Düsterburg") this doesn't matter.

### DIFF
--- a/src/filesystem_zip.cpp
+++ b/src/filesystem_zip.cpp
@@ -61,10 +61,11 @@ ZipFilesystem::ZipFilesystem(std::string base_path, FilesystemView parent_fs, St
 	uint32_t central_directory_size = 0;
 	uint32_t central_directory_offset = 0;
 
-	ZipEntry entry;
+	ZipEntry entry = {};
 	entry.is_directory = false;
-	std::vector<char> filepath_arr;
 	std::string filepath;
+	std::string filepath_cp437;
+	bool is_utf8;
 
 	encoding = ToString(enc);
 	if (FindCentralDirectory(zipfile, central_directory_offset, central_directory_size, central_directory_entries)) {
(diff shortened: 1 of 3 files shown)


fdelapena (approved):

Ghabry:
The new datafile works, the CP437 filenames are found

carstene1ns (approved):
Looks good to me.

PR #2592, opened 2021-07-20
Some fixes

carstene1ns:
- fmmidi fallback
- sdl(2)_mixer/openal


.gitattributes
```diff
@@ -0,0 +1,26 @@
...
+
+tests/doctest.h -diff linguist-vendored
+builds/autoconf/m4/ax_*.m4 -diff linguist-vendored
+src/dr_wav.h -diff linguist-vendored
+src/midiprogram.h -diff linguist-vendored
+src/midisequencer.* -diff linguist-vendored
+src/midisynth.* -diff linguist-vendored
+src/picojson.h -diff linguist-vendored
+builds/android/app/src/main/java/org/libsdl/app/SDL*.java -diff linguist-vendored
+builds/android/app/src/main/java/org/libsdl/app/SDLActivity.java diff -linguist-vendored
```
fdelapena, line 20: So, is SDLActivity overriding the previous wildcard for making use of `diff` without the `-`?

fdelapena (approved):

.gitattributes
carstene1ns: Right.
> When more than one pattern matches the path, a later line overrides an earlier line. This overriding is done per attribute.
> see https://git-scm.com/docs/gitattributes


Also explanation of the linguist stuff: https://github.com/github/linguist/blob/master/docs/overrides.md#summary

Ghabry (approved):

PR #2594, opened 2021-07-22
Improve ALSA, workaround issues

Ghabry:
This PR workarounds a crash in Yume2kki by clamping values (this is incorrect but better than crashing).

Also it removes all the volume adjustment code. There are multiple issues in here, especially when combined with "Fade".

Another problem is that MIDI volume messages are already logarithmic, so they are adjusted twice making MIDI almost silent when volume != 100.

So the logic varies depending on whether it is MIDI or Sampled and we did not expect this when adding AdjustVolume... And seems nobody tested MIDI with volumes != 100 so this was completely missed in the Midi PR. (so also only 100 -> 0 fade was tested which worked as expected.)

Fixing this properly needs at least a day and I'm on vacation by next week so I cannot fix this properly now. But I also do not want this broken code for 1 month in the continuous build...

But you can decide what you prefer. Then I remove the last commit. I will fix this properly in September.

### DIFF
--- a/src/audio_decoder_midi.cpp
+++ b/src/audio_decoder_midi.cpp
@@ -155,8 +155,7 @@ int AudioDecoderMidi::GetVolume() const {
 		return static_cast<int>(volume * 100);
 	}
 
-	// Lie about the volume as this is handled by the Midi messages internally
-	return 100;
+	return static_cast<int>(volume * 100);
 }
 
 void AudioDecoderMidi::SetVolume(int new_volume) {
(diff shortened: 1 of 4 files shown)


fdelapena (approved):

Ghabry:
As correctly observed in the chat the Adjust can stay for SE, as there are no side effects and no Midi there :).

Somehow testing Fade was completely forgotten. There is also a Fade In option for BGM. This was also broken for MidiOut because Update was never called... Quickly resolved this now.

Also the calculation of the Delta looks wrong for sample-based buffers. It always passes in 1/60s as the delta which makes 0 sense considering this runs on an extra thread. Must be some really old legacy code...

carstene1ns (approved):
Looks good to me.

Ghabry:
@krmbn0576 
except for the Drums which need Fluidsynth this PR should have fixed the worst audio issues you found

PR #2595, opened 2021-07-22
libxmp: Fix VFS support

### LINKED ISSUE #2570: libxmp: Fix VFS support (possible with 4.5.0)
libxmp was recently updated to 4.5.0.

This contains the new function ``xmp_test_module_from_memory`` which is essential for proper VFS support 👍

Ghabry:
Requires 4.5.0 because this is the first version that supports callbacks.

Fix #2570

-------

Contrary to WildMidi supporting <4.5.0 makes zero sense here:

For WildMidi older versions only lose wildmidi.cfg VFS support but loading Midi already worked.

Here without VFS it is not even possible to decider whether a module is supported. Also almost no game uses Tracker Files so not having this enabled on all distributions does not really hurt the game support.


fdelapena (approved):

carstene1ns (approved):
Looks good to me.

PR #2599, opened 2021-07-27
WIP: Fix translation loading

Ghabry:
Is a WIP because there is another bug in Filesystem "Subtree" which is only exposed in translation.cpp:84

```cpp
translation_root_fs = fs.Subtree(TRDIR_NAME);
```

That is the only case in the entire code where a non-properly cased path is passed to subtree. This means it crashes in an assert when the folder is not named "languages" (lowercase).


Ghabry:
Also fixed asset loading and made the parser more error tolerant (ported over from lcftrans).

ChangeHeroName and ChangeHeroTitle event commands can be now translated

PR #2608, opened 2021-07-31
Colorize terminal output, if possible

carstene1ns:
Also, allow the browser to filter/colorize the console log.

Needs testing on:
- [x] Windows (debug build opens console)
- [x] Web
- [x] Linux
- ~~Wii (usbgecko, may be emulated by dolphin)~~
- ~~3DS (debug build or with 3dslink)~~
- ~~Switch (nxlink)~~

Obligatory screenshot:

### DIFF
--- a/resources/unix/bash-completion/easyrpg-player
+++ b/resources/unix/bash-completion/easyrpg-player
@@ -1,5 +1,5 @@
 # bash completion script for easyrpg-player
-# (c) carstene1ns <dev f4ke de> 2016-2017
+# (c) carstene1ns <dev f4ke de> 2016-2021
 # available under the MIT license
 
 _easyrpg-player ()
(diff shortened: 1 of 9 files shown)


carstene1ns:
test this please.

Ghabry:
Btw fmtlib also supports colours but it lacks all the Pty etc detection and only uses ansi escape codes

fdelapena (approved):
Works for me, thanks!

The emscripten console log levels worked, too: https://easyrpg.org/play/pr2608/

resources/unix/bash-completion/easyrpg-player
```diff
@@ -72,7 +72,7 @@ _easyrpg-player ()
       return
       ;;
     # these have no argument and shall be used exclusively
-    -@(v|-version|h|-help))
+    -@(v|-version|h|-help|-no-log-color))
```
carstene1ns, line 75: This is obviously wrong... 😅
Will remove when I am at home later.

carstene1ns:
After further thinking about it, it does not make sense to have this on non-desktop platforms, so these just get the current behaviour.

fdelapena:
Captura de pantalla de 2021-08-03 12-57-35

It works as expected on Windows.

carstene1ns:
Thank you, I guess this is ready then 👍

src/player.cpp
```diff
@@ -150,6 +150,15 @@ void Player::Init(int argc, char *argv[]) {
 	// Must be called before the first call to Output
 	Graphics::Init();
 
+	// FIXME: actual command line parsing is too late for setting this,
+	// it should be refactored after release to not break things now
+	for (int i = 1; i < argc; ++i) {
+		if (strcmp(argv[i], "--no-log-color") == 0) {
+			Output::SetTermColor(false);
+			break;
```
Ghabry, line 158: this is really a general problem that should be addressed in a further PR.

This is also a problem for --help and --version: Before printing there stuff the console already contains "EasyRPG" and "CLI".

Ghabry (approved):
LGTM, Fedora will just be unhappy because we vendor too much stuff ;)

PR #2618, opened 2021-08-25
Audio improvements

Ghabry:
Fixed Fade In and added GetTicks to DrWav and XMP. Now "Get Midi Ticks" works for all types of audio.

-----

The looping code for ogg/opus is based on the Midi code, including "loops_to_end" to keep the BGM alive etc.

Not verified what AudioRemony does but this is the most consistent approach.

VX Ace and MV differ in how the tags are used: VX Ace version loops when LOOPLENGTH is reached.
MV uses WebAudio which plays the entire track once and then respects loop start and length.


Ghabry:
that the fluidsynth system_reset reenables some notes (fades them out) seems to be an expected side-effect and not a bug.

So I removed the call as all_sounds_off and controller reset are enough (hopefully all Midi playing nice :))

Ghabry:
@krmbn0576  can you test FluidSynth with this branch please?

krmbn0576 (approved):
@Ghabry 
I tested this and it sounded beautiful when I changed the MIDI. This seems to have solved all MIDI problems. Thank you very much! 😇

src/audio_decoder.cpp
```diff
@@ -170,38 +181,32 @@ void AudioDecoder::Update(std::chrono::microseconds delta) {
...
 int AudioDecoder::GetVolume() const {
-	return (int)volume;
+	return static_cast<int>(log_volume);
 }
 
 void AudioDecoder::SetVolume(int volume) {
-	this->volume = (double)volume;
+	this->volume = Utils::Clamp<double>(static_cast<double>(volume), 0.0, 100.0);
+	this->log_volume = AdjustVolume(this->volume);
 }
```
carstene1ns, line 197: I think this would read easier without `this->`.

```suggestion
void AudioDecoder::SetVolume(int new_volume) {
	volume = Utils::Clamp<double>(static_cast<double>(new_volume), 0.0, 100.0);
	log_volume = AdjustVolume(volume);
}
```

Sugestion feature merges somehow old and new code 🤦🏼

src/audio_generic.cpp
```diff
@@ -233,7 +233,7 @@ bool GenericAudio::PlayOnChannel(BgmChannel& chan, Filesystem_Stream::InputStrea
 	if (chan.decoder && chan.decoder->Open(std::move(filestream))) {
 		chan.decoder->SetPitch(pitch);
 		chan.decoder->SetFormat(output_format.frequency, output_format.format, output_format.channels);
-		chan.decoder->SetFade(0, volume, std::chrono::milliseconds(fadein));
+		chan.decoder->SetFade(volume, std::chrono::milliseconds(fadein));
```
carstene1ns, line 236: Is volume guaranteed to be 0 at this point?

Ghabry: good find. I forgot a "SetVolume" call here.

Ghabry:
Now everything should work.
The only kinda-problem I see (or hear xD) is that the BGM volume for Midi is slightly higher when using fluidsynth or FmMidi compared to the synth of the system (for that case it is perfect).
But this is not something I want to solve here, the problem can be anywhere...

PR #2623, opened 2021-08-28
Maniac Patch: Implement Load, Save, SaveInfo and Mouse input

Ghabry:
This implements the imo most useful commands of the Maniac Patch (except for String picture, this is also pretty cool!)

_Still have to do careful testing of this again. There shouldn't be regressions as the code is usually in ManiacPatch checks._

Discussion:

### Load

This does a Async Load (yields the interpreter), in Maniac it loads directly, this is incompatible with how our code works.

The parameter for checking whether the save exists is ignored, in RPG_RT loading a non-existant save simply crashes.

### Save

This does a Async save (Yields the interpreter), in Maniac this saves directly resulting in saving at an undefined state because there is stuff in temporaries. So yielding here is safer.

### Get Save Info

In Maniac this is implemented by setting the arguments (or not if the pic exists) and then doing a jump just before the 2k3E picture code that handles layers etc. (so save game pictures shown inherit the attributes of the current picture in the slot)
As we cannot really do this I added a function for retrieving the current Show Params instead.
There will be likely minor incompatibilities here, but when they do not do anything fancy with the picture it should work...

### Key Input Proc (Mouse)

Had to add "& 2"-checks where applicable as Maniac uses bit 1 of certain keys for this:

- Confirm: Mouse left
- Cancel: Mouse right
- Shift: Mouse middle
- Up: Wheel Up
- Down: Wheel Down

To reduce chaos in the code this will only work properly with RPG2k3E, KeyInput is already messy enough, so sorry, no fully Maniac mouse input in Player for 2k ;).

### New return values in Control Variables

Maniacs added tons of stuff here, also new ops like sin and pow. For now I only add the absolute minimum:

- Current Time (Maniac reuses the savegame time function here, I use stdlib time functions)
- Frame counter
- Maniac version: Returns the highest version available before the engine was replaced with this replace

### ValueOrVariable

Maniac Patch adds support for "Variable indirect", "Switch" and "Switch Indirect" to some event commands (and I guess with TPC and the new engine to even more). 

To make this easier for me I simply added this handling to ValueOrVariable which is our resolver for this indirection stuff.

Has the side-effect that all event commands that support indirection got these features now with a minimal code change.

This e.g. implements for **Call Event**:  "Added method to specify common event by variable number variable."

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -1642,14 +1642,25 @@ bool Game_Interpreter::CommandChangeLevel(lcf::rpg::EventCommand const& com) { /
 }
 
 int Game_Interpreter::ValueOrVariable(int mode, int val) {
-	switch (mode) {
-		case 0:
-			return val;
-		case 1:
-			return Main_Data::game_variables->Get(val);
-		default:
-			return -1;
+	if (mode == 0) {
+		return val;
+	} else if (mode == 1) {
+		return Main_Data::game_variables->Get(val);
+	} else if (Player::IsPatchManiac()) {
+		// Maniac Patch does not implement all modes for all commands
+		// For simplicity it is enabled for all here
+		if (mode == 2) {
+			// Variable indirect
+			return Main_Data::game_variables->Get(Main_Data::game_variables->Get(val));
+		} else if (mode == 3) {
+			// Switch (F = 0, T = 1)
+			return Main_Data::game_switches->Get(val) ? 1 : 0;
+		} else if (mode == 4) {
+			// Switch through Variable indirect
+			return Main_Data::game_switches->Get(Main_Data::game_variables->Get(val)) ? 1 : 0;
+		}
 	}
+	return -1;
 }
 
 bool Game_Interpreter::CommandChangeParameters(lcf::rpg::EventCommand const& com) { // Code 10430
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -21,6 +21,7 @@
 #include <iostream>
 #include <sstream>
 #include <cassert>
+#include <ctime>
 #include "game_interpreter.h"
 #include "audio.h"
 #include "dynrpg.h"
@@ -43,6 +44,7 @@
 #include "sprite_character.h"
 #include "scene_gameover.h"
 #include "scene_map.h"
+#include "scene_save.h"
 #include "scene.h"
 #include "game_clock.h"
 #include "input.h"
@@ -51,6 +53,7 @@
 #include "player.h"
 #include "util_macro.h"
 #include <lcf/reader_util.h>
+#include <lcf/lsd/reader.h>
 #include "game_battle.h"
 #include "utils.h"
 #include "transition.h"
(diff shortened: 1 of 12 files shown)


Ghabry:
Already found some bugs:

### Load

Forgot to remove the warning (upps)

Different behaviour:
ManiacPatch goes from black to fade-in here, this is not implemented

### Key Input Proc

Wheel Up and Right click not working (Copy paste error)

----

The rest appears to work. In the new, rewritten Maniac Patch GetSaveInfo seems to be kinda broken o_O

src/game_interpreter.cpp
```diff
@@ -1185,8 +1229,38 @@ bool Game_Interpreter::CommandControlVariables(lcf::rpg::EventCommand const& com
...
+					if (Player::IsPatchManiac()) {
+						std::time_t t = std::time(0);
+						std::tm* tm = std::localtime(&t);
+						value = (tm->tm_year - 100) * 10000 + (tm->tm_mon + 1) * 100 + tm->tm_mday;
+					}
+					break;
+				case 11:
+					// Current time (HHMMSS)
+					if (Player::IsPatchManiac()) {
+						std::time_t t = std::time(0);
```
carstene1ns, line 1246: minor nit: `std::time(nullptr);`

src/game_interpreter.cpp
```diff
@@ -2866,10 +2999,10 @@ bool Game_Interpreter::CommandKeyInputProc(lcf::rpg::EventCommand const& com) {
 		} else {
 			// For Rpg2k >=1.50
 			_keyinput.keys[Keys::eShift] = com.parameters[5] != 0;
-			_keyinput.keys[Keys::eDown] = param_size > 6 ? com.parameters[6] != 0 : false;
-			_keyinput.keys[Keys::eLeft] = param_size > 7 ? com.parameters[7] != 0 : false;
-			_keyinput.keys[Keys::eRight] = param_size > 8 ? com.parameters[8] != 0 : false;
-			_keyinput.keys[Keys::eUp] = param_size > 9 ? com.parameters[9] != 0 : false;
+			_keyinput.keys[Keys::eDown] = param_size > 6 && com.parameters[6] != 0;
```
carstene1ns, line 3002: Maybe rewrite this as `check_param(X)` and do the size check there as well?

src/game_interpreter.cpp
```diff
@@ -3456,39 +3610,132 @@ bool Game_Interpreter::CommandToggleFullscreen(lcf::rpg::EventCommand const& /*
...
+	if (!save) {
+		Output::Debug("ManiacGetSaveInfo: Save not found {}", save_number);
 		return true;
 	}
 
-	Output::Warning("Maniac Patch: Command Save not supported");
+	std::time_t t = lcf::LSD_Reader::ToUnixTimestamp(save->title.timestamp);
+	std::tm* tm = std::gmtime(&t);
+
+	// YYMMDD
```
carstene1ns, line 3640: Since this is repeated (return `YYMMDD`/`HHMMSS` as integer from time), might move to common function? (in utils?)

src/game_interpreter.cpp
Ghabry: found check_param vs check_key too confusing, added check_key where necessary instead

src/game_interpreter.cpp
Ghabry: replaced the calculation with ``strftime``, much simpler this way ;)

src/game_interpreter.cpp
```diff
@@ -2880,17 +3016,19 @@ bool Game_Interpreter::CommandKeyInputProc(lcf::rpg::EventCommand const& com) {
 				_keyinput.keys[Keys::eRight] = true;
 				_keyinput.keys[Keys::eUp] = true;
 			}
-			_keyinput.keys[Keys::eNumbers] = param_size > 5 ? com.parameters[5] != 0 : false;
-			_keyinput.keys[Keys::eOperators] = param_size > 6 ? com.parameters[6] != 0 : false;
-			_keyinput.time_variable = param_size > 7 ? com.parameters[7] : false;
-			_keyinput.timed = param_size > 8 ? com.parameters[8] != 0 : false;
+			_keyinput.keys[Keys::eNumbers] = check_key(5);
+			_keyinput.keys[Keys::eOperators] = check_key(6);
+			_keyinput.time_variable = param_size > 7 && com.parameters[7]; // not a bool
```
carstene1ns, line 3021: This looks wrong.

src/game_interpreter.cpp
```diff
@@ -1235,17 +1235,17 @@ bool Game_Interpreter::CommandControlVariables(lcf::rpg::EventCommand const& com
 				case 10:
 					// Current date (YYMMDD)
 					if (Player::IsPatchManiac()) {
-						std::time_t t = std::time(0);
+						std::time_t t = std::time(nullptr);
 						std::tm* tm = std::localtime(&t);
-						value = (tm->tm_year - 100) * 10000 + (tm->tm_mon + 1) * 100 + tm->tm_mday;
+						value = Main_Data::game_variables->Set(com.parameters[2], atoi(Utils::FormatDate(tm, Utils::DateFormat_YYMMDD).c_str()));
```
carstene1ns, line 1240: This sets the variable twice now?

src/game_interpreter.cpp
carstene1ns: Thought about it too, however needs ugly conversion this way.

src/game_interpreter.cpp
Ghabry: Was not sure how to handle this because this is the only argument that is not converted to a bool 😅

src/game_interpreter.cpp
Ghabry: copy paste error...

src/game_interpreter.cpp
carstene1ns: IMO just like before, but false will become 0 anyways:
`param_size > 7 ? com.parameters[7] : 0;`

Ghabry: Oooh now I see what you mean. Yeah this is wrong.

Ghabry: Now I see the problem: This was incorrectly transformed by clang-tidy because the expression is ``param_size > 7 ? com.parameters[7] : false;``. This does not break the code here but the ``false`` must be actually a ``0``...

carstene1ns: (what I said earlier) ;)

src/utils.h
```diff
@@ -327,6 +331,15 @@ namespace Utils {
...
 
+	/**
+	 * Formats a date.
+	 *
+	 * @param tm Time structure
+	 * @param format Format string (See strftime)
+	 * @return formatted date
+	 */
+	std::string FormatDate(const std::tm* tm, StringView format);
+
```
carstene1ns, line 342: Not sure if it is worth, but https://github.com/EasyRPG/Player/blob/10a3bbe2fcc32196765e7e1a034e5248cdf9736b/src/output.cpp#L83-L92 could use this too.

Ghabry: yeah I will change it.
btw fmtlib also has a nice time-formater but this needs fmtlib 6.0 and you know how far behind some distributions are :/

carstene1ns (approved):

fdelapena (approved):

PR #2632, opened 2021-09-03
Add support for negative terrain damage

rueter37:
This PR adds support for negative terrain damage. RPG_RT supports this and the RPG Maker 2003 documentation says that the
editor allows setting negative terrain damage values.

### DIFF
--- a/src/game_player.cpp
+++ b/src/game_player.cpp
@@ -610,16 +610,18 @@ bool Game_Player::Move(int dir) {
 	bool red_flash = false;
 
 	if (terrain) {
-		if (!terrain->on_damage_se || (terrain->on_damage_se && (terrain->damage > 0))) {
-			Main_Data::game_system->SePlay(terrain->footstep);
-		}
-		if (terrain->damage > 0) {
+		if (terrain->damage != 0) {
 			for (auto hero : Main_Data::game_party->GetActors()) {
-				if (!hero->PreventsTerrainDamage()) {
-					red_flash = true;
+				if (terrain->damage < 0 || !hero->PreventsTerrainDamage()) {
+					if (terrain->damage > 0) {
+						red_flash = true;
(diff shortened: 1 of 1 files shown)


fdelapena (approved):

carstene1ns (approved):

src/game_player.cpp
```diff
@@ -613,10 +613,12 @@ bool Game_Player::Move(int dir) {
 		if (!terrain->on_damage_se || (terrain->on_damage_se && (terrain->damage > 0))) {
 			Main_Data::game_system->SePlay(terrain->footstep);
 		}
-		if (terrain->damage > 0) {
+		if (terrain->damage != 0) {
 			for (auto hero : Main_Data::game_party->GetActors()) {
-				if (!hero->PreventsTerrainDamage()) {
-					red_flash = true;
+				if (terrain->damage < 0 || !hero->PreventsTerrainDamage()) {
```
Ghabry, line 618: did you verify that:

- "Preventing damage" does not prevent heal? (we are talking about RPG Maker 2003 here!)
- "Negative damage" does not flash the screen in a red color?

src/game_player.cpp
```diff
@@ -613,10 +613,12 @@ bool Game_Player::Move(int dir) {
 		if (!terrain->on_damage_se || (terrain->on_damage_se && (terrain->damage > 0))) {
 			Main_Data::game_system->SePlay(terrain->footstep);
```
Ghabry, line 614: this seems to match. Can you quickly check that the condition when the SE plays is correct?

rueter37:
@Ghabry I have updated the condition when the SE is played. The SE is played if `on_damage_se` is unset or if at least one party member takes terrain damage.

Ghabry (approved):

Issue #2574, opened 2021-07-12
Incorrect scroll speed in some windows


Ghabry:
see #2264

**Title screen, main menu screen, party/actor battle menus**
- Scrolling stops at end of menu. In RPG_RT, holding down arrow scrolls endlessly.
- Scrolling might be _slightly_ too slow.

**Equip screen**
![](https://i.imgur.com/kWS4l0y.png)
- Same as above
- Scrolling stops at end of equip slots. In RPG_RT, holding down arrow scrolls endlessly.

**Shop screen**
![](https://i.imgur.com/MgLoO2y.png)
- No smooth scrolling (not in other menus?)

rueter37:
Hasn't the endless scroll issue already been fixed? Or do I miss something?

Ghabry:
I do not understand "task lists". I added the issue but the task was still not flagged as "done" after merge -_-

@rueter37  is anything else of this already fixde?

rueter37:
@Ghabry Yes, the checkboxes 1 and 3 are already fixed by #2361. Only "smooth scrolling" (checkbox 4) is remaining.

PR #2634, opened 2021-09-04
Change menu cursor scrolling speed

rueter37:
This PR decreases the interval between two menu cursor scrolls to 4 frames and increases the initial delay to 23 frames.

This is a partial fix for #2574. Still missing is the "smooth scrolling" stuff in the shop window.

### DIFF
--- a/src/input.cpp
+++ b/src/input.cpp
@@ -35,14 +35,14 @@ namespace Input {
 	 * to be maintained pressed before being
 	 * repeated for fist time.
 	 */
-	constexpr int start_repeat_time = 20;
+	constexpr int start_repeat_time = 23;
 
 	/**
 	 * Repeat time (in frames) a key has to be
 	 * maintained pressed after the start repeat time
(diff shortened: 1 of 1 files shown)


fdelapena (approved):

rueter37:
Did more tests and found that a initial delay of 25 frames is too slow. But the old value (20 frames) is too fast, so I went with a initial delay of 22 frames. It feels right for me, but getting the absolutely correct value requires decompilation of RPG_RT.

Ghabry:
decompiling this is hard because this is not a part we ever investigated.

I did a 60 FPS recording instead and then counted this multiple times (framedrops...)

In most cases I got "23 frames" for the first and "4 frames" for the repeating.

rueter37:
@Ghabry I have changed the initial delay to 23 frames now. So everything should be right now.

carstene1ns (approved):

Ghabry (approved):

Issue #2589, opened 2021-07-19
Title BGM playback is incorrect


krmbn0576:
If you go to the load scene and press the cancel button to return to the title scene, the background music will continue. This is the desired behavior. However, if you start the game, go back to the title scene, then go to the load scene as before, and press the cancel button to go back to the title scene, the BGM will be replayed from the beginning. This is undesirable. This happens both before and after #2302.

PR #2637, opened 2021-09-05
Fix BGM play upon returning to the title screen from load screen

rueter37:
This PR makes the BGM no longer being restarted upon returning to the title screen from the load screen.

Fixes: #2589


fdelapena (approved):

Ghabry (approved):
this makes sense - restart_title_cache is never reset to false

PR #2640, opened 2021-09-06
Clear system graphic when clearing the cache

### LINKED ISSUE #2601: Android: Improper cleanup of system graphic
When launching a 2nd game it will show a warning that the System graphic of the previous game was not found.

This looks harmless though, the 2nd game shows the correct graphic on the title screen.

Nothing serious but should be fixed

Ghabry:
Fix #2601

Bet this was omitted before due to some sideeffects but it seems to work by now.

(Not tested on Android yet though)

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -511,6 +511,7 @@ void Cache::Clear() {
 
 	cache_tiles.clear();
 
+	system_name.clear();
 	system2_name.clear();
 }
 
--- a/src/graphics.cpp
+++ b/src/graphics.cpp
@@ -57,6 +57,9 @@ void Graphics::Quit() {
 	message_overlay.reset();
 
 	Cache::Clear();
+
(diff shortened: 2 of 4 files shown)


Ghabry:
okay this made it worse on Android: Now it crashes when switching the game -_-

Good that we have working Android C++ debugging by now :)

Ghabry:
Combined with the second commit it works but what worries me here is that I cannot figure out why it worked before o_O

Ghabry:
Assert was not..... Triggered.

CMake defines NDEBUG.

All PR builds since the CMake migration were broken because they triggered the assert and nobody noticed😂

fdelapena (approved):

carstene1ns:
> CMake defines NDEBUG.

Only in release mode. Pull request builds are in debug mode.

carstene1ns (approved):

PR #2643, opened 2021-09-10
Add small battle system

rueter37:
This PR adds the small battle system in RPG Maker 2003 battles. For UX reasons the menus in small windows show 4 menu items while RPG_RT only shows 3 menu items (too much unused space).

Fixes: #2575


src/scene_battle_rpg2k3.cpp
```diff
@@ -321,6 +321,46 @@ void Scene_Battle_Rpg2k3::CreateUi() {
 		SetupSystem2Graphics();
 	}
 
+	if (lcf::Data::battlecommands.window_size == lcf::rpg::BattleCommands::WindowSize_small) {
+		int height = 68;
+		int y = SCREEN_TARGET_HEIGHT - height;
+
```
Ghabry, line 327: Except for ``status_window`` this looks the same for everything.

Did not launch a compiler but something like this should work:

```cpp
for (auto* window: { options_window, command_window, skill_window, item_window, target_window }) {
  if (window) {
     // Height, Y, BorderY, SetContents, SetMenuItemHeight
  }
}
```

Ghabry: static typing is in the way here, lets use a lambda instead:

```cpp
auto small_window  = [&](auto* window) {
  if (window) {
     // Height, Y, BorderY, SetContents, SetMenuItemHeight
  }
};

small_window(options_window);
small_window(command_window);
...
```

rueter37: Updated.

Ghabry (commented):
This PR nicely shows that we have a huge problem with magic numbers in our window code :/

code-wise looks fine to me, just needs a test

fdelapena (approved):

rueter37:
This cannot be merged yet; I have encountered some weird stuff if you use the small battle system with the alternative battle type: The graphics are sometimes glitching and I had even once a crash occurring. Interestingly it only occurs in the alternative battle type, the traditional and gauge battle types are not affected. I don't even know why this is happening in the first place.

Ghabry:
Do you use Linux?
Best way to catch such stuff is Address sanitizer (-fsanitize=address).
Windows MSVC also has experimental support for ASAN by now.

rueter37:
@Ghabry Thank you for mentioning the address sanitizer. I was able to track down the cause for the memory corruption and fix it. This PR can be merged now after approval.

Ghabry:
Sorry for the delay. Reviewing it this weekend :)

Ghabry:
btw in the RPG_RT.exe I used the small window also showed 4 items. Maybe was fixed by BingShan in ManiacPatch.

PR #2648, opened 2021-09-12
Battle 2k3: Consider acting battler for certain page conditions

### LINKED ISSUE #2535: Grimm's Hollow: Swing meter appears at the wrong times in battle
#### Name of the game: Grimm's Hollow (https://ghosthum.itch.io/grimms-hollow)

[GrimmsHollowSave.zip](https://github.com/EasyRPG/Player/files/6489869/GrimmsHollowSave.zip)

Load the save and start the compat tutorial. Just always pick the normal attack, even when you are asked later to use a skill. The "swing meter" will randomly appear alot even when no attack is executed (while menuing).

Initially reported to the core: https://github.com/libretro/easyrpg-libretro/issues/53

Ghabry:
fmatthew missed something ;) (well partially, there was a FIXME in the code but only for ``command_actor``)

When evaluating conditions RPG Maker 2003 passes the current active battler. This battler is evaluated for the page conditions:

- Battle Command: checked battler == current_battler (NULL-check missing because it makes no sense here)
- Actor / Enemy turn: battler is NULL (battle start) or checked battler == current battler

Test cases:

- Battle Command: #2535 . A page has run condition "Hero uses Attack". When the enemy attacks this page was activated again. This is _wrong_. It is only activated when the hero acts. **Works now**
- Actor / Enemy turn: Create two pages, one with "Actor 1 Turn Count x1" and one with "Enemy 1 Turn Count x1" (= run page everytime)
  - When battle starts: Both pages run (good, thats the NULL case)
  - When Actor attacks: Before: Both pages run. **wrong**. Now: Only actor page runs. **Works**
  - When Enemy attacks: Before: Both pages run. **wrong**. Now: Only enemy page runs. **Works**

(The 3 affected conditions are 2k3 exclusive, so this one-line change in 2k is completely harmless.)

Fix #2535

### DIFF
--- a/src/game_interpreter_battle.cpp
+++ b/src/game_interpreter_battle.cpp
@@ -68,13 +68,21 @@ bool Game_Interpreter_Battle::AreConditionsMet(const lcf::rpg::TroopPageConditio
 	if (condition.flags.turn && !Game_Battle::CheckTurns(Game_Battle::GetTurn(), condition.turn_b, condition.turn_a))
 		return false;
 
-	if (condition.flags.turn_enemy &&
-		!Game_Battle::CheckTurns((*Main_Data::game_enemyparty)[condition.turn_enemy_id].GetBattleTurn(),	condition.turn_enemy_b, condition.turn_enemy_a))
-		return false;
+	if (condition.flags.turn_enemy) {
+		const auto& enemy = (*Main_Data::game_enemyparty)[condition.turn_enemy_id];
+		if (source && source != &enemy)
+			return false;
+		if (!Game_Battle::CheckTurns(enemy.GetBattleTurn(), condition.turn_enemy_b, condition.turn_enemy_a))
+			return false;
+	}
 
-	if (condition.flags.turn_actor &&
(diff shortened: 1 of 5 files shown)


fdelapena (approved):

fmatthew5876 (approved):

src/game_interpreter_battle.cpp
```diff
@@ -68,13 +68,22 @@ bool Game_Interpreter_Battle::AreConditionsMet(const lcf::rpg::TroopPageConditio
...
+	}
 
-	if (condition.flags.turn_actor &&
-		!Game_Battle::CheckTurns(Main_Data::game_actors->GetActor(condition.turn_actor_id)->GetBattleTurn(), condition.turn_actor_b, condition.turn_actor_a))
-		return false;
+	if (condition.flags.turn_actor) {
+		const auto* actor = Main_Data::game_actors->GetActor(condition.turn_actor_id);
+		if (source && source != actor)
+			return false;
+		if (!Game_Battle::CheckTurns(actor->GetBattleTurn(), condition.turn_actor_b, condition.turn_actor_a)) {
```
fdelapena, line 83: Nit: inconsistent curly braces style compared to other.

PR #2649, opened 2021-09-13
Maniac: Implement Loop and Break Loop enhancements

Ghabry:
_This will be the final Maniac Patch-patch for 0.7_. I still have _ControlVarArray_ and some stuff for _ControlVariable_ to offer but this stuff will be half-broken because the variable range is wrong (Maniac has 32bit range without overflow), so makes no sense to add this when calculation will get stuck at 9999999.

-----------

I still have to carefully recheck all the loop commands. There could be off-by-1 errors I missed :)

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -3495,7 +3495,7 @@ bool Game_Interpreter::CommandLoop(lcf::rpg::EventCommand const& com) { // code 
 	auto& frame = GetFrame();
 	auto& index = frame.current_command;
 	frame.maniac_loop_info.resize((com.indent + 1) * 2);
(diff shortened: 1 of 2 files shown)


Ghabry:
Good that this was delayed. Found two save incompatibilites vs. Maniac Patch when using "X Times". Is correct now.

Ghabry:
This PR is not urgent so I give till end of month for more reviews before merge :)

Unfortunately hard to test on specific games because they usually use more than just the loops of maniac :/

Ghabry:
I'm slowly considering to simply call the next release 0.7.1 instead of 0.7.0.1. There are already lots of enhancements again and the additions by rueter and BlisterB (SAF) are almost merge ready (a matter of a few weeks).

This would also unlock this PR and allow merging it.

Ghabry:
This one is also rdy for merge

PR #2659, opened 2021-09-30
Implement smooth scrolling in menu windows

rueter37:
Depends on #2643 due to different menu item heights (normally always 16, but is 14 in RPG Maker 2003 battles with small battle system)

This PR makes menu windows do smooth scrolling now if the menu cursor exceeds the edge.

Fixes: #2574


rueter37:
Rebased this and #2643 now.

rueter37:
I went with the easy route and disabled the smooth scroll upon using Page Up and Page Down as RPG_RT does not support page scrolling.

Issue #2314, opened 2020-09-01
Implement abort event hotkey in test play mode


fmatthew5876:
If you run RPG_RT in test play mode, you can use the F10 key to abort the currently active foreground event at any time. 

We don't implement this feature at all. We currently use F10 for screenshots.

DynRPG address: 000B203A

Ghabry:
Imo instead of using a hotkey for this it could be a debug menu option instead.

fdelapena:
Because this is TestPlay only, F10 hotkey by default in this mode wouldn't hurt, and non-keyboard ports likely won't require the key mapping, but this is anyways possible once settings scene lands. Debug menu sounds fine as an option for all ports, also suitable for ports without F10 mapping but F9 mapped.

Fishaphrodite:
Maybe I shouldn't poke my nose in this, however, as far as I understand, using the same shortcut as the debug menu would prevent accessing the debug menu during the event itself (while it doesn't sound problematic, I don't know how that would react during battles initiated by an event, also the test-play debug menu is a feature I like to be able to trigger and undo without cancelling the current event).

F10 is used for screenshots, so it'd be a bit off to change it in that regard as well (I spam the screenshot key a lot, if I were to break a chain of event by doing this, I would be a bit disturbed).

Debug menu option would be fine, but the debug menu cannot be accessed during a message box, so if for some reason the message box is stuck, that might be troublesome (However, that is assuming this issue could happen, so maybe that is a non-issue.)

I am just an humble user of the player however, so you may disregard my comments if they are too intrusive/wrong.

fdelapena:
I agree the take screenshot key should be moved to another function key to prevent clashing with the F10 key to match RPG_RT behavior, despite this function is fairly unknown for most game developers.

fmatthew5876:
I would be in favor of moving the screenshot key and replicating this behavior for F10.

rueter37:
I would suggest to move the screenshot key to F7, as @fmatthew5876 already reserved the F8 key for the planned game settings scene (see #2035).

fdelapena:
As mentioned on IRC, RPG Maker 2003 1.10+ uses  the F1 key, so it could be reused for settings and available in Player.

carstene1ns:
[...]

PR #2660, opened 2021-10-03
Implement abort event hotkey in debug mode

rueter37:
This PR enables aborting the active foreground event in debug mode by pressing the F10 key and moves the screenshot hotkey to the F7 key.

Fixes: #2314


src/input_buttons.h
```diff
@@ -73,6 +73,7 @@ namespace Input {
 		FAST_FORWARD_PLUS,
 		TOGGLE_FULLSCREEN,
 		TOGGLE_ZOOM,
+		DEBUG_ABORT_EVENT,
```
Ghabry, line 76: could you move this next to the other DEBUG stuff?

rueter37: Fixed.

Ghabry: sorry, just noticed you must also add it to ``kButtonNames`` and ``kButtonHelp``. They will be used by the settings scene later

rueter37: Added `DEBUG_ABORT_EVENT` to `kButtonNames`and `kButtonHelp` now. While this PR doesn't add the settings menu, `SETTINGS_MENU` has been added to `kButtonNames` because `kButtonNames` and `kButtonHelp` were in a inconsistent state (39 vs. 40 entries). In `kButtonHelp` was already a description for `SETTINGS_MENU` while this one was missing in `kButtonNames`.

Ghabry: yeah, I noticed something similiar with the key names when I readded the input display to the settings scene: When I added additional keyboard keys I forgot to add them to names and help, making the settings-input scene instant-crash...

Ghabry (commented):
Checked in RPG_RT:

1. Calls an unknown function in Window Message (purpose likely the same)
2. Unconditional call to ``FinishMoveRoute`` (so no if-check)
3. Ends event processing

So you are pretty accurate ^^

PR #2678, opened 2021-11-01
Cancel the Inn music after 10 seconds.

### LINKED ISSUE #2611: Inn: Harmony skips silence at the end of MIDIs

[POSADA.zip](https://github.com/EasyRPG/Player/files/6986970/POSADA.zip) (from the game [The Story Of Egroj](https://www.hellsoft.net/H5/component/remository/func-startdown/357/?Itemid=527)).

The attached MIDI file contains an inn jingle with notes for around 4 seconds, but the total MIDI length is around 28 seconds because it contains a marker as a comment after around 28 seconds. It does not contain event loop (111).

marker

Harmony respects the silence on maps, but skips the silence on Inn.

Player does not skip silence for libfluidsynth and native MIDI on Inn.

Note: if the MIDI loops and the sequence does not end, it is a different issue, fixed since #2612.

- [ ] Check if Harmony sends some kind of event to the MIDI device. cc @tyrone-sudeium 
- [ ] Test behavior with MIDIs with 111 event loop.

Related: #2584

Ghabry:
Fix #2611

### DIFF
--- a/src/scene_map.cpp
+++ b/src/scene_map.cpp
@@ -41,6 +41,8 @@
 #include "output.h"
 #include "dynrpg.h"
 
+using namespace std::chrono_literals;
+
 static bool GetRunForegroundEvents(TeleportTarget::Type tt) {
 	switch (tt) {
 		case TeleportTarget::eForegroundTeleport:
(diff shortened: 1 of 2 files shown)


fdelapena (approved):

carstene1ns (approved):
I do not understand how the inn(async) works, but codewise looks good to me.

fmatthew5876 (approved):

Issue #2679, opened 2021-11-02
Setting --fps-limit 0 causes floating point exception crash


fmatthew5876:
Try `--fps-limit 0` to disable fps limit, you get an immediate crash. Probably a divide by 0 bug somewhere.

rueter37:
The crash is triggered [here](https://github.com/EasyRPG/Player/blob/master/src/game_clock.h#L137) when `fps-limit` is set to 0, the cause is indeed a division by 0. Does it make sense to add a check if `fps` is 0 and if yes, then set `ns` to a sufficiently low value to emulate (near) infinite FPS?

Ghabry:
You mean set fps to a very high value? Like numeric_limit<int>::max() or how that was called... imo that is the easiest solution here...

rueter37:
Your suggestion sounds even better. Is better than setting `ns` to a certain value if `fps` is 0.

fmatthew5876:
Using an arbitrarily high number could work.

Otherwise, I believe https://github.com/EasyRPG/Player/blob/master/src/baseui.cpp#L63 setting this frame_limit to 0 if the fps_limit is 0 would also work. Needs testing

rueter37:
@fmatthew5876 Your suggestion works for me. I have added a PR (#2684) to fix this issue.

PR #2684, opened 2021-11-04
Set frame_limit to zero if fps_limit is set to zero

rueter37:
This PR sets the frame_limit explicitly to zero if the fps_limit is set to zero to avoid a division by zero.

Fixes: #2679


fdelapena (approved):

fmatthew5876 (approved):

Ghabry (approved):
Works well (when Vsync is off).

For later:
Maybe there should be an option that disables the framelimit when vsync is on because vsync already does the timing.

PR #2686, opened 2021-11-07
Move AudioSeCache handling to Game_System.

### LINKED ISSUE #2582: VFS: Sounds are opened even when they are cached
Just noticed a minor performance regression (at least on devices where file opening is slow):

Before the VFS branch no file handle was opened when "SeCache" had a cache hit. This doesn't work anymore because a stream is passed in into the Audio handler (wasn't possible otherwise because RTP and Game have different VFS).

In case of a cache hit the file is opened, cache hit, closed. Kinda useless.

Can be likely fixed by moving the cache handling from Audio Generic to Game_System...

Ghabry:
Fixes a performance regression (unnecessary opening of file handles)

Fix #2582

### DIFF
--- a/src/audio.h
+++ b/src/audio.h
@@ -21,6 +21,7 @@
 // Headers
 #include <string>
 #include "filesystem_stream.h"
+#include "audio_secache.h"
 
 /**
  * Base Audio class.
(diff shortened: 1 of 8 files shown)


Ghabry:
This is ready now. Shouldn't break anything (famous last words) but in general the caching was just moved one frame higher in the callstack.

Still recommend testing ;)

fdelapena (approved):

Ghabry:
No feedback on this for 2 weeks so imo good to merge?

PR #2688, opened 2021-11-11
Translations: Fix reading of Meta.ini

rueter37:
The Meta.ini file in the translations has never been read because of path not being complete. This PR fixes this.

### DIFF
--- a/src/translation.cpp
+++ b/src/translation.cpp
@@ -43,7 +43,7 @@
 #define TRFILE_RPG_RT_COMMON "rpg_rt.ldb.common.po"
 #define TRFILE_RPG_RT_BATTLE "rpg_rt.ldb.battle.po"
 #define TRFILE_RPG_RT_LMT    "rpg_rt.lmt.po"
-#define TRFILE_META_INI      "META.INI"
+#define TRFILE_META_INI      "meta.ini"
 
 // Message box commands to remove a message box or add one in place.
 // These commands are added by translators in the .po files to manipulate
@@ -98,10 +98,10 @@ void Translation::InitTranslations()
 			item.lang_dir = tr_name.second.name;
 			item.lang_name = tr_name.second.name;
 
-			// If there's a manifest file, read the language name and help text from that.
-			std::string meta_name = translation_root_fs.FindFile(item.lang_dir, TRFILE_META_INI);
-			if (!meta_name.empty()) {
-				lcf::INIReader ini(meta_name);
+			// If there's a manifest file, read the language name and help text from that
+			auto meta_file = translation_root_fs.OpenFile(item.lang_dir, TRFILE_META_INI);
+			if (meta_file) {
+				lcf::INIReader ini(meta_file);
 				item.lang_name = ini.GetString("Language", "Name", item.lang_name);
 				item.lang_desc = ini.GetString("Language", "Description", "");
 			}
--- a/src/filesystem.cpp
+++ b/src/filesystem.cpp
@@ -243,7 +243,7 @@ Filesystem_Stream::InputStream FilesystemView::OpenFile(StringView name, Span<St
 
 Filesystem_Stream::InputStream FilesystemView::OpenFile(StringView dir, StringView name, Span<StringView> exts) const {
 	assert(fs);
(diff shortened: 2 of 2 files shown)


Ghabry (commented):
Oh that code is not VFS aware. Must have missed it.
It will not work e.g. in zip archives.
Use OpenFile in translation_fs and type auto instead of string.
And in the if condition that stream (is false of it failed)

rueter37:
@Ghabry Can confirm that this doesn't work with zip archives. I tried out your proposal but it doesn't work for me (but maybe I'm missing something obvious).

rueter37:
@Ghabry Found a solution which should work now.

src/translation.cpp
```diff
@@ -99,9 +99,9 @@ void Translation::InitTranslations()
 			item.lang_name = tr_name.second.name;
 
 			// If there's a manifest file, read the language name and help text from that.
-			std::string meta_name = translation_root_fs.FindFile(item.lang_dir, TRFILE_META_INI);
-			if (!meta_name.empty()) {
-				lcf::INIReader ini(meta_name);
+			auto meta_file = translation_root_fs.Subtree(item.lang_dir).OpenFile(TRFILE_META_INI);
```
Ghabry, line 102: strange, imo ``OpenFile(item.lang_dir).OpenFile(TRFILE_META_INI)`` should also work o_O

rueter37: This one unfortunately fails for me.

Ghabry: maybe you found a bug in the Filesystem code. Have to check this

Ghabry: Yeah you found a bug :)

(The ``TRFILE_META_INI`` to lowercase is just for consistency)

```patch
diff --git a/src/filesystem.cpp b/src/filesystem.cpp
index 7bb1359ae..c1bec3a41 100644
--- a/src/filesystem.cpp
+++ b/src/filesystem.cpp
@@ -243,7 +243,7 @@ Filesystem_Stream::InputStream FilesystemView::OpenFile(StringView name, Span<St
 
 Filesystem_Stream::InputStream FilesystemView::OpenFile(StringView dir, StringView name, Span<StringView> exts) const {
 	assert(fs);
-	return fs->OpenFile(MakePath(name), dir, exts);
+	return fs->OpenFile(MakePath(dir), name, exts);
 }
 
 Filesystem_Stream::InputStream FilesystemView::OpenFile(const DirectoryTree::Args &args) const {
diff --git a/src/translation.cpp b/src/translation.cpp
index 5229f921a..6f22c2fdf 100644
--- a/src/translation.cpp
+++ b/src/translation.cpp
@@ -43,7 +43,7 @@
 #define TRFILE_RPG_RT_COMMON "rpg_rt.ldb.common.po"
 #define TRFILE_RPG_RT_BATTLE "rpg_rt.ldb.battle.po"
 #define TRFILE_RPG_RT_LMT    "rpg_rt.lmt.po"
-#define TRFILE_META_INI      "META.INI"
+#define TRFILE_META_INI      "meta.ini"
 
 // Message box commands to remove a message box or add one in place.
 // These commands are added by translators in the .po files to manipulate
@@ -98,10 +98,10 @@ void Translation::InitTranslations()
 			item.lang_dir = tr_name.second.name;
 			item.lang_name = tr_name.second.name;
 
-			// If there's a manifest file, read the language name and help text from that.
-			std::string meta_name = translation_root_fs.FindFile(item.lang_dir, TRFILE_META_INI);
-			if (!meta_name.empty()) {
-				lcf::INIReader ini(meta_name);
+			// If there's a manifest file, read the language name and help text from that
+			auto meta_file = translation_root_fs.OpenFile(item.lang_dir, TRFILE_META_INI);
+			if (meta_file) {
+				lcf::INIReader ini(meta_file);
 				item.lang_name = ini.GetString("Language", "Name", item.lang_name);
 				item.lang_desc = ini.GetString("Language", "Description", "");
 			}
```

rueter37: Your patch works for me. :+1:

Ghabry (approved):

fdelapena (approved):

Issue #2690, opened 2021-11-14
Language select: Help Window ignores Transparency setting


Ghabry:
You know that small one line window at the top that we use in multiple scenes

![screenshot_0](https://user-images.githubusercontent.com/1331889/141683968-d7e523fe-54cc-40d5-8c4c-3d8a9b4f32db.png)

PR #2693, opened 2021-11-14
Translation selection: Make help window respect transparency setting

rueter37:
This PR makes the help window in the translation selection scene check the transparency setting of the database if RM2k3E is used.

Fixes: #2690


Ghabry (approved):
Reminds me to replace some numbers with constants

fdelapena (approved):

PR #2696, opened 2021-11-16
Emscripten related fixes

Ghabry:
Luckily this was debuggable without using Emscripten :D

The first commit fixes e.g. a folder called "music" not being found when launching directly from the game dir. So this was not found by our Android users just because the project-path there is a path with one directory component in it.

-------

The 3rd commit is not an emscripten fix but it resolves an assert on Linux

### DIFF
--- a/src/directory_tree.cpp
+++ b/src/directory_tree.cpp
@@ -73,24 +73,22 @@ DirectoryTree::DirectoryListType* DirectoryTree::ListDirectory(StringView path) 
 		std::string parent_dir, child_dir;
 		std::tie(parent_dir, child_dir) = FileFinder::GetPathAndFilename(fs_path);
 
-		if (!parent_dir.empty()) {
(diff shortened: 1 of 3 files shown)


Ghabry:
Can also confirm that the # works btw. Though this was already resolved by a different commit.

Works for me

Screenshot_20211116_232827-fs8

fdelapena (approved):

fdelapena:
Not sure if worth to include a quick one to prevent the "Save" issue like this:

https://easyrpg.org/play/pr2696/?game=thisgamedoesnotexist

fdelapena (approved):

Ghabry:
This PR is kinda battle tested as it is part of Yume2kki Multi**P**layer since a week :D

carstene1ns (approved):
Looks good to me.

PR #2699, opened 2021-11-19
Android: Support of SAF

### LINKED ISSUE #1041: Can't change button layout in portrait
Whenever I try to change the buttons, the screen is forced to landscape. Any changes made there completely screw up the layout in portrait.

I don't even know what version I have installed because there's no "about" section or anything in the program, but it should be the latest available on the play store.

### LINKED ISSUE #1330: Investigate Android Game Browser ANR (Application Not Responding)
These are happening quite often in the current version (may be an issue of the updated Android ndk/sdk/build tools):
**Input dispatching timed out (Waiting because no window has focus but there is a focused application that may eventually add a window when it finishes starting up.)**

```
"main" tid=1 Native
"main" prio=5 tid=1 Native
  | group="main" sCount=1 dsCount=0 obj=0x746264b0 self=0xf4f36500
  | sysTid=27481 nice=0 cgrp=default sched=0/0 handle=0xf71c5b34
  | state=S schedstat=( 4707946272 249162681 25230 ) utm=144 stm=326 core=5 HZ=100
  | stack=0xff1f7000-0xff1f9000 stackSize=8MB
  | held mutexes=

   #00  pc 0000000000040f40  /system/lib/libc.so (__getdents64+12)
   #01  pc 000000000001a0c5  /system/lib/libc.so (_ZL16__readdir_lockedP3DIR+20)
   #02  pc 000000000001a19f  /system/lib/libc.so (readdir+16)
   #03  pc 00000000000141e7  /system/lib/libjavacore.so (???)
   #04  pc 000000000002ddf1  /data/dalvik-cache/arm/system@framework@boot.oat (Java_java_io_File_listImpl__Ljava_lang_String_2+92)
   at java.io.File.listImpl (Native method)
   at java.io.File.list (File.java:740)
   at java.io.File.listFiles (File.java:782)

   at org.easyrpg.player.game_browser.GameScanner.scanFolder (GameScanner.java:144)
   at org.easyrpg.player.game_browser.GameScanner.scanGames (GameScanner.java:92)
   at org.easyrpg.player.game_browser.GameScanner.getInstance (GameScanner.java:51)
   at org.easyrpg.player.game_browser.GameBrowserActivity.displayGameList (GameBrowserActivity.java:161)
   at org.easyrpg.player.game_browser.GameBrowserActivity.onResume (GameBrowserActivity.java:92)

   at android.app.Instrumentation.callActivityOnResume (Instrumentation.java:1259)
   at android.app.Activity.performResume (Activity.java:6399)
   at android.app.ActivityThread.performResumeActivity (ActivityThread.java:3110)
   at android.app.ActivityThread.handleResumeActivity (ActivityThread.java:3152)
   at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:2499)
   at android.app.ActivityThread.access$900 (ActivityThread.java:153)
   at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1347)
   at android.os.Handler.dispatchMessage (Handler.java:102)
   at android.os.Looper.loop (Looper.java:148)
   at android.app.ActivityThread.main (ActivityThread.java:5451)
   at java.lang.reflect.Method.invoke! (Native method)
   at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:756)
   at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:646)
```

### LINKED ISSUE #2434: Update to Android API 30 by November 2021
> From November 2021, updates to existing apps will be required to target API level 30 or above

Woohoo -_-

- [x] **BLOCKER**: Load time of the Game Browser is beyond inacceptable
- [x] Select Game directory through SAF
- [x] Game launching through SAF including JNI layer
- [x] XYZ-Image through SAF in the Game Browser 
- [x] Game saving and log file writing
- [ ] Standalone mode (see also #2560)
- [x] BUG: RTP folder is created multiple times on startup
- [x] RTP support through SAF
- [x] Soundfont (FluidSynth) support through SAF
- [ ] Loading of SF2 from APK (half-way to standalone mode), see also #2560
- [x] When suspending the app while playing and resuming it crashes (some important Java handle destroyed?)
- [x] Bug Report feature

BlisterB:
This PR add the support of Android SAF in order to be compatible with API 30 and the new rules about scoped storage permissions.

What has been done : 
- At the start of the app, it checks storage permissions and ask the right to access a Games Folder if necessary.
     - On API < 30, we just need to select the folder one time, on API >= 30, we have to ask for this permission at each start (permanent rights doesn't seems to work)
     - Drop the "multi games folder" feature, not suitable with this new paradigm
- Rewrite most of the code referring to File class (deprecated and broken) and replaced with DocumentFile class.

Important notes :
- Minimum API is now 21
~~- SAF is really slow and using tricks to avoid this system would probably lead to future incompatibilities.~~

What remains to do :
- [x] Fast asynchronous recursive games scanning & result cache
- [x] Check/save of the encoding configuration in the `.ini` file
- [x] Support for soundfout files (waiting for the player to take into account `--soundfont` argument in the command line)
- [x] Rework inputs layouts
- [x] Fix the long loading times at the start of a game
- [x] Fix the crash during the start of some games (ex. Ara Fell)
- [x] Change the messages explaining how the app works (at the end)
- [x] Bug Report feature
- [x] Add a button to open soundfont folder

Fix #1041
Fix #1330
Fix #2434


Ghabry:
I can already see the async loading: Searching for games please come back in 2 hours. Send hate to Google kthxbye

BlisterB:
It is now possible to use a custom soundfont by putting the .sf2 in the games folder (only on the android side).
The path is sent to the native code through command line with argument `--soundfont`.

As discussed with Ghabry, the idea is to also provide a default sounfont in the apk and drop Timidity.
A message will be placed in Settings to inform the player of this possibility and inform what is the current used soundfont.

BlisterB:
The browser now scans games asynchronously. I will later add a visual feedback explaining that it's processing.
The app is much more snappier, but games scanning remains pretty slow.

BlisterB:
The games list is now cached and only refreshed when the player click the "refresh" button, or change the games folder in settings.
With the cache system and asynchronous scans, the only thing we can improve is the scan speed, this is the next things I will be working on.

BlisterB:
The speed of game scanning has been dramatically improved, it's not a problem anymore.
By design the scan can't be recursive without a huge performance loss (we use the heavy `listFile()` for the games folder to obtain various information about files/folder, and quick searches in subfolders with `query()` to obtains only file names).
It may be possible to use `query()` for all the search, but other things are more important for now.

BlisterB:
The user doesn't need anymore to choose the games folder at each startup, even on API 30.

BlisterB:
The favorite ("like") feature is now instant. It previously used to rescan the games folder. The app has never been so snappy.

Ghabry:
@BlisterB 

I researced the query stuff a bit more (good job btw)
As a second column in the ``query`` you can ask for ``DocumentsContract.Document.COLUMN_MIME_TYPE`` and then compare if the string is equal to ``DocumentsContract.Document.MIME_TYPE_DIR``.

If yes it is a folder. Should make recursive scanning possible again :)

Ghabry:
Here is a patch to make the native part faster and to fix some API warnings

```
diff --git a/builds/android/app/src/main/java/org/easyrpg/player/player/SafFile.java b/builds/android/app/src/main/java/org/easyrpg/player/player/SafFile.java
index f26e09f5..050d7ea6 100644
--- a/builds/android/app/src/main/java/org/easyrpg/player/player/SafFile.java
+++ b/builds/android/app/src/main/java/org/easyrpg/player/player/SafFile.java
@@ -1,14 +1,18 @@
 package org.easyrpg.player.player;
 
+import android.content.ContentResolver;
 import android.content.Context;
+import android.database.Cursor;
 import android.net.Uri;
 import android.os.ParcelFileDescriptor;
+import android.provider.DocumentsContract;
 import android.util.Log;
 
 import androidx.documentfile.provider.DocumentFile;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.List;
 
 /**
  * A wrapper around SAF for use from JNI
@@ -82,7 +86,7 @@ public class SafFile {
         }
 
         // No difference between read mode and binary read mode
-        try (ParcelFileDescriptor fd = context.getContentResolver().openFile(root.getUri(), "r", null)) {
+        try (ParcelFileDescriptor fd = context.getContentResolver().openFileDescriptor(root.getUri(), "r")) {
             return fd.detachFd();
         } catch (IOException e) {
             return -1;
@@ -119,7 +123,7 @@ public class SafFile {
             actualFile = df.getUri();
         }
 
-        try (ParcelFileDescriptor fd = context.getContentResolver().openFile(actualFile, mode, null)) {
+        try (ParcelFileDescriptor fd = context.getContentResolver().openFileDescriptor(actualFile, mode)) {
             return fd.detachFd();
         } catch (IOException e) {
             return -1;
@@ -131,13 +135,33 @@ public class SafFile {
             return null;
         }
 
+        Uri root_uri = root.getUri();
+
         ArrayList<String> files = new ArrayList<>();
         ArrayList<Boolean> is_dir = new ArrayList<>();
 
-        for (DocumentFile file: root.listFiles()) {
-            files.add(file.getName());
-            is_dir.add(file.isDirectory());
+        final ContentResolver resolver = context.getContentResolver();
+        final Uri childrenUri = DocumentsContract.buildChildDocumentsUriUsingTree(root_uri, DocumentsContract.getDocumentId(root_uri));
+
+        Cursor c = resolver.query(childrenUri,new String[] {
+                DocumentsContract.Document.COLUMN_DOCUMENT_ID,
+                DocumentsContract.Document.COLUMN_MIME_TYPE },
+                null, null, null);
+        while (c.moveToNext()) {
+            // The name can be something like ``primary:path/of/the/folder/the_file.txt
+            // Get rid of all that junk
+            String file_path = c.getString(0);
+            int slash_pos = file_path.lastIndexOf('/');
+            if (slash_pos != -1) {
+                file_path = file_path.substring(slash_pos + 1);
+            }
+
+            String mime_type = c.getString(1);
+
+            files.add(file_path);
+            is_dir.add(mime_type.equals(DocumentsContract.Document.MIME_TYPE_DIR));
         }
+        c.close();
 
         return new DirectoryTree(files, is_dir);
     }
```

builds/android/app/src/main/java/org/easyrpg/player/player/EasyRpgPlayerActivity.java
```diff
@@ -354,11 +354,7 @@ public Game getProjectPath() {
      * @return Full path to the RTP
      */
     public String getRtpPath() {
-        // TODO : Return RTP
-        return "";
-        //String str = SettingsManager.getEasyRPGFolder() + "/rtp";
-        // Log.v("SDL", "getRtpPath " + str);
-        //return str;
+        return SettingsManager.getRtpFolder().toString();
```
Ghabry, line 357: will this return a valid SAF Uri? (content:// and stuff)

BlisterB: Yep :)

Ghabry: I just tested it.
You must use ``return SettingsManager.getRtpFolder().getUri().toString();``, then it will work.Yours gives back the class name + pointer (Standard toString impl)

BlisterB:
The recursive games search is back and is fast. The game browser is officially better than before. :)

I've also moved all the file's related operation to the Helper class in order to centralized them, in case Google decides to change all the rules again.

Ghabry:
@BlisterB  added my two commits are requested

Ghabry:
Jenkins: test this please

Ghabry:
@BlisterB You can now grab the APK from the PR builder in case you want to do some native code testing.

BlisterB:
The app can now retrieve the encoding of the game, it can also save another encoding settings.
I dropped the JSON preference file and use the .ini file directly.

Thank you @Ghabry ;)

Ghabry:
>  Fix the crash during the start of some games (ex. Ara Fell)

That is due to missing Midi support. It still uses "get_timidity_path" JNI stuff on Android. Have to patch this out and hook it in with fluidsynth.

BlisterB:
The Gamebrowser activity now recovers from the crash of a child activity (a game crashed for example).
It used to don't display the game list after coming back from the crash of a child activity.
I discovered that after a crash, the app will display the previous activity, but static field of the SettingsActivity are set to null (it's probably linked to the cancel of the Context and the fact that DocumentFile are linked to this context).
This is extremely bizarre but I'll remember that it is not safe to store user's preferences in static fields.

BlisterB:
The input layouts feature has been reworked. You can continue to create multiple input layouts and choose an active layout.
It's no more possible to associate a specific layout to a game folder.
I've also done a lot of code cleaning.

BlisterB:
The Audio Settings activity now display a checkbox to enable/disable custom SoundFounts. There is also some text explaining the feature and the .sf2 currently used (the first .sf2 fount in the games folder).
An improvement of the system would be to display a list of all the .sf2 with a RadioButton system to select the one to use. Maybe for later.

Ghabry:
@BlisterB 
Sorry for the slow response time: Is there anything missing here except for the soundfont handling on Player side?

BlisterB:
Hello @Ghabry
It need the soundfont handling in order to work. And a rework of information messages explaining how the app works.
I think that it would be a good idea to make the use of the RTP folder optional (to help improve the initial loading) but it can be done in a futur PR.

Ghabry:
I guess "Report a bug" is also kinda broken? Because attaching the files does not really work.

On the good news: We are down to ca. 5 bug reports per month so looks a bit as if we do not really need it anymore (and just link e.g. Github)

Ghabry:
Dropped now Timidity support completely. When no soundfont is provided Player uses now FmMidi.
Wildmidi has no advantage on Android anymore (is mostly useful for low power devices that cannot handle soundfont or fmmidi).
This resolves also the crash on Android when playing Midi.

Ghabry:
About the "TODO: Convert to SAF" functions: They can be all removed. I expect that SAF makes the folder the user selects writable solving this permission mess.

Ghabry:
@BlisterB Soundfont stuff works now properly. When a game ran before then after changing the setting the app must be restarted due to some global state in C++ Player but this is no big deal imo. Is not like this is changed alot ^^.

-------------

Some nits about the terms:

I suggest changing the following terms:

Settings -> Game Directories: Change to "EasyRPG Folder"

Then the wording in "EasyRPG folder":

```
[SELECT EASYRPG FOLDER] <-- Button

This folder may only be used for RPG Maker 2000/2003 games. Do not select folders that contain other things, such as the download folder.

[OPEN RTP FOLDER] <-- Launches the file browser of Android and shows RTP

The RTP provides shared assets used by some games. Put the RPG Maker 2000 RTP in ($PATH_TO_2000) and the RPG Maker 2003 RTP in ($PATH_TO_2003).
```

Audio:

```
[x] Use Custom Soundfont

You can provide a custom soundfont to alter how MIDI music sounds. Place a .sf2 file in the EasyRPG folder.
Soundfont used: $SOUNDFONT_NAME

-----

Also the paths displayed for RTP and Soundfont should be trimmed so they do not contain garbage like ``primary:``. The soundfont could be only the filename.

**Comment by BlisterB** [comment]
The soundfont feature works great! 👍

**Comment by BlisterB** [comment]
@Ghabry The app now creates these folders: games, soundfonts, saves (for futur support of .zip files).
I've also done the modification of the wording as you asked (EasyRPG is indeed way better, thanks).
I'll change the soundfont system in the futur commit (and also drop the pre-SAF functions).

**Comment by BlisterB** [comment]
The app now allow to select a soundfont among a list of .sf2 files located in the "soundfonts" folder.

Next step : I'll clean the UI and remove unnecessary features.

**Comment by BlisterB** [comment]
RTP Scanning is now an optional feature. When disabled, the initial loading of a game is around 5/10 seconds less longer.

Unnecessary audio features have been deleted (enable audio).

**Comment by BlisterB** [comment]
(TODO : Add a button for opening soundfont folder)

**Comment by Ghabry** [comment]
Except for the broken "Report a bug" feature I only have minor nit-picks about the wording left.

These are just suggestions:

----------

No game found error message:

```
No RPG Maker 2000/2003 games found. Put them in subdirectories of the "games" folder.
The EasyRPG folder can be changed in the settings.
```

----------

**The "How to use" Dialog:**

You rewrote the dialog, this is fine. Some suggested improvements / rewording:

```
Put your RPG Maker 2000/2003 games in subdirectories of the "games" folder.
Games have to be uncompressed.
ZIP, RAR, LZH and EXE files are NOT supported.

Optional features (Please refer to the settings):
- The RTP (shared assets used by some games) can be provided
- A custom soundfont to alter how MIDI music sounds can be provided
```

-----------------

**Audio settings:**

About the custom soundfont. Imo it should be explained that this is about the MIDI output. Not everybody knows what a soundfont is.

```
Custom Soundfont:
Put soundfonts in SF2-format in the "soundfonts" folder. Soundfonts alter how MIDI music sounds.
```

``Please quit and restart...``: Make this sentence an additional localization entry. Otherwise all languages must be updated when this bug is fixed :)

Is ``SoundfontS found:`` necessary? Imo it is clear from the context that this is the soundfont list.

-------------

**EasyRPG Folders**:

This was my suggestion but I actually still consider our EasyRPG folder explanation confusing :(

There is a risk that users select the ``games`` folder which is totally wrong and the sentence does not make this clear.

New idea:

```
SELECT EASYRPG FOLDER

For using EasyRPG Player you must select a folder. This folder is exclusive to EasyRPG Player. Do not select a folder that contains other things, such as the download folder. RPG Maker 2000/2003 games must be put in the "games" subdirectory of this folder.
```
(How about a ``OPEN GAMES FOLDER`` button?)

------

Can you call the "RTP scanning" instead ``Enable RTP support``? Do not think scanning is a term people will understand. Same for ``Please note that enabling RTP support...``.

**Comment by BlisterB** [comment]
I've rework how the games browser works.
It now shows a loading icon when scanning for games. When displaying errors, it show a button "open the games folder" to help the user to understand that they have to put games in this folder now.

I just have to fix the bug report feature and take all the @Ghabry suggestions into account in order to finalize this PR.

**Comment by BlisterB** [comment]
The bug report feature is now fixed, and the app only suggest apps that handles emails.
Attention: if you use bug report without sending the report, the next time Gmail will show you the first draft without the proper attached files. Android. ¯\_(ツ)_/¯

**Comment by BlisterB** [comment]
Just need to take last comment from @Ghabry into account and this PR will be good.

**Comment by BlisterB** [comment]
@Ghabry I took your suggestions into account, except this one : 
> Is SoundfontS found: necessary? Imo it is clear from the context that this is the soundfont list.
I find this label useful, especially for the case where there is no soundfont found.

The only thing that is left is the // TODO SAF function to clear. But they seem to be used, I have to dig this point.

**Comment by Ghabry** [comment]
One minor issue I noticed: when I change the easyrpg folder in the settings it will not change until I reopen the app

**Comment by BlisterB** [comment]
I deleted the TODO SAF function.I'll recode them the right way if I implement zip reading.

@Ghabry Oh it's a nasty regression, I'll dig that. Thank you for the bug report. :)

**Comment by BlisterB** [comment]
The regression is fixed. :)

**Comment by BlisterB** [comment]
Bug report: Going back from Settings to Games Browser using the back arrow on the screen crashes the app.

**Comment by BlisterB** [comment]
Vertical Input Layout is now implemented.
It's totally separated from the Horizontal Input Layout (no more auto configuration), I've provided default configuration for both modes. I've added the Fast Forward button by default (a lot of users doesn't know that this feature exists).
I've deleted all the JSON files reading/writing, we now store input layout in configuration (it won't be affected by future changes of SAF).
I've also fix the regression that I've previously reported.

**Comment by BlisterB** [comment]
The screen now display square pixels, for a far better and cleaner image. :)

**Comment by Ghabry** [comment]
@BlisterB 

While you add default buttons maybe also add SHIFT button by default. Also too many users have no idea how to add this one.

Guess you are 99% finished now? Excellent!

Only needs a rebase. decoder_wildmidi conflicts (remove timidity stuff and the define must be ``#elif defined(__vita__)``).

**Comment by Ghabry** [comment]
And have slight layout problems on Portrait mode. In general great to have this, but is not perfect yet xD

The B is a bit out of bounds and the Fast Forward button is floating on top of the game window.

Maybe move FF in the top-right corner. There it does not overlay anything important.

**Comment by BlisterB** [comment]
The user is now able to choose the image size : 
- Uniform pixel size (best image)
- Stretch the image (wider image)

**Comment by BlisterB** [comment]
While editing input layout, buttons are now moved into a grid.
I've added an explanation for adding new buttons.

@Ghabry You might want to check the wording lol

File: builds/android/app/src/main/res/values/strings.xml [hunk 2]
```diff
@@ -90,6 +93,7 @@ Please tell us in detail what went wrong.\n\n
     <string name="default_layout">Default</string>
 	<string name="horizontal_input_layout">Horizontal (Landscape) Input Layout</string>
 	<string name="vertical_input_layout">Vertical (Portrait) Input Layout</string>
+	<string name="settings_edit_layout_explanation">When editing an input layout, a button can be added (ex. a Shift button).\n Just press the back button and select \"add a button\".</string>
```
Ghabry, line 96: When editing an input layout, buttons can be added (e.g. a Shift button).\n To add a button press the back button and select \"Add a button\".

BlisterB:
- I've fixed the wording of editing layout.
- I've fixed a regression in the games browser caching.
- I've fixed a long standing bug: in EasyRPG Activity, the drawer use to not close when clicking outside the drawer.

BlisterB:
@Ghabry How is the scan speed on your device?
I realize that we could dramatically improve it by skipping subdirectory scans when the folder name is like "Charset", "Title" etc

Ghabry:
You mean the Game Browser scans subdirectories that are not Title? Are you sure about this? o_O

That shouldn't happen.


About Player: The Player must scan all the subdirectories.

And yes everything is pretty fast here. Cannot really complain.

BlisterB:
Oh I was talking about the scan of available games.
Currently we just scan with a depth of 2 folders. By skipping all folder like "Charset", we could set infinite depth and keeping a scan speeds very low.
But there would be some edge cases like not scanning a game called "Title", or a game contained in a "Music" folder.

What do you think of this idea?

Ghabry:
When a game was found in a folder the scanner should not traverse in subdirectories of it. If you put a game inside a game you are stupid (not you, you know what I mean :D).

This is how the built-in Game Browser works: You cannot visit game directories, they are directly booted.

BlisterB:
You are absolutely right, I forgot this part of the implementation. ^^"
Maybe a sign that the code is becoming mature. ^^""""

builds/android/app/src/main/java/org/easyrpg/player/game_browser/Game.java
```diff
@@ -0,0 +1,127 @@
...
+import androidx.annotation.NonNull;
+import androidx.documentfile.provider.DocumentFile;
+
+import org.easyrpg.player.settings.SettingsManager;
+
+public class Game implements Comparable<Game> {
+	private final String title;
+    private final String gameFolderPath;
+    private final String savePath;
+	private boolean isFavorite;
```
fdelapena, line 16: nit: stray indented space

fdelapena (approved):
Thanks, looking great! :+1:

Ghabry:
About the rebase:

I guess you run ``git rebase upstream/master``

You will get a merge conflict:

```
automatischer Merge von src/decoder_wildmidi.cpp
KONFLIKT (Inhalt): Merge-Konflikt in src/decoder_wildmidi.cpp
```

Open ``src/decoder_wildmidi.cpp`` and search for ``======``

The stuff from ``<<< HEAD`` to ``=====`` is the upstream code. The stuff from ``=====`` to ``>>>`` is our code.

To resolve the conflicts:

Delete everything (including the ``<=>`` markers) so that the result is ``#elif defined(__vita__)``:

```
#elif defined(__SWITCH__)
        // Only wildmidi paths, no timidity because it was never used on Switch
        config_file = "./wildmidi.cfg";
        found = FileFinder::Root().Exists(config_file);
        if (!found) {
                config_file = "/switch/easyrpg-player/wildmidi.cfg";
                found = FileFinder::Root().Exists(config_file);
        }
#elif defined(__vita__)
        // Only wildmidi paths, no timidity because it was never used on PSVita

        // Shipped
        config_file = "app0:/wildmidi.cfg";
        found = FileFinder::Root().Exists(config_file);

        // Preferred global directory
        if (!found) {
                config_file = "ux0:/data/easyrpg-player/wildmidi.cfg";
                found = FileFinder::Root().Exists(config_file);
```

Save the edited file.

Run:

```
git add src/decoder_wildmidi.cpp 
git rebase --continue
```

(Confirm the commit message as is)

The rebase was successful.

builds/android/app/src/main/AndroidManifest.xml
```diff
@@ -1,19 +1,14 @@
...
     android:versionCode="1"
-    android:versionName="1.0"
-    android:installLocation="auto">
-
+    android:versionName="1.0">
     <!-- OpenGL ES 2.0 -->
-    <uses-feature android:glEsVersion="0x00020000" />
-
-    <!-- Touchscreen support -->
+    <uses-feature android:glEsVersion="0x00020000" /> <!-- Touchscreen support -->
```
carstene1ns, line 8: Looks like a lot of blank lines got stripped here, leading to misaligned comments.

builds/android/app/src/main/java/org/easyrpg/player/button_mapping/ButtonMappingActivity.java
```diff
@@ -187,16 +204,12 @@ public void addAButton(String s) {
         } else if (s.equals("9")) {
             keyCode = KeyEvent.KEYCODE_9;
         } else if (s.equals("+")) {
-            // API11: KeyEvent.KEYCODE_NUMPAD_ADD
```
carstene1ns, line 190: We should actually use the keycodes, not the magic numbers.

carstene1ns (approved):
Just few nits.

We probably should document somewhere, that the android files have a different license.

Ghabry:
The basic app functionality appears to work 👍 

Can you quickly reformat the files IniFileManager.java and Game.java with your IDE? They look very messy indentationn wise.

builds/android/app/src/main/AndroidManifest.xml
fdelapena: This could be formatted by Android Studio IDE itself, maybe is not worth to restore positioning by hand if later gets rewritten again

Ghabry (approved):

PR #2703, opened 2021-12-13
Unequip shield on equipping a two-handed weapon and vice versa

rueter37:
The shield is unequipped now upon equipping a two-handed weapon and vice versa. While the unequip check worked on actors which can wield two weapons, it did not work on actors which can wield a weapon and a shield.


fdelapena (approved):

Ghabry (commented):
This will fail when you have a shield that is set to "two handed" (that can happen when you check it while it is a weapon and then change the type to shield)

We had similiar classes of bugs before here.

rueter37:
Thanks for the heads up with those edge cases. This should work for now. There is more stuff to fix (stat preview with shield which have the `two_handed` flag set) but the stat preview fix should be postponed for now.

rueter37:
I have added a separate PR for the stat preview stuff.

Ghabry (approved):

Issue #2590, opened 2021-07-19
Yume2kki: Song playback speed incorrect


coinsandboins828626:
#### Name of the game:

Yume2kki (Recent version)

#### Player platform:

Windows

#### Attach files (as a .zip archive or link them)

[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/6844064/easyrpg_log.txt)
[Save01.zip](https://github.com/EasyRPG/Player/files/6844065/Save01.zip)

#### Describe the issue in detail and how to reproduce it:

While playing, I've noticed that certain songs do not have the correct playback speed that they normally would. Instead, they play at full speed, 100%, with no alterations. This usually applies to songs that usually would play at around 25% speed or lower. Example/How to reproduce: In the sound room, songs 127D, 329C, and 340A are all songs set to play at low speed, yet will experience the aforementioned bug.

Also, I've noticed that midi files do not react to speed changes at all, no matter what speed they are set to play at.
Example/How to reproduce: In the sound room, song 346B is a midi song set to play slightly lower than normal, yet plays at the regular file speed.

All of these errors do not persist when using the regular rpg maker, just to clarify. The linked save file has all of the example tracks unlocked in the sound room if you would like to reproduce quickly.

Ghabry:
This is semi-fixed in the "continuous build":

Out-of-range Music values are now clamped instead of reset to default (you will get 50% now, still not 25% but closer)

Pitch changes for MIDI work since yesterday ;).

coinsandboins828626:
Downloaded the continuous build and the issues seem to be fixed! There are some playback errors on that build but I'm aware it is in development and that they don't occur in the current stable release anyway.

Ghabry:
Can you quickly list for us what you mean with playback errors?

To get MIDI working we had to rewrite huge parts of our audio code so it is possible that we broke stuff by accident.

Thanks!

coinsandboins828626:
A couple of things I noticed during a short play session:
-Certain audio tracks had low volume (mostly midi) or were completely silent
-Sometimes the audio would cut out when entering certain areas
-Graveyard World's theme (in the music room, track 26), had its pitch changes sort of freak out and appear to be going above the amount they should.

Ghabry:
> Certain audio tracks had low volume (mostly midi) or were completely silent

This is known

> Sometimes the audio would cut out when entering certain areas

This could be a known bug in our fade-out code, but we will check.

[...]

PR #2706, opened 2021-12-21
Extend allowed tempo range for BGMs and SEs

rueter37:
This PR extends the allowed tempo range for BGMs and SEs from 50-200% to 10-400%.

Fixes: #2590

### DIFF
--- a/src/game_system.cpp
+++ b/src/game_system.cpp
@@ -84,8 +84,9 @@ void Game_System::BgmPlay(lcf::rpg::Music const& bgm) {
 		Output::Debug("BGM {} has invalid fadein {}", bgm.name, bgm.fadein);
 	}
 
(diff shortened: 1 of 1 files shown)


Ghabry:
Did you make some tests to ensure that this still works fine? (I couldn't see why not but... :))

rueter37:
Well, while the entire range works for me, I don't recommend using values below 25%. This is way too slow then (but still works).

Ghabry:
Could you add comments to both changes?

"Normal pitch is 50 to 200 but Yume2kki uses out of range values"

rueter37:
Comments have been added now.

Ghabry (approved):

fdelapena (approved):

Issue #2597, opened 2021-07-26
Savegame made with EasyRPG can contain invalid “battleanim_target” id and will crash RPG_RT.exe


florianessl:
#### Name of the game:
Sternenkindsaga (1.1.3)

#### Player platform:
Windows

#### Attach files (as a .zip archive or link them)
[Sternenkindsaga - Version 1.1.3 - Saves.zip](https://github.com/EasyRPG/Player/files/6880809/Sternenkindsaga.-.Version.1.1.3.-.Saves.zip)

#### Describe the issue in detail and how to reproduce it:
(Original thread with more details can be found here: https://community.easyrpg.org/t/save-game-made-with-easyrpg-contains-invalid-battleanim-target-id-and-will-crash-rpg-rt-exe/992)

This seems to be caused by a map transition while a BattleAnimation is playing. A currently running BattleAnimation from a previous map might be saved into the .lsd with battleanim_active=T and the battleanim_target is set to an Event ID from the previous map. After loading the save, RPG_RT.exe will try to execute the animation on the given ID and will crash with “Invalid event specified” if no Event is found..
**Steps to reproduce:**
In Sternenkindsaga, enter the Map 0109 "Vorplatz" from the west and immediately go through the door north, go into the custom menu and save your game. If done quickly enough, this will create a corrupt save file which cannot be loaded in RPG_RT.exe. The game is trying to display the fountain animation inside with an event target that doesn't exist.

Ghabry:
@fmatthew5876 

As the RPG_RT.exe internals expert:
Do you know anything about the behaviour of Battle Anims on map change?
Are they just set to 0?

CherryDT:
`battleanim_active` is simply set to `false`.

rueter37:
Would it be a possible fix for this issue to add `data.battleanim_active = false;` after [line 88 of game_screen.cpp](https://github.com/EasyRPG/Player/blob/master/src/game_screen.cpp#L88)?

PR #2708, opened 2021-12-28
Set battleanim_active to false on map change

rueter37:
This PR makes sets the `battleanim_active` flag to false on map change. This change should avoid RPG_RT crashes with savegames made with EasyRPG which have an active battle animation which targets an invalid event because of a map change.

Fixes: #2597

### DIFF
--- a/src/game_screen.cpp
+++ b/src/game_screen.cpp
@@ -86,6 +86,7 @@ void Game_Screen::OnMapChange() {
 	movie_res_x = 0;
 	movie_res_y = 0;
 
+	data.battleanim_active = false;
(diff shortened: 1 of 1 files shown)


Ghabry (approved):

Ghabry:
@rueter37 

As this was discussed in IRC and this is also minor:

Can you add a commit that removes the std::sort from ChangeBattleCommands?
There are other issues with the command but that fixes the most obvious problem.

Ghabry:
nvm have 2 PRs open so will do it myself ^^'

fdelapena (approved):

PR #2711, opened 2021-12-29
Fix stat preview with "two handed" non-weapon equipment

rueter37:
This PR should make the stat preview work correctly now even with non-weapon equipment which has the "two handed" flag set.


Ghabry (approved):

Ghabry:
Jenkins: Test this please

fdelapena (approved):

PR #2714, opened 2022-01-05
Yume2kki: Fix Panorama handling

### LINKED ISSUE #2195: Yume2kki: Maiden Outlook background not showing properly
#### Name of the game:
Yume2kki

#### Player platform:
Linux

[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/4572958/easyrpg_log.txt)

The Maiden Outlook area in 2kki looks weird in EasyRPG:

And here's how it's supposed to look:

To reproduce:
Start/Load a game and head to the Maiden Outlook. (Sleep > Nexus > Library (Entrance looks like a roof) > read the book on the second bookcase to the left of the Library's entrance.
(Sorry if this ends up being a duplicate.)

Ghabry:
Panorama positioning depends on width and height of the panorama. 

In some cases (not scrolling or looping) the position is reset after loading the panorama.
Here was the bug: Multiple ChangePanorama commands overwrote the reset indicator.

Added a async yield op to the event and made the panorama image important to ensure that the image is loaded.

Refactored Spriteset_Map to prevent deletion as the panorama change is otherwise a dangling reference.

Fix #2195

### DIFF
--- a/src/async_op.h
+++ b/src/async_op.h
@@ -40,7 +40,8 @@ class AsyncOp {
 			eExitGame,
 			eTerminateBattle,
 			eSave,
-			eLoad
+			eLoad,
+			eYield
 		};
 
 		AsyncOp() = default;
@@ -71,6 +72,9 @@ class AsyncOp {
 
 		/** @return a Load async operation */
 		static AsyncOp MakeLoad(int save_slot);
+
+		/** @return a Yield for one frame to e.g. fetch an important asset */
+		static AsyncOp MakeYield();
 
 		/** @return the type of async operation */
 		Type GetType() const;
@@ -214,5 +218,8 @@ inline AsyncOp AsyncOp::MakeLoad(int save_slot) {
 	return AsyncOp(eLoad, save_slot);
 }
 
-#endif
+inline AsyncOp AsyncOp::MakeYield() {
+	return AsyncOp(eYield);
+}
 
+#endif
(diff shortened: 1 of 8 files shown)


Ghabry:
Appears to be incomplete.

Asked some to test this and it is still broken on at least 3 otther maps.

Waiting for the list

fdelapena (approved):

Ghabry:
This needs careful testing now in other games.

I added some new test cases to the panorama bug report.

They are all fixed.

Ghabry:
Found another Yume2kki map where the offset is wrong m(

Ghabry:
Looks like the algorithm changed between 2k and 2k3.

I used 2k3 DynRPG in Yume2kki for the testing. When I use the bundled RPG2k RPG_RT the result is different -_-

Ghabry:
As stated in the chat the behaviour is slightly different for 2k vs 2k3. I checked different engine versions and this is indeed a 2k vs. 2k3 difference, was not fixed in a later version or something.

Combined this commit with a refactor because this ``const int w = parallax_width * TILE_SIZE * 2;`` happens alot in the code and my ``did_teleport`` I _just_ added did not work here because 2k does this logic _after_ instead of _before_ so this would have been too messy.

Also found a bug in my code: This repositioning logic is not depending on the loop state of the map.

Ghabry:
For reference the old panorama test case: https://github.com/EasyRPG/Player/files/2705008/panoramatest.zip

Ghabry:
Well I would say: One more week of Yume2kki online testing. Then this should be safe to merge.

Ghabry:
The only reports I received were about misaligned panoramas on first time entering (Async issue) and these are fixed by the last commit (And I retested most of the reports, they are fixed).

So this is as good as it can get for a merge :+1:

PR #2723, opened 2022-02-16
Translation enhancements

### LINKED ISSUE #2691: Reloading translations crashes
Translating works by rewriting the database strings, so after one translation you must reload the database.

The reloads causes danglings pointers everywhere into the old database.

Needs some callback feature across the scene stack, e.g. ``OnTranslationChange`` that refreshes all pointers :/

Ghabry:
Also do not unload the system graphic string anymore here.

Fix #2691

~~Waiting for Jan (Deep8) to confirm that this is working as translation change on map is not something that is normally possible :)~~

https://user-images.githubusercontent.com/1331889/154381192-1fa4377d-5417-4e2c-99a2-56dc09c5ed30.mp4

### DIFF
--- a/resources/easyrpg-player.6.adoc
+++ b/resources/easyrpg-player.6.adoc
@@ -138,6 +138,9 @@ NOTE: Incompatible with *--load-game-id*.
 
 NOTE: Incompatible with *--load-game-id*.
 
+*--language* 'LANG'::
+  Loads the game translation in language/LANG folder.
+
 *--test-play*::
   Enable TestPlay mode.
 
--- a/src/async_handler.cpp
+++ b/src/async_handler.cpp
@@ -268,6 +268,15 @@ void FileRequestAsync::Start() {
 		}
 	}
 
+	if (graphic && !Player::translation.GetCurrentLanguageId().empty()) {
+		// FIXME: Asset replacement will only work once for translations
+		std::string modified_path_trans = FileFinder::MakePath(lcf::ReaderUtil::Normalize(Tr::GetCurrentTranslationFilesystem().GetFullPath()), modified_path);
+		auto it = file_mapping.find(modified_path_trans);
+		if (it != file_mapping.end()) {
+			modified_path = modified_path_trans;
+		}
+	}
+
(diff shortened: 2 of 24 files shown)


fdelapena:
jenkins: test this please

rueter37:
While testing this PR I unfortunately still get a crash when I change the translation, switch back to the default language and start a new game.

Ghabry:
@rueter37
I cannot reproduce this. Guess you hit some edge case.
Can you provide a sample game?

rueter37:
Try this one: [TestGame-2000.zip](https://github.com/EasyRPG/Player/files/8083405/TestGame-2000.zip). This is a modified version of TestGame-2000 which contains a small german translation which translates some strings in the starting map to German.

Ghabry:
@rueter37  found what you meant. Happened after resetting the game and then switching the translation. Fixed now.

Also added translation support for the web player. Though the asset translating is limited to "only once per asset" so if you e.g. loaded already the original (like the Title graphic) it is not replaced anymore.

Ghabry:
well who cares about API limitations.
Only had to remove an assert that triggers a false-positive after the request clear.

The only thing that is not replaced is the system graphic but this should be acceptable ^^

rueter37:
Tested it again and the crash on starting a new game after switching the language back to default no longer occurs. :+1:

Ghabry:
(now I'm done)

fdelapena (approved):

lychees:
Most of the localization games contains not only texts, but also images. (e.g. the mannual in yumenikki)
Do we have a plan to support this feature in the future?

image
image

rueter37:
@lychees As far as I know this is already supported. According to the [manual](https://easyrpg.org/player/guide/game_translation) it should work like this:

> **Translating assets**
> 
> Many games also contain text on graphics, and a few even have voiceover.
> These assets can also be translated. All you have to do is mirror the corresponding path in the translation folder.
> For example, the translated version of Picture/Intro.png is located in languages/LANGUAGE_NAME/Picture/Intro.png.

lychees:
Thanks, that would be helpful.

PR #2725, opened 2022-02-17
Raise the priority of WQY when encoding is Chinese Simplified.

### LINKED ISSUE #2724: Confusion between Simplified Chinese and Japanese fonts.
Hello EasyRPG Player team. We are a community of Chinese yumenikki fan games. Many of us like the idea of the multiplayer version, it brings a whole new experience. While working on translation those games and setup our own local multiplayer version, we found there are some mistake when displaying Chinese characters in EasyRPG.

Long story short, today in East Asia, the Chinese characters used in different regions are different. E.g. the Japanese language makes many of [their own Chinese characters](https://ja.wikipedia.org/wiki/%E5%92%8C%E8%A3%BD%E6%BC%A2%E5%AD%97) in history.

image

Both Japanese and Chinese fonts appear in a same paragraph is not a good idea. Note that this bug only occurs in EasyRPG and did not reproduced in the original rpgmaker 2000 and 2003.

S~NY{QG786F_{ A6DT(6H5S

UPOA6DJEY`A{}YJM@TNRKWG

We have successfully resolve the bug in our local multiplayer version, still need a few days to improvements the code in order to merge them to the master branch.

Regards.

lychees:
As mentioned in https://github.com/EasyRPG/Player/issues/2724, this is not a perfect solution, but do solve the problem.
(Anyway it costs no side effect for other encoding) Note that some punctuations don't exist in WQY, so we still need to check Mincho when there is no code in WQY.

Tested passed in [webplayer](https://easyrpg.org/player/guide/webplayer/).

Fix #2724

### DIFF
--- a/src/font.cpp
+++ b/src/font.cpp
@@ -71,11 +71,23 @@ namespace {
 	}
 
 	BitmapFontGlyph const* find_gothic_glyph(char32_t code) {
+		if (Player::IsCP936()) {
+			auto* wqy = find_glyph(BITMAPFONT_WQY, code);
+			if (wqy != NULL) {
+				return wqy;
+			}
+		}		
 		auto* gothic = find_glyph(SHINONOME_GOTHIC, code);
 		return gothic != NULL ? gothic : find_fallback_glyph(code);
 	}
 
 	BitmapFontGlyph const* find_mincho_glyph(char32_t code) {
+		if (Player::IsCP936()) {
+			auto* wqy = find_glyph(BITMAPFONT_WQY, code);
+			if (wqy != NULL) {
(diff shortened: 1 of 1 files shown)


Ghabry:
Jenkins: Test this please

Ghabry:
@lychees  does it also make sense to do this for Traditioinal Chinese? (Player::IsBig5())

lychees:
> @lychees does it also make sense to do this for Traditioinal Chinese? (Player::IsBig5())
No, this bug does not appear in traditional Chinese.

image

fdelapena (approved):

PR #2726, opened 2022-02-17
Add "easyrpg_affected_by_row_modifiers" attribute for skills

rueter37:
This PR adds the "easyrpg_affected_by_row_modifiers" attribute for skills. If this attribute is set, then the damage and hit rate is affected by row modifiers. This feature is needed if you want to add skills which should act like normal physical attacks in RPG Maker 2003 games.

### DIFF
--- a/src/algo.cpp
+++ b/src/algo.cpp
@@ -106,7 +106,7 @@ int CalcNormalAttackToHit(const Game_Battler &source,
 }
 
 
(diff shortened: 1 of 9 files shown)


Ghabry:
Completely forgot about the unit tests so maybe adding this is less dangerous then I thought :D

fdelapena (approved):

Ghabry (approved):
Sorry for the long delay

PR #2729, opened 2022-02-18
Use wqy by default for Chinese encoding.

lychees:


Ghabry:
I'm not really sure why this is needed. The old code should already work just fine.

With this patch you disable all the fallback handling in Chinese?

PR #2734, opened 2022-02-24
Proper overflow handling

### LINKED ISSUE #2650: Incorrect handling of overflow
Test case:

```
@> Control Variables: V[0001] = 9999999
@> Control Variables: V[0001] *= 300
```

This overflow a signed 32bit integer.

RPG_RT: 9999999
Player: -9999999

Ghabry:
When variables overflow the value is now correctly clamped.

Mostly required by Maniac Patch because it uses the entire integer range but also broke with multiply in normal maker (see fixed issue).

Also implemented the "Input numbers via numpad" extension from newer Maniac versions.

Fix #2650

### DIFF
--- a/src/game_variables.cpp
+++ b/src/game_variables.cpp
@@ -616,6 +616,6 @@ StringView Game_Variables::GetName(int _id) const {
 }
 
 int Game_Variables::GetMaxDigits() const {
-	auto val = std::max(std::abs(_max), std::abs(_min));
+	auto val = std::max(std::llabs(_max), std::llabs(_min));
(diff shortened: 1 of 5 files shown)


fdelapena (approved):

Ghabry:
Undefined behaviour at its finest:
``abs(INT_MIN)`` is again ``INT_MIN`` because it overflows and ``max(abs(INT_MAX), abs(INT_MIN))`` returns a max value of ``INT_MIN``... wtf?! :)

Changed to long long to ensure it fits ^^

fdelapena (approved):

Issue #1818, opened 2019-06-14
Support Maniac Patch (Metabug for all releases)


Ghabry:
Because Maniac Patch is a moving target I decided to close the old issues and just update this one issue here when new features are added, otherwise this slowly becomes messy.

Old issues: #1424, #1557, #1608
Some of these old issues contain more information not moved here yet.

Known versions:

TODO: Figure out which version added which feature

- [ ] v180811: Adds Get Save Info, Save, Load, Get/Set Mouse Position, ~~End Load~~
- [ ] v181209
- [ ] v190122: Adds Show String Picture
- [ ] v190526

(Tick when all features of this version are supported)

---

**New commands**

- [ ] **Get Save Info** (since v180811), Code 3001

This command gets the same information as what is displayed on the default load screen.
Store the date, level, and HP in a variable and the face graphic in a picture.
Character names are not currently supported.
    
If no save data is present, the variable for the date contains 0.
Face graphics are treated as 4 * 4 sprite sheets.
Therefore, it is not displayed when it is a file like arranging 16 or more patterns.
It is not a problem even if it comes off from the standard as long as it is four divisions oppositely.
The picture uses the original settings, except for files and sprite sheet-related items.
    
In the English version, the Order of the date (YYMMDD) is the same.

- String arguments: None
0. How to specify a save number (0: constant 1: variable)
1. Save Number
2. Variable to receive save date (YYMMDD)
3. Variable to receive save time (HHMMSS)
4. Variable to receive the first character level
5. The variable that receives the top character's HP
6. Reservation item (0 fixed)
7. How to specify the picture number (0: constant 1: variable)
8. Picture number to receive the face graphic of member 1 (first character)
9. Picture number to receive member 2 face graphics
10. Picture number to receive member 3 face graphics
11: Picture number to receive member 4 face graphics

- [ ] **Save**  (since v180811), Code 3002

A command that performs the equivalent of the default save function.
_This command will not yield which means the game state is undefined while saving, in Player this should return to the Update loop like a Transition and reenter_

Save number 0 or less is not available.
You can optionally store execution results in a variable.
The value is 1 when it succeeds, and 0 if it fails.
    
Note that assignment to a variable is done after saving.

String arguments: None
0. How to specify a save number (0: constant 1: variable)
1. Save Number
2. Whether to receive the success or failure of a save (0: Receive 1: Received)
3: Variable to receive success

- [ ] **Load**  (since v180811), Code: 3003

[...]

PR #2735, opened 2022-02-25
Maniac Patch: Implement GetPictureInfo

Ghabry:
One of the more basic commands.
Can be used to query picture positions.
Tested and confirmed working.

PR #2741, opened 2022-02-28
Maniac Pictures: Support Origin, Fixed Angle and Frames for duration

Ghabry:
Okay kinda clickbait. Origin is not supported yet because I cannot figure out some rounding errors.
But it will come this week ;)

Wanted an issue number already for the metabug (#1818)

- [x] Fixed Angle standalone
- [x] Switching from Fixed Angle to other effects (this looks bad because maniac uses 360°, rest 256°, not my fault)
- [x] How many frames Move picture lasts
- [x] Origin


Ghabry:
Origin works now for Show and Move Picture.
Depends now on #2735 due to conflicts.

PR #2743, opened 2022-03-01
Emscripten: basic Web Gamepad support

### LINKED ISSUE #1156: Emscripten: gamepad support
HTML5 has a Gamepad API. It looks like Emscripten [supports it](https://kripken.github.io/emscripten-site/docs/api_reference/html5.h.html#gamepad), however it does not seem to abstract SDL Joystick, so it needs port specific work. The [HTML5 test](https://github.com/kripken/emscripten/blob/master/tests/test_html5.c) has some useful example code.

tyrone-sudeium:
This PR adds _very_ rudimentary Web Gamepad support.

Some notes:
- It seems like Joystick support was broken because `SDL_JoystickOpen(0)` at init time doesn't work when running in emscripten.
- I haven't touched any of the button maps, but it's not great right now. The dpad is completely nonfunctional, and the X button on the Xbox is mapped to arrow down, Y is Esc, B is Enter, and so on. The left stick does work as expected though. I think we'll need to have a special default button map for when `#ifdef EMSCRIPTEN`...
- If there's any gamepad connected and active in the browser window, the on-screen touch controls will immediately disappear. They should reappear when the gamepad disconnects. I moved the logic for displaying the touch controls into JS since the logic is now more complicated than what can be expressed by a simple media query.

The motivation for this change is obviously supporting game controllers out of the box for web deployments of EasyRPG Player, but it's particularly notable since web is the only way to deploy to iOS users, who'll for the first time be able to use MFi controllers instead of clunky touch controls, and stopping those touch controls getting in the way of the game UI.

Fixes #1156

### DIFF
--- a/resources/emscripten/emscripten-shell.html
+++ b/resources/emscripten/emscripten-shell.html
@@ -115,13 +115,6 @@
       opacity: 0.4;
     }
 
-    @media (hover: hover) and (pointer: fine) {
-      #apad,
-      #dpad {
-        display: none;
-      }
-    }
-
     @media (min-width: 320px) and (min-height: 240px) { #canvas { width: 320px; height: 240px; } }
     @media (min-width: 640px) and (min-height: 480px) { #canvas { width: 640px; height: 480px; } }
     @media (min-width: 960px) and (min-height: 720px) { #canvas { width: 960px; height: 720px; } }
@@ -249,7 +242,7 @@ function bindKey(node, key, keyCode) {
 
     keysDown.delete(event.target.id);
     node.classList.remove('active');
-  
+
     if (lastTouchedId) {
       document.getElementById(lastTouchedId).classList.remove('active');
     }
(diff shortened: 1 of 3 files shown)


Ghabry:
Cool!
Is it possible to forward the "raw keys" from the gamepad into emscripten instead of transforming them into keyboard codes?
SDL2 has already gamepad support (and the input mapping UI comes soon)

tyrone-sudeium:
> Is it possible to forward the "raw keys" from the gamepad into emscripten instead of transforming them into keyboard codes? SDL2 has already gamepad support (and the input mapping UI comes soon)

How far out is the input mapping UI? What you describe definitely sounds better, but it also sounds like exactly what SDL should already be doing. SDL [does have an implementation of Web Gamepad in it](https://github.com/libsdl-org/SDL/blob/main/src/joystick/emscripten/SDL_sysjoystick.c) but for whatever reason it's not working in Player.

Ideally, whatever misconfiguration or perhaps bug we have that prevents this from working we should fix.

I didn't know how to do that. Doing some dodgy JavaScript hacks, that I know ;)

Ghabry:
You can test the input mapping here: https://easyrpg.org/play/pr2730/  (start game, in the menu is a new option "Settings")

But Joypad events are indeed not recognized _at all_. Very strange. Works locally.

Ghabry:
actually my Xbox controller doesn't even work with this PR here o_O

tyrone-sudeium:
> though the touch controls hiding isn't working like expected on iPadOS

This is now fixed

tyrone-sudeium:
I have a theory as to why our SDL Joystick events don't work in Emscripten.

If you read the SDL2 [tests](https://github.com/libsdl-org/SDL/blob/2f7b885df954c130252e1a98b49ea0a7013c1dec/test/testjoystick.c#L307), you'll see that when running in emscripten, they use [`emscripten_set_main_loop`](https://emscripten.org/docs/api_reference/emscripten.h.html#c.emscripten_set_main_loop) to set up the "game" loop, rather than a plain ol' `while(true)`. I suspect an emscripten main loop is required for joysticks due to the fact that it needs to poll the web gamepad API.

Haven't tested this assumption since I'm not really sure where would be the appropriate place to set this up in the Player, but also this requirement isn't documented anywhere too, so, could be bogus. Worth investigating if someone can point me to the right place in the C++ to slap this

Ghabry:
Cool. I wonder if this also helps with Bluetooth controllers on Android.
Cannot test them due to lack of controllers.

Ghabry (approved):
Works excellent! Including hotplugging. Nice.

We must really fix the layout to fit a typical xbox controller but this is not your fault. :D

fdelapena (approved):

PR #2745, opened 2022-03-02
Compatible with cpp17

lychees:
Make it easier for multiplayer version to include some cpp17 headers.
No side effects.

### DIFF
--- a/src/pending_message.cpp
+++ b/src/pending_message.cpp
@@ -76,10 +76,10 @@ std::string PendingMessage::ApplyTextInsertingCommands(std::string input, uint32
 
 	std::string output;
 
-	auto iter = input.data();
+	const char* iter = input.data();
 	const auto end = input.data() + input.size();
 
-	auto start_copy = iter;
+	const char* start_copy = iter;
 	while (iter != end) {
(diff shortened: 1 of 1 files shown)


Ghabry:
Jenkins: Test this please <3

src/pending_message.cpp
```diff
@@ -132,7 +132,7 @@ std::string PendingMessage::ApplyTextInsertingCommands(std::string input, uint32
 		// Fast path - no substitutions occured, so just move the input into the return value.
 		output = std::move(input);
 	} else {
-		output.append(start_copy, end);
+		output.append(start_copy, end - start_copy);
```
Ghabry, line 135: No side effects but doesn't this substraction change the API?

lychees: No, it is same.
https://www.cplusplus.com/reference/string/string/append/

Ghabry: Verified this now also in my IDE: This picks a different overload... C++ is a mess :)

Ghabry (approved):

fdelapena (approved):

PR #2749, opened 2022-03-14
Windows: Support mpg123 1.28.1 and newer

Ghabry:
Untested, lets see what Jenkins says...

### DIFF
--- a/src/decoder_mpg123.cpp
+++ b/src/decoder_mpg123.cpp
@@ -30,7 +30,13 @@ static void Mpg123Decoder_deinit(void) {
 	mpg123_exit();
 }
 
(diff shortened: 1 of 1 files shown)


Ghabry:
It fixed Windows but breaks everything else 😊

Ghabry:
Stupid: This is a breaking change but they did not bump ``MPG123_API_VERSION`` making it impossible to use ``#ifdef`` here.

Instead I hardcoded the underlying type.

Similiar to the SDL2_Mixer solution: https://github.com/libsdl-org/SDL_mixer/commit/ff4eac94b48c111b31ca124e2cbb4577fef72bd8

fdelapena (approved):

PR #2752, opened 2022-03-21
Improve Gamepad / Refactor Input Api

Ghabry:
- [x] 3ds: Add Input buttons
- [x] 3ds: Fix Analog handling
- [x] libretro: Fix Analog handling
- [x] PS Vita: Add input buttons
- [x] PS Vita: Fix Analog handling
- [x] Switch: Add Input buttons
- [x] Switch: Fix Analog handling
- [x] SDL: Fix Analog handling

----------

This rewrites parts of the input API related to gamepads based on how most Gamepads these days are:

When the initial API was designed it was inspired by SDL Joystick. These use 32 arbitrary keys that and are very low level: You can handle any kind of controller with it.

These days the typical controller is usually:

- 2x Analog Sticks
- Every Analog Stick can be pressed down
- 1x Dpad
- 4 Buttons (A,B,X,Y)
- Start, Select
- 2 Left and 2 Right Triggers (one analog)

This is an assumption made by Microsoft (XInput), SDL2 (Game Controller), HTML5 Gamepad API and libretro (Retropad).

I also removed the keyboard input in libretro. They do not forward all keyboard input anymore except if you are in "Focus Mode" which disables the RetroPad. Libretro wants exclusive access to the keyboard, so I fulfill this wish... Is kinda crap but nothing we can do here.

-----------

Changes for SDL1.0:

I kinda hacked this in. We must really drop this :/

- Invented some constants as is not possible to use the Gamepad constants properly here
- HAT is mapped to right analog stick
- This makes it still work for the Wii

----------

New default mapping for Game Controllers to make them usable. (Custom remapping in settings PR)

(Note: Does not work on Linux in Browsers, works everywhere else)

- Dpad: Directions
- Left Stick: Directions
- Right Stick: Numbers matching directions on Numpad
- Left Stick Pressed: 0
- Right Stick Pressed: 5
- A: Decision
- B: Cancel
- X: Cancel
- Y: Shift
- Left Trigger: Debug Through + Debug Menu (Full press)
- Right Trigger: Fast Forward x3 + x10 (Full press) (same as "accelerate in racing games" xD)
- Shoulder Left: Debug event abort
- Shoulder Right: Toggle FPS
- Back: Reset
- Start: _Settings Menu_ (Not used yet)
- 4 Rear Buttons (Almost no controller, Steamdeck has them): +, -, *, /


carstene1ns:
Looks really good so far! Nice workaround to support wii (uses hat) by emulating joystick.

What is missing yet is a way to rename buttons. Currently we have forced keyboard keys for homebrew platforms (which is wrong, but was easy). Wii is a little better, as it has custom mapping, however still using "joy other 4" as name instead of "Button 2". That will break as soon as we have custom mapping scene. I need this for the custom WiiUI.

Ghabry:
> Currently we have forced keyboard keys for homebrew platforms (which is wrong, but was easy)

Yeah that was a lazy solution that must be fixed for the Input Settings scene.

> Wii is a little better, as it has custom mapping, however still using "joy other 4" as name instead of "Button 2".

Yeah, this was a quick workaround.

The lack of button renaming is already bad at the PC. When remapping the user will currently see stuff like "JOY_STICK_LEFT_Y_UP" instead of "Joy Left Y+" in the WIP remapping scene.

We need a new data structure, something like:

```cpp
Array_Of_Struct {
  int8_t key,
  char* internal_name,
  char* readable_name
}
```

Then on startup a port can overwrite the relevant strings in ``readable_name``.

------------

btw, do you want to have individual mappings for the different Wii-Accessoires or will you make "Minus on Wiimote is same as Minus on CC"?

If you want to support individual could ues the provided ``JOY`` things for the Classic Controller (as this is the closest to a controller ^^) and then just 10 ``OTHER`` fields for the Wiimote.

When they can be renamed it doesn't really matter that they are called ``OTHER`` in code.

carstene1ns:
I want to allow using all buttons individually.

src/platform/psp/psp_input_buttons.cpp
```diff
@@ -23,20 +23,19 @@
...
-	};
+		{CANCEL, Keys::JOY_A}, // Cross
+		{DECISION, Keys::JOY_B}, // Circle
+		{FAST_FORWARD, Keys::JOY_X}, // Square
+		{SHIFT, Keys::JOY_Y}, // Triangle
+
+		{N1, Keys::JOY_SHOULDER_LEFT}, // Left trigger
+		{N2, Keys::JOY_SHOULDER_RIGHT}, // Right trigger
+
+		{DOWN, Keys::JOY_DPAD_UP}, // Down
```
fdelapena, line 34: typo?

Ghabry:
As asked by carstene1ns added a stub-api for adding platform specific names that can be queried by the Input Scene:

Example usage (in ``PLATFORM_input_buttons.cpp``):

```cpp
Input::ButtonNamesArray Input::GetInputKeyNames() {
	return {
		{Keys::JOY_DPAD_DOWN, "Dpad Down"},
		{Keys::JOY_A, "Classic Controller A"},
		{Keys::JOY_Z, "Nunchuck Z"},
		{Keys::JOY_TRIGGER_RIGHT_FULL, "Right Trigger Down"},
		{Keys::JOY_OTHER_10, "Balance Board Left"},
		{Keys::JOY_OTHER_20, "Bongo Right"}
	};
}
```

Ghabry:
Wow this was a suuuuper boring task. Moved now all our ports to use a dedicated "Input buttons" instead of depending on input_buttons_desktop.

Also attempted to add suitable input layouts based on the desktop gamepad layout.

Likely broken and has tons of compile errors. Not tested yet but at least the boring part is done...

Ghabry:
Hm this will be hard to test on most ports. There is also other breakage it seems:

- 3ds build throws an assert in AudioSeCache (must be older, not by this build)
- Vita3k freezes with "No device for ."
- Wii build has now completely borked input :/
- Yuzu does not run on my device anymore (too old?)

Ghabry:
Small test case (Input test on the village)

[input.zip](https://github.com/EasyRPG/Player/files/8647434/input.zip)

Grab the last working PR build for the system from Jenkins.

For Vita please test:

- Touch buttons are all properly detected (Logo is cancel)
- D-Pad and left stick for walk
- Right stick (all axis incl diag) for 1-9 without 5 (that would be press stick down, not possible on Vita)
- Other buttons according to psp2_input_buttons.cpp file

For Switch please test:

- Touch buttons are all properly detected (Logo is cancel)
- D-Pad and left stick for walk
- Pressing Left stick is 0
- Pressing Right stick is 5
- Right stick (all axis incl diag) for 1-9
- Other buttons according to switch_input_buttons.cpp file

Ghabry:
Should now work everywhere, _except_:

- Untested: Vita (no working emulator available)
- Wii: Semi-broken. sdl_wii has weird button mappings not matching with wiibrew. carstene1ns will replace this with a ``wii_ui``. Then SDL1 can be removed completely.

Ghabry:
This is now ready.

The Semi-broken Wii input is already a problem in master and some ports have additional features on the shoulder triggers. I will move them to the config scene in the next PR.

src/platform/psp/psp_input_buttons.cpp
Ghabry: uhm, yes. Will fix it but consider this file unsupported.

fdelapena (approved):

src/platform/3ds/3ds_ui.cpp
```diff
@@ -120,6 +116,13 @@ CtrUi::CtrUi(int width, int height, const Game_ConfigVideo& cfg) : BaseUi(cfg)
 	top_image.tex = tex;
 	top_image.subtex = &subt3x;
 
+	// FIXME: Add option to settings Ui
+	if (cfg.stretch_width.Get()) {
```
carstene1ns, line 120: This has nearly no effect, since you disable stretching in `UpdateDisplay()`.
The idea here was to handle this like fullscreen for other platforms. So, no real need for config option here. Same is true for switch and vita, where you only made this a FIXME without option.

carstene1ns (approved):
Looks good to me.

src/platform/3ds/3ds_ui.cpp
Ghabry: Yeah this change makes no sense currently because there must be a function for toggling this. This is similiar to a "keep aspect ratio" setting. Could be even provided for SDL2 in case somebody likes it to have everything stretched... (that's why it is not the same as fullscreen)

Wanted to finish this as part of the Settings PR as I cannot do it in this PR properly (And breaking that feature for a while isn't toooo bad)

carstene1ns:
The only problem I have with it is that it completely breaks Wii input.

PR #2762, opened 2022-04-03
Resampler: Fix crackles when looping the audio

### LINKED ISSUE #2666: Yume 2kki: Certain music crackles/pops audibly when looping
#### Name of the game: 
Yume 2kki

#### Player platform:
Windows

tut
[music.zip](https://github.com/EasyRPG/Player/files/7430799/music.zip)
[easyrpg_log.txt](https://github.com/EasyRPG/Player/files/7430804/easyrpg_log.txt)

Certain music in the game audibly pops when it loops, compared with playing on RPG Maker 2000 where the same music, whilst still making a popping sound if listening carefully, is significantly quieter. Above is a GIF showing how to quickly get to two areas in the game where this occurs, with text of the music filenames appearing when in the area where that music plays. I've also included a zip file that contains the two different music examples - one that is a recording of the music taken in EasyRPG, then for comparison that same music recorded in RPG Maker, and then the original music files that both programs are reading from.

Ghabry:
I "only" needed 4 hours to come up with this solution. Well, after debugging for 4 hours I finally understood how our resampler works. I was also able to simplify the code alot until I discovered where the crackles are coming from: The wrapped_decoder is not looping. 

Refactoring the resampler to get it merged with the decoder will take a while, so I provide here a workaround to get rid of the crackles for now.

------------

For a smooth looping without gaps the wrapped_decoder must be able to loop.

The solution is more of a hack / short-term solution:
All loop-related methods were forwarded to the wrapped_decoder

The better, more complex solution is refactoring the code by merging Decoder+Resampler
to give more control about the decoding process without all this ugly forwarding.

Fix #2666

### DIFF
--- a/src/audio_decoder_base.h
+++ b/src/audio_decoder_base.h
@@ -88,7 +88,7 @@ public:
 	 *
 	 * @return if looping
 	 */
-	bool GetLooping() const;
+	virtual bool GetLooping() const;
 
 	/**
 	 * Enables/Disables audio stream looping.
@@ -97,14 +97,14 @@ public:
 	 *
 	 * @param enable Enable/Disable looping
 	 */
-	void SetLooping(bool enable);
+	virtual void SetLooping(bool enable);
 
 	/**
 	 * Gets the number of loops
 	 *
 	 * @return loop count
 	 */
-	int GetLoopCount() const;
+	virtual int GetLoopCount() const;
 
 	// Functions to be implemented by the audio decoder
 	/**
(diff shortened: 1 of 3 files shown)


fdelapena (approved):

Ghabry:
The result in emscripten is better than before but the "blop" sounds are still kinda noticable.

Could be also a problem of the shity Web Audio API. This must be revisited when the audio stuff is refactored. Still better than before with this PR.

carstene1ns (approved):
Looks good to me.

PR #2774, opened 2022-04-26
Add git revision info to player version

### LINKED ISSUE #87: Use "git describe" to display a nice rev-number

carstene1ns:
Replaces #2070, fixes #87.

Works both on git checkouts and release tarballs.
A checked out git tag behaves the same like a release tarball.
`make distcheck` works

For our builds we can easily inject a tag (like CI, PR, AUR, OBS) with
`-DPLAYER_VERSION_APPEND` for cmake or `--enable-append-version` for autotools.


.github/workflows/stable-compilation.yml
```diff
@@ -52,12 +52,13 @@ jobs:
 
       - name: Compile
         run: |
+          VER="(GA, `date +%Y-%m-%d`)"
```
fdelapena, line 55: Looking good! Is there a simple way to override this date value later to support reproducible builds? Or an alternate, more deterministic way by getting the tag/commit date from git, for example (the release date could be different, however).

fdelapena (approved):

.github/workflows/stable-compilation.yml
carstene1ns: Had a look at this before. Currently the date is just the default string if not provided. Maintainers can just overwrite it to empty for example. 

Some time in the future we can add support for `SOURCE_DATE_EPOCH`, which adds up on that. Then `PLAYER_DATE` could get introduced to implement this.

CMakeLists.txt
```diff
@@ -447,6 +448,46 @@ if (NOT CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
...
+			string(APPEND PLAYER_VERSION_GIT "-${GIT_DIRTY}")
+			set(GIT_DIRTY_MESSAGE ", you have uncommitted changes")
+		endif()
+
+		message(STATUS "Found git info: ${GIT_COMMITS} commits since tag"
+			" \"${GIT_TAG}\", object hash is ${GIT_HASH}${GIT_DIRTY_MESSAGE}")
+	endif()
+endif()
+
+string(TIMESTAMP PLAYER_DATE "(%Y-%m-%d)")
```
carstene1ns, line 480: CMake supports forcing timestamp since 3.8:
https://cmake.org/cmake/help/latest/command/string.html#timestamp

So this works ootb.

Ghabry (approved):

carstene1ns:
Do not merge yet.

.github/workflows/stable-compilation.yml
carstene1ns: Using an autoconf archive macro solves this for now.

CMakeLists.txt
```diff
@@ -1159,9 +1174,11 @@ set(DOXYGEN_STATUS "Unavailable")
 if(DOXYGEN_FOUND)
 	set(DOXYGEN_STATUS "Available (target \"doc\")")
 	# fake autotools variables
+	set(PACKAGE_VERSION ${PLAYER_VERSION})
```
carstene1ns, line 1177: This has been successive broken in 2016 and 2018 `m(`

doxygen used the version of the expat package here (because of config file providing `PACKAGE_VERSION`)

carstene1ns:
test this please

CMakeLists.txt
```diff
@@ -898,7 +951,10 @@ if(${PLAYER_BUILD_EXECUTABLE} AND ${PLAYER_TARGET_PLATFORM} STREQUAL "SDL2")
 		set_target_properties(${EXE_NAME} PROPERTIES WIN32_EXECUTABLE $<NOT:CONFIG:Debug>)
 
 		# Add resources
-		target_sources(${EXE_NAME} PRIVATE "resources/player.rc")
+		string(REPLACE "." "," RC_VERSION ${PROJECT_VERSION})
```
carstene1ns, line 954: Something wrong here.
`PRODUCTVERSION 0,7` missing patch and tweak.

carstene1ns:
This is ready now.

Ghabry:
Not a full review due to lack of time.
Wow you really pushed this forward. Even updates all the version information in all of the resource files etc. Great job!

Also the oldest issue closed :tada:

PR #2776, opened 2022-05-03
Fixup for building libretro core and Windows port

carstene1ns:
- The `player.rc` file lists the resource files from current directory.
  This was not seen in the PR, since the builder does an in-source build.

- retroarch should show a full version string now (untested).

- `CMAKE_FORCE_COLORS` is a developer option, building with ninja disables all colors from gcc output, which makes it less readable. This enables them back again, like with Makefiles.


fdelapena (approved):

CMakeLists.txt
```diff
@@ -957,6 +968,7 @@ if(${PLAYER_BUILD_EXECUTABLE} AND ${PLAYER_TARGET_PLATFORM} STREQUAL "SDL2")
 		string(REPLACE "." "," RC_VERSION ${PROJECT_VERSION})
 		string(TIMESTAMP RC_YEAR "%Y")
 		configure_file(resources/windows/player.rc.in resources/windows/player.rc @ONLY)
+		file(COPY resources/windows DESTINATION resources PATTERN "*.in" EXCLUDE)
```
Ghabry, line 971: will try to figure out if this is solvable without the copy

Ghabry: According to the docs (which is peak MSDN again translating names of Resource fields xD) the resources can also be absolute paths.

So this should work:
```
23456 ICON "@CMAKE_CURRENT_SOURCE_DIR@/resources/windows/player.ico"
```

Same for the XML line (which is currently MinGW only but will need the manifest for HighDPI later):

```
1 24 "@CMAKE_CURRENT_SOURCE_DIR@/resources/windows/player.xml"
```

CMakeLists.txt
```diff
@@ -20,6 +20,17 @@ set(CMAKE_CXX_EXTENSIONS ON)
...
+# ninja/parallel builds disable output colors, forcing them here
+option(CMAKE_FORCE_COLORS "Always produce ANSI-colored output" OFF)
+if(${CMAKE_FORCE_COLORS})
+	if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+		add_compile_options("-fdiagnostics-color=always")
+	elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+		add_compile_options("-fcolor-diagnostics")
+	endif()
+	set(DOCTEST_COLOROPT "--force-colors")
+endif()
```
Ghabry, line 32: FYI: CMake just added https://github.com/Kitware/CMake/blob/cbd36eac23628f74ae3b50f9d1b8a0478c3e317a/Help/release/dev/color-diagnostics.rst#id1  which introduces ``CMAKE_COLOR_DIAGNOSTICS``.

Not usable yet as this is a dev feature.

carstene1ns: I guess I can wait until 3.24, other idea is to wrap this into a CMAKE_VERSION conditional and use the official name, so it will work for older versions too. At least for doctest there needs to be a way to set it.

Ghabry: Maybe keep your code and call it ``CMAKE_COLOR_DIAGNOSTICS`` and wrap it already in a version check so the var name stays the same across versions.

carstene1ns: This is exactly what I meant :D

carstene1ns: applied.

Ghabry:
Jenkins: Test this please

Made the Windows PR builds consistent with the normal one (out of tree) and enabled libretro build for 64bit Win-PR-Builder to ensure at least minimal coverage here :)

Ghabry:
Jenkins: Test this please <3

src/platform/libretro/libretro_ui.cpp
```diff
@@ -565,11 +565,8 @@ RETRO_API unsigned retro_api_version() {
  * Can be called at any time, even before retro_init(). */
 RETRO_API void retro_get_system_info(struct retro_system_info* info) {
 	memset(info, 0, sizeof(*info));
-	info->library_name = "EasyRPG Player";
-	#ifndef GIT_VERSION
-	#define GIT_VERSION ""
-	#endif
-	info->library_version = PLAYER_VERSION GIT_VERSION;
+	info->library_name = GAME_TITLE;
+	info->library_version = Version::GetVersionString(true, false).c_str();
```
Ghabry, line 569: C-Api crap: This returns a pointer to a temporary and will crash.

Possible workaround:

```cpp
static std::string version = Version::GetVersionString(true, false);
info->library_version = version.c_str();
```

Pointers to ``.c_str()`` are also kinda unsafe but this works when the std::string is not modified afterwards.

carstene1ns: I figured it would not work :D 
Thank you for the fix.
Other solution would be replicating the definitions in cmake like for version.cpp.

Ghabry (approved):

PR #2787, opened 2022-05-20
FileSystem: Support / and drive:/

Ghabry:
_Havn't done any hardware tests yet, only operated on ``/`` and ``--project-path romfs:/`` to verify this behaviour_

When using ``/`` as the ``--project-path`` the ``/`` was removed making this _empty string_ which is then handled as the working directory (is silently converted to ``.`` at the important places). This is completely wrong.

Similiar problem with drive letters (``drive:/``). Windows does not care if you use ``C:`` or ``C:/``, which is what I used for testing here.
Unfortunately ctrulib and others are more strict here and want the ``/``.

-------

Fixing this was not _that_ terrible.

Functions that are drive-aware now: ``MakeCanonical``, ``GetPathAndFilename``, the substring calculation in ``FindFile``.

Also added Unit Tests to verify this behaviour.


carstene1ns (approved):
The unit tests prove it works ;-)

fdelapena (approved):

PR #2793, opened 2022-05-24
Fixup CMake generator expressions

carstene1ns:


carstene1ns:
The reason why the message worked before was an SDL bug, video was not deinitialized.

To fix this properly we should likely do it in JS after player exits.

fdelapena (approved):

Ghabry:
temporarely fixed the vcpkg installation.

Proper fix follows during the week.

Ghabry (approved):
The generator syntax is really terrible :sweat_smile:

PR #2796, opened 2022-05-28
Migrate AmigaOS4/Morphos to CMake

carstene1ns:
This is completely untested, but the `Makefile` was not adapted before (→ so not working anyway).
However, I took great care to carry everything over.

- AmigaOS4 has the same custom SDL1 route like Wii
- MorphOS uses SDL2
- AROS may work, SDL2 port is unfinished (at least this is what google tells me)


Ghabry:
Hard to review obviously. Could ask @BSzili for help

BSzili:
I'll try to look into this later, I think @BeWorld2018 used autotools to build the MorphOS version.

carstene1ns:
https://github.com/BeWorld2018/Player

Looks like we can upstream at least the patch. Looking at the build script, this could be easier in cmake

For reference, I used these https://github.com/AmigaPorts/AmigaCMakeCrossToolchains

Should work, when the toolchain is installed.

carstene1ns:
Patches incorporated.

Ghabry (commented):
Ugh another platform that uses absolute paths that do not look like drive letters 🙈. Well if they exist DirectoryTree will not fall back to working directory.

I agree with " merge and forget" when all builders are green

Ghabry (approved):

PR #2797, opened 2022-05-29
Maniac Patch: Final contribution

### LINKED ISSUE #491: Support more keys for KeyInputProc
It was suggested to support more than 0-9, Shift, NumPad.
Because keycodes are higher then the original return value of this event (highest is 29) it will probably not even break games ;).

Ghabry:
This concludes my Maniac Patch contributions for now.

New stuff:

- Most Conditional Branch enhancements except "Game was just loaded"
- The Keyboard part of KeyInputProcEx (not very useful, see it as an alternative to Ineluki). Joypad mapping part not supported as we have our own Joypad code and remapper (soon ^^')
- Control Global Save Command
- Control Variable enhancements including the Expression parser
  - Unsupported are assignment operators (+=, etc.) because they assign in expressions which violates the purpose of the ControlVariables command
  - Array operations are unsupported as I consider them not too useful

-------------

The final one for later this year is Show String Picture. This is lots of work to add.

Fix #491


Ghabry:
Needs some more testing but is otherwise finished.

fdelapena (approved):

gadesx:
I've heard that the latest maniacs patch it's a bit broken, atleast the editor.
I don't know what release version works fine to use "show string picture"
Latest actually: patch_maniacs_211010

carstene1ns (approved):
I have no idea about manic patch games, that use all features to test.

Nice refactor, though.

PR #2804, opened 2022-06-09
Improve DirectoryTree & FileFinder_RTP performance

Ghabry:
As observed by @BlisterB and @carstene1ns the RTP detection is incredible slow on Android but even on other systems we are wasting time in the detector.

Added three layers of _speed_ to make the RTP detection on startup faster:

1. When the RTP folder is empty the folder is considered invalid and not scanned (Android will hit this one)
2. When not empty the RTP-type-scan will abort when for a certain RTP type there are 10 missing files (this makes RTP2k detect ``30%`` faster, RTP2k3 detect ``59%`` faster).
3. When a folder is missing this is cached making the lookup much faster  because the filesystem is not hit again (this makes the ``RTP::Detect`` Unit Test where almost all folders are missing and "abort after 10 misses" is disabled ``20%`` faster). On slow systems with shitty IPC (hello Android and 3DS) the speedup in this rare error case should be gigantic.

---------

Redesigned the data structure of the DirectoryTree from std::unordered_map to sorted std::vector as suggested by fmatthew in #2306

This gives a ``5%`` speedup for ``FindFile`` lookups in my benchmarks and reduces memory usage of the tree by ``15%``. Could be more on systems with a smaller cpu cache.

---------

Faster startup when a RTP exists (or the folder is empty) should be noticable on slower systems. Faster ``FindFile`` likely not (microbenchmarks...).

---------

Android loading times (old):

- RTP Off: 2s 
- RTP On & installed: 5s
- RTP On & empty: 16s

Android loading times this branch: _Note that the PR build has debug on so you will not hit these times_

- RTP Off: 2s
- RTP On & installed: 5s (could be a few ms faster but can't properly measure this on Android)
- RTP On & empty: **2s**

### DIFF
--- a/src/filefinder_rtp.cpp
+++ b/src/filefinder_rtp.cpp
@@ -178,6 +178,12 @@ void FileFinder_RTP::AddPath(StringView p) {
 	using namespace FileFinder;
 	auto fs = FileFinder::Root().Create(FileFinder::MakeCanonical(p));
 	if (fs) {
+		auto files = fs.ListDirectory();
+		if (files->size() == 0) {
+			Output::Debug("RTP path {} is empty, not adding", p);
+			return;
+		}
+
 		Output::Debug("Adding {} to RTP path", p);
 
 		auto hit_info = RTP::Detect(fs, Player::EngineVersion());
--- a/bench/bitmap.cpp
+++ b/bench/bitmap.cpp
@@ -272,7 +272,7 @@ static void BM_ToneBlit(benchmark::State& state) {
 	auto rect = src->GetRect();
 	auto tone = Tone(255,255,255,128);
 	for (auto _: state) {
-		dest->ToneBlit(0, 0, *src, rect, tone, opacity, false);
+		dest->ToneBlit(0, 0, *src, rect, tone, opacity);
 	}
 }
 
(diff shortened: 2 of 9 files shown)


Ghabry:
@BlisterB I think the times are good enough to have RTP enabled by default now. It is still few seconds slower (when installed) but when not installed it will not impact startup.

fdelapena (approved):

Ghabry:
I could delay the RTP check until the first asset is requested.
Then you would get a hang while in-game but only for RTP games. Would this be preferred?

src/filefinder_rtp.cpp
```diff
@@ -178,6 +178,12 @@ void FileFinder_RTP::AddPath(StringView p) {
 	using namespace FileFinder;
 	auto fs = FileFinder::Root().Create(FileFinder::MakeCanonical(p));
 	if (fs) {
+		auto files = fs.ListDirectory();
+		if (files->size() == 0) {
```
carstene1ns, line 182: This change alone likely fixes the issue I had... 😉

Ghabry: Yes. The rest is just for hypothetical other things or such a small improvement that you need tools to see. 😅

carstene1ns:
I think doing the RTP detection while the game is running will involve stutters in audio. So, merging.

PR #2819, opened 2022-06-26
Use 64 bit z-layer

Ghabry:
Found in Il Mito. Reported by mokodeoz. Thanks!

------

This changes the Z-layer value from int32_t to uint64_t.

For correct render order of events it is necessary to put in Y, X and the ID in the Z value and this simply does not fit in ~28 Bit properly.

Now Z-Value is:

- Upper 8 bit = Layer
- Next 8 Bit = Unused yet :)
- Next 48 Bit: Layer specific, Event uses 16 Bit Y, 16 Bit X and 16 Bit ID.

It is possible that I missed a Z variable somewhere which will then overflow. So please test some games if you find rendering order bugs.

------

Performace impact: Likely a non-relevant amount of nanoseconds slower on 32 bit hardware

Also optimized the Tile cache structure, is now 4 byte instead of 8.

### DIFF
--- a/src/drawable.h
+++ b/src/drawable.h
@@ -137,6 +137,7 @@ inline void Drawable::SetVisible(bool value) {
 	_flags = value ? _flags & ~Flags::Invisible : _flags | Flags::Invisible;
 }
 
+// Upper 8 bit are reserved for the layer 
 static constexpr uint64_t z_offset = 64 - 8;
 
 // Lower 56 Bit are free to use
--- a/src/drawable.h
+++ b/src/drawable.h
@@ -27,33 +27,13 @@ class Drawable;
 template <typename T>
 static constexpr bool IsDrawable = std::is_base_of<Drawable,T>::value;
 
-enum Priority {
-	Priority_Background = 5 << 24,
-	Priority_TilesetBelow = 10 << 24,
-	Priority_EventsBelow = 15 << 24,
-	Priority_Player = 20 << 24, // In Map, shared with "same as hero" events
-	Priority_Battler = 20 << 24, // In Battle (includes animations)
-	Priority_TilesetAbove = 25 << 24,
-	Priority_EventsAbove = 30 << 24,
-	Priority_EventsFlying = 35 << 24,
-	Priority_Weather = 36 << 24,
-	Priority_Screen = 40 << 24,
-	Priority_PictureNew = 45 << 24, // Pictures in RPG2k Value! and RPG2k3 >=1.05, shared
-	Priority_BattleAnimation = 50 << 24,
-	Priority_PictureOld = 55 << 24, // Picture in RPG2k <1.5 and RPG2k3 <1.05, shared
-	Priority_Window = 60 << 24,
-	Priority_Timer = 65 << 24,
-	Priority_Frame = 70 << 24,
-	Priority_Transition = 75 << 24,
-	Priority_Overlay = 80 << 24,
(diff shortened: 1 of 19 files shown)


zell180:
Ok the bug seems fixed but two new problem happens:
1) slow perfomance reach title screen, load menu, return title screen after press F12
2) The screen posted below
Immagine
[Save08.zip](https://github.com/EasyRPG/Player/files/8990354/Save08.zip)

To reproduce the bug: Talk with Ikki (the boss in front of you) and select "Parla" in Battle System Option

Ghabry:
@zell180 

Thanks. Is fixed. Missed a variable.

Ghabry:
Thanks. Another stupid oversight: Implicit fallthrough in switch-case -_-

carstene1ns:
GCC actually warns in this case...

Ghabry:
But not with the default CMake Debug setting. ^^'

carstene1ns:
Debug/optimization level != Warnings enabled.
We build on jenkins with `-Wall -Wextra` and have chatched this: https://ci.easyrpg.org/job/player-linux-pr/5602/console

zell180:
now everything seems work correctly

src/game_character.cpp
```diff
@@ -116,7 +116,11 @@ int Game_Character::GetScreenZ(bool apply_shift) const {
 		z = Priority_EventsAbove;
 	}
 
-	z += GetScreenY(apply_shift, false);
+	Drawable::Z_t y = static_cast<Drawable::Z_t>(GetScreenY(apply_shift, false));
+
+	Drawable::Z_t x = static_cast<Drawable::Z_t>(GetScreenX(apply_shift));
+
+	z += (y << 32) + (x << 16);
```
carstene1ns, line 123: again magic numbers

src/game_player.cpp
```diff
@@ -64,10 +64,11 @@ lcf::rpg::SavePartyLocation Game_Player::GetSaveData() const {
 	return *data();
 }
 
-int Game_Player::GetScreenZ(bool apply_shift) const {
+Drawable::Z_t Game_Player::GetScreenZ(bool apply_shift) const {
 	// Player is always slightly above events
 	// (and always on "same layer as hero" obviously)
-	return Game_Character::GetScreenZ(apply_shift) + 1;
+	// Set all "ID bits" (first 16) to ensure this
```
carstene1ns, line 70: magic number, no explanation
(sure is `1 << 16`, former `z_mask`)

src/drawable.h
```diff
@@ -157,4 +137,30 @@ inline void Drawable::SetVisible(bool value) {
 	_flags = value ? _flags & ~Flags::Invisible : _flags | Flags::Invisible;
 }
 
+static constexpr uint64_t z_offset = 64 - 8;
```
carstene1ns, line 140: magic number: comment about 8 bit tile cache?

Ghabry: This is unrelated to the tile cache. It is used below by the layers to enforce the right rendering order (Is the same as the hardcoded ``<< 24`` in the old code, just now for 64 bit instead of 32)

Ghabry:
@carstene1ns
I added some comments to the marked locations. Hope this makes it more clear :).

Most of that stuff is shifting bits around to squeeze multiple smaller numbers into one big 64 bit variable.

fdelapena (approved):

PR #2825, opened 2022-07-11
Translation/Encoding improvements

### LINKED ISSUE #2724: Confusion between Simplified Chinese and Japanese fonts.

### LINKED ISSUE #2731: Inconsistent code logic when setting encoding in local .ini file.
#### Name of the game:
LcdDem V0.03 CHS

#### Player platform:
Web

 - The `RPG_RT.ini` file (you can find it in the game directory)

"LcdDem V0.03 CHS" seems has some problem and the web browser cannot identify the correct encoding. (It is weird since in rpgmaker and the latest easyrpg_windows_x64 version is fine) So I followed [https://wiki.easyrpg.org/user/player/special-features](https://wiki.easyrpg.org/user/player/special-features) to setting the encoding in the .ini file mannully. Then the game itself is fine, but the multiplayer menu and chatbox remain in English.

image

I use Player::IsCP936() as the single source of truth to handle the in-game translation for menu and chatbox.
So I believe there could be some inconsistent code logic when setting encoding in local .ini file.

Ghabry:
Invented a new field ``Code`` in the meta.ini to provide the locale/iso-code of the language.

Currently supported:

- ja_JP
- ko_KR
- zh_TW
- zh_CN
- ru_RU

Fixes problems with the font selection (full width cyrillic for CJK fonts, Japanese in Chinese etc.). Allows to pick the proper input scene.

When not specified it will use the current encoding.

When an unknown locale is specified result will be Not Russian and Not CJK (Western Europe?).

Fix #2731
Fix #2724


Ghabry:
Showcase (Had to restart the Player multiple times because the languages are sometimes not properly loaded, but otherwise it works) :D

_Note_: In the language box the text renderer used is always the current active locale, not the one of the language. So the rendering here is incorrect. Fixing this is lots of work so I'm ignoring this minor problem for now. Sorry :)

Default language: Japanese (Full width cyrillic and Japanese Kanji used instead of Chinese), this is unfixable without lots of work.

screenshot_0


Before: Russian uses full-width everywhere

screenshot_1

After: Russian with correct code

screenshot_2

While Russian is active the language box is also rendered using Russian so no full-width (just a side-effect of the implementation)

screenshot_3

Before: Chinese using incorrect glyphs from Japan

screenshot_4

After: Chinese with proper glyphs

screenshot_6

While Chinese is active the language box is also rendering using the correct chinese glyphs (just a side-effect of the implementation)

screenshot_7

Ghabry:
The ugly umlauts was a problem with ttyp0. I tried various versions until Player 0.6 and the all have this problem. Somehow I never saw this issue before. Maybe most German games use the other font setting (which uses RMG2000).

fdelapena (approved):

src/translation.cpp
```diff
@@ -180,14 +189,10 @@ void Translation::SelectLanguageAsync(FileRequestResult* result, StringView lang
 	if (!ParseLanguageFiles(lang_id)) {
 		return;
 	}
-	current_language = ToString(lang_id);
-
-	// We reload the entire database as a precaution.
-	Player::LoadDatabase();
```
Ghabry, line 192: Oh I deleted too many lines here. Fixing this asap

carstene1ns (approved):
Allrighty then!

PR #2833, opened 2022-08-20
Update Android translation

Ghabry:
There are some translations by now so lets add them upstream to prevent conflicts when editing Android files :)


fdelapena (approved):

Mimigris:
> There are some translations by now so lets add them upstream to prevent conflicts when editing Android files :)

When checking the French translation I've seen some errors in the string.xml so I'm gonna report them here:

Line 33
Raffraichir -> Rafraîchir

Line 42
ont été placé -> ont été placés

Line 68
Shift -> Maj

Ghabry:
Imo these translation updates could be squashed and merged. There is not much value in having them as individual commits

PR #2855, opened 2022-10-29
Add freetype font rendering

### LINKED ISSUE #2857: TTF color font support
There were suggestions to add color font support to the FreeType renderer.

This is possible and actually not much work and I showed with "color ExFont" that our code is suitable to add such stuff.

Problem is more the available fonts and the tooling: The Emoji in typical fonts are far too large (128x128) and downscaling them to 12x12 will look terrible. So own fonts should be provided which have 12x12 color glyphs as a tech demo.

The only tool I was able to find is FontTools for Python https://pypi.org/project/fonttools/ . Everything else appears to be commercial.

How to do Color Emoji: There are three _standards_:

- Microsoft uses CPAL/COLR. These are vector graphics (SVG) embedded into the font. (Limited FreeType support, likely because SVG rasterisation is complecated)
- Apple uses [SBIX](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6sbix.html). FreeType only supports PNG glyphs.
- Google uses CBDT/CBLC which appears to be an extension of EBDT/EBLC table which is used for Monochrome glyphs. Freetype supports this.

-------------

Note to self: It was suggested to support color TTF fonts. Problem is that no font with such small image size exists. The ones from Google etc. don't work.

How to get color out with FreeType:

- Instead of ``FT_Set_Pixel_Sizes`` call ``FT_Select_Size(face, 0)``
- Instead of ``FT_LOAD_MONOCHROME`` use ``FT_LOAD_COLOR``
- Instead of ``FT_RENDER_MODE_MONO`` use ``FT_RENDER_MODE_NORMAL``
- Check for color font with ``FT_HAS_COLOR(face)``
- To render it something like this should work:

```
if (ft_bitmap->pixel_mode == FT_PIXEL_MODE_BGRA) {
    bm = Bitmap::Create(ft_bitmap->buffer, width, height, 0, format_B8G8R8A8_a().format());
```

--------

cbdt font dumper:

```
import sys

from fontTools.ttLib import TTFont

font = TTFont(sys.argv[1])

cbdt = font["CBDT"]

for glyphName, bitmap in cbdt.strikeData[0].items():
  print(glyphName)
  bitmap.decompile()
  filename = f"glyph-{glyphName}.png"
  with open(filename, "wb") as f:
    f.write(bitmap.imageData)
```

Ghabry:
This adds experimental freetype rendering support.

Still marked as draft as I must test some stuff. But the basic rendering works.

When fonts are provided as "Font/Font.ttf" (gothic) or "Font/Font2.ttf" (mincho) these fonts are used instead of the built-in. Should also work for the web player and translations can provide custom fonts.

Also supports complex stuff like Thai (not using Harfbuzz though as Freetype is good enough to render Thai properly).

~~TODO: Testing testing testing. Fallback logic?~~

screenshot_0
screenshot_1
screenshot_2

Fix #2857

### DIFF
--- a/src/filefinder.cpp
+++ b/src/filefinder.cpp
@@ -470,6 +470,12 @@ Filesystem_Stream::InputStream FileFinder::OpenSound(StringView name) {
 	return open_generic("Sound", name, args);
 }
 
+Filesystem_Stream::InputStream FileFinder::OpenFont(StringView name) {
+	auto FONT_TYPES = Utils::MakeSvArray(".ttf", ".ttc", ".otf", ".fon");
+	DirectoryTree::Args args = { MakePath("Font", name), FONT_TYPES, 1, false };
+	return open_generic("Font", name, args);
+}
+
 bool FileFinder::IsMajorUpdatedTree() {
 	auto fs = Game();
 	assert(fs);
--- a/src/filefinder.cpp
+++ b/src/filefinder.cpp
@@ -61,6 +61,13 @@ namespace {
 	std::shared_ptr<Filesystem> root_fs;
 	FilesystemView game_fs;
 	FilesystemView save_fs;
+
+	constexpr const auto IMG_TYPES = Utils::MakeSvArray(".bmp",  ".png", ".xyz");
+	constexpr const auto MUSIC_TYPES = Utils::MakeSvArray(
+			".opus", ".oga", ".ogg", ".wav", ".mid", ".midi", ".mp3", ".wma");
+	constexpr const auto SOUND_TYPES = Utils::MakeSvArray(
+			".opus", ".oga", ".ogg", ".wav", ".mp3", ".wma");
+	constexpr const auto FONTS_TYPES = Utils::MakeSvArray(".ttf", ".ttc", ".otf", ".fon", ".fnt", ".bdf");
 }
 
 FilesystemView FileFinder::Game() {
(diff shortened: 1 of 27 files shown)


Ghabry:
Fixed the rendering of FON. The metrics are wrong but this is a font issue. Will see if I can detect the font and then just force correct metrics...

screenshot_3

Ghabry:
Added a detection for RM2000 and RMG2000: When these fonts are used the offset is overwritten with a good one.

screenshot_4

BlisterB:
As discussed in the Discord : 
- The file "Font.xxx" doesn't seem to be detected. When selected, EasyRPG uses the default font. "Font2.xxx" works correctly.

I haven't saw any other problem for now, everything seem to work nicely. Awesome feature. ^^

jetrotal:
More discussion from discord:


Some fonts are not optimized for EasyRPG proportion Settings:

image
[Fonts Used.zip](https://github.com/EasyRPG/Player/files/9900098/Fonts.Used.zip)
FindersKeepers font is better displayed at `size = 16` than at the default `size = 12`


To solve that, I'd like to suggest an optional settings layer, that could be something like:
```JSON
{
  "fontConfig": {
    "0": {
      "source": "./MSGOTHIC.TTF",
      "size": 12,
      "spacing": "auto",
      "etc": "etc"
    },
    "1": {
      "source": "auto",
      "size": "auto",
      "spacing": "auto",
      "etc": "etc"
    }
  }
}
```

I would like to also sugest `Font Fallback for Missing Glyphs`.

I Tried a font called Material Design Icons, that adds a lot of custom characters:
image

But the lack of `A,B,C...` glyphs made the font blank:
image


Maybe, A character going back to the default font whenever it fails loading its glyph could help on this.

---------------------------

Last feature I'd like to suggest is a changing font mid Text input.

like in this mockup:
```
Helen: Yo Don Miguel!
I have got \fon[1]Cool Compass\fon[0]!
I can sell it to ya!
```
image

Ghabry:
All this font config stuff is a good idea but honestly I would prefer to have this as part of the database which means this is a job for the editor.

Just want to avoid having lots of config files for stuff that should be in the database. But you are totally right that we need something like this.

For now I will only add a fallback to our built-in fonts when a glyph is missing.

Ghabry:
> The file "Font.xxx" doesn't seem to be detected

This is fixed now.

Also not crasing anymore when closing the player ;)

Ghabry:
Note to self: It was suggested to support color TTF fonts. Problem is that no font with such small image size exists. The ones from Google etc. don't work.

How to get color out with FreeType:

- Instead of ``FT_Set_Pixel_Sizes`` call ``FT_Select_Size(face, 0)``
- Instead of ``FT_LOAD_MONOCHROME`` use ``FT_LOAD_COLOR``
- Instead of ``FT_RENDER_MODE_MONO`` use ``FT_RENDER_MODE_NORMAL``
- Check for color font with ``FT_HAS_COLOR(face)``
- To render it something like this should work:

```
if (ft_bitmap->pixel_mode == FT_PIXEL_MODE_BGRA) {
    bm = Bitmap::Create(ft_bitmap->buffer, width, height, 0, format_B8G8R8A8_a().format());
```

Ghabry:
The system graphic color mask is now correctly applied:

screenshot_2

Bonus feature: Colorful ExFonts

screenshot_3

Open issues:

- Ugly half-width space in Help Window
- FreeType requires Fallback to built-in fonts

Ghabry:
This is now ready.

There is one minor breaking change: ExFont was relocated to "Font/ExFont" for consistency. Emscripten will enforce this but local Player will check both locations.

All changes outside of the FreeType code will return the same font metrics as before so no code behaviour should change when no FreeType Font is loaded.


New features:

- FreeType font loading (obviously)
- Font/Font and Font/Font2 are loaded depending on Font in the system graphic setting (currently hardcoded to 12 px size, font config should be an editor setting later)
- Support for TTF and Pixel Fonts (FON, BDF, etc.). Everything that FreeType supports
- _Complex shaping (Harfbuzz) is currently not supported due to how our code is designed but Thai appears to work good enough without it_
- Support for RGB Color in ExFont. Like you know with Emojis this will not recolor the glyph with the system graphic but the drop-shadow is added.
- Support for Color Glyphs in Fonts (the same feature Color Emojis use). **EXPERIMENTAL (!)**. There are no fonts in-the-wild that have 12x12 color glyphs. I'm currently preparing a Python script to make generating of such fonts possible. :)
- FreeType falls back to built-in Bitmap when a glyph is not found

The renderer likely has bad performance because there is no glyph caching etc. involved. If this becomes an issue this can be improved later.

Also I do not recommend using this on any low memory platforms because the entire font file is loaded into memory. So this wastes many megabyte.

Ghabry:
Example of creative (ab)use ;)

screenshot_9

src/player.cpp
```diff
@@ -993,6 +1000,24 @@ void Player::LoadDatabase() {
 	}
 }
 
+void Player::LoadFonts() {
+	Font::SetDefault(nullptr, true);
+	Font::SetDefault(nullptr, false);
```
fdelapena, line 1005: 🤔

src/filefinder.cpp
```diff
@@ -61,6 +61,13 @@ namespace {
 	std::shared_ptr<Filesystem> root_fs;
 	FilesystemView game_fs;
 	FilesystemView save_fs;
+
+	constexpr const auto IMG_TYPES = Utils::MakeSvArray(".bmp",  ".png", ".xyz");
+	constexpr const auto MUSIC_TYPES = Utils::MakeSvArray(
+			".opus", ".oga", ".ogg", ".wav", ".mid", ".midi", ".mp3", ".wma");
+	constexpr const auto SOUND_TYPES = Utils::MakeSvArray(
+			".opus", ".oga", ".ogg", ".wav", ".mp3", ".wma");
+	constexpr const auto FONTS_TYPES = Utils::MakeSvArray(".ttf", ".ttc", ".otf", ".fon", ".fnt", ".bdf");
```
fdelapena, line 70: .woff and .woff2 OpenType formats are useful for significantly smaller emscripten asset downloads (may require extra freetype dependencies, e.g. brotli).

fdelapena (approved):

src/player.cpp
Ghabry: The ``bool`` is for "mincho or gothic" :D. This strange API is used everywhere in Font.

Ghabry:
Jenkins: Test this please

Ghabry:
(This is now really, really finished. Backported all issues I found while working on Harfbuzz)

src/filefinder.cpp
Ghabry: The FileFinder code aborts early when the directory is missing so for most games adding more extensions won't hurt. I add a commit :)

PR #2859, opened 2022-11-14
Add Harfbuzz support

Ghabry:
~~WIP PR to test harfbuzz shaping support :)~~

First I refactored the ``Font`` and ``Text`` Api: ``Font`` is now only responsible for rendering single glyphs (``char32_t``). For any high level usage the ``Text`` Api must be used.

The exception to "single glyph" is the shaping Api which takes a ``u32string`` because this conditionally converts codepoints depending on the shaping rules. When Harfbuzz is enabled the ``Text`` Api will automatically handle this.

A special case is ``Window_Message``. It renders glyph-by-glyph instead of all-in-one-go. This requires usage of the low level ``Font`` Api directly because the shaping information must be known.

### DIFF
--- a/src/text.h
+++ b/src/text.h
@@ -120,7 +120,7 @@ namespace Text {
 	 * For continuous rendering use the "width" property.
 	 *
 	 * @param font the font used to render.
-	 * @param ch the character to mesaure.
+	 * @param glyph the character to mesaure.
 	 * @param is_exfont if true, treat ch as an exfont character. Otherwise, a utf32 character.
 	 *
 	 * @return Rect describing the rendered string boundary
--- a/src/text.h
+++ b/src/text.h
@@ -54,7 +54,7 @@ namespace Text {
 	 *
 	 * @return Where to draw the next glyph when continuing drawing. See Font::GlyphRet.advance
 	 */
-	Point Draw(Bitmap& dest, int x, int y, Font& font, const Bitmap& system, int color, StringView text, Text::Alignment align = Text::AlignLeft);
+	Point Draw(Bitmap& dest, int x, int y, const Font& font, const Bitmap& system, int color, StringView text, Text::Alignment align = Text::AlignLeft);
 
 	/**
 	 * Draws the text onto dest bitmap with given parameters. Does not draw a shadow.
@@ -68,7 +68,7 @@ namespace Text {
 	 *
 	 * @return Where to draw the next glyph when continuing drawing. See Font::GlyphRet.advance
 	 */
-	Point Draw(Bitmap& dest, int x, int y, Font& font, Color color, StringView text);
+	Point Draw(Bitmap& dest, int x, int y, const Font& font, Color color, StringView text);
 
 	/**
 	 * Draws the character onto dest bitmap with given parameters.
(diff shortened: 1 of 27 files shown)


Ghabry:
The Yume Nikki Online project is asking about Arabic text.

What impresses me here is that Harfbuzz can render right-to-left scripts correctly without changing a single line of code :thinking: 

a

Ghabry:
Somebody asked for [Times New Bastard](https://github.com/weiweihuanghuang/Times-New-Bastard). This is a font that makes every 7th character Sans Serif (Ligature magic).

When you count the glyphs you can see that it works. Though the font looks terrible. No good low res glyphs provided by the font:

screenshot_2

Ghabry:
Linking libretro on macOS is broken since a while. Fits with Harfbuzz as its the issue:

```diff
diff --git a/builds/cmake/Modules/FindHarfbuzz.cmake b/builds/cmake/Modules/FindHarfbuzz.cmake
index 1f5a1fd0..654600e0 100644
--- a/builds/cmake/Modules/FindHarfbuzz.cmake
+++ b/builds/cmake/Modules/FindHarfbuzz.cmake
@@ -64,6 +64,30 @@ if(HARFBUZZ_FOUND)
			INTERFACE_INCLUDE_DIRECTORIES "${HARFBUZZ_INCLUDE_DIRS}"
			INTERFACE_LINK_LIBRARIES Freetype::Freetype
			IMPORTED_LOCATION "${HARFBUZZ_LIBRARY}")
+
+		if(APPLE)
+			# Framework list taken from Harfbuzz CMakeLists
+			if(IOS)
+				find_library(COREFOUNDATION CoreFoundation REQUIRED)
+				find_library(CORETEXT CoreText REQUIRED)
+				find_library(COREGRAPHICS CoreGraphics REQUIRED)
+
+				set_property(TARGET Harfbuzz::Harfbuzz APPEND PROPERTY
+					INTERFACE_LINK_LIBRARIES ${COREFOUNDATION} ${CORETEXT}
+						${COREGRAPHICS})
+
+				mark_as_advanced(COREFOUNDATION)
+				mark_as_advanced(CORETEXT)
+				mark_as_advanced(COREGRAPHICS)
+			else()
+				find_library(APPLICATION_SERVICES ApplicationServices REQUIRED)
+
+				set_property(TARGET Harfbuzz::Harfbuzz APPEND PROPERTY
+					INTERFACE_LINK_LIBRARIES ${APPLICATION_SERVICES})
+
+				mark_as_advanced(APPLICATION_SERVICES)
+			endif()
+		endif()
	endif()
 endif()
 
```

Ghabry:
Reran all the message tests of our TestGame for both with and without Freetype/Harfbuzz. Found one bug concerning the speed which is resolved now.

Ready to review. I consider this "kinda low risk" because except for "GetSize" the new code is only executed when a font is specified. The built-in font will not hit the new code.

src/text.h
```diff
@@ -100,6 +100,31 @@ namespace Text {
...
+	 */
+	Rect GetSize(const Font& font, StringView text);
+
+	/**
+	 * Determines the size of a bitmap required to render a single character.
+	 * The dimensions of the Rect describe a bounding box to fit the character.
+	 * For continuous rendering use the "width" property.
+	 *
+	 * @param font the font used to render.
+	 * @param ch the character to mesaure.
```
fdelapena, line 123: s/ch/glyph/

fdelapena (approved):

fdelapena:
> The Yume Nikki Online project is asking about Arabic text.
> 
> What impresses me here is that Harfbuzz can render right-to-left scripts correctly without changing a single line of code thinking
> 
> a

Works great, adding a note regarding a comment in chat regarding limited height for drawing glyphs: the shadow in the Arabic script above shows it's cropped (16 px height limit), and this could limit Thai scripting and "Zalgo" text, but I think it's good enough as is until some user complains ^^.

PR #2863, opened 2022-11-22
Android: Support SDL 2.26 & more lib fixes

Ghabry:
(for later)


Ghabry:
Android app is currently crashing. Will check with a debugger

Ghabry:
Jenkins: Test this please

Ghabry:
correction: Forgot to recompile. Android works.

Switch and 3DS also compiling now.

Can't reproduce this gamepad is undefined bug on Firefox anymore. Maybe some Firefox update already fixed this again. What they didn't fix is that the "gamepad added" event fires multiple times per second m( and the button mapping is still wrong.

So still Chrome only for proper gamepad: Only fires it once and it works correctly.

fdelapena (approved):

Issue #1640, opened 2019-02-17
Player crashes during transitions when color depth is forced to 16bpp


gameblabla:
#### Name of the game: 
Wadanohara...Also got the crash with Pom gets Wifi and i believe any game using screen transitions is affected.

#### Player platform:
Linux : MIPS32 mipsel (little endian) and ARMel (Armv5). SDL 1.2, sound support was disabled. 
(Doesn't matter as i can confirm sound support does not affect the crash in any way and works fine)

#### Attached files
 - The `easyrpg_log.txt` log file
[easy_log.txt](https://github.com/EasyRPG/Player/files/2872768/easy_log.txt)
(Note that the log got corrupted. I had to extract its content with a hex editor)
- GDB output for crash
[GDB_log.txt](https://github.com/EasyRPG/Player/files/2872766/GDB_log.txt)
- Makefile i use for MIPS32
[Makefile.txt](https://github.com/EasyRPG/Player/files/2872767/Makefile.txt)
#### Describe the issue in detail and how to reproduce it:

Every time a game fades in (i've tested it with Wadanohara, Pom Gets Wifi), the EasyRPG Player crashes horribly. It is reproduceable and it always crashes at the exact same place.
(In the case of Wadanohara, game fades to black in the opening. It's supposed to show the picture for the 1st chapter, the very beginning but it crashes before it even goes there)

Note that i could only reproduce this issue on MIPS32r1 (little endian) and ARMv5 (little endian) SoCs, operating system is Linux (2.6.31, 3.14 for MIPS32, 4.14 for ARMv5) and i'm using SDL 1.2.
The same configuration, Makefile on a PC (x86_64) works just fine, including with SDL 1.2.
On those platforms though, it crashes.
It also crashes on QEMU-MIPS the same way, which allowed me to get a gdb output, i have attached the log as GDB_output.txt and the Makefile as Makefile.txt. (see above in Attached files)

If you need my patched buildroot toolchain for compiling it yourselves on MIPS, you can get the source here
[https://github.com/rs-97-cfw/buildroot](https://github.com/rs-97-cfw/buildroot)

Let me know if you are interested in running the binary yourself (or need extra info, just provide me the instructions).
Note that i am unable to give the SDL 2 backend a try on those platforms as it is unavailable there. (Only framebuffer is available, no hardware acceleration and SDL2 does not support fbcon)

[...]

PR #2864, opened 2022-11-22
Improve SDL1.2 (a bit)

Ghabry:
This PR gives a final breath of life to the deprecated SDL1.2 library because crappy hardware still uses this :P.

- Always use a 32 bit drawing surface which is ideal for Pixman and then blit onto the screen surface
    - This is slightly slower because of an additional blit but solves issues such as #1640 @gameblabla 
    - Tested Wadanohara intro 8 Bit, 16 Bit and 32 Bit without crashing
- Limited CMake support (works on Linux only)

Tests pending:

- [ ] Wii


Ghabry:
The Wiibrew documentation about SDL is totally wrong. Likely devkitPro breaking backwards compatibility as usual :P

Input works now, including D-Pad and Classic controller :)

fdelapena (approved):

PR #2870, opened 2022-12-07
WIP: Add "Show String Picture"

Ghabry:
Depends on #2859 

WIP PR to get some feedback from testers in :)

### Incompatible change

Maniac Patch loads from the Windows font directory. Loading fonts in a cross platform way for all operating systems is alot of work so I decided to not support this.

Instead the Font is searched for in the ``Font`` folder of the game. Supported are all fonts that FreeType supports. The name must be the same as the one specified in the Show String Picture command.

To match with typical font naming conventions the search order is:

1. (only when bold & italic) ``FONT_NAME-BoldItalic``
2. (only when bold) ``FONT_NAME-Bold``
3. (only when italic) ``FONT_NAME-Italic``
4. ``FONT_NAME-Regular``
5. ``FONT_NAME``

### Known issues, do not report them, I work on them:

- Gradient looks broken for larger font sizes
- Disable Text Gradient option unimplemented
- Emscripten Async loading missing
- When using the same font with different sizes the rendering looks wrong (the font is shared to save memory)
- The "pad to N digits" (I think this was \d[n]) is not supported yet

### New features

You can translate the text with lcftrans :)

Full Harfbuzz shaping support for rendering complex scripts.
 
### Not planned

Loading fonts installed on the current system.

The letter spacing option is not supported because this is not supported by HarfBuzz and doing it manual will fail for any non-trivial case (that needs some kind of shaping logic).

Bold/Italic: Will only work when the font provides a bold/italic typeface. No automatic transformation is happening.


Ghabry:
Jenkins: Test this please

Needs liblcf changes so our CI has some issues building it :)

Windows (x64) build for testing: [Player.zip](https://github.com/EasyRPG/Player/files/10178145/Player.zip)

Ghabry:
New exe with this patch (Old text lines were not removed,  wups :)):

[Player.zip](https://github.com/EasyRPG/Player/files/10178665/Player.zip)

```
diff --git a/src/game_windows.cpp b/src/game_windows.cpp
index cad84c1b..3cae08ad 100644
--- a/src/game_windows.cpp
+++ b/src/game_windows.cpp
@@ -87,6 +87,8 @@ Game_Windows::Window_User* Game_Windows::GetWindowPtr(int id) {
 }

 bool Game_Windows::Window_User::Create(const WindowParams& params) {
+       Erase();
+
        data.width = params.width;
        data.height = params.height;
        data.system_name = lcf::DBString(params.system_name);
```

BlisterB:
Display position Top-Left doesn't seem to be correctly positioned 
RPG_RT
EasyRPG
Code

MackValentine:
Hi !
I found 2 bug with StringPic :

the origin is not taken into consideration
To fix it I just add "ApplyOrigin(false);" inside the "AttachWindow" function

If you use a StringPic, then a normal Picture with the same ID, the Picture will keep the StringPic inside the "new" Picture
To fix it I just add "data.easyrpg_type = lcf::rpg::SavePicture::EasyRpgType_default;" at the beginning of the "Show" function

( Sorry I'm really a noob with Git, I don't know how I can explain better >< )

Ghabry:
@MackValentine  thanks I added your fixes

MackValentine:
I found some other bugs :
1/ ExFont is broken
2/ When the Picture Size is on Constant / Variable / Variable ID, if width or height is a 0, it should be calculated, not 0
3/ if you use the Auto size, and you write \c[X], the size is misscalculated
4/ Position of the text when the window back is disabled ( no margin / no draw frame ) is not correct ( It's a little too high, and if you select a "Right" origin, it's too far to the left )

https://github.com/MackValentine/easyrpg-player-maniacs/blob/master/src/game_windows.cpp
In the Game_Windows::Window_User::Refresh i made some change to try a correction for 2/3

Ghabry:
Pretty annoying to add but StringPic works now in the web player:

image

Ghabry:
@MackValentine 

/1 is now resolved

/3 is imo a bug in Maniac Patch. It adds space for a command that should be zero space. So I won't implement this.

Ghabry:
more bugs in Maniac ShowStringPic: ExFont adds a full-width space after it :shrug: 

```
$ABlabla
blabla$a
```

image

Player does not because because there is no space in the text:

image

Ghabry:
jenkins: test this please

Ghabry:
I'm still totally confused concerning the font metrics.

As already noticed earlier there are multiple confusions about pt/px in the layouting of the text.

Found another one:

The ExFont is automatically rescaled but because it is not a font (so not handled by WinApi) it uses the "pt" value verbatim as "px" without converting it.

So for larger fonts with ExFont you get nonsense like this (much larger than the rest):

grafik

fdelapena (approved):
A great foundation on free text rendering :+1:.

PR #2882, opened 2023-01-22
Battle 2k3: (2nd try) Fix hang in traditional layout because no actor became ready

Ghabry:
Battle system.. Fix one issue, add two more ;)

The previous fix was incomplete. In some cases the battle was still stuck in traditional layout because the actor stayed permanently active :/.

rueter suggest moving the fix to line 1284. This worked only partially: It was better than my original fix but still triggered the assert.

My new solution: Moved the check now before ``IsBattleActionPending()``. Everything appears to work now.

### DIFF
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -50,6 +50,8 @@
 #include <memory>
 #include "feature.h"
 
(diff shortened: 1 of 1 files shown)


fdelapena (approved):

Ghabry:
Jenkins: test this please <3

PR #2900, opened 2023-02-26
Emscripten viewport

Ghabry:
This removes the media-queries and instead uses the entire available area.

Our SDL2 code handles it internally now since the settings PR

for best comparison I recommend opening https://easyrpg.org/play/master and https://easyrpg.org/play/pr2900 on a phone or any other very high dpi display and toggle fullscreen. The difference is extreme.

### DIFF
--- a/resources/emscripten/emscripten-shell.html
+++ b/resources/emscripten/emscripten-shell.html
@@ -117,7 +117,6 @@
   </style>
 </head>
 <body>
-
   <div id="controls">
     <button id="controls-fullscreen" class="unselectable">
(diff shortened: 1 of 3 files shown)


Ghabry:
two problems:

- Looks good in Chromium mobile phone simulation but terrible on my phone in Firefox :thinking: 
- "It is now safe" easteregg renders in wrong aspect ratio

Ghabry:
Rendering looks now much better because the size of the canvas matches the sdl2 window:

Bilinear: left old, right new

grafik

Integer: left old, right new

grafik

fdelapena (approved):

PR #2917, opened 2023-03-10
Flatpak: Make EasyRPG logo square and minified

fdelapena:
This should fix `/export/share/icons/hicolor/scalable/apps/org.easyrpg.player.svg is not a valid icon: Expected a square icon but got: 626x608`


Ghabry (approved):
works now. thanks

PR #2921, opened 2023-03-15
Use Unicode saves when translations are loaded

### LINKED ISSUE #2887: Data loss when saving in translated games
Even for games that use the translation feature savegames still use the local encoding of the game.

Assuming the original game is "Western European" and the translation is "Japanese"

When saving the savegame can contain Japanese game from active events or e.g. a translated hero name.

All the Japanese data in that savegame is lost when converted to "Western European".

---------

Maybe add a new chunk ``unicode`` to the ``easyrpg`` section of the savegame. When set the savegame is saved as UTF-8 and loaded as UTF-8 encoding preventing data loss :thinking:

Ghabry:
When a translation is loaded always store in Unicode to prevent data loss

Something for @Mimigris

Fix #2887

----------

It is still possible to lose data but this cannot be really avoided without becoming incompatible to RPG_RT:

1. Activate a translation
2. Save the game while some event with messages etc. is loaded (or custom hero name etc.)
3. Change back to native language setting
4. Load the previous savegame and save it again
5. The custom hero name will be lost because "native language" always writes in the game encoding, not in utf-8

The typical use case will be that they stay on the same translation so this will be usually safe

### DIFF
--- a/src/scene_save.cpp
+++ b/src/scene_save.cpp
@@ -41,6 +41,7 @@
 #include "output.h"
 #include "player.h"
 #include "scene_save.h"
+#include "translation.h"
 #include "version.h"
 
 Scene_Save::Scene_Save() :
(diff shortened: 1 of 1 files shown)


Ghabry:
jenkins: test this please

fdelapena (approved):

PR #2937, opened 2023-03-26
Android fixes

Ghabry:
-


Ghabry:
the PR build is a debug build so the signature is wrong. Used root permissions to put an old XML settings file into the data folder. Crash is fixed. It uses now a new key for the favourites.

fdelapena (approved):

PR #2954, opened 2023-04-12
Do not support 2k3 features in 2k

Ghabry:
First commit is unrelated to 2k/2k3 but it stops reporting the loop state for WAV. You know for what ;)

Ineluki Key Patch is now locked behind a "harmony.dll" file or ``KeyPatch=1`` in EasyRPG.ini.

When emulating RPG Maker 2000 then RPG Maker 2003 commands are not supported anymore. This makes sense in general. Could solve some bugs in games that were initially developed in 2k3 and moved to 2k. Opt-in is available: ``RPG2k3Commands=1`` in EasyRPG.ini.


Ghabry:
😏

grafik

fdelapena (approved):

PR #2955, opened 2023-04-12
Final fixes

Ghabry:
just some random stuff that doesn't fit anywhere else


Ghabry:
@Mimigris please grab the built libretro core here and test if it works.

I dispise the RetroArch interface so I won't set this up here.

Mimigris:
> @Mimigris please grab the built libretro core here and test if it works.
> 
> I dispise the RetroArch interface so I won't set this up here.

I've already started to test it one hour ago and it seems to work great, thank you for the fix!

fdelapena (approved):

PR #2958, opened 2023-04-15
Windows fixes

Ghabry:
WinApi for directory reading appears to work but would be good to have some others launch some games for testing.


Ghabry:
mimigris reported that no dump file is created for the OFF battle crash (and I can confirm this) but in other cases the dmp is created.

Likely the heap corruption preventing the execution of our crash handler.

fdelapena (approved):

PR #2962, opened 2023-04-18
Windows: Fix Manifest for x64

Ghabry:
Not on Windows right now.

Can someone quickly test if 32bit and 64bit Player from this PR start?


jetrotal:
both x86 and x64 opened fine here:
image

Ghabry:
also confirmed working on my system now

PR #2985, opened 2023-05-04
Android: Make SAF slightly faster by getting rid of DocumentFile

Ghabry:
…in the native code wrapper

Also do not query anything when the entire goal is to open a file descriptor.

This should make opening of files for reading slightly faster as the DocumentFile roundtrip is gone.

However in my tests it doesn't feel faster. There are still awful lagspikes when a new file is being opened m(

------------

This also could break stuff so have fun sharing and testing ;).

### DIFF
--- a/builds/android/app/src/main/java/org/easyrpg/player/player/SafFile.java
+++ b/builds/android/app/src/main/java/org/easyrpg/player/player/SafFile.java
@@ -10,6 +10,8 @@ import android.util.Log;
 
 import androidx.documentfile.provider.DocumentFile;
 
+import org.easyrpg.player.Helper;
+
 import java.io.IOException;
 import java.util.ArrayList;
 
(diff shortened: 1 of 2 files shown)


builds/android/app/src/main/java/org/easyrpg/player/player/SafFile.java
```diff
@@ -150,31 +141,25 @@ public int createOutputFileDescriptor(boolean append) {
...
-
         ArrayList<String> files = new ArrayList<>();
         ArrayList<Boolean> is_dir = new ArrayList<>();
 
         final ContentResolver resolver = context.getContentResolver();
-        final Uri childrenUri = DocumentsContract.buildChildDocumentsUriUsingTree(root_uri, DocumentsContract.getDocumentId(root_uri));
+        String documentId = DocumentsContract.getTreeDocumentId(rootUri);
+        if (DocumentsContract.isDocumentUri(context, rootUri)) {
+            documentId = DocumentsContract.getDocumentId(rootUri);
+        }
```
Ghabry, line 151: no idea why this is needed. I copied this from the ``DocumentFile`` constructor. When its missing it doesn't work :shrug:

Ghabry:
Looking for some testers here: Please install the APK and test if games still work (in ZIP and extracted).

Also curious about the "feeling" if it is for you the same performance or slower/faster (also try measuring the game load time).

Ghabry:
No idea why but this breaks custom SF when the game is not inside a zip

Ghabry:
I think this works now. The problem here was that SAF transforms invalid URIs to a root URI (extremely stupid). Then opening this root URI gives a valid file descriptor to a directory.

I detect this case now...

fdelapena (approved):

PR #2986, opened 2023-05-04
Require C++17

Ghabry:
__has_include and [[maybe_unused]] are c++17 features so this will build-fail already when the compiler does not know it ;)


fdelapena (approved):

Ghabry:
@fdelapena the mix of C++14 and C++17 (lcf / Player) makes the build fail for old Debian/Ubuntu versions. I think its better to just merge this and to cherry-pick the other commits in a ``0-8-0-stable`` branch as we did before.

PR #3004, opened 2023-05-14
Emscripten: retry download on uncommon failure

aleck099:
From ynoclient: https://github.com/ynoproject/ynoclient/commit/095cf5bddb902478573b3b57a0fc4d4284a00587
TCP connections aren't reliable!

### DIFF
--- a/src/async_handler.cpp
+++ b/src/async_handler.cpp
@@ -71,17 +71,72 @@ namespace {
 	}
 
 #ifdef EMSCRIPTEN
-	void download_success(unsigned, void* userData, const char*) {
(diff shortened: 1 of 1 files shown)


Ghabry:
Thanks for the patch.

A good test for this is likely simulating a bad wi-fi connection / mobile connection.

E.g. disconnect the WiFi and then reconnect it after a few seconds.

Will test this later :)

fdelapena:
jenkins: ok to test

Ghabry (approved):
When the connection drops all 10 retries fire pretty fast and I can't find a simple solution for waiting. So lgtm for now. Thanks.

fdelapena (approved):

Issue #2915, opened 2023-03-09
Android: Multi-Touch broken on Android 13 and certain Samsung Devices


Ghabry:
This is reported alot and I found issues documenting it in the Lemuroid repository (an app supporting libretro cores that is not RetroArch :heart: )

https://github.com/Swordfish90/Lemuroid

https://github.com/Swordfish90/Lemuroid/blob/96da5eddb8d2602adf8c82f57b34518e64a309c6/lemuroid-touchinput/src/main/java/com/swordfish/touchinput/radial/LemuroidTouchConfigs.kt#L131-L133

```java
        // Using standard MotionEvents.getX/Y() is broken on Samsung devices running Android 11,
        // but Android 13 broke MotionEvents.getRawX/Y() events in landscape mode for every device.
        val useRawTouchCoordinates = Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU
```

Relevant issues:

- https://github.com/Swordfish90/Lemuroid/issues/582
- https://github.com/Swordfish90/Lemuroid/issues/558
- https://github.com/Swordfish90/Lemuroid/issues/620

Ghabry:
The real issue here is btw dynamic resolution scaling by Samsung. They run games at lower resolutions by default to give higher performance and to not break the games they fake the X/y coordinates of touch events.
But ui components are not repositioned so all the calculations are off 🤦

hsoftxl:
what is "btw dynamic resolution"?

fdelapena:
> what is "btw dynamic resolution"?

"btw" in that context means "By the way, "

The automatic resolution change is likely some sort of a battery optimization feature that could be disabled somewhere in settings. This is vendor specific and version specific, so can't tell where is this placed in devices.

Ghabry:
This is not reproducible on a Samsung Galaxy A52 with Android 13.

Everything works fine.

Additionally got a report that Gamepads have a huge latency: Imo there is some noticable input lag that is not suitable for fast action games but RPGs are perfectly playable with it.

Ghabry:
Recording of the issue:

Single touch works, multi touch behaves... interesting.

https://github.com/EasyRPG/Player/assets/1331889/0a1dbd5d-5b68-4b0a-a1a3-d219466cd63a

Ghabry:
No idea how to fix the button mapping activity but this is not urgent.

However found a workaround for the touch-stuff:

[...]

PR #3016, opened 2023-05-22
Fix Samsung Multitouch bug

### LINKED ISSUE #2830: Andoird: FF strange behaviour
Name of the game: indipendent from the game
Player platform: Android

Using Fast Forward (FF) on Android Player have strange behaviour, i use FF in mode "keep pressed" and:
1) if i push FF and keep it pressed and then i push UP, UP not work and RIGHT become UP (DOWN works well)
2)  if i push FF and keep it pressed if i push A,B or other button then FF stop work

Ghabry:
I'm pretty annoyed right now but I think this is working >.>

Please also test on other devices to ensure my heuristic does not cause fall-positives.

-------

This happens when a game is running (GlSurface exists) and a touch happens across different views.
Our touch ui uses one view per button and is affected.

The culprit is a game optimisation that reduces the size of the surface.

Fix is based on https://github.com/F0RIS/SamsungMultitouchBugSample/tree/Fix1

See also: https://github.com/Swordfish90/Lemuroid/issues/178

Fix #2830
Fix https://github.com/EasyRPG/Player/issues/2915

### DIFF
--- a/builds/android/app/src/main/java/org/easyrpg/player/Helper.java
+++ b/builds/android/app/src/main/java/org/easyrpg/player/Helper.java
@@ -8,6 +8,7 @@ import android.database.Cursor;
 import android.graphics.Color;
 import android.graphics.Paint;
 import android.graphics.Paint.Style;
+import android.graphics.Point;
 import android.net.Uri;
 import android.os.ParcelFileDescriptor;
 import android.provider.DocumentsContract;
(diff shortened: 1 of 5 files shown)


Ghabry:
Received positive feedback from another Samsung user (mokodeoz). Waiting for non-Samsung feedback, too :)

fdelapena (approved):

PR #3023, opened 2023-05-25
Various minor fixes

### LINKED ISSUE #2998: Minor stuff to fix
Stuff so small it is not worth a seperate issue but before I forget it:

- [x] Font API has inconsistent return value for GetSize (TTF)
- [x] Remove everything except ``code`` from shell file keyboard event handler
- [x] Viewport is not correctly centered when having an odd tilesize (?) (seems that using ``std::ceil`` instead of ``RoundTo`` works
- [x] opusfile not found anymore on windows
- [x] UBSan issues

Ghabry:
Fixes various minor stuff.

- [x] Opus works again on Windows
- [x] Emscripten touch still works
- [x] TTF Fonts still render correctly

Fix #2998

### DIFF
--- a/src/game_player.cpp
+++ b/src/game_player.cpp
@@ -755,11 +755,11 @@ void Game_Player::SetTotalEncounterRate(int rate) {
 }
 
 int Game_Player::GetDefaultPanX() {
-	return (std::ceil(static_cast<float>(Player::screen_width) / TILE_SIZE / 2) - 1) * SCREEN_TILE_SIZE;
+	return static_cast<int>(std::ceil(static_cast<float>(Player::screen_width) / TILE_SIZE / 2) - 1) * SCREEN_TILE_SIZE;
 }
 
 int Game_Player::GetDefaultPanY() {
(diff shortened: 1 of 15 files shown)


src/game_player.cpp
```diff
@@ -755,11 +755,11 @@ void Game_Player::SetTotalEncounterRate(int rate) {
 }
 
 int Game_Player::GetDefaultPanX() {
-	return (Utils::RoundTo<int>(static_cast<float>(Player::screen_width) / TILE_SIZE / 2) - 1) * SCREEN_TILE_SIZE;
+	return static_cast<int>(std::ceil(static_cast<float>(Player::screen_width) / TILE_SIZE / 2) - 1) * SCREEN_TILE_SIZE;
```
Ghabry, line 758: oh this is already in. forgot to rebase :thinking:

Ghabry:
Battle damage popup should now work again for all three cases: Built-in font, Bitmap font, True Type font. Please test.

fdelapena (approved):

PR #3030, opened 2023-06-05
Widescreen improvements

Ghabry:
Like everything else the map is now centered when fake resolution is active. Should improve the UX because it gets rid of some glitches :).

First two commits are not changing any behaviour.

3rd commit allows to change the resolution while in the title scene. Until now this didn't work without: Start game, press F12.

4th and 5th commit: 4th results in the screenshot in the middle and 5th results in the screenshot at the bottom. (top is before) Cropping the panorama makes sense here because technically it is drawn outside of the size of the map, nothing should be drawn their.

out2

Another example using Fake Res for a custom resolution. That map is again 20x15. Cropping works.

out

### DIFF
--- a/src/battle_animation.cpp
+++ b/src/battle_animation.cpp
@@ -31,6 +31,8 @@
 #include "player.h"
 #include "options.h"
 #include "drawable_mgr.h"
+#include "scene_map.h"
+#include "spriteset_map.h"
 
 BattleAnimation::BattleAnimation(const lcf::rpg::Animation& anim, bool only_sound, int cutoff) :
 	animation(anim), only_sound(only_sound)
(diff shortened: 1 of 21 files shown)


Ghabry:
There is one annoyance left: On looping maps that are smaller than the screen the camera is positioned incorrectly. Unfortunately cannot figure out how to solve this. The scrolling code is complicated ^^'

fdelapena:
Looking good besides the mentioned by Mimigris, however there are some cases for new games where some picture behaviour is desired with the legacy way, e.g. for custom HUD positioned on screen corners. Maybe getting this under some configurable feature later is worth.

fdelapena (approved):

Ghabry:
Fixed centering for battle animations and zoom on actor.

2k battle Ui is not scrolling anymore in widescreen mode. For 2k3 I will postbone this for later as this requires implementing it 3 times for all these modes...

PR #3033, opened 2023-06-12
Support games in LZH archives

Ghabry:
That is the archive format used by the "Create Game Disk" feature of the editor

This is totally not production ready. Needs code cleanup, (android support?), encoding detection etc.

Also new library needed (lhasa), so infrastructure work etcpp.

https://github.com/EasyRPG/Player/assets/1331889/b8d3dff1-a036-4bef-b4fc-953e6b54e0b3

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,6 +122,8 @@ add_library(${PROJECT_NAME} OBJECT
 	src/fileext_guesser.h
 	src/filesystem.cpp
 	src/filesystem.h
+	src/filesystem_lzh.cpp
+	src/filesystem_lzh.h
 	src/filesystem_native.cpp
 	src/filesystem_native.h
 	src/filesystem_root.cpp
@@ -827,7 +829,7 @@ endif()
 player_find_package(NAME PNG TARGET PNG::PNG REQUIRED)
 player_find_package(NAME fmt TARGET fmt::fmt REQUIRED)
 
-# Do not use player_find_package. enable_language used by pixman on Android does work properly inside function calls
+# Do not use player_find_package. enable_language used by pixman on Android does not work properly inside function calls
 find_package(Pixman REQUIRED)
 target_link_libraries(${PROJECT_NAME} PIXMAN::PIXMAN)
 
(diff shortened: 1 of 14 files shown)


Ghabry:
it was mentioned that LZH files created through the game disk feature contain duplicates when the RTP is embedded.

Open source tools do not show the first entry in the archive (but Ark becomes confused with the stats and shows compressed size of first entry and decompressed size of second entry).

The rule appears to be "Latest File Wins".

The RTP assets are also the first things added to the archive so this could be even a _feature_ of the format? Similiar to tar which was originally designed for archives on tape where seeking is hard to do. So instead of overwriting the file you just append it again...

jetrotal:
Cool! Will test it right now!

Random question, Is possible to tell player.exe to autoload a lzh file, instead of displaying a list of games?

Someone asked me recently if it was possible to hide better the contents of their game, and I thought about something similar to the `default game loading` when using the webplayer.

That + allowing people to customize the extension of the file to be something other than .lzh could be cool for that.

Ghabry:
There is currently no way to test it because our buildscripts lack the library (and vcpkg does not have the package merged yet but I got it approved this week).

Better readable with a non-whitespace diff: https://github.com/EasyRPG/Player/pull/3033/files?w=1

---------

> Random question, Is possible to tell player.exe to autoload a lzh file, instead of displaying a list of games?

Some silly apprach (That will need a launch script) is: Append the ZIP to the EXE file, then use ``zip -A``.

Linux code for this:

```
cat Player.exe game.zip > Player_with_game.exe
zip -A Player_with_game.exe
```

Then launch this through a start script:

```
./Player_with_game.exe --project-path Player_with_game.exe
```

fdelapena (approved):
Likely needs to update the README.md before merging. Thanks!

lhasa will need to be provided with enough time by downstream packagers if want to support this feature on existing distros that package easyrpg-player. Pinging earlier because this new package process usually takes some time as it is requested by a maintainer. It is required at least on Fedora. cc @suve

Ghabry:
Oh and maybe an explanation why ``libarchive``, which is packaged almost everywhere, is not used:

It's heavy usage of C locales and charset conversion makes it unusable for us: C locales are badly supported on some platforms we support and the charset conversion are a problem.

We do not know the charset beforehand and have to guess it and ``libarchive`` has no option to get a _raw_ filename back (and pass them to ICU for guessing the charset). It will always send the string through some conversion code. That API design makes it unfortunately unusable for us.

There is a longstanding issue from 2015 about this https://github.com/libarchive/libarchive/issues/587

PR #3049, opened 2023-06-29
Battle: Fix broken animation when the sprite is flipped

Ghabry:
While fixing flipped spritesheets I forgot to consider that battle sprites can be also flipped.

The battle sprites contained adjustment for the src_rect when they were flipped. As I added a fix to the sprite class this means it was double-inverted and broke.

Solved this by removing all the adjustment code from the battle sprites.

(seeing these workarounds this means that this was a long-standing bug in the flip-code)

------

**Please test various 2k3 games and check if the battles look correct!**

I tested:

- Lufia V For the Savior

### DIFF
--- a/src/battle_animation.cpp
+++ b/src/battle_animation.cpp
@@ -107,10 +107,9 @@ void BattleAnimation::DrawAt(Bitmap& dst, int x, int y) {
 			continue;
 		}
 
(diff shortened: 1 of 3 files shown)


Ghabry:
Jenkins: test this please

fdelapena (approved):

PR #3051, opened 2023-07-05
String Variables

enewey:
implements all major maniacs string variable operations and modifiers, with the exception of the `.sjis` and `.utf8` encoding modifiers for file reading/writing.

.asg & .cat
    t[..]
    v[..]
    s[..]
    .actor [a] .name 
    .skill [a] .name 
    .item [a] .name 
    .enemy [a] .name 
    .troop [a] .name 
    .terrain [a] .name 
    .element [a] .name 
    .state [a] .name 
    .anim [a] .name 
    .tileset [a] .name
    .s [a] .name 
    .v [a] .name 
    .t [a] .name
    .cev [a] .name             
    .class [a] .name           
    .anim2 [a] .name           
    .map [a] .name             
    .mev [a] .name             
    .member [a] .name 
    .actor [a] .desc 
    .skill [a] .desc
    .item [a] .desc
    .member [a] .desc 
    .cat
    .ins
    .rem
    .rep
    .exrep
    .subs
    .join
    .file

.toNum
.getLen
.inStr
.exInStr
.exMatch
.split
.toFile
.popLine

also adds support for string comparators in conditional statements
.eq
.neq
.contains
.notContains

2k3 test project used for developing these features:
https://rpgmaker.net/users/narcodis/locker/easyrpg_testing.zip
proper unit testing could be added 😓


Ghabry:
Jenkins: test this please <3

Thanks for this huge contribution!

You received an invite for the "Player Team". This whitelists you for the pull request builder (will auto-build your stuff).

Also add yourself and all other authors that contributed to this to the Authors.md if you havn't done already.

src/filefinder.cpp
```diff
@@ -68,6 +68,7 @@ namespace {
 	constexpr const auto SOUND_TYPES = Utils::MakeSvArray(
 			".opus", ".oga", ".ogg", ".wav", ".mp3", ".wma");
 	constexpr const auto FONTS_TYPES = Utils::MakeSvArray(".fon", ".fnt", ".bdf", ".ttf", ".ttc", ".otf", ".woff2", ".woff");
+	constexpr const auto TEXT_TYPES = Utils::MakeSvArray(".txt", ".csv");
```
jetrotal, line 71: Silly request: Could we cover more file formats?
.svg .xml .json .yml... Don't need to add any code to interpret or format those files.

That would help me with some other projects.

Ghabry:
Not going to review this right now as is a bit too long xD.

But one remark already: I don't like the commit text. Could you prefix all of them with something like ``StringVar: `` to make it clear what the context is?

The easiest way to do this should be:

```
git rebase -i HEAD~23
```

Then change all the ``pick`` to ``reword``.

Now git will open a text editor for every single commit (so 23 times). Here prefix all the commits with something like ``StringVar: ``.

Then push: ``git push --force-with-lease origin``.

Is a bit tedious but shouldn't take more than a few minutes.

src/filefinder.cpp
enewey: Absolutely possible. I think .svg, .xml, .json, .yaml are all a good start, I can add those.

Ghabry:
Jenkins: test this please

Ghabry:
Uploaded the string var test project to the web player: https://easyrpg.org/play/pr3051/?game=string-var  (note that txt file loading won't work as the code for "async asset fetching" is not implemented currently)

enewey:
Looks like String vars in a string pic isn't fully supported -- while you can display a stringvar through message codes (e.g. `@pic[1].strpic { "\t[1]", ... }` ), you cannot just plug in the string variable directly (e.g. `@pic[1].strpic { t[1], ... }`). Need to update string pics, and any command that takes a string parameter (like a filename).

Ghabry:
except for StringPics (where this appears to be used alot(?)) I would prefer to have support for string-vars in the other commands as another PR.

Adding support to all other commands is alot of additional complexity and lots of more testing that nothing breaks. So not required for the MVP. ;)

enewey:
I agree -- don't want to explode the scope of this. Isolating to supporting string pics sounds reasonable.

src/game_strings.cpp
```diff
@@ -0,0 +1,223 @@
...
+		case 5:  params.string_id += Split(params, static_cast<std::string>(string), args[1], args[2]); break;
+		case 8:  break; // range case not applicable for popLine; see case in game_interpreter.cpp
+		case 9:  ExMatch(params, static_cast<std::string>(string), args[1] + (params.string_id - start), args[2]); break;
+		case 10: ExMatch(params, static_cast<std::string>(string), args[1] + (params.string_id - start), args[2], args[3]); break;
+		}
+	}
+	return GetData();
+}
+
+Game_Strings::Str_t Game_Strings::PrependMin(Str_t string, int min_size, char c) {
```
Ghabry, line 215: why is the ``PrependMin`` required?

src/game_strings.cpp
```diff
@@ -0,0 +1,223 @@
...
+	std::stringstream ss(str);
+
+	while (offset >= 0 && std::getline(ss, result)) { offset--; }
+
+	offset = ss.rdbuf()->in_avail();
+
+	Set(params, static_cast<Str_t>(ss.str().substr(str.length() - offset)));
+	params.string_id = string_out_id;
+	return Set(params, static_cast<Str_t>(result));
+}
```
Ghabry, line 148: What kind of linebreak is Maniacs emitting when writing to files? Is it ``\r\n`` or ``\n``?

Can Maniacs process strings correctly that have both types of line breaks?  (also relevant for PopLine as there can be different line ending).

EasyRPG Player usually handles both types of line breaks independent of the platform. There is ``Utils::ReadLine`` that has the same interface as ``std::getline``.

--------

Also needs checking if Maniacs reads in binary mode or in text mode by just reading a binary file.

A good test here is a PNG image read into a string and then written to the drive again. If its still an image it must be binary mode.

PNG also has ``0x1A`` very early in the header which is "End Of File" in text mode so it will read 0-bytes right after it encountered ``PNG``.

src/game_interpreter.cpp
```diff
@@ -4626,6 +4663,340 @@ bool Game_Interpreter::CommandManiacSetGameOption(lcf::rpg::EventCommand const&)
...
+			args[2] = Main_Data::game_variables->GetWithMode(args[2], modes[2]);
+			result = static_cast<Game_Strings::Str_t>(static_cast<std::string>(Main_Data::game_strings->GetWithMode(com.string, args[0], modes[0])).substr(args[1], args[2]));
+			break;
+		case 10: //Join Variables (join) <fn(string delimiter, int var_id, int size)>
+		{
+			std::string op_string = "";
+			std::string delimiter = (std::string)Main_Data::game_strings->GetWithMode(com.string, args[0], modes[0]);
+			// mode of arg[1] is weird, seems offset by +2, not sure what the intent is
+			args[1] = Main_Data::game_variables->GetWithMode(args[1], modes[1] - 2);
+			args[2] = Main_Data::game_variables->GetWithMode(args[2], modes[2]);
```
Ghabry, line 4875: There is already ``ValueOrVariable(mode, val)`` available here in ``Game_Interpreter``. No need to create a new method in ``game_variables``.

For string its imo fine as this is new and there are far less usages of this across the interpreter... (yet)

src/game_interpreter.cpp
```diff
@@ -4626,6 +4663,340 @@ bool Game_Interpreter::CommandManiacSetGameOption(lcf::rpg::EventCommand const&)
...
+		string_id_1 = Main_Data::game_variables->Get(string_id_1);
+	}
+
+	int op    = (com.parameters[3] >>  0) & 255;
+	int fn    = (com.parameters[3] >>  8) & 255;
+	int flags = (com.parameters[3] >> 16) & 255;
+
+	int hex_flag     = (flags >> 1) & 1;
+	int extract_flag = (flags >> 2) & 1;
+	int first_flag = (flags >> 3) & 1;
```
Ghabry, line 4714: So stupid that this is a new command so there was full freedom of implementation and it still contains all these stupid bit-masking shit :facepalm:

src/game_interpreter.cpp
```diff
@@ -4264,16 +4285,32 @@ bool Game_Interpreter::CommandManiacShowStringPicture(lcf::rpg::EventCommand con
...
-	auto components = Utils::Tokenize(com.string, [](char32_t ch) {
-		return ch == '\x01';
+	// maniacs stores string parsing modes in the delimiters
+	// x01 -> literal string
+	// x02 -> direct reference
+	// x03 -> indirect reference
+	// for the displayed string, the id argument is in com.parameters[22]
+	// here we are capturing all the delimiters, but currently only need to support reading the first one
+	int i = 0;
+	int delims[3] = {};
```
Ghabry, line 4295: This will crash when there are for whatever reason more than 3 delimiters. Make this a vector or stop adding the modes when you are at 3.

Ghabry (changes requested):
Only quickly looked through the code and these are the largest issues I'm currently seeing.

Sorry but text processing and encodings is tricky so we must think here of all the edge-cases as we are multi-platform, multi-language and Unicode aware, so a "lol, works in Japan for me" is not good enough.

All in all for your first PR (which solves a complex problem) this is already pretty good work :).

src/game_interpreter.cpp
```diff
@@ -4626,6 +4663,340 @@ bool Game_Interpreter::CommandManiacSetGameOption(lcf::rpg::EventCommand const&)
...
+			result = Main_Data::game_strings->PrependMin(result, args[1], '0');
+			break;
+		case 3: //Database Names <fn(int database_id, int entity_id, bool dynamic)>
+			args[1] = Main_Data::game_variables->GetWithMode(args[1], modes[1]);
+
+			switch (args[0])
+			{
+			case 0:  //.actor[a].name
+				if (args[2]) result = static_cast<Game_Strings::Str_t>(Main_Data::game_actors->GetActor(args[1])->GetName());
+				else         result = static_cast<Game_Strings::Str_t>(lcf::ReaderUtil::GetElement(lcf::Data::actors, args[1])->name);
```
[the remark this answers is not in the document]
enewey: Ah, I will add safeguards around these to check for `nullptr`. If the element doesn't exist, Maniacs will crash with an error message (such as `Actor [99] does not exist`), similar to if trying to open a file that does not exist.

src/game_strings.h
```diff
@@ -0,0 +1,162 @@
...
+#include "player.h"
+#include "string_view.h"
+
+/**
+ * Game_Strings class.
+ */
+class Game_Strings {
+public:
+	using Str_t = lcf::DBString;
+	using Strings_t = std::vector<Str_t>;
```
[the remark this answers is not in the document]
enewey: Nice, yeah I wasn't sure what `DBString` was exactly, but it makes a lot more sense to use just standard strings I think. 

As far as the data structure, a map makes a lot of sense to me. Shouldn't require much in terms of changes as access has already been pretty abstracted behind the `Get` and `Set` methods.

src/game_strings.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+int Game_Strings::ToNum(Str_Params params, int var_id) {
+	if (!ResizeWithId(params.string_id)) return -1;
+	std::string str = static_cast<std::string>(Get(params.string_id));
+
+	int num;
+	if (params.hex)
+		num = std::stoi(str, 0, 16);
+	else
+		num = std::stoi(str);
```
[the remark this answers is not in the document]
enewey: Did not know this about `std::stoi`, good note! I will look into the alternatives suggested.

src/game_strings.cpp
enewey: Some string assignment operations have an option for creating a "minimum length" for the saved string. For example, when assigning a string as a variable, such as `t[1] .asg v[1]`, you can also optionally specify a `.min`, like `t[1] .asg v[1] .min 5`, to give the string a minimum length. In the case of a number, it will prepend `0`s, and for a normal string, it will prepend whitespace.

e.g.
```
v[1] = 3
t[1] .asg v[1] .min 5
@msg.show "\t[1]"
```
will result in displaying `00003`.

src/game_strings.cpp
enewey: Ah, I did not know that about the `Utils` package -- I will check that out! As for line endings I will admit I did not test too extensively with return carriages -- will report back as I experiment with stuff.

Maniacs will only read files with a `txt` or `csv` extension when assigning to a string, so I renamed a `image.png` to `image.txt`. When reading the file as `sjis`/`ansi`, it read only the `PNG`, but reading it as `utf8` appeared to read nothing at all (perhaps it failed? not sure). 

I'm pretty dumb when it comes to all this encoding stuff so your input is extremely valuable here.

src/game_interpreter.cpp
enewey: Yep, I didn't realize this existed already until I went and sniffed around the StringPics code... I can refactor this away no problem.

src/game_interpreter.cpp
enewey: Oh good point -- consistency is not maniac's strong suit, haha

jetrotal:
Ghabry asked to list the issues pointed at rm2k3 channel:

image

----------


image

`t[1] .asg s[1]`

-----------


image



------------

image image


```js
@loop 20 {
    v[1] += 1
    t[2] .asg v[1]
    t[1] .asg  .cat t[2], "
", t[1]
    
}

@pic[1].strpic {
    "\t[1]"
    .pos 160, 120 .center
    .size 0, 0    .chromakey 1
    .scale 100
    .trans 0
    .rgbs 100, 100, 100, 100
    .font "", 12
    .spacing 0, 4
    .skin "" .stretch
    .mapLayer 7
    .eraseWhenTransfer
    .affectedByFlash
    .affectedByShake
}
```

this last one may have been fixed already.

src/game_interpreter.cpp
enewey: ✅

src/game_interpreter.cpp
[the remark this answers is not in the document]
enewey: Made changes to address this

src/game_strings.h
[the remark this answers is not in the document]
enewey: Updated to use `std::string`

src/game_variables.h
```diff
@@ -205,6 +206,17 @@ inline Game_Variables::Var_t Game_Variables::GetIndirect(int variable_id) const
...
+// mode 0: pass through id value
+// mode 1: get by id
+// mode 2: get indirect by id
+inline Game_Variables::Var_t Game_Variables::GetWithMode(int id, int mode) const {
+	switch (mode) {
+	case 1: return Get(id); break;
+	case 2: return GetIndirect(id); break;
+	}
+	return id;
+}
```
Ghabry, line 218: (just a reminder to use ``ValueOrVariable`` available in the interpreter)

src/maniac_patch.cpp
```diff
@@ -600,4 +606,162 @@ bool ManiacPatch::GetKeyState(uint32_t key_id) {
...
+
+	switch (op) {
+	case 0: // eq
+	case 2: // contains (l contains r)
+		return std::regex_match(str_l, search);
+	case 1: // neq
+	case 3: // notContains (l does not contain r)
+		return !std::regex_match(str_l, search);
+	default:
+		return false;
```
Ghabry, line 631: My review comment was lost here: Pls do not use ``std::regex`` for such comparisons as the implementing is slow and/or broken on some platforms.

use Utils::LowerCase or Utils::StrIcmp instead.

Ghabry (changes requested):
Looks pretty good, I think we are 98% there :)

Ghabry:
I rebased the PR but I cannot push to the branch (because it is master I guess?).

Will likely close this PR later and create my own to resolve the last pending issues.

src/maniac_patch.cpp
enewey: done

jetrotal:
Thanks for the updates!

The webplayer not loading text from files is a glitch?
image


Just listing it here to let you guys know about it...

enewey:
Have resolved most issues, still a couple hanging threads to resolve. 

An issue has cropped up with displaying string pictures, though. Seems to be related to some code changes I had to accommodate on rebase in the `PendingMessage` class. It appears when a string is assigned a string containing a string param (i.e. `t[1] .asg "\t[2]"`), it will strip out all newline characters from the original string. Not sure what else it might be stripping out. 

Maniacs rpg_rt:
image

EasyRPG player:
image

Player also appears to incorrectly(?) color text when maniacs does not, but I don't know how concerned I am about that particular thing.

For the newline stripping piece, I will try to find time to dig into the PendingMessage class in more depth to deduce why this is happening.

src/game_strings.cpp
```diff
@@ -0,0 +1,223 @@
...
+	return num;
+}
+
+int Game_Strings::GetLen(Str_Params params, int var_id) {
+	if (!ResizeWithId(params.string_id)) return -1;
+
+	int len = static_cast<std::string>(Get(params.string_id)).length();
+	Main_Data::game_variables->Set(var_id, len);
+	return len;
+}
```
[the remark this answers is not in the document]
enewey: When testing the string `My name is Đức`:
Assigned to a string directly, length is 14.
When read from a file specified as utf8, length is 14.
When read from a file specified as ansi, length is 17.
image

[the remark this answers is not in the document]
enewey: With the PR as it is, the player does not behave the same way.
With files read as utf8, we are saying its length is 17
With files read as ansi, we are saying its length is 17
And when assigned directly to a string, we are saying its length is 15... I _believe_ this is because `tpc` doesn't allow the `ứ` character (either due to my computer's locale or tpc itself, no idea).
image
image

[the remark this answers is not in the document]
enewey: Looks like BingShan recently released an update to `tpc` on his private discord that does not encode the text-editor as shift-jis, so I may be able to test this a little more comprehensively... I'll check this out.

Ghabry:
I have a hard time going through the tests because I'm already stuck for a long time at the first one: Bottom left. Just this bottom left test took me already 90 minutes to check what is going in...

First discovery: Maniac changed how unknown command codes are parsed. E.g. a message with ``\A[0] \B[1] \e[2]`` is rendered as ``[0] [1] [2]`` by the Player. This is also what RPG_RT does, Maniac changed this and it renders ``\A[0] \B[1] \e[2]`` instead. I believed for 45 minutes that this is a player bug until I finally started an original RPG_RT to notice that it is a Maniac change -_-

-----

Top Maniac, Bottom Player:

I wouldn't say that we are wrong here. This is just how it worked in original RPG_RT. Also is an edge-case nobody will usually encounter. Notice the missing ``\`` at the last line before the ``s[1]``. The Maniac parser has some bug here and is inconsistent.

image

image

Imo test passed :heavy_check_mark: 

-----

Looks like what we need for ``\t`` in a message box is to evaluate the expression? Though enabling the evaluation will also enable it for the String Picture because they share the same code. Is this a problem? Where is the ``y`` in Maniac coming from?

image

image

Test failed :heavy_minus_sign:

Ghabry:
The 2nd test at the bottom row is also completely different.

Left is Maniac, Right is Player. Am I using the wrong version of Maniac patch? I use the RPG_RT.exe provided in the ZIP archive of the test game.

image

------

Sorry I cannot continue this review under these circumstances. My time is limited and something is completely broken.

jetrotal:
Some direct comparations EasyRPG x Maniacs_RT (bigger windows being maniacs). 
I may repeat the reports you guys gave:

Read PNG:
image

Uncommon Characters:
image
image

var, name, string, switch:
image
image
image

Database stuff (error only on easyRPG - Seems correct?):
image

"\t[1]" reading:
image

Ghabry:
About the remaining issues according to the screenshots:

- [X] Read PNG: I guess the difference is that we read in "Binary mode" and Maniac_RT in "Text mode". Text mode is not portable across operating systems so I would suggest to ignore this edge case (actually with our implementation it should be possible to read a PNG in "utf-8" mode and write the PNG in "utf-8" mode and it should survive :sweat_smile:)
- [ ] _Has Patch_ Uncommon Characters: I implemented this now locally and I get UTF-8 17 and ANSI 14, so the opposite of Maniac. The problem here is that our Player uses UTF-8 internally and Maniac_RT likely the local codepage (as the old RPG_RT). That an implementation detail and imo not worth to fix.
- [ ] var, name, string, switch: This is a difference in how we evaluate variables in Messages and String Pics. Fixing this will break it in the other case though... So I would be fine with keeping this _broken_ until a later PR.
- [ ] "\t[1]" fails to read new lines, but interprets \c sintaxes: Another implementation difference of Messages vs. String Pics: String Pics evaluate \n while in Messages they don't.

The last two cannot be fully implemented because we use shared code for both messages and string pics compared to Maniacs which seems to use 2 implementations......

Ghabry:
@enewey I figured out solutions for most of the remaining problems. Just need to clean it up a bit. Will send you a patch on Sunday or Monday that will resolve them.

jetrotal:
one thing we forgot to take a look is loading text files with webplayer

Ghabry:
@enewey Creating commits was easier than explaining but I cannot push to your master branch (permission denied).

Please grab my commits manually:

```
git remote add ghabry https://github.com/ghabry/easyrpg-player
git fetch ghabry
git rebase ghabry/strvar --autostash --committer-date-is-author-date
```

Now you have my commits and you can push.

Ghabry:
@jetrotal  In case you already want to retest here is an emscripten instance with my changes:

http://server.mastergk.de:8000/easyrpg-player.html

src/game_interpreter.cpp
```diff
@@ -4634,6 +4671,307 @@ bool Game_Interpreter::CommandManiacSetGameOption(lcf::rpg::EventCommand const&)
...
+			args[1] = ValueOrVariable(modes[1], args[1]);
+			args[2] = ValueOrVariable(modes[2], args[2]);
+			result =  Main_Data::game_strings->GetWithMode(str_param, args[0], modes[0]).substr(args[1], args[2]);
+			break;
+		case 10: //Join (join) <fn(string delimiter, int id, int size)>
+		{
+			std::string op_string = "";
+			std::string delimiter = (std::string)Main_Data::game_strings->GetWithMode(str_param, args[0], modes[0]);
+			Output::Debug("args {} {} {}", args[0], args[1], args[2]);
+			Output::Debug("modes {} {} {}", modes[0], modes[1], modes[2]);
```
Ghabry, line 4840: Please remove these two debug lines

src/game_strings.cpp
```diff
@@ -0,0 +1,225 @@
...
+int Game_Strings::InStr(Str_Params params, std::string search, int var_id, int begin) {
+	if (!ResizeWithId(params.string_id)) return -1;
+
+	if (params.extract) {
+		search = Extract(search, params.hex);
+	}
+
+	std::string str = Get(params.string_id);
+
+	Output::Debug("Searching for {} in {}", search, str);
```
Ghabry, line 77: And this one

Ghabry (approved):
This PR is already complex enough and it is a very solid implementation of String Vars. Thanks again for this huge contribution.

There is only nitpicky stuff left (e.g. lack of unit tests, lots of copying instead of string view, code formatting in some parts) but this can be all addressed later.

fdelapena (approved):

PR #3052, opened 2023-07-06
Battle 2k3: fixed battler facing direction + horizontal ally/enemy select

Primekick:
Those features were originally intended for **Lufia V: For the Savior** to make it work on EasyRPG since it relied on some small EasyRPG plugin before, but I've seen a few people wishing they could do something similar, so I thought it might be a good idea include them in the original Player distribution.

### Fixed battler facing direction 
Normally battlers turn around to face the unit they're currently attacking, which is undesirable if you're build an old school RPG like the aforementioned **Lufia V**. This feature adds an option to make sure the battlers only ever face a single direction no matter what. It only affects how sprites are displayed and is purely cosmetic, no changes to the actual battle processing.

### Horizontal ally/enemy select
This one is a logical consequence to the previous feature - if you're lining up all units horizontally next to each other it feels weird to select them with up/down buttons. It adds an option to select allies/enemies by pressing left/right instead of up/down.

### DIFF
--- a/src/feature.cpp
+++ b/src/feature.cpp
@@ -56,6 +56,3 @@ bool Feature::HasFixedEnemyFacingDirection() {
 	return HasRpg2k3BattleSystem() && lcf::Data::battlecommands.easyrpg_fixed_enemy_facing_direction > 0;
 }
 
-bool Feature::HasHorizontalAllyEnemySelecting() {
-	return HasRpg2k3BattleSystem() && lcf::Data::battlecommands.easyrpg_horizontal_ally_enemy_selection;
-}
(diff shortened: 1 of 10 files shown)


Ghabry:
This raises the question if "no flip" is really what you want for all battle types because some of them will break completely like "Pincer" or "Sourround" or "Back attack".

Obviously this "don't flip" is designed for the "Normal" and the "Initiative" battle system only because with manual placement they can be modified to be kinda like "Front facing" battle.

Ghabry:
random idea: will it break to just always allow "left/right" for enemy selection combined with "up/down"? Will be _incompatible_ with RPG_RT (but not in a breaking way) and avoids a chunk.

Ghabry:
The facing looks fine. For the left/right navigation I figured out a better solution but I couldn't get it working and I'm on vacation by tomorrow. So this has to wait until I'm back. sorry.

fdelapena:
jenkins: ok to test

fdelapena (approved):

fdelapena:
jenkins: test this please

PR #3060, opened 2023-07-25
CMake: Fix Big endian detection when crosscompiling

Ghabry:
Needs at least CMake 3.20 to work properly

Removed runtime check as everything uses a build system now.

(untested)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -767,11 +767,15 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
 endif()
 
 # Endianess check
-if(NOT DEFINED CMAKE_TOOLCHAIN_FILE)
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.20)
+	if (CMAKE_CXX_BYTE_ORDER STREQUAL "BIG_ENDIAN")
+		target_compile_definitions(${PROJECT_NAME} PRIVATE WORDS_BIGENDIAN=1)
+	endif()
+else()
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -767,7 +767,11 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
 endif()
 
 # Endianess check
-if(NOT DEFINED CMAKE_TOOLCHAIN_FILE)
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.20)
+	if (CMAKE_CXX_BYTE_ORDER STREQUAL "BIG_ENDIAN")
+		target_compile_definitions(lcf PRIVATE WORDS_BIGENDIAN=1)
```
carstene1ns, line 772: Copy paste error:
```suggestion
		target_compile_definitions(${PROJECT_NAME} PRIVATE WORDS_BIGENDIAN=1)
```

Also the liblcf comment applies here as well.
There is an inconsistency: We could change `PUBLIC`to `PRIVATE` in line 778 or use `PUBLIC` here.
It does not propagate to the executable this way, but we do not need endianess there.

carstene1ns (changes requested):

CMakeLists.txt
Ghabry: right sdl_ui is built in target ``PROJECT_NAME`` so ``PRIVATE`` works here

carstene1ns (approved):

PR #3074, opened 2023-08-14
Debug scene: Add option to set the hero movement speed

Ghabry:
because why not.

Legal values are from 1-7. Max speed via events is 6.

7 still works but 8 and higher break the screen scrolling :sweat_smile:

### DIFF
--- a/src/scene_debug.cpp
+++ b/src/scene_debug.cpp
@@ -401,6 +401,15 @@ void Scene_Debug::vUpdate() {
 					PushUiVarList();
 				}
 				break;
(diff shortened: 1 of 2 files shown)


jetrotal:
I have the same feature implemented as a command, maybe it can be useful: 

https://github.com/EasyRPG/Player/commit/f4a940a949f966f6f4b4c04802c1d9a6898e0035#diff-6201ae4886638abb44984fb7d69b851b9596edd6cd6615c7a7738060150acab8R4660

fdelapena (approved):

PR #3080, opened 2023-09-07
First attempt at a fix for multi-game functionality

sorlok:
Hi all,

I implemented the "multi game" functionality a while ago, and it seems that a recent-ish commit broke it. I've done a cursory fix of the functionality, and confirmed that it now works on "The Way: Episode 2". The fix uses "FileSystemView::Exists()" (rather than "FindFile") since it needs to access files outside of the save directory (e.g., from "Episode1" we need to read "../Episode2/<save_files>". 

This also fixes a few smaller issues: infinite looping in Update(), and failing to call Start() in the parent class. Tested on Windows.

Definitely open to suggestions and comments, but I'd like to get this feature back to (at least) what it was doing before.

Thanks!

### DIFF
--- a/src/meta.cpp
+++ b/src/meta.cpp
@@ -142,30 +142,27 @@ std::vector<Meta::FileItem> Meta::SearchImportPaths(const FilesystemView& parent
 
 
 std::vector<Meta::FileItem> Meta::BuildImportCandidateList(const FilesystemView& parent_fs, StringView child_path, StringView parent_game_name, int pivot_map_id) const {
-	(void)parent_fs;
-	(void)child_path;
-	(void)parent_game_name;
-	(void)pivot_map_id;
-
 	// Scan each folder, looking for an ini file
 	// For now, this only works with "standard" folder layouts, since we need Game files + Save files
 	std::vector<Meta::FileItem> res;
-#if 0
-	FIXME
+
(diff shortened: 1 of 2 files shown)


Ghabry:
Oh my bad. I completely forgot about this feature and that it is still broken :/

Thanks for the suggested fix. I will review it next week

fdelapena:
jenkins: add to whitelist

sorlok:
Any idea why Android is failing? I see:

`> [CXX1409] /var/lib/jenkins/workspace/player-android-pr/builds/android/app/.cxx/Debug/303n5v6u/arm64-v8a/android_gradle_build.json debug|arm64-v8a : expected buildFiles file '/var/lib/jenkins/workspace/player-android-pr/builds/cmake/Modules/Findlhasa.cmake' to exist`

...but I definitely didn't change that file.

Ghabry:
That's a caching issue. Ignore it

Ghabry:
I think the problem here was that for a proper "go up" the full path to the directory is required instead of "." and getting the path was never implemented.

Though considering that this feature is quite exotic I think your workaround is fine for now. (also the saves would be somewhere else when the game is in a ZIP making the problem even harder ^^')

Ghabry (approved):

sorlok:
Thanks, that all makes sense to me.

PR #3087, opened 2023-09-13
Ineluki MP3 patch improvements

### LINKED ISSUE #2598: Ineluki .link files are only supported for BGM's but not for SE's
#### Name of the game:
Any game with Inelukis MP3 patch (Sternenkindsaga uses it for SE's)

#### Player platform:
Any

I guess this was an oversight when the MP3 patch was implemented:

Inelukis “link.wav” files work fine as BGM but any time a sound effect with the ending .link.wav is played, EasyRPG Player just states: _“Couldn’t play SE . Format not supported”_

### LINKED ISSUE #2609: *.link.wav is not just reserved for Ineluki's MP3 Patch
#### Name of the game:
Les aventures d'un avatar 6

#### Player platform:
Windows

The french community used a tool called "Piez-o-matic" back in the day, to easily convert *.mp3 files to *.wav on the user side. This allowed them create smaller game downloads without having to resort to a MP3 patch.

I just found a game that was delivered with music files ending with "*.link.mp3" which are then converted to "*.link.wav". The EasyRPG Player does have a problem with these files and just states: "BGM file not readable:"
screenshot_0
This game doesn't make use of Ineluki's MP3 patch but still uses the extension. I checked a few other games that were bundled with Piez-o-matic and this one seems to be the only one where that is the case. (But there could be more out there)

Ghabry:
Only consider files as links when they are small (< 500 Byte). "Les aventures d'un avatar 6" uses files with such names for real audio files.

Support of links for sound effects with the same logic as BGM. Note that the "loop" statement is not supported.

Fix #2598
Fix #2609

### DIFF
--- a/src/game_system.cpp
+++ b/src/game_system.cpp
@@ -245,7 +245,6 @@ void Game_System::ResetSystemGraphic() {
 	ReloadSystemGraphic();
 }
 
-
 template <typename T>
 static const T& GetAudio(const T& save, const T& db) {
 	return save.name.empty() ? db : save;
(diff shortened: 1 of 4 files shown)


src/game_system.cpp
```diff
@@ -580,6 +580,38 @@ void Game_System::OnSeReady(FileRequestResult* result, lcf::rpg::Sound se, bool
...
+			se_request_ids[line] = request->Bind(&Game_System::OnSeInelukiReady, this, se);
+			request->Start();
+			return;
+		}
+
+		se_cache = AudioSeCache::Create(std::move(stream), result->file);
+	}
+
+	if (!se_cache) {
+		Output::Warning("Sound {}: Format not supported", result->file);
```
fdelapena, line 599: Maybe this brings warnings where RPG_RT.exe don't? Use Output::Debug instead?

fdelapena (approved):

src/game_system.cpp
Ghabry: this is copy-pasted from the non-Ineluki SE handler.
The warning will only trigger when a SE file is unsupported format wise. Missing files won't trigger it so should be save imo

PR #3092, opened 2023-09-22
Midi: Properly shutdown and reinitialize

Ghabry:
First commit fixes an oversight in the shutdown logic of the Player: Most audio stuff was not deleted because they were static variables... 

This fixes the fluidsynth crashes on shutdown.

_Should also fix the libretro crash when playing a second game but I havn't tested this (yet)_.

-------------

Second commit is less serious: Fluidsynth and WildMidi are now reinitialized when loading a second game. This saves RAM (unloaded when not used by the game) and allows per-game soundfonts. Before only the first was loaded.

### DIFF
--- a/src/audio_generic.cpp
+++ b/src/audio_generic.cpp
@@ -22,31 +22,20 @@
 #include <memory>
 #include "audio_decoder_midi.h"
 #include "audio_generic.h"
-#include "audio_generic_midiout.h"
-#include "filefinder.h"
 #include "output.h"
-
-GenericAudio::BgmChannel GenericAudio::BGM_Channels[nr_of_bgm_channels];
-GenericAudio::SeChannel GenericAudio::SE_Channels[nr_of_se_channels];
-bool GenericAudio::BGM_PlayedOnceIndicator;
-
-std::vector<int16_t> GenericAudio::sample_buffer = {};
-std::vector<uint8_t> GenericAudio::scrap_buffer = {};
-unsigned GenericAudio::scrap_buffer_size = 0;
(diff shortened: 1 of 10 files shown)


src/audio_generic.cpp
```diff
@@ -22,31 +22,20 @@
 #include <memory>
 #include "audio_decoder_midi.h"
 #include "audio_generic.h"
-#include "audio_generic_midiout.h"
```
antecrescent, line 25: 7269d325035f547907ec352bff5d39bc611abc88 breaks compilation with for platforms libretro, psvita, sdl and switch when compiling audio.o because they include `audio_generic.h`. Since this commit, said header includes `audio_generic_midiout.h` which does not provide a definition for `AudioDecoderMidi` itself.
This used to be no problem because `audio_generic_midiout.h` was only included from `audio_generic.cpp` which also includes the missing header.

The fix is to include `audio_decoder_midi.h` either in `audio_generic.h` or in `audio_generic_midiout.h`.

src/audio_generic.cpp
```diff
@@ -22,31 +22,20 @@
 #include <memory>
 #include "audio_decoder_midi.h"
```
antecrescent, line 23: This line can be deleted if  you move `audio_generic_midiout.h` (see next comment).

antecrescent (changes requested):

src/audio_generic.cpp
Ghabry: C++ Header mess :/. Thanks for testing this!

fdelapena (approved):

PR #3095, opened 2023-09-23
Detect the engine version for "Ahriman's Prophecy" correctly

Ghabry:
It is RPG Maker 2003 1.0.8.0 but their exe file lacks a VERSIONINFO.

The code segment size is checked instead to detect the version correctly.

----------

Segment size taken from this Makerpendium page: https://www.makerpendium.de/index.php/RPG_RT.exe

Finally a simple commit to review ;)

----------

Reported by samuel in the forum https://community.easyrpg.org/t/ahrimans-prophecy-battle-menu-problem-android/1332 Thanks!

### DIFF
--- a/src/exe_reader.cpp
+++ b/src/exe_reader.cpp
@@ -327,7 +327,14 @@ int EXEReader::FileInfo::GetEngineType(bool& is_maniac_patch) const {
 			// VALUE! or Rpg2k3 < 1.0.2.1
 			// Check CODE segment size to be sure
 			if (code_size > 0xB0000) {
-				return Player::EngineRpg2k3;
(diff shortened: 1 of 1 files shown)


fdelapena (approved):

Ghabry:
To have this documented:

UPX compressing preserves the VERSIONINFO and the LOGO resources. However the CODE segment is gone.

This means a UPX compressed RPG Maker 2003 < 1.02 game will be incorrectly detected as RPG Maker 2000 VALUE!

All other UPX'ed RPG Maker 2003 games (except with hacked exe that deleted the VERSIONINFO :P) will be correctly detected as "code size" is only used for the VALUE! vs. ancient 2k3 game check.

PR #3096, opened 2023-09-25
Custom Logos at Startup

jetrotal:
To add custom logos, you need to add image files named LOGO1, LOGO2, LOGO3, LOGO`n` to the Font folder. 
It will always load the next logo file, while sequential files exist in the Font folder.

~~You can skip having LOGO0, it loads default easyRPG logo when LOGO0 doesn't exists.~~
EasyRPG logo is always displayed at bootup, to avoid a blank screen while loading a heavy game.

Some reference files for testing:

LOGO1
LOGO2
LOGO3

### DIFF
--- a/src/scene_logo.cpp
+++ b/src/scene_logo.cpp
@@ -160,7 +160,7 @@ std::vector<BitmapRef> Scene_Logo::preloadLogos() {
 
 	for (int logoIndex = 0; ; logoIndex++) {
 		Filesystem_Stream::InputStream logoStream;
-		if (logoIndex != 0) logoStream = FileFinder::OpenImage("Font", "LOGO" + std::to_string(logoIndex));
+		if (logoIndex != 0) logoStream = FileFinder::OpenImage("Logo", "LOGO" + std::to_string(logoIndex));
 		//TODO: Maybe get LOGO1,LOGO2,LOGO3 from rpg_rt too?
 
 		if (!logoStream) {
--- a/src/scene_logo.cpp
+++ b/src/scene_logo.cpp
@@ -40,16 +40,39 @@ Scene_Logo::Scene_Logo() :
 	type = Scene::Logo;
 }
 
+Filesystem_Stream::InputStream logo_stream;
+Filesystem_Stream::InputStream next_logo;
+
 void Scene_Logo::Start() {
 	if (!Player::debug_flag && !Game_Battle::battle_test.enabled) {
 		std::time_t t = std::time(nullptr);
 		std::tm* tm = std::localtime(&t);
 
-		if (Rand::ChanceOf(1, 32) || (tm->tm_mday == 1 && tm->tm_mon == 3)) {
-			logo_img = Bitmap::Create(easyrpg_logo2, sizeof(easyrpg_logo2), false);
-		} else {
-			logo_img = Bitmap::Create(easyrpg_logo, sizeof(easyrpg_logo), false);
+		if (FileFinder::Game()) logo_stream = FileFinder::OpenImage("Font", "LOGO" + std::to_string(Player::current_logo));
+		//TODO: Maybe get LOGO1,LOGO2,LOGO3 from rpg_rt too?
+
+		if (!logo_stream && Player::current_logo == 0) {
+
+			if (Rand::ChanceOf(1, 32) || (tm->tm_mday == 1 && tm->tm_mon == 3)) {
+				logo_img = Bitmap::Create(easyrpg_logo2, sizeof(easyrpg_logo2), false);
+			}
+			else {
+				logo_img = Bitmap::Create(easyrpg_logo, sizeof(easyrpg_logo), false);
+			}
+
 		}
+		else {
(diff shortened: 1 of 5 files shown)


Mimigris:
From a quick test:
- I'm unable to replace the LOGO0 on my side, it is completely ignored and not read at all
- Holding Shift while the other logos are displayed will wait until you no longer hold it to go to the next scene, as a leftover of the version details present on the default scene, I'm not sure if it should be kept?
- The v0.8 line without the details is not erased when pressing Shift, so it is then weirdly displayed (look at v0.8)
screenshot_2

Didn't tested more outside of that, but that seems promising 👍

src/scene_logo.cpp
```diff
@@ -40,16 +40,39 @@ Scene_Logo::Scene_Logo() :
...
 void Scene_Logo::Start() {
 	if (!Player::debug_flag && !Game_Battle::battle_test.enabled) {
 		std::time_t t = std::time(nullptr);
 		std::tm* tm = std::localtime(&t);
 
-		if (Rand::ChanceOf(1, 32) || (tm->tm_mday == 1 && tm->tm_mon == 3)) {
-			logo_img = Bitmap::Create(easyrpg_logo2, sizeof(easyrpg_logo2), false);
-		} else {
-			logo_img = Bitmap::Create(easyrpg_logo, sizeof(easyrpg_logo), false);
+		if (FileFinder::Game()) logo_stream = FileFinder::OpenImage("Font", "LOGO" + std::to_string(Player::current_logo));
```
Ghabry, line 51: Using ``FileFinder::Game`` in ``Start`` has multiple problems:

1. The Game filesystem is not configured in all cases. Actually only when you use ``--project-path`` but not when you double click the Player. In that case (see ``vUpdate``) the Game FS is ``nullptr`` and it uses ``GetDefaultProjectPath``. This could be moved to ``Start`` but then there is the next problem:

2. You parse the filesystem while no logo is visible. This gives a bad user experience because reading the filesystem can be slow (that's the main reason why the heavy lifting happens in ``Update``). This means there can be seconds of black screen before the logo appears. This will be noticable on slow homebrew platforms but also on Android where parsing the filesystem can take seconds (thanks Google). In the web player you would need to wait for an entire download to finish before a logo can be displayed.

What would obviously work is that the EasyRPG logo is displayed and if there are more logos it cycles through them.

jetrotal: I'm trying to change how the logos are loaded, by storing them in a Vector before start loading them:

```cpp
void Scene_Logo::LoadLogos() {
	Player::current_logo = 0;
	Player::logos_container.clear();

	bool defaultLogoLoaded = false;

	while (true) {
		Filesystem_Stream::InputStream next_logo = FileFinder::OpenImage("Font", "LOGO" + std::to_string(Player::current_logo));

		if (!next_logo) {
			// If the first logo is missing, load the default logos
			if (Player::current_logo == 0 && !defaultLogoLoaded) {
				std::time_t t = std::time(nullptr);
				std::tm* tm = std::localtime(&t);

				if (Rand::ChanceOf(1, 32) || (tm->tm_mday == 1 && tm->tm_mon == 3)) {
					Player::logos_container.push_back(Bitmap::Create(easyrpg_logo2, sizeof(easyrpg_logo2), false));
				}
				else {
					Player::logos_container.push_back(Bitmap::Create(easyrpg_logo, sizeof(easyrpg_logo), false));
				}
				defaultLogoLoaded = true;
			}
			else {
				// Break the loop if OpenImage fails for other logos
				break;
			}
		}
		else {
			Player::logos_container.push_back(next_logo);
		}

		Player::current_logo++;
	}
}
```


but not idk what exactly needs to be done, and the instructions are too complex for asking help on chatGPT.

but yeah, logo0 could always be the easyRPG one.

Ghabry:
Also for discussion: Is ``Font\LogoN`` a good location for this? Are there better folder suggestions? Seperate ``Logo`` folder?

Ghabry (approved):
I rewrote the logo loading logic to always only load one logo and I got rid of global state. The remaining technique (such as invoking another logo scene) remained the same.

Is now much faster when loading 200 logos :P. There is no upper limit.

jetrotal:
> I rewrote the logo loading logic to always only load one logo and I got rid of global state.

Thanks! When I use the flag --new-game with the 200+ logo collection, it gets stuck at the logos screen for some time.

Could a skip after the first be added in this specific case?

Ghabry:
Right. The same will happen when running in test game mode. Will add a fix.

Ghabry:
The logos are now correctly skipped when ``TestPlay`` or ``BattleTest`` are used. ``--new-game`` will only skip it in combination with one of them (like before).

fdelapena (approved):

MesaBlack:
how about a custom video at startup?

Mimigris:
> how about a custom video at startup?

At this point, you could probably skip the title screen in your game and manually create a video on your side when starting the game, it would probably be better to do it this way.

PR #3106, opened 2023-09-29
Add ExecuteCommand(com);

jetrotal:
Minor Change.
It's a variation of ExecuteCommand(); that accepts `com` as a parameter.

Ghabry commented that this would make the code faster. But, this also open possibilities to create a scripting language, since we can create `com` parameters from strings.

------------------

This implementation could also be used to attach code from dynRPG patches to new `com` stuff. Having both @comments and `com` commands share the same code.

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -615,7 +615,7 @@ lcf::rpg::MoveCommand Game_Interpreter::DecodeMove(lcf::DBArray<int32_t>::const_
 bool Game_Interpreter::ExecuteCommand() {
 	auto& frame = GetFrame();
 	const auto& com = frame.commands[frame.current_command];
-	return Game_Interpreter::ExecuteCommand(com);
+	return ExecuteCommand(com);
 }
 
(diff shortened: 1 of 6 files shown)


src/game_interpreter.cpp
```diff
@@ -615,7 +615,10 @@ lcf::rpg::MoveCommand Game_Interpreter::DecodeMove(lcf::DBArray<int32_t>::const_
 bool Game_Interpreter::ExecuteCommand() {
```
Ghabry, line 615: You can remove the no argument ``()`` version of this function from map and battle as they have the same code. And in game_interpreter.h you can delete the virtual as the function is not supposed to be overridden by sub classes

You said in chat this was needed because of a stackoverflow but I don't see why. Was probably a different issue.

Ghabry:
> Ghabry commented that this would make the code faster.

For context: When the event is in ``Game_Interpreter`` this will avoid fetching the command a second time via

```cpp
auto& frame = GetFrame();
const auto& com = frame.commands[frame.current_command];
```

You can likely not measure the speedup but this code can be executed 10000s of times per second so every nanosecond counts :D.

Ghabry (approved):

Issue #3104, opened 2023-09-28
Settings: Incorrect input help text UwU


MikuChan03:
🎶 Hi 🎶

When watashi was using the new settings menu fow advanced stuffs, the title line was all wwong O.O
But aftew some weseawch, the pwoblem is wathew obvious.
In input_buttons.h, the kButtonHelp has 44 entwies, while InputButton and kButtonNames have only 42 (<- answew
to evewything ✨✨✨✨✨(ﾉ◕ヮ◕)ﾉ*:･ﾟ✧✨✨✨✨✨)

The double entwies awe:
"Move up one page in menus",
"Move down one page in menus",
"Page up key",
"Page down key",

These awe all supposed to be fow
Input -> Key/Button mapping -> Engine -> Page Up
Input -> Key/Button mapping -> Engine -> Page Down

Miku wants to humbly suggest wemoving the latew two entwies, because only entwies in
Input -> Key/Button mapping -> Game
evew have the "key" suffix in the descwiption.

(Btw, what's the likelihood of none of these paths containing any 'w's ????)

This about fixie wixies the pwoblem xD
Bye! 💕

jetrotal:
![image](https://github.com/EasyRPG/Player/assets/45118493/bb30c6f7-fafb-42ab-8f3c-fdc12bb55e61)

twanwkws fwow twew wepwowt

PR #3107, opened 2023-09-30
Update key descwiptions

MikuChan03:
Hi~~

Watakushi (said when sewious), doesn't want to be a bothew,
but watakushi believes that the wwong key descwiptions could make usews dizzy.

screenshot_1

So hewe's a tiny PW that changes what I said in #3104. While at it, we also wemove the x3 and x10
pawt in the fast fowward descwiptions, because that's configuwable now


jetrotal:
Heyo! 

Maybe renaming FF buttons to
"Fast Foward A" and "Fast Foward B" could help, since nowdays it's possible to make the second slower than the first.

Not necessary, but I wonder if it would be possible to concat the speed value from the FF variables to help showing what they do?

Thanks again!

MikuChan03:
@jetrotal sure, will do ^.^
Though if we weally wanted to go all out, we would make the cweation of these
speed modifiews and the keybindings of them dynamic, meaning the usew could have
like 10 diffewent speed modifiews?

jetrotal:
hm... unlimited speed modifiers? I wonder...

Though, there are games that uses a lot of buttons already, and we plan to reserve more shortcuts to when we add support do multiplayer from https://github.com/monokotech/EasyRPG-Multiplayer-Native

Maybe 2 is already a good amount 🤔

Mimigris:
If you want my opinion, 2 buttons to edit the speed is likely good enough for now, it could still be changed later in any case if there's a request. 👍

MikuChan03:
KK,

watashi did all that good stuff! Meaning watashi changed not only the usew facing descwiptions of fast forward,
but also the source code. Only the actual Enums wewe too scawy fow watashi, because they awe being used in platfowms
as well.

screenshot_0

Made it so that the descwiption is dynamically updated too:

screenshot_1

Everything is happening so much 🤪

jetrotal:
Thanks for the help, very sugoi and very kakkoii.

Ghabry:
Jenkins: test this please

MikuChan03:
Sorry, forgot to exchange a variable name in platforms. Will do so right away.

MikuChan03:
Jenkins: test this please

Ghabry:
Jenkins: test this please

Sowwy, only admins can do this

Ghabry:
Imo this looks fine. I just want to do a chance on the Android Java Ui side so it reads/writes the value directly through the INI like we already do for some other settings.

Will do this next week, I don't have all the Android dev env crap setup on this system.

fdelapena (approved):

Ghabry:
Currently not motivated to do the Android stuff. Merging this for now and fixing that minor problem later

TruncatedDinoSour:
a

DeveloLongScript:
github in a nutshell

**trolls**

PR #3109, opened 2023-10-01
New Command: 2049 - SetGameSpeed

jetrotal:
Minor experiment, our first new Event Command.

---------------------------

To use it, write in Maniacs TPC:
`@raw 2049, "", speedIsVar, speed, speedLimitIsVar,speedLimit, hideSpeedMultiplierIsVar, hideSpeedMultiplier `

e.g.:
`@raw 2049, "", 0, 10, 0, 50, 0, 1`

The command must be inserted inside a parallel process, because `Game_Clock::SetGameSpeedFactor(speed);` only changes the speed at current frame.

image image

Any refactor (to improve how it's called or to hide the speed counter) is welcome.

### DIFF
--- a/src/fps_overlay.cpp
+++ b/src/fps_overlay.cpp
@@ -85,7 +85,7 @@ void FpsOverlay::Draw(Bitmap& dst) {
 	}
 
 	// Always drawn when speedup is on independent of FPS
-	if (last_speed_mod > 1) {
+	if (!Game_Clock::GetSpeedOverlayMode() && last_speed_mod > 1) {
 		if (speedup_dirty) {
 			std::string text = "> x" + std::to_string(last_speed_mod);
 
--- a/src/game_clock.cpp
+++ b/src/game_clock.cpp
@@ -25,6 +25,8 @@
 constexpr bool Game_Clock::is_steady;
 Game_Clock::Data Game_Clock::data;
 
+int hideDisplay = 0;
+
 // Damping factor fps computation.
 static constexpr auto _fps_smooth = 2.0f / 121.0f;
 
--- a/src/game_clock.h
+++ b/src/game_clock.h
@@ -36,6 +36,8 @@ public:
 	using duration = clock::duration;
 	using time_point = clock::time_point;
 
+	int hideDisplay;
+
 	static constexpr bool is_steady = clock::is_steady;
 
 	/** Get current time */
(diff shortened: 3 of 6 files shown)


Ghabry:
What is the motivation behind this command?

I don't want to be too negative but I already see the Yume2 gatekeepers adding this command so you are not playing the game incorrectly like the already did with the ini setting to kill widescreen.

jetrotal:
I was thinking about cases where a dev would that attach it to a custom button, or select the speed of dialogs only, or even in a minigame that requires faster game speed.

lemme know if this is too specific or if it can annoy other people.



Lemme know if that seems useless...

Mimigris:
> What is the motivation behind this command?

On some Maniac games like [THUNDER WINDY -Castle of the UndeadKnigh-](https://viprpg2020kohaku.xxxxxxxx.jp/entry16.html) or [Suika's drunk runaway](https://poyouli.itch.io/suikas-drunk-runaway), the Maniac command Set Game Option was used with the setting Set FPS (this changes the speed of the game, a bit like what is done here though it allows another range of values): in the first case, it was done to speed up a game that was created as being slow, while in the latter case, it was done to speed up a loading (which is the generation of a map with the Set Tile command of the patch). This could serve as examples of what could be done with a command like this one.

> And yeah, mimigris comented that the 2kki devs already discovered other ways of detect if a game is using easyRPG.

Hopefully they no longer try to detect the Player to do different things when using it nowadays, they just list sometimes the issues that they have in the changelog of the game and on the [wiki page of differences](https://wikiwiki.jp/yumenikki-g3/%E3%83%84%E3%82%AF%E3%83%BC%E3%83%AB%E8%B1%86%E7%9F%A5%E8%AD%98/EasyRPG%E3%81%A8%E3%82%AA%E3%83%AA%E3%82%B8%E3%83%8A%E3%83%AB%E3%81%AE%E5%B7%AE%E7%95%B0), though we still need to keep our eyes on them from time to time to avoid a situation like this to reoccur intentionally.

Ghabry:
The explanations sounds reasonable so this PR can stay open until I have some time to think about it more

jetrotal:
> The explanations sounds reasonable so this PR can stay open until I have some time to think about it more

Now the command multiplies the current speed value, protecting the FF buttons from any 2kki intentions.
And it also caps at 100x speed, to avoid lagging the engine.

Ghabry:
@jetrotal I think I will close this PR. I will instead implement this FPS control from the Maniac Game Option command that you documented in #3147 . The Maniac one has the advantage that it can be a fraction and even slow down (and fast forward will be a multiple of it)

jetrotal:
Ok, that one seems better

PR #3110, opened 2023-10-02
New Command: 99999 - Barebones Scripting Language

jetrotal:
TPC commands for testing purposes:
```js
@raw 99999, "@10110(""test"",[1 ,2, 3,4 ,5,6],0) //show dialog" 
@raw 99999, "@10310 ("""",[0 ,0, 100],0) //add 100 money" 
@raw 99999, "@10110(""@10110(\""test\"",[1 ,2, 3,4 ,5,6],0)"",[1 ,2, 3,4 ,5,6],0) //a collection of problematic syntax just to test how it deals with it"
```
Work in progress. I need a lot of help with this one.

It injects commands from 99999's com.string.

I can't parse proper syntax, since it isn't mapped anywhere. 
But it can be useful alongside stringvars to load commands from txt files.

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -66,6 +66,7 @@
 #include "baseui.h"
 #include "algo.h"
 #include "rand.h"
+#include <regex>
 
 enum BranchSubcommand {
 	eOptionBranchElse = 1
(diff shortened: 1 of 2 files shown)


fdelapena:
Looks promising, just some observations:
- Because regex has portability issues, maybe is worth to replace it with some str functions.
- Is 99999 used by an existing command? Otherwise maybe is better to use the maximum "reserved" value from easyrpg command IDs.

jetrotal:
Not needed anymore, #3137 covers and improves this idea.

PR #3124, opened 2023-10-15
10220 - Control Variables New Entry: GetTime

jetrotal:
This one was requested a lot by the Creative Unconscious devs,
They want to include special events related to christmas date or other holydays:

### 10220 - Control Variables New Entry: GetTime
Gets current date information and attach it to a variable.
image

Usage:
```js
@raw 10220, "Type_of_Date", targetVar_IsVar, targerVar

```
```js
// "set year to variable 1:"
@raw 10220, "getYear", 0, 1
// "set month to variable 2:"
@raw 10220, "getMonth", 0, 2
// "set day to variable 3:"
@raw 10220, "getDay", 0, 3
// "set hour to variable 4:"
@raw 10220, "getHour", 0, 4
// "set minute to variable 5:"
@raw 10220, "getMinute", 0, 5
// "set second to variable 6:"
@raw 10220, "getSecond", 0, 6
// "set day of week to variable 7:"
@raw 10220, "getWeekDay", 0, 7
// "set days in year to variable 8:"
@raw 10220, "getYearDay", 0, 8
// "set daylight savings flag to variable 9:"
@raw 10220, "getdaylightsavings", 0, 9
// "set the entire timestamp to variable 10:"
@raw 10220, "gettimestamp", 0, 10
```

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -821,8 +821,6 @@ bool Game_Interpreter::ExecuteCommand(lcf::rpg::EventCommand const& com) {
 			return CommandManiacSetGameOption(com);
 		case Cmd::Maniac_CallCommand:
 			return CommandManiacCallCommand(com);
-		case static_cast<Game_Interpreter::Cmd>(2054): //Cmd::EasyRpg_GetTime
-			return CommandGetTime(com);
 		default:
 			return true;
 	}
@@ -1074,6 +1072,33 @@ bool Game_Interpreter::CommandControlSwitches(lcf::rpg::EventCommand const& com)
 }
 
 bool Game_Interpreter::CommandControlVariables(lcf::rpg::EventCommand const& com) { // code 10220
+
+	if (!com.string.empty()){
+	std::string periodName = Utils::LowerCase(ToString(com.string));
+	int32_t outputVariable = ValueOrVariable(com.parameters[0], com.parameters[1]);
+
+	std::time_t t = std::time(nullptr);
+	std::tm* tm = std::localtime(&t);
+
+	int32_t dateOutput{};
+
+	if (periodName == "getyear") dateOutput = tm->tm_year + 1900;
+	if (periodName == "getmonth") dateOutput = tm->tm_mon + 1;
+	if (periodName == "getday") dateOutput = tm->tm_mday;
+	if (periodName == "gethour") dateOutput = tm->tm_hour;
+	if (periodName == "getminute") dateOutput = tm->tm_min;
+	if (periodName == "getsecond") dateOutput = tm->tm_sec;
+	if (periodName == "getweekday") dateOutput = tm->tm_wday + 1;
+	if (periodName == "getyearday") dateOutput = tm->tm_yday + 1;
+	if (periodName == "getdaylightsavings") dateOutput = tm->tm_isdst + 1;
+	if (periodName == "gettimestamp") dateOutput = t;
+
+	Main_Data::game_variables->Set(outputVariable, dateOutput);
+	Game_Map::SetNeedRefresh(true);
+
+	return true;
+	}
+
 	int value = 0;
 	int operand = com.parameters[4];
 
@@ -4642,32 +4667,6 @@ bool Game_Interpreter::CommandManiacCallCommand(lcf::rpg::EventCommand const&) {
 	}
 
 	Output::Warning("Maniac Patch: Command CallCommand not supported");
-	return true;
-}
-
-bool Game_Interpreter::CommandGetTime(lcf::rpg::EventCommand const& com) {
-
-	std::string periodName = Utils::LowerCase(ToString(com.string));
-	int32_t outputVariable = ValueOrVariable(com.parameters[0], com.parameters[1]);
-
-	std::time_t t = std::time(nullptr);
-	std::tm* tm = std::localtime(&t);
-
-	int32_t dateOutput{};
-
-	if (periodName == "getyear") dateOutput = tm->tm_year + 1900;
-	if (periodName == "getmonth") dateOutput = tm->tm_mon + 1;
-	if (periodName == "getday") dateOutput = tm->tm_mday;
-	if (periodName == "gethour") dateOutput = tm->tm_hour;
-	if (periodName == "getminute") dateOutput = tm->tm_min;
-	if (periodName == "getsecond") dateOutput = tm->tm_sec;
-	if (periodName == "getweekday") dateOutput = tm->tm_wday +1;
-	if (periodName == "getyearday") dateOutput = tm->tm_yday +1;
-	if (periodName == "gettimestamp") dateOutput = t;
-
-	Main_Data::game_variables->Set(outputVariable, dateOutput);
-	Game_Map::SetNeedRefresh(true);
-
 	return true;
 }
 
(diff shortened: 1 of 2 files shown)


Mimigris:
Hmmm, I didn't tested the pull request, though I think that having the possibility to directly have a way to get these info is a really good idea. Though, while I know that in terms of writing the format of a command, you may be limited by how the Maniac patch allows you to write things, but wouldn't it make more sense to extend the Control Variables command instead of creating a new command for that? I think it would be a weird thing to create a command just for that instead of extending the Control Variables command, which would allow a bit more possibilities, as well as not creating thousands of commands each time a new command to get a value is created.

jetrotal:
>wouldn't it make more sense to extend the Control Variables command instead of creating a new command for that?

It would be harder to create and maintain commands, since neither vanilla nor maniacs have a way to allow this type of change on a pre existing command.

The user would have to edit a xml version of his map or ldb, then editing the command on any way would break it.

carstene1ns:
> since neither vanilla nor maniacs have a way to allow this type of change on a pre existing command.

Just use EasyRPG Editor for this? 😅

Also, should be possible to patch this in. Cherry has added a lot of functionality through binary patches in the official editor. ~~We~~Someone could add a compatibility shim to allow such things.

jetrotal:
> Just use EasyRPG Editor for this? 😅

Having an unfinished editor just for command snippets sounds worse than using `@raw` commands from a vastly used patch.

The editor is not fully usable rn, we would have to work on it for 2 or 3 years to have it fully working, and let the CU community use it. 

> Also, should be possible to patch this in. Cherry has added a lot of functionality through binary patches in the official editor. ~We~ Someone could add a compatibility shim to allow such things.

Again, we already have have maniacs patch `@raw commands` letting us access and write new functions. Since the patch It's widelly used, it makes more sense to use it instead creating yet another patch.

IMHO, We should bring users of different patches closer to easyRPG, instead of creating more and more alternatives to tweak the editor.

----------------

The common response for asking for new features was "no, wait for the editor". 
Those raw commands is finally a way to do it ourselves and let those new features be future proof, since all new commands are using the same `executeCommand(com)` functions that every other command vanilla command uses.

src/game_interpreter.cpp
```diff
@@ -4643,6 +4645,34 @@ bool Game_Interpreter::CommandManiacCallCommand(lcf::rpg::EventCommand const&) {
...
+	if (periodName == "gethour") dateOutput = tm->tm_hour;
+	if (periodName == "getminute") dateOutput = tm->tm_min;
+	if (periodName == "getsecond") dateOutput = tm->tm_sec;
+	if (periodName == "getweekday") dateOutput = tm->tm_wday +1;
+	if (periodName == "getyearday") dateOutput = tm->tm_yday +1;
+
+	std::vector<int32_t> updateVarParams = { 0, outputVariable, 0, 0, 0, dateOutput };
+	outputCommand.parameters = lcf::DBArray<int32_t>(updateVarParams.begin(), updateVarParams.end());
+
+	return ExecuteCommand(outputCommand);
```
Ghabry, line 4673: Is there a reason why you delegate this to the ControlVariables command instead of setting the variables directly?

jetrotal: I thought it would be better to not rewrite existing code. 
and that would also cover setting up savefiles.

Though I could simply use `CommandControlVariables(outputCommand);` instead of `executeCommand(outputCommand);`

jetrotal: fixed using

```cpp
Main_Data::game_variables->Set(outputVariable, dateOutput);
Game_Map::SetNeedRefresh(true);
```

Ghabry:
There are valid concerns raised if this should be part of the "Control Variables" command. I will check first if there is a non-shitty way to achieve this currently...

With lcf2xml for sure but most are using this closed source TPC thing.

jetrotal:
> There are valid concerns raised if this should be part of the "Control Variables" command. I will check first if there is a non-shitty way to achieve this currently...
> 
> With lcf2xml for sure but most are using this closed source TPC thing.

All I can think is whitelisting some of DynRPG commands to work even without it enabled.
Then create non destructive ways to run @raw commands from those comments, similar to #3110.

Ghabry:
With that ``raw`` workaround proposed this can move into ControlVariables.

Suggested boilerplate:

```patch
diff --git a/src/game_interpreter.cpp b/src/game_interpreter.cpp
index c0c1aa9ec..02c6af39d 100644
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -1250,6 +1250,26 @@ bool Game_Interpreter::CommandControlVariables(lcf::rpg::EventCommand const& com
 			// Expression (Maniac)
 			value = ManiacPatch::ParseExpression(MakeSpan(com.parameters).subspan(6, com.parameters[5]), *this);
 			break;
+		case 200: { // EasyRPG Command
+			auto cmd_size = com.parameters.size();
+			int op, subop;
+			int arg1 = 0;
+			int arg2 = 0;
+			if (cmd_size >= 7) {
+				op = com.parameters[5];
+				subop = com.parameters[6];
+				if (cmd_size >= 8) {
+					arg1 = com.parameters[7];
+				}
+				if (cmd_size >= 9) {
+					arg2 = com.parameters[8];
+				}
+				value = ControlVariables::EasyRpgCommand(op, subop, arg1, arg2);
+			} else {
+				value = 0;
+			}
+			break;
+		}
 		default:
 			Output::Warning("ControlVariables: Unsupported operand {}", operand);
 			return true;
diff --git a/src/game_interpreter_control_variables.cpp b/src/game_interpreter_control_variables.cpp
index d8c0590dd..903970038 100644
--- a/src/game_interpreter_control_variables.cpp
+++ b/src/game_interpreter_control_variables.cpp
@@ -488,3 +488,24 @@ int ControlVariables::Divmul(int arg1, int arg2, int arg3) {
 int ControlVariables::Between(int arg1, int arg2, int arg3) {
 	return (arg1 >= arg2 && arg2 <= arg3) ? 0 : 1;
 }
+
+int ControlVariables::EasyRpgCommand(int op, int subop, int arg1, int arg2) {
+	(void)arg1;
+	(void)arg2; // Remove when args are used
+
+	enum class Op {
+		GetDate = 0
+	};
+
+	switch (static_cast<Op>(op)) {
+		case Op::GetDate:
+			switch (subop) {
+				case 0: // TODO
+					break;
+			};
+			return 0;
+			break;
+		default:
+			return 0;
+	}
+}
diff --git a/src/game_interpreter_control_variables.h b/src/game_interpreter_control_variables.h
index a534b83b7..1fcbc717f 100644
--- a/src/game_interpreter_control_variables.h
+++ b/src/game_interpreter_control_variables.h
@@ -45,6 +45,7 @@ namespace ControlVariables {
 	int Muldiv(int arg1, int arg2, int arg3);
 	int Divmul(int arg1, int arg2, int arg3);
 	int Between(int arg1, int arg2, int arg3);
+	int EasyRpgCommand(int op, int subop, int arg1, int arg2);
 }
 
 #endif
```

jetrotal:
Implemented changes you guys asked in a straight forward way, since controlVariables doesn't care about strings at all...


It works fine with Maniacs @raw command, but looks ugly AF:
image


that will be solved once we implement #3137, The code will look like:
`@easyrpg_raw @ControlVariables, "getDay", 0, 1`

PR #3131, opened 2023-10-25
Destiny Patch support

drxgb:
I'm going to start implementing Destiny Patch support on EasyRPG.
I hope can help.
Awaiting for a feedback. Thanks! =D


jetrotal:
Nice!

Some Documentation, to help contextualize this patch:
Original Destiny Patch, made by Bananen-Joe:
[DestinyScript_Eng.pdf](https://github.com/EasyRPG/Player/files/13160734/DestinyScript_Eng.pdf)
image

Destiny Patch Maniacs Update, made by Kotatsu Akira:
https://dev.makerpendium.de/docs/destiny_help/main-de.htm?page=object_command
image

Ghabry:
Hello. As adding the rest of destiny will be a huge task I have to ask:

Your code looks good but do you feel confident enough in C++ to implement the rest of it?

1. You will have to write a parser for the entire destiny syntax (imo this is the hard part)
2. Implement the functions offered by destiny (this is more tedious than hard imo, the tricky ones like sockets can be skipped for now)

Ghabry:
Hm actually the grammar doesn't look too hard to parse because the language lacks flow control like if and for. Very straightforward.

I can offer my help here if you encounter problems.

src/destiny.cpp
```diff
@@ -0,0 +1,160 @@
+/*
```
Ghabry, line 1: Imo this is a good start. Just a little nit-pick:

Can you rename it to ``Game_Destiny`` similiar to how Ineluki patch is implemented (which is ``Game_Ineluki``).

Then make it a class and put it the variable in ``main_data.h/cpp`` and instantiate it in ``ResetGameObjects``.

Also please update ``Makefile.am`` and ``CMakeLists.txt`` by adding the new files to it (put them at the correct location in alphabetical order).

-----------

Yes I know that DynRPG does it in a different way but I plan to rewrite this as less global state makes the cleanup easier when switching the game.

drxgb:
> Hello. As adding the rest of destiny will be a huge task I have to ask:
> 
> Your code looks good but do you feel confident enough in C++ to implement the rest of it?
> 
> 1. You will have to write a parser for the entire destiny syntax (imo this is the hard part)
> 2. Implement the functions offered by destiny (this is more tedious than hard imo, the tricky ones like sockets can be skipped for now)

Thanks for the feedback!
I'm basing on the Assembly source code made by Bananen-joe. There's a lot of work to do yet. But I'm going step by step while I'm studying the EasyRPG infrastructure.

Maybe the interpreter I could use OOP.
Besides we can create some parsing interfaces (like parser, scanner and handler) that can be implemented by another patches like DynRPG and Maniacs. Idk if this idea can be acceptable but I have this wish.

Ghabry:
If you think you can implement such a parser then go for it :).

When it is flexible enough it can also be used for other parts as you proposed

(Maniac has nothing to parse on the player side but there is an expression parser needed for the editor.)

src/game_destiny.cpp
```diff
@@ -0,0 +1,350 @@
...
+	{
+		code += '\n';
+		code += ToString((*it++).string);
+		++current;
+	}
+
+	length = code.length() + 1;
+	destinyScript = new char[length];
+	strcpy_s(destinyScript, length, code.c_str());
+
```
Ghabry, line 239: ``strcpy_s`` is a Microsoft extension.

In general you do not need ``new`` or ``malloc`` in modern C++ code. You could just change the type of ``_destinyScript`` to ``std::string`` and do ``destinyScript = code``.

You can still do the ``_scriptPtr`` stuff via ``_scriptPtr = &destintyScript.data()``.

src/player.cpp
```diff
@@ -826,8 +827,14 @@ void Player::CreateGameObjects() {
 		if (!FileFinder::Game().FindFile("accord.dll").empty()) {
 			game_config.patch_maniac.Set(true);
 		}
+
+		if (!FileFinder::Game().FindFile(DESTINY_DLL).empty()) {
+			game_config.patch_destiny.Set(true);
+		}
 	}
 
+	Output::Debug("Patch configuration: dynrpg={} maniac={} key-patch={} common-this={} pic-unlock={} 2k3-commands={} destiny={}",
```
Ghabry, line 836: This ``Output::Debug`` was removed and relocated to https://github.com/EasyRPG/Player/blob/2cb2956b055907ca5adbe842adec6f19261e502b/src/game_config_game.cpp#L234

Simply add a ``add_bool(patch_destiny);`` their

Ghabry:
Honestly when these two issues are resolved I would be fine with merging it for now. Just to avoid regular rebasing because of conflicts in the other files.

Of course its currently not useful as is a stub and the coding style is different but most of the upcoming patch commits will be pretty isolated (only in ``game_destiny``) and style issues can be fixed at the end as this will be _huuuge_ when completed.

(Also we had a DynRpg interpreter but 0 implemented plugins for years xD)

src/game_destiny.cpp
drxgb: Like this?
```cpp
private:
	// Member data
	std::string _destinyScript;
```
```cpp
const char* Interpreter::MakeString(SaveEventExecFrame& frame)
{
	std::string code;

	int32_t& current = frame.current_command;
	const std::vector<EventCommand>& cmdList = frame.commands;
	std::vector<EventCommand>::const_iterator it = cmdList.begin() + current++;

	code = ToString((*it++).string);

	while (it != cmdList.cend() && it->code == static_cast<int32_t>(EventCommand::Code::Comment_2))
	{
		code += '\n';
		code += ToString((*it++).string);
		++current;
	}

	_destinyScript = code;
	return _scriptPtr = _destinyScript.data();
}
```

Ghabry: Yes, this ``_scriptPtr`` assignment is legal here because ``_destinyScript`` is a member variable.

drxgb: Nice! I'll push again to see whether the code passes.

drxgb:
I forgot to note: later I want to do unit tests on `test_runner_player`. So I'm studying its structure to apply tests to `game_destiny`.

Ghabry:
Thats fine. Just create the file and add the file at the correct location in ``Makefile.am``. CMake grabs it automatically. There shouldn't be any conflicts later as we usually don't edit that part of the file :)

Ghabry:
Jenkins: test this please

src/game_destiny.cpp
```diff
@@ -0,0 +1,342 @@
...
+
+// =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+//				Game_Destiny implementations
+// =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+
+void Game_Destiny::Load()
+{
+	// Do not load Destiny whether player cannot find "Destiny.dll"
+	if (! FileFinder::Game().Exists(DESTINY_DLL))
```
Ghabry, line 41: Can you add ``#include "filefinder.h"`` to the headers?  This fails to build on emscripten. Guess ``exe_reader`` indirectly pulls that header in and that include is skipped for the emscripten build.

src/game_destiny.cpp
```diff
@@ -0,0 +1,342 @@
...
+		exe.seekg(1, std::ios_base::cur);
+		exe.read(reinterpret_cast<char*>(&language), sizeof(uint32_t));
+		exe.seekg(1, std::ios_base::cur);
+		exe.read(reinterpret_cast<char*>(&dllVersion), sizeof(uint32_t));
+		exe.seekg(1, std::ios_base::cur);
+	}
+#else
+	// TODO [XGB]: Find to manage Destiny initialization parameters on Emscripten
+	dllVersion = 0x20000;
+	language = ENGLISH;
```
Ghabry, line 78: This must be ``Destiny::ENGLISH``.

src/game_destiny.cpp
drxgb: You mean this:
```cpp
// game_destiny.cpp

// Headers
#include "game_destiny.h"
#include "filefinder.h"
```

or this?

```cpp
// game_destiny.h

#ifndef EP_GAME_DESTINY_H
#define EP_GAME_DESTINY_H

#include <cstdint>
#include <sstream>
#include <vector>

#include "filefinder.h"
#include "lcf/rpg/saveeventexecframe.h"
```

src/game_destiny.cpp
drxgb: Oh, that's true. My bad.
The compiler will never know where this keyword came from if I do not specify its namespace 😅

src/game_destiny.cpp
Ghabry: add it to the .cpp file (so first option)

drxgb: Done! =D

fdelapena (approved):

Ghabry (approved):

PR #3133, opened 2023-10-26
Game_Browser - Autorun a specific game

jetrotal:
If a game filename contains `gamedata.easyrpg` and there's only one game in player's folder, the player will autorun it.
image image image image
image

It would be better if I could skip the game_browser scene or hide the screen with default system. But this is working fine.

### DIFF
--- a/src/filefinder.cpp
+++ b/src/filefinder.cpp
@@ -320,6 +320,40 @@ bool FileFinder::IsRPG2kProjectWithRenames(const FilesystemView& fs) {
 	return !FileExtGuesser::GetRPG2kProjectWithRenames(fs).Empty();
 }
 
+bool FileFinder::OpenViewToEasyRpgFile(FilesystemView& fs) {
+	auto files = fs.ListDirectory();
+	if (!files) {
+		return false;
+	}
+
+	int items = 0;
+	std::string filename;
+
+	for (auto& file : *files) {
+		if (StringView(file.second.name).ends_with(".easyrpg")) {
+			++items;
+			if (items == 2) {
+				// Contains more than one game
+				return false;
+			}
+			filename = file.second.name;
+		}
+	}
+
+	if (filename.empty()) {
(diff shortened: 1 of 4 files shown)


fdelapena:
A problem with Windows shortcuts is, they are harder than expected to create them containing relative paths. IF absolute, a usual breaking case happens when the user moves the game folder somewhere else. "Portable" apps are quite common as well, and drive letters change often.

Exe embedding is nice but they trigger low quality antivirus heuristics due to the hacky PE stuff that makes it work.

Despite this is mostly  a Windows specific solution, this patch looks simple enough and harmless behaviour on other platforms compared to implement Windows specific code.

Ghabry:
okay, fine. Lets add this. But I will do some small UX improvements before:  :)

- The autoboot should also work in the Logo scene
- Maybe relax the name requirement a bit? E.g. "Must end on .easyrpg" and the part in front of it doesn't matter.

Ghabry:
Relaxed it a bit and made it work from the logo scene: The player will now autorun a game if it is the only game in the directory and ends on ``.easyrpg``. Does not matter if folder or (zip) archive.

fdelapena (approved):

jetrotal:
image
Testing this fake 3D game with the memory leak PR, looks like it went from unplayable to very laggy...
Link to it:
https://biloumaster.fr/download/[BConcept]-Vaisseau3D.zip

It has a version optimized for easyRPG too,
but it may be interesting to check the rm2k3 one.

Ghabry:
With very very agressive caching (never delete it) I can even reach 166 FPS (refresh rate of my display) but it needs 10 GB of RAM xD. Though the blend effect will still frame drop because it causes a cache miss.

Guess the issue is similiar to Aedemphia: Lots of tone blit effects that are not efficient to do on 32 bit graphics without GPU support compared to the simple palette editing 8 bit can do.

PR #3137, opened 2023-10-30
DynRPG whitelist + easyrpg_raw command

jetrotal:
Whitelist selected dynRPG commands, so people can use them without enabling the dynRPG plugin.
image

------------------------------

Invoke a command using raw input, similar to how maniacs patch raw work.
easyRPG raw syntax:
```js
@easyrpg_raw 10110, "test", [1,2,3,4,V1],1 
//Message box
@easyrpg_raw $ShowMessage, T1, [1,2,3,4,V1],1
//Also Message box
```


Ghabry:
We need to parse deeper

grafik

Ghabry:
The crossover nobody asked for

grafik

Ghabry:
btw that "frame pushing" looks like a safe solution for this ``Call Command`` thing of Maniacs. Is this event command hard to implement?

jetrotal:
> btw that "frame pushing" looks like a safe solution for this `Call Command` thing of Maniacs. Is this event command hard to implement?

https://jetrotal.github.io/CSA/#MNC%20|%20Call%20Command
image


It's similar but uses an array of variables in sequence for the arguments

Ghabry:
That PR is kinda the wrong location for this, so one final question before we move this maybe somewhere else:

I think that CSA example needs more permutations and a 3rd argument to figure out how this works. E.g. 

``@cmd 1, "", .args v[2], 3`` gives parameters: ``0, 1, 2, 3``. How is ``v[2]`` encoded? o_O

jetrotal:
@Ghabry I also set up a collection of useful constants, that came from destinyScript. Those could be useful for parsing common values:
image


made in two formats, the second one is a Map, I guess that's the one that would work with current code:
https://www.diffchecker.com/YAS7V0Mi/

copy the text at the bottom.

jetrotal:
image
image

jetrotal:
image

jetrotal:
Hey guys, Removed a lot of repeated code, including using some functions we already have in Utils.

I think this is good enough for a working raw command interpreter.
I wanted to break it into multiple lines of code, but Ghabry suggested writing in the future an eval command for that
(maybe `easyrpg_eval "prefix", "string"` ?)

Other things that may be important for this in the future is setting up the "constants.h" content as
auto-generated entries from liblcf.

--------------------------------------

As soon as this this pr is aproved, I can rework #3124 as you guys asked.

jetrotal:
Ok, multi-line is done. 
I had to add something similar to the "," on the part of the code that parses text to arg[n]:
It parses ';' as ',' too with an extra input in case of being "easyrpg_raw". I hope that's not too invasive...

Ghabry:
I talked with Jetrotal: I will help getting this in when I'm done with my remaining issues for 0.8.1 milestone. Likely in one or two weeks.

My current plan is to split it into two PRs. One for the DynRPG refactor, one for easyrpg_raw. Makes reviewing this a bit easier :)

Ghabry:
Would you mind when I swap the argument order to:

```
$ControlVars "A", 1, 0, 3, 4, 6, 8"
```

Makes "indent" the 2nd parameter and the numeric list everything from 3rd to the end. Makes the parsing simpler and is not ambigious.

jetrotal:
> Would you mind when I swap the argument order to:
> 
> ```
> $ControlVars "A", 1, 0, 3, 4, 6, 8
> ```
> 
> Makes "indent" the 2nd parameter and the numeric list everything from 3rd to the end. Makes the parsing simpler and is not ambigious.

hm... I made it this way to let people ignore the indent parameter, and read each type of parameter by group

jetrotal:
@Ghabry asked for me to post the event2string command I was working on, here it is:
https://gist.github.com/jetrotal/e8b35118b94ddbe8ee10885e220f706c

PR #3138, opened 2023-10-30
Multiplayer

monokotech:


jetrotal:
Thanks for the huge update!
We are gonna take some time to review and test it all, probably we will finish after the 8.1 release 🤔

By a quick look, I can see some small stuff, like F9 button that is already used for rpg maker's --test-play menu 
image


that said, thanks again for this cool update!

Ghabry:
can you explain what Virtual3D is?

monokotech:
> can you explain what Virtual3D is?

This is an attempt to synchronize 3D visual maze-like map in Yume2kki, and you can enable this feature by adding "ServerVirtual3DMaps=1881,33,3,7" in the ini file. However, since Yume2kki has various implementations of this type of 3D map, I'm currently only basing it on the 3D Red Maze (MAP1881).

The name Virtual3D comes from the VirtualImage event in the MAP1881 map.

ServerVirtual3DMaps=1,2,3,4
1. map_id => Make these codes only valid on this map.
1. character_event_id => This is a character that will be moved by events.
1. character_terrain_id => What kind of terrain id is used to represent the character?
1. refresh_switch_id => Which switch id decides to refresh the screen?

I hope to provide a standard for convenient synchronization, and I think qxy's implementation is a good example.

monokotech:
> Thanks for the huge update! We are gonna take some time to review and test it all, probably we will finish after the 8.1 release 🤔
> 
> By a quick look, I can see some small stuff, like F9 button that is already used for rpg maker's --test-play menu image
> 
> that said, thanks again for this cool update!

Thank you for your response!

About the small stuff:

I use shortcut keys similar to minetest. This conflicts with the existing debug function, but I found that this can distinguish --test-play and multiplayer mode, that is, --test-play will not be enabled in multiplayer.

I also accept EasyRPG shortcut keys.

Ghabry:
Jenkins: test this please

Thanks for this huge contribution. This is too much to quickly review so I will take a look at this when I have a calm week. Based on my work schedule this will be the first week of January and the first week of February.

Maybe I find time earlier but these are the dates where I know that I have some freetime.

PR #3140, opened 2023-10-30
Maniac: Implement CallCommand and String Vars

Ghabry:


fdelapena (approved):

Ghabry:
Pending final change that depends on #3141 

```cpp
if (com.parameters.size() >= 5) {
    cmd.string = Main_Data::game_strings->GetWithModeBitfield(com.string, com.parameters[0], 3, com.parameters[4]);
} else {
    cmd.string = com.string;
}
```

btw, such code will be also needed for other commands just with a different index:

```cpp
if (Player::IsPatchManiac() && com.parameters.size() >= A) {
    cmd.string = Main_Data::game_strings->GetWithModeBitfield(com.string, com.parameters[0], B, com.parameters[A - 1]);
} else {
    cmd.string = com.string;
}
```

fdelapena (approved):

PR #3141, opened 2023-11-02
Refactor String Vars

Ghabry:
Noteworthy changes:

- unordered_map instead of vector to reduce huge waste of memory for sparse string usage (e.g. starting at 1000)
- StringView used where possible to reduce copying
- Bugfix: Setting variables didn't refresh the interpreter


fdelapena (approved):

jetrotal:
Hi, we spoke about supporting more file types for the StrVars load/save files,
maybe those change could also fit this pr:

image
image

PR #3142, opened 2023-11-02
Fix issue 3135 (Fluidsynth: Turn off all sound when the music changes)

### LINKED ISSUE #3135: MIDI with soundfont - interrupting a MIDI with BGM (OFF) and playing another MIDI will play a note of the previous MIDI before playing the new one
#### Player platform:

While using a custom soundfont (the default setting on Windows does not work for this issue), interrupt a MIDI track with the command `Play BGM: (OFF)`, and then play another MIDI track (can be the same one). When the new MIDI track will play, some notes of the previous MIDI track will be heard while they should not.

You can test this with testgame: interact with the NPC on the left or right to play a MIDI, then the one at the center to stop it, and then the one on the left or right again to encounter the issue (make sure to have a custom soundfont to encounter the issue).
[MIDIissue.zip](https://github.com/EasyRPG/Player/files/13192033/MIDIissue.zip)

Reported by ajie on X, with the following footage as an example:
https://twitter.com/aj_108fn/status/1716463030903652845

Ghabry:
Solves an issue that old notes finish fading out when using BGM Stop followed by a BGM Play later.

Fix #3135

### DIFF
--- a/src/audio_decoder_midi.cpp
+++ b/src/audio_decoder_midi.cpp
@@ -398,6 +398,8 @@ void AudioDecoderMidi::reset() {
 		midi_msg = midimsg_make(midi_event_control_change, channel, midi_set_reg_param_lower, 0);
 		mididec->SendMidiMessage(midi_msg);
 	}
+
+	mididec->Reset();
 }
 
 void AudioDecoderMidi::reset_tempos_after_loop() {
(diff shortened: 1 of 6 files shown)


Ghabry:
confirmed working via a message on twitter. I just have to fix the FluidLite compat issues.

Ghabry:
closing in favor of #3148

Issue #1038, opened 2016-09-18
Support webm (VP8+Opus)


Ghabry:
Just for completeness, the only true Video format :+1:

fdelapena:
And the only true audio format :P

carstene1ns:
What about VP9 video?
![img](https://p.f4ke.de/2qi4rGqFHa.png)

fdelapena:
> What about VP9 video?

What about AV1? IETF NetVC candidate from AOMedia (featuring Xiph's Daala + Cisco's Thor + Google's VP10).

> (...) the only true Video format :+1:

_[citation needed]_ :laughing:

![the real only true video format](https://cloud.githubusercontent.com/assets/174885/19413962/3827a4b2-92fb-11e6-96ce-960efe2feb6b.png)
[Source](https://people.xiph.org/~jm/daala/revisiting/)

It looks like it will be mostly ready when we will implement this feature :yum:. What about using a frontend library which supports multiple codecs such gstreamer or ffmpeg?

Ghabry:
What about working, uncompressed avi before we think about anything more complex ;)

lmgennetay:
What about Aëdemphia video? The only thing missing is video support to make the game 100% playable with a more efficient rendering than RPG_RT.

Tontonnonor:
Hello,
I'm the Aëdemphia developper. I admire the work you've done with EasyRPG which is very impressive. I've been told that posting here would increase the chances of getting one day the video feature so... Yes, I'd be very interested if this feature would was done, and would be very happy to use EasyRPG for my game. :)
Best regards and thanks for the great job already done.

Falcossj:
Wanted to follow that previous message from Tontonnonor, as rpg maker gamer, easy RPG is almost perfect, I was able to play perfectly at Velsarbor on my phone, and what a great pleasure. 
Such a shame some games like Aedemphia and others using videos can't work... Hope you'll be able to add the video feature one day! That's probably one of the only missing thing to consider Easy RPG as a perfect Alternarive. 
Good job for the good work you already did, and thanks !

Ghabry:
Okay okay, I will take a look.

The most likely video format that will be supported is "webm" which contains VP8/VP9 video and a OGG audio feed.

Will need to get libvpx working first (and a parser for webm which is just a simpler mkv container. The library libwebm is not packaged alot and has a CVE since 3 years).

FalcoRPGMaker:
[...]

PR #3143, opened 2023-11-03
 Video Player - Proof of Concept

jetrotal:
Windows Only.

Ok, to make this feature work:
Create a Folder called "Plugin" (not Plugins as older tests) in your Project, and paste ffplay.exe from ffmpeg inside it.
[ffplay.zip](https://github.com/EasyRPG/Player/files/13246249/ffplay.zip).

It's basically a new .exe that runs over easyRPG one.

It doesn't have all the functionality from the og command, since I don't know how program them inside FFplay.
So, you won't be able to move, resize, etc...
Neither you'll be able to speed it up with FastForward keys.
Any key or mouse click will close it instantly (would be better a hold to close).

---------------

To cover other platforms, Ghabry suggested working with .dll files instead.

Here's a DLL version of ffplay I got from @YepSfx:
https://github.com/YepSfx/FFPlay-1.2.12

And here's a chinese tutorial I translated from @piaodazhu repo: 
https://github.com/jetrotal/simpleFFplay

---------

We don't know anything about video player, but that workaround may help us.

### DIFF
--- a/src/filefinder.cpp
+++ b/src/filefinder.cpp
@@ -54,7 +54,7 @@
 
 namespace {
 //#ifdef SUPPORT_MOVIES
-	constexpr const auto MOVIE_TYPES = Utils::MakeSvArray( ".mov", ".mp4", ".webm", ".ogv"); //".avi", ".mpg" not supported by web standards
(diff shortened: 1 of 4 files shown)


fdelapena:
Related (also recent comments): #1036, #1038.

Issue #2490, opened 2021-03-17
Feature request: Text skip


zelkd:
A text skipping option could be  quite handy for text heavy games when dying or missing saves, it's one of the main feature of fatalmix.

Example: https://files.catbox.moe/oewhmw.mp4
(game is https://www.freem.ne.jp/win/game/22650 btw)

PR #3147, opened 2023-11-13
Maniacs Feature - "Skip Dialogs" from Fatal Mix.

jetrotal:
Hold ESC to skip all dialogs when in Debug Mode.

Things to consider:
 
- Need an alternative to `strlen(text_index) < 3`. Since the textboxes can stop while being bigger than 3 characters, 
if the com.string has more than 4 lines of text.
- We need a flag that works as an alternative to `Player::debug_flag`, so people can use it in their games 
(both shift and this skip should be activated?)

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -4667,13 +4667,64 @@ bool Game_Interpreter::CommandManiacChangePictureId(lcf::rpg::EventCommand const
 	return true;
 }
 
-bool Game_Interpreter::CommandManiacSetGameOption(lcf::rpg::EventCommand const&) {
+bool Game_Interpreter::CommandManiacSetGameOption(lcf::rpg::EventCommand const& com) {
+	// #ManiacSetGameOption, "",[optIsVar, optType, opt, bitshift(moreOpt)]
 	if (!Player::IsPatchManiac()) {
 		return true;
 	}
 
-	Output::Warning("Maniac Patch: Command SetGameOption not supported");
-	return true;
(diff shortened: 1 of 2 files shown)


Ghabry:
FYI: There is already "message appears without delay" when holding shift. So you can already hold shift and hammer decision for the same result.

jetrotal:
yeah, what I did was similar to #2490

jetrotal:
After a second look, they want something similar do Maniac's Fatal Mix as a flag:
image


Weird part, is that it's attached to Right Shift. That could be a problem

Ghabry:
Do you want this auto-enter behaviour available in general or just when this fatal mix skip option is enabled?

jetrotal:
Oh boy, lemme try to remember how I set this up...

If I remember correctly, my code was very similar to the fatal mix behavior.
But I found it useful on vanilla games as well. 
It was a faster skip withtout hammering enter.

Right now, maybe we should stick it to the fatal mix?

Maybe in the future, would be interesting to have a way of accessing maniacs commands even with it disabled...?

Ghabry:
Considering it is a debug feature it could be an option in the settings whether you want "Instant text" with Shift or "Instant auto enter text" with Shift. :thinking:

jetrotal:
> Considering it is a debug feature it could be an option in the settings whether you want "Instant text" with Shift or "Instant auto enter text" with Shift. 🤔

Sure, it could also come with a setting

Ghabry:
There is another PR #3487 which also implements this and is more feature complete

PR #3148, opened 2023-11-16
MacOS Fixes

### LINKED ISSUE #3135: MIDI with soundfont - interrupting a MIDI with BGM (OFF) and playing another MIDI will play a note of the previous MIDI before playing the new one

tyrone-sudeium:
Fixes three issues:

- `5-beatles-lustig.mid` from Unterwegs in Duesterburg causes a crash in `MusicDeviceMIDIEvent` due to a bogus MIDI message. We're in some good company here, both VLC (at least on macOS) and Sekaiju also crash attempting to play it.
- I also noticed Unterwegs in Duesterburg in macOS has a _lot_ of stuck notes issues in macOS (i.e when changing BGM, some notes from the previous track will be permanently stuck on). I tracked it down to #3090 removing the `midimsg_all_sound_off` from reset, to be more accurate with Harmony. Harmony never supported macOS, and the macOS synthesizer being prone to stuck notes is why those messages were there in the first place. I'm open to other solutions to fix this, but without it, MIDI is pretty scuffed on Macs.
- The `-G Xcode` generator in cmake will completely fail to build (at least on Xcode 15), due to a fundamental disconnect between cmake and Xcode when you have multiple code files with the same filename (like `src/audio.cpp` vs `src/platform/sdl/audio.cpp`). Simplest workaround is to remove the duplication by renaming the two files that had duplicated names.

Fix #3135


Ghabry (changes requested):
1. You inserted space instead of tab for the indent.
2. Can you add this ``NeedsSoftReset`` to Decoder Fluidsynth? Has a similiar issue and your fix is more universal.

fdelapena (approved):

src/audio_decoder_midi.cpp
```diff
@@ -378,6 +402,10 @@ void AudioDecoderMidi::meta_event(int event, const void* data, std::size_t size)
 
 void AudioDecoderMidi::reset() {
 	// MIDI reset event
+	if (mididec->NeedsSoftReset()) {
+		// SoundOff every channel: necessary for synths like macOS which tend to get stuck
+		SendMessageToAllChannels(midimsg_all_sound_off(0));
```
Ghabry, line 407: sorry have to ask you for another small change:

Can you move this new code block between the GM reset and the pitch bend block? (I hope this doesn't break macOS?)

Tested this again in fluidsynth and the all sound off will not work properly when executed before the GM reset (same audio glitch remains).

```cpp
        // GM system on (resets most parameters)
        const unsigned char gm_reset[] = { 0xF0, 0x7E, 0x7F, 0x09, 0x01, 0xF7 };
        mididec->SendSysExMessage(gm_reset, sizeof(gm_reset));

        // MIDI reset event
        if (mididec->NeedsSoftReset()) {
                // SoundOff every channel: necessary for synths like macOS which tend to g>
                SendMessageToAllChannels(midimsg_all_sound_off(0));
        }

        // Set the Pitch bend range to 256
        for (int channel = 0; channel < 16; channel++) {
```

Ghabry (approved):
thanks, works now with fluidsynth :+1:

PR #3152, opened 2023-11-19
Load Logos from RPG_RT

Ghabry:
Logos are now loaded from RPG_RT on startup.

The original logos are skipped by default. You can change this in the Settings under Engine.

Risky change!!! Please test many games here and report back! An error in the parser can crash the Player :/.


Mimigris:
> Risky change!!! Please test many games here and report back! An error in the parser can crash the Player :/.

So far I have encountered no crashes, though the parsing of the logo seems to not work with the latest Japanese 2k RPG_RT (with the setting all logos displayed) and probably some others, would need to check that once I have time.

jetrotal:
Random quirk I noticed on windows:
The EasyRPG logo is the only one that doesn't have a fade in animation.

Maybe, the first logo scene is called without having the EasyRPG logo properly loaded yet?

Ghabry:
> So far I have encountered no crashes, though the parsing of the logo seems to not work with the latest Japanese 2k RPG_RT (with the setting all logos displayed) and probably some others, would need to check that once I have time.

Oh. Yeah it became late and I only tested with Vampires Dawn.

I thought the newer engines use "LOGO1" as the resource name but it is "LOGO". Easy fix, coming soon :)

Ghabry:
Loading the Logo from newer RPG_RT is now in.

Mimigris:
> Loading the Logo from newer RPG_RT is now in.

It seems that those changes broke the loading of logos from RPG Maker 2003 games for some reason (Japanese, English, Maniac, patch...), now everything that I have work outside of the RPG Maker 2003 games that I have.

Ghabry:
@Mimigris another case of sloppy testing on my side. 2k VALUE! uses ``LOGO``. 2k3 uses ``LOGO1``. Old 2k uses ``LOGO1`` to ``LOGO3``. But now its working.

Mimigris:
> @Mimigris another case of sloppy testing on my side. 2k VALUE! uses `LOGO`. 2k3 uses `LOGO1`. Old 2k uses `LOGO1` to `LOGO3`. But now its working.

Everything seems to work properly on my side now, the only thing not working properly are Maniac games where the logo is not loaded (due to the fact that the engine is coded differently), so except if you think that it is something really important, it seems good to me. 👍

Ghabry:
Don't plan to figure it out for maniac. So if the rest works I consider this done

Mimigris:
Actually I found a new issue, though it is not related to the RPG_RT detection this time:
By opening and closing the settings menu of the Player a lot of times during the display of the logo (I'm not sure if it's even intended to allow it there), it is possible that, when selecting the exit option of the game once on the title screen, instead of exiting, the game will either
a) display again the logo scene of the game (though it will wait for a input instead of fading after around 3 seconds) and then relaunch your game
b) open the settings menu (without the exit game option) while still using the title picture, system and system sounds from the game, and closing it will return you to the game browser

It can be a bit hard to reproduce, though you just need to spam the keys during the display of the logo and you may be able to reproduce it.

Ghabry:
That you can exit from the logos using the settings is intentional to leave troll games with 100 logos :sweat_smile: . But yeah this sounds like some kind of race condition, likely logo pushed while at the same time settings pushed.


> Another thing that I don't remember if I have mentioned it or not is that holding the Shift key on a logo will make it wait until the key is no longer held until going to the next scene as a leftover of the scene with the EasyRPG logo, I'm not sure if it makes sense to keep it.

This is intentional. Some logos of games contain introduction text and this way it is easier to read :)

Issue #3054, opened 2023-07-11
Allow EasyRPG Player on Wii to run from USB without crashing


brunochanrio:
Hello, is possible to make EasyRPG Player on Wii to run from USB without crashing? is because i don't use a SD on my Wii, i run all the homebrew from USB

carstene1ns:
Please attach a photo of the stack trace.
And we need the exact version used, if not 0.8.

carstene1ns:
Debugging the crashes @Mimigris reported:

```
80530720
0x80530720: _free_r at ??:?
80530578
0x80530578: _free_r at ??:?
8021c600
0x8021c600: std::unique_ptr<midisynth::synthesizer, std::default_delete<midisynth::synthesizer> >::~unique_ptr() at /opt/devkitpro/devkitPPC/powerpc-eabi/include/c++/13.1.0/bits/unique_ptr.h:403
 (inlined by) FmMidiDecoder::~FmMidiDecoder() at /var/lib/jenkins/workspace/player-wii/src/decoder_fmmidi.h:31
 (inlined by) FmMidiDecoder::~FmMidiDecoder() at /var/lib/jenkins/workspace/player-wii/src/decoder_fmmidi.h:31
80013528
0x80013528: std::unique_ptr<midisequencer::sequencer, std::default_delete<midisequencer::sequencer> >::~unique_ptr() at /opt/devkitpro/devkitPPC/powerpc-eabi/include/c++/13.1.0/bits/unique_ptr.h:403
 (inlined by) AudioDecoderMidi::~AudioDecoderMidi() at /var/lib/jenkins/workspace/player-wii/src/audio_decoder_midi.cpp:95
8001365c
0x8001365c: AudioDecoderMidi::~AudioDecoderMidi() at /var/lib/jenkins/workspace/player-wii/src/audio_decoder_midi.cpp:95
80019308
0x80019308: std::default_delete<AudioDecoderBase>::operator()(AudioDecoderBase*) const at /opt/devkitpro/devkitPPC/powerpc-eabi/include/c++/13.1.0/bits/unique_ptr.h:99 (discriminator 1)
 (inlined by) std::unique_ptr<AudioDecoderBase, std::default_delete<AudioDecoderBase> >::~unique_ptr() at /opt/devkitpro/devkitPPC/powerpc-eabi/include/c++/13.1.0/bits/unique_ptr.h:404 (discriminator 1)
[...]

Issue #3126, opened 2023-10-20
Wii - A and B buttons are reversed in the default mapping, and the name of the keys is not consistent


Mimigris:
#### Player platform:
Wii, continuous build of the Player.

#### Describe the issue in detail and how to reproduce it:
On the Wii version, the default mapping for the A button is Cancel and the B button is Decision: compared to usual games released on Wii, where A is accept and B is cancel, this is counter-intuitive and as such should be reversed. This is likely a regression from the version 0.8.0, where the settings menu was added.
![screenshot_1](https://github.com/EasyRPG/Player/assets/108797333/2f6f222f-7327-4f10-bea8-ced0cd39c619)

In addition to this, in the settings menu, the name of the keys is not consistent everywhere on the Wii version: the content in parentheses is not always written the same way, with missing or additional spaces present, or missing characters (e.g. `1 (Wiimote`, `- (Wiimote /CC )`), so this may need to be looked at (Note that some keys may have been remapped on these two screens and should be ignored; the incorrect colours on these screenshots is due to another issue, please ignore that).
![screenshot_2](https://github.com/EasyRPG/Player/assets/108797333/eee9cbd8-7e20-4d0a-89e5-5dcc8dfbad10)
![screenshot_3](https://github.com/EasyRPG/Player/assets/108797333/5706cf91-068b-4e9c-b2dd-94b37c171283)

Mimigris:
Closing it as it is now superseded by #3594 and that some of the name issues have since been fixed.

Issue #1348, opened 2018-04-08
Add Wii VC for Wii U support? label:Wii label:Enhancement


CatmanFan:
Wii U GamePad support for vWii has already been out, and quite a few apps have the support for it from the very beginning. I am just wondering if there is a way that EasyRPG can implement this type of GamePad support in the Wii version.

I have already tried injecting an unknown version of EasyRPG Wii as a Wii VC channel and I installed it onto my Wii U, but it never detects any games from the SD (probably because it only reads from the game package and nothing else) and any Wii U GamePad input doesn't work. I can still power it off using the GamePad, though.

Ghabry:
EasyRPG Player for the Wii doesn't use libogc directly and uses SDLWii instead which isn't actively developed anymore. Which library supports WiiU Gamepad in Wii mode?

There were also plans for a real WiiU port of Player but by now all attempts ended in crashes.

fdelapena:
> There were also plans for a real WiiU port of Player but by now all attempts ended in crashes.

#3167

PR #3167, opened 2023-12-03
Initial WiiU support

carstene1ns:
Simple and straightforward.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1076,16 +1076,7 @@ if(${PLAYER_AUDIO_BACKEND} MATCHES "^(SDL2|SDL1|libretro|psvita|3ds|switch|wii|a
 endif()
 
 # Executable
-if(NINTENDO_WIIU)
-	add_executable(easyrpg-player src/platform/sdl/main.cpp)
-	target_link_libraries(easyrpg-player ${PROJECT_NAME})
-	wut_create_rpx(easyrpg-player)
-	wut_create_wuhb(easyrpg-player
-		NAME "EasyRPG Player ${PLAYER_VERSION_FULL}"
-		SHORT_NAME "Player"
-		AUTHOR "EasyRPG Team"
-	) # todo icon, splashs
-elseif(${PLAYER_BUILD_EXECUTABLE} AND ${PLAYER_TARGET_PLATFORM} MATCHES "^SDL(1|2)$")
+if(${PLAYER_BUILD_EXECUTABLE} AND ${PLAYER_TARGET_PLATFORM} MATCHES "^SDL(1|2)$" AND NOT NINTENDO_WIIU)
 	if(APPLE)
 		set(EXE_NAME "EasyRPG-Player.app")
 		set_source_files_properties(${${PROJECT_NAME}_BUNDLE_ICON} PROPERTIES MACOSX_PACKAGE_LOCATION "Resources")
@@ -1175,7 +1166,8 @@ elseif(${PLAYER_BUILD_EXECUTABLE} AND ${PLAYER_TARGET_PLATFORM} MATCHES "^SDL(1|
 		endif()
 		install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${PLAYER_JS_OUTPUT_NAME}.wasm DESTINATION ${CMAKE_INSTALL_BINDIR})
 	endif()
-elseif(${PLAYER_TARGET_PLATFORM} MATCHES "^(psvita|3ds|switch|wii)$")
+elseif(${PLAYER_TARGET_PLATFORM} MATCHES "^(psvita|3ds|switch|wii)$" OR NINTENDO_WIIU)
+	set(CPACK_PLATFORM "${PLAYER_TARGET_PLATFORM}")
 	if(NINTENDO_3DS)
 		add_executable(easyrpg-player src/platform/3ds/main.cpp)
 		target_link_libraries(easyrpg-player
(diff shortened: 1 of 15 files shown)


CMakeLists.txt
```diff
@@ -686,11 +686,12 @@ elseif(${PLAYER_TARGET_PLATFORM} STREQUAL "amigaos4")
 		src/platform/sdl/axis.h
 		src/platform/sdl/sdl_ui.cpp
 		src/platform/sdl/sdl_ui.h)
+else()
 	message(FATAL_ERROR "Invalid target platform")
 endif()
 
 # Shared by homebrew platforms
-if(${PLAYER_TARGET_PLATFORM} MATCHES "^(3ds|psvita|switch|wii)$")
+if(${PLAYER_TARGET_PLATFORM} MATCHES "^(3ds|psvita|switch|wii)$" OR NINTENDO_WIIU)
```
Ghabry, line 694: I was wondering first why you didn't add "wiiu" here but I get it now:

The target platform is SDL2 with a few WIIU adjustments.

------

This looks still a bit WIP but: Can I merge this or are more commits planned?

carstene1ns:
#3126 - first part fixed
#3054 - should be fixed, tested here
#1348 - bit related, but unfixed.

carstene1ns:
This is ready now.

carstene1ns:
Something is fishy in big endian land: The debug build asserts in `sdl_format_to_format()`.

We get [LSB](https://github.com/devkitPro/SDL/blob/7b058bd042a7c624c277aaddbca6fb2f6ae5233d/include/SDL_audio.h#L94C9-L94C21), but the WiiU SDL2 sets explicitly [MSB](https://github.com/devkitPro/SDL/blob/7b058bd042a7c624c277aaddbca6fb2f6ae5233d/src/audio/wiiu/SDL_wiiuaudio.c#L110).

Here is a patch that makes Player start, but is a bit ugly:
```patch
diff --git a/src/platform/sdl/sdl_audio.cpp b/src/platform/sdl/sdl_audio.cpp
index adf56dfb4..5e184c7fa 100644
--- a/src/platform/sdl/sdl_audio.cpp
+++ b/src/platform/sdl/sdl_audio.cpp
@@ -53,9 +53,11 @@ AudioDecoder::Format sdl_format_to_format(Uint16 format) {
 			return AudioDecoder::Format::U8;
 		case AUDIO_S8:
 			return AudioDecoder::Format::S8;
-		case AUDIO_U16SYS:
+		case AUDIO_U16LSB:
+		case AUDIO_U16MSB:
 			return AudioDecoder::Format::U16;
-		case AUDIO_S16SYS:
+		case AUDIO_S16LSB:
+		case AUDIO_S16MSB:
 			return AudioDecoder::Format::S16;
 #if SDL_MAJOR_VERSION > 1
 		case AUDIO_S32:
@@ -64,6 +66,7 @@ AudioDecoder::Format sdl_format_to_format(Uint16 format) {
 			return AudioDecoder::Format::F32;
 #endif
 		default:
+			Output::Warning("Couldn't convert SDL Audio Format! Got: {x}", format);
 			assert(false);
 	}
 ```

I had this problem with the buildscripts before and added big endian defines manually. Not sure if this is a packaging error or some HLE in cemu maybe.

Ghabry:
okay lets try to figure this one out (worth checking ``SDL_BYTEORDER``).

Cannot do an hardware test before xmas. Is something for after merge. Will take the WiiU from my parents attic when I visit them and take it to my flat. Will make testing easier xD.

Ghabry:
@carstene1ns after lots of tracing about the preprocessor values: They are all correct.

The issue is much simpler. 

We do ``want.format = AUDIO_S16;``

And the code in SDL2 says: ``#define AUDIO_S16       AUDIO_S16LSB``

So changing this to ``AUDIO_S16SYS`` should resolve it...

carstene1ns:
This seems to be a difference in SDL1.2 vs 2, otherwise we would have noticed this on Wii and Amiga.

Ghabry (approved):
I'm merging this for now to make the PR builder green. Will amend this when I have the WiiU around, will set mine up in January :)

carstene1ns:
Yay

AlbertoDiazC:
Seems to error about the folder sdmc:/data/easyrpg-player/ , has to be created beforehand manually.

carstene1ns:
@AlbertoDiazC Thank you for the feedback, did you try if it does run games? I could not test it, because I do not own a Wii U.

AlbertoDiazC:
Yes it does run games quite well.
Settings can't be saved when that folder doesn't exist, which is a minor issue. A fix should be easy imho, the player could create it at first if it's missing.

Ghabry:
Hmm the setting folder is automatically created when missing. Must be a WiiU exclusive bug 🤔. Thanks for testing!

AlbertoDiazC:
Found another small issue, this time is about the rtp path.
This is the relevant fragment of the logfile:

[2024-01-21 18:35:29] Debug: EXEReader: VERSIONINFO resource found (DE 0xea180; 0x10e770; len 0x2ec)
[2024-01-21 18:35:29] Debug: RPG_RT information: version=1.1.1.0 logos=1 code=0xc9000 cherry=0xb800 geep=0x0 i386=true easyrpg=false
[2024-01-21 18:35:29] Debug: Engine configured as: 2k=false 2k3=true MajorUpdated=true Eng=true
[2024-01-21 18:35:29] Debug: RTP path ./rtp/2003 is invalid, not adding
[2024-01-21 18:35:29] Debug: RTP path /data/easyrpg-player/rtp/2003 is invalid, not adding
[2024-01-21 18:35:29] Debug: Patch configuration: dynrpg=false maniac=false key-patch=false common-this=false pic-unlock=false 2k3-commands=false
[2024-01-21 18:35:29] Debug: Cannot find: Logo/LOGO1
[2024-01-21 18:35:29] Debug: EXEReader: LOGO1 resource found (DE 0xe9d60; 0xea398; len 0x3d38)
[2024-01-21 18:35:29] Debug: Resolution changed to 320x240
[2024-01-21 18:35:29] Debug: Loaded Map Map0001.lmu

I keep good copies of the rtp at
./rtp/2003    (not usable in aroma cfw because a proccess cant know its own path)
/data/rtp/2003    (and 2000)
/data/easyrpg-player/rtp/2003   (and 2000)

I figure this bug could be fixable with one of these workarounds:
Prepend smdc:/   as in sdmc:/data/rtp/2003
Use the full path /vol/external01/data/rtp/2003

Issue #2961, opened 2023-04-17
Show String Picture - System/image.png displays incorrect transparent color 


jetrotal:
Please fill in the following fields before submitting an issue:
#### Name of the game:
Maniacs - custom life bar using Show StringPic Tutorial: 
https://www.rmteka.pl/maniacs-patch-nie-gryzie-10-prosty-hud-wersja-alternatywna/

#### Player platform:
Windows 64x

#### Attach files (as a .zip archive or link them)
https://www.rmteka.pl/wp-content/uploads/2023/04/prosty-HUD-przez-show-string-picture.zip

#### Describe the issue in detail and how to reproduce it:

That's a weird one. In this tutorial, the user created a HP bar using show string Picture and a custom System Graphic background(the images from the ./system/something.png). 

In the image bellow there are 2 examples, one displays correctly the opacity of the image (RPG_RT),
and another shows the player interpreting the color-index-0 as black instead of transparent:

![image](https://user-images.githubusercontent.com/45118493/232593782-3ef71496-541e-45b5-8cad-75ef7124291a.png)
![image](https://user-images.githubusercontent.com/45118493/232593835-92f74fef-d46c-4716-b696-24400606bbb5.png)

-------------------------------------

their hack is basically avoiding the use a system graphic like this:
![image](https://user-images.githubusercontent.com/45118493/232595306-433f5fab-344c-449c-9c12-5984b6638a55.png)

and instead, using it like this:
![image](https://user-images.githubusercontent.com/45118493/232595391-8ef5e214-779f-42d2-b9e8-29178b7ef1b5.png)

Ghabry:
wtf?

Why even use Show String Pic here. This is a Picture without text.

florianessl:
Similar to both this and the other mentioned issues:
The blitting logic of the font renderer is taking information from outside of the indexed color gradient defined in the system graphic.
For comparison, here's how a big font is handled by both Maniacs & EasyRPG (font sizes do also differ)

Maniacs:
![rm2k3](https://github.com/EasyRPG/Player/assets/42896817/07be8b13-33b7-4f7a-884f-7c6e3f34664e)

EasyRPG:
![easyrpg](https://github.com/EasyRPG/Player/assets/42896817/760765db-573d-4680-8cac-220c36136e28)

System Graphic used:
![systemgraphic](https://github.com/EasyRPG/Player/assets/42896817/d4672b69-c255-4c1e-885a-41e062555528)

[...]

Issue #2880, opened 2023-01-06
Custom font: Crash with error "Couldn't load FreeType character 0x81"


Mimigris:
#### Name of the game:
This is a general issue.

#### Player platform:
Windows and Online Player, latest continuous build.
The font used is the font extracted from the ultimate_rt_eb.dll file.

#### Describe the issue in detail and how to reproduce it:
When using a custom font, some characters may crash the game if they are loaded, like "", instead of trying to fallback on the default font used by EasyRPG. It does not happen with every font.
![image](https://user-images.githubusercontent.com/108797333/211030916-4feb0780-8072-4373-98ee-05544767e14e.png)

Here's a test game if you want to test the issue. Make sure to use a continuous build of EasyRPG Player. Please note that the RTP of RPG Maker 2003 may be needed. The game should crash on the title screen on EasyRPG Player while this is not the case if you try to use the RPG_RT.exe.
![image](https://user-images.githubusercontent.com/108797333/211031067-17cd36d6-edab-4e54-8784-94680e6e2056.png)

[Font Crash Test.zip](https://github.com/EasyRPG/Player/files/10360561/Font.Crash.Test.zip)

Ghabry:
What are you trying to do here? 0x81 is a control character. This is a non-printable character. It doesn't belong into the string.

---------------

I should add the error code to the error message.

The error code is 3 which corresponds to "broken file". :laughing:

This means the file has a character map entry for 0x81 but it doesn't point to a valid character in the file.

RPG_RT appears to ignore this.

I'm not sure what to do with this as the font is obviously broken garbage.

Mimigris:
> What are you trying to do here? 0x81 is a control character. This is a non-printable character. It doesn't belong into the string.

Well the game seems to crash once this character needs to be displayed, which should not be an issue... but it is used in some games like the official English translation of Yume Nikki for example (getting the Kitchen Knife effect "                   Kitchen Knife" or changing Menu Theme while the time is stopped with the Traffic Light effect "@@@@@@šƒƒjƒ…[ƒ^ƒCƒvƒ`ƒFƒ“ƒWš" - this second message was left untranslated by error though).
![screenshot_44](https://user-images.githubusercontent.com/108797333/211048799-5e462a25-a285-4f56-ac72-20b218a92296.png)
![image](https://user-images.githubusercontent.com/108797333/211048917-27f965a6-7eb0-400e-8eaa-43994aa67580.png)
[...]

PR #3168, opened 2023-12-04
Config for Soundfont, Font and "Focus Lost"

Ghabry:
Pause when focus lost is now a setting, not a compile time option:

screenshot_0 png

More Audio options

screenshot_2 png

Information about MIDI drivers

screenshot_1 png

The soundfont can be selected

screenshot_3 png

Font configuration with preview:

screenshot_4 png

screenshot_5 png

### DIFF
--- a/src/filefinder.h
+++ b/src/filefinder.h
@@ -37,6 +37,14 @@
  * insensitive files paths.
  */
 namespace FileFinder {
+	constexpr const auto IMG_TYPES = Utils::MakeSvArray(".bmp",  ".png", ".xyz");
+	constexpr const auto MUSIC_TYPES = Utils::MakeSvArray(
+			".opus", ".oga", ".ogg", ".wav", ".mid", ".midi", ".mp3", ".wma");
+	constexpr const auto SOUND_TYPES = Utils::MakeSvArray(
+			".opus", ".oga", ".ogg", ".wav", ".mp3", ".wma");
+	constexpr const auto FONTS_TYPES = Utils::MakeSvArray(".fon", ".fnt", ".bdf", ".ttf", ".ttc", ".otf", ".woff2", ".woff");
+	constexpr const auto TEXT_TYPES = Utils::MakeSvArray(".txt", ".csv", ""); // "" = Complete Filename (incl. extension) provided by the user
+
 	/**
 	 * Quits FileFinder.
 	 */
--- a/src/filefinder.h
+++ b/src/filefinder.h
@@ -37,7 +37,7 @@
  * insensitive files paths.
  */
 namespace FileFinder {
-	constexpr const auto IMG_TYPES = Utils::MakeSvArray(".bmp",  ".png", ".xyz");
+	constexpr const auto IMG_TYPES = Utils::MakeSvArray(".bmp", ".png", ".xyz");
 	constexpr const auto MUSIC_TYPES = Utils::MakeSvArray(
 			".opus", ".oga", ".ogg", ".wav", ".mid", ".midi", ".mp3", ".wma");
 	constexpr const auto SOUND_TYPES = Utils::MakeSvArray(
--- a/src/window_settings.h
+++ b/src/window_settings.h
@@ -42,6 +42,7 @@ public:
 		eInputButtonRemove,
 		eVideo,
 		eAudio,
+		eAudioMidi,
 		eLicense,
 		eEngine,
 		eSave,
@@ -60,8 +61,10 @@ public:
 		std::string text;
 		std::string value_text;
 		std::string help;
+		std::string help2;
 		std::function<void(void)> action;
 		OptionMode mode;
+		Font::SystemColor color = Font::ColorDefault;
 		int current_value;
 		int original_value;
 		int min_value;
@@ -106,6 +109,7 @@ public:
 	 */
 	void Refresh();
 
+	Window_Help* help_window2 = nullptr;
 private:
 
 	/**
@@ -134,6 +138,7 @@ private:
 	void RefreshButtonList();
 	void RefreshVideo();
 	void RefreshAudio();
+	void RefreshAudioMidi();
 	void RefreshEngine();
 	void RefreshLicense();
 
--- a/src/window_settings.h
+++ b/src/window_settings.h
@@ -169,6 +169,7 @@ private:
 		English,
 		French,
 		German,
+		Spanish,
 		Chinese,
 		Japanese_Hiragana,
 		Japanese_Kanji,
@@ -176,14 +177,15 @@ private:
 		Diacritics
 	};
 
-	EnumConfigParam<SampleText, 8> sample_text{
+	EnumConfigParam<SampleText, 9> sample_text{
 		"<Sample Text>", "Text to show in the preview", "", "", SampleText::English,
-		Utils::MakeSvArray("English", "French", "German", "Chinese", "Japanese (Hiragana)", "Japanese (Kanji)", "Korean", "Various Diacritics"),
-		Utils::MakeSvArray("", "", "", "", "", "", "", ""),
+		Utils::MakeSvArray("English", "French", "German", "Spanish", "Chinese", "Japanese (Hiragana)", "Japanese (Kanji)", "Korean", "Various Diacritics"),
+		Utils::MakeSvArray("", "", "", "", "", "", "", "", ""),
 		Utils::MakeSvArray(
 			"The quick brown fox jumps over the lazy dog 1234567890.?!",
-			"Voix ambiguë d’un cœur qui au zéphyr préfère les jattes de kiwis",
+			"Voix ambiguë d'un cœur qui au zéphyr préfère les jattes de kiwis",
 			"Victor jagt zwölf Boxkämpfer quer über den großen Sylter Deich",
+			"Jovencillo emponzoñado de whisky, ¡qué figurota exhibe!",
 			"天地玄黃宇宙洪荒。日月盈昃辰宿列張。寒來暑往秋收冬藏。閏餘成歲律呂調陽。雲騰致雨露結為霜。金生麗水玉出崑岡。劍號巨闕珠稱夜光。果珍李柰菜重芥薑。海鹹河淡鱗潛羽翔。",
 			"いろはにほへと ちりぬるを わかよたれそ つねならむ うゐのおくやま けふこえて あさきゆめみし ゑひもせす",
 			"色は匂えど散りぬるを我が世誰ぞ常ならん有為の奥山今日越えて浅き夢見じ酔いもせず",
--- a/src/window_settings.h
+++ b/src/window_settings.h
@@ -82,6 +82,7 @@ public:
 		int arg = -1;
 		int scratch = 0;
 		int scratch2 = 0;
+		std::vector<Option> options;
 	};
 
 	/** Constructor  */
@@ -90,13 +91,13 @@ public:
 	/** @return true if the index points to an enabled action */
 	bool IsCurrentActionEnabled() const {
 		return (index >= 0
-				&& index < static_cast<int>(options.size())
-				&& static_cast<bool>(options[index].action));
+				&& index < static_cast<int>(GetFrame().options.size())
+				&& static_cast<bool>(GetFrame().options[index].action));
 	}
 
 	/** Execute the action pointed to by index */
 	Option& GetCurrentOption() {
-		return options[index];
+		return GetFrame().options[index];
 	}
 
 	UiMode GetMode() const;
@@ -149,8 +150,6 @@ private:
 	void RefreshLicense();
 
 	void UpdateHelp() override;
-
-	std::vector<Option> options;
 
 	struct Memory {
 		int index = 0;
--- a/src/window_settings.h
+++ b/src/window_settings.h
@@ -43,8 +43,11 @@ public:
 		eVideo,
 		eAudio,
 		eAudioMidi,
+		eAudioSoundfont,
 		eLicense,
 		eEngine,
+		eEngineFont1,
+		eEngineFont2,
 		eSave,
 		eEnd,
 		eAbout,
@@ -110,6 +113,7 @@ public:
 	void Refresh();
 
 	Window_Help* help_window2 = nullptr;
+	RangeConfigParam<int> font_size { "<Font Size>", "Font size to use. Not supported for the built-in font.", "", "", 12, 6, 16};
 private:
 
 	/**
@@ -139,7 +143,9 @@ private:
 	void RefreshVideo();
 	void RefreshAudio();
 	void RefreshAudioMidi();
+	void RefreshAudioSoundfont();
 	void RefreshEngine();
+	void RefreshEngineFont(bool mincho);
 	void RefreshLicense();
 
 	void UpdateHelp() override;
--- a/src/window_settings.h
+++ b/src/window_settings.h
@@ -164,6 +164,34 @@ private:
 
 	void SavePosition();
 	void RestorePosition();
+
+	// The sample text for the font preview. This option is not saved.
+	enum class SampleText {
+		English,
+		French,
+		German,
+		Chinese,
+		Japanese_Hiragana,
+		Japanese_Kanji,
+		Korean,
+		Diacritics
+	};
+
+	EnumConfigParam<SampleText, 8> sample_text{
+		"<Sample Text>", "Text to show in the preview", "", "", SampleText::English,
+		Utils::MakeSvArray("English", "French", "German", "Chinese", "Japanese (Hiragana)", "Japanese (Kanji)", "Korean", "Various Diacritics"),
+		Utils::MakeSvArray("", "", "", "", "", "", "", ""),
+		Utils::MakeSvArray(
+			"The quick brown fox jumps over the lazy dog 1234567890.?!",
+			"Voix ambiguë d’un cœur qui au zéphyr préfère les jattes de kiwis",
+			"Victor jagt zwölf Boxkämpfer quer über den großen Sylter Deich",
+			"天地玄黃宇宙洪荒。日月盈昃辰宿列張。寒來暑往秋收冬藏。閏餘成歲律呂調陽。雲騰致雨露結為霜。金生麗水玉出崑岡。劍號巨闕珠稱夜光。果珍李柰菜重芥薑。海鹹河淡鱗潛羽翔。",
+			"いろはにほへと ちりぬるを わかよたれそ つねならむ うゐのおくやま けふこえて あさきゆめみし ゑひもせす",
+			"色は匂えど散りぬるを我が世誰ぞ常ならん有為の奥山今日越えて浅き夢見じ酔いもせず",
+			"키스의 고유조건은 입술끼리 만나야 하고 특별한 기술은 필요치 않다",
+			"ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖØÙÚÛÜÝÞßàáâãäåæçèéêëìíîïðñòóôõöøùúûüýþÿ"
+		)
+	};
 };
 
 #endif
--- a/src/window_settings.h
+++ b/src/window_settings.h
@@ -113,16 +113,16 @@ public:
 	 */
 	void Refresh();
 
-	Window_Help* help_window2 = nullptr;
-	RangeConfigParam<int> font_size { "<Font Size>", "Font size to use. Not supported for the built-in font.", "", "", 12, 6, 16};
-private:
-
 	/**
-	 * Draws an item together with the quantity.
+	 * Draws an item.
 	 *
 	 * @param index index of item to draw.
 	 */
 	void DrawOption(int index);
+
+	Window_Help* help_window2 = nullptr;
+	RangeConfigParam<int> font_size { "<Font Size>", "Font size to use. Not supported for the built-in font.", "", "", 12, 6, 16};
+private:
 
 	template <typename Param, typename Action>
 	void AddOption(const Param& p,
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1052,7 +1052,7 @@ if(${PLAYER_AUDIO_BACKEND} MATCHES "^(SDL2|SDL1|libretro|psvita|3ds|switch|wii|a
 		player_find_package(NAME FluidLite
 			CONDITION PLAYER_WITH_FLUIDLITE
 			DEFINITION HAVE_FLUIDLITE
-			TARGET FluidLite::fluidlite)
+			TARGET "fluidlite::fluidlite;fluidlite::fluidlite-static")
 	endif()
 
 	# xmp (lite)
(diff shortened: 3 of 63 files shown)


src/filefinder.h
```diff
@@ -37,6 +37,14 @@
  * insensitive files paths.
  */
 namespace FileFinder {
+	constexpr const auto IMG_TYPES = Utils::MakeSvArray(".bmp",  ".png", ".xyz");
```
fdelapena, line 40: nit: two spaces before ".png"

Mimigris:
I know that it is still WIP but here's what I found in terms of issues while testing:

- The `Pause when focus lost` text changed since the picture that you sent and is now too long to properly fit on-screen (a custom font is used here but it's also the case with the default font)
screenshot_5

- The `MIDI Drivers` menu seems to be laggy when selecting an option if no MIDI has been played during the current session and in some other occasions (e.g. game browser menu - the lag also occurs when selecting `> Information <` even if it doesn't do anything outside of playing a sound)
- The `MIDI Drivers` menu falsely report `Native MIDI` as Not working if attempted to be used in-game (tested with the Windows x64 version - the setting work as intended outside of that, it's just the display in the menu that is incorrect) with a warning too long to be displayed and read
screenshot_6
- Unlike changing `MIDI Drivers`, changing the soundfont on the fly with the setting menu will not change it without reloading the game, which can be annoying to properly test

- If you want to change the size of the font currently being used, you need to change the size and then reselect the entry, you cannot just change the size to see and apply the changes, which is a bit annoying
- Since the font change is directly applied once a new character has been rendered, you can change it while displaying a message to have a message with different character sets, as well as having an incorrect spacing being used: it doesn't seem to cause gameplay issues since it only impacts the current message and doesn't change anything in terms of gameplay so I don't think that it's a huge problem
image


Now more on a feedback side, feel free to take that into account or not:
- Disabling all MIDI entries display a "Couldn't play BGM music.mid. Format not supported" warning (that's intended), but if possible, maybe it could be better to specify that no MIDI entry is available instead in said case
- If I remember correctly, there is a thing on the RPG_RT where the two fonts used are not the same depending on the region/encoding, maybe having something to reverse the fonts used (either as a toggle directly in the menu or automatically depending on the encoding) could be used
- Having the possibility to maybe add an X number of pixels after a character could be useful for some fonts where all of the letters are sticked together, making them unreadable

Also, the webplayer doesn't seem to like the entries Open Font/Sound directory and seems to randomly crash outside of that, I'm not entirely sure why for the second part.

Ghabry:
Thanks for the exhaustive testing. I won't respond to all but they all sound reasonable.

> The Pause when focus lost text changed since the picture that you sent and is now too long to properly fit on-screen (a custom font is used here but it's also the case with the default font)

Hm I already fixed this one locally, maybe forgot to commit it.

But yeah some messages do not fit. I think about implementing a "scrolling" feature that scrolls the text back and forth so the rest is visible.

> Unlike changing MIDI Drivers, changing the soundfont on the fly with the setting menu will not change it without reloading the game, which can be annoying to properly test

Yeah I would prefer to have some kind of "hot reloading" here so the audio changes immediately. I just couldn't find a way yet that does not crash xD.

> Having the possibility to maybe add an X number of pixels after a character could be useful for some fonts where all of the letters are sticked together, making them unreadable

Good idea. This is already required for StringPic so can be exposed.

Ghabry:
Instead of "Open (Sound)Font" emscripten got now "Upload (Sound)Font".

Can already see a certain game developers demanding that this is patched out but not my problem :P

Ghabry:
Removed the "preload in a thread" commit as it appears to crash sometimes and have no time to figure out where the race condition is. Here the code for posterity:

```patch
commit ea340f22dea3571fa4988734b93ab4c4b90cc413
Date:   Thu Dec 21 15:07:15 2023 +0100

    Fluidsynth: Preinit soundfont in a thread on our major platforms.
    
    Reduces lag

diff --git a/src/audio_generic.cpp b/src/audio_generic.cpp
index cd45e13d..0839dcaa 100755
--- a/src/audio_generic.cpp
+++ b/src/audio_generic.cpp
@@ -218,7 +218,7 @@ bool GenericAudio::PlayOnChannel(BgmChannel& chan, Filesystem_Stream::InputStrea
 		chan.decoder.reset();
 
 		// Order is Fluidsynth, WildMidi, Native, FmMidi
-		bool fluidsynth = Audio().GetFluidsynthEnabled() && MidiDecoder::CreateFluidsynth(true);
+		bool fluidsynth = Audio().GetFluidsynthEnabled() && MidiDecoder::CreateFluidsynthWithLock(true);
 		bool wildmidi = Audio().GetWildMidiEnabled() && MidiDecoder::CreateWildMidi(true);
 
 		if (!fluidsynth && !wildmidi && Audio().GetNativeMidiEnabled()) {
diff --git a/src/audio_midi.cpp b/src/audio_midi.cpp
index 9827be97..91878f7c 100644
--- a/src/audio_midi.cpp
+++ b/src/audio_midi.cpp
@@ -26,7 +26,18 @@
 #ifdef USE_AUDIO_RESAMPLER
 #include "audio_resampler.h"
 #include "audio.h"
+#endif
+
+#if (defined(HAVE_FLUIDSYNTH) || defined(HAVE_FLUIDLITE)) && defined(SUPPORT_STD_SYNCHRONIZATION)
+#include <atomic>
+#include <mutex>
+#include <thread>
 
+namespace {
+	std::mutex sf_load_mutex;
+	std::thread sf_load_thread;
+	std::atomic<bool> sf_load_thread_done;
+}
 #endif
 
 void MidiDecoder::GetFormat(int& freq, AudioDecoderBase::Format& format, int& channels) const {
@@ -53,7 +64,7 @@ std::unique_ptr<AudioDecoderBase> MidiDecoder::Create(bool resample) {
 	std::unique_ptr<AudioDecoderBase> mididec;
 
 	if (Audio().GetFluidsynthEnabled()) {
-		mididec = CreateFluidsynth(resample);
+		mididec = CreateFluidsynthWithLock(resample);
 	}
 
 	if (!mididec && Audio().GetWildMidiEnabled()) {
@@ -90,6 +101,19 @@ std::unique_ptr<AudioDecoderBase> MidiDecoder::CreateFluidsynth(bool resample) {
 	return mididec;
 }
 
+std::unique_ptr<AudioDecoderBase> MidiDecoder::CreateFluidsynthWithLock(bool resample) {
+#if (defined(HAVE_FLUIDSYNTH) || defined(HAVE_FLUIDLITE)) && defined(SUPPORT_STD_SYNCHRONIZATION)
+	const std::lock_guard<std::mutex> lock(sf_load_mutex);
+
+	if (sf_load_thread_done) {
+		sf_load_thread.join();
+		sf_load_thread_done.store(false);
+	}
+#endif
+
+	return CreateFluidsynth(resample);
+}
+
 std::unique_ptr<AudioDecoderBase> MidiDecoder::CreateWildMidi(bool resample) {
 	std::unique_ptr<AudioDecoderBase> mididec;
 
@@ -133,6 +157,10 @@ std::unique_ptr<AudioDecoderBase> MidiDecoder::CreateFmMidi(bool resample) {
 }
 
 bool MidiDecoder::CheckFluidsynth(std::string& status_message) {
+#if (defined(HAVE_FLUIDSYNTH) || defined(HAVE_FLUIDLITE)) && defined(SUPPORT_STD_SYNCHRONIZATION)
+	const std::lock_guard<std::mutex> lock(sf_load_mutex);
+#endif
+
 	if (works.fluidsynth && works.fluidsynth_status.empty()) {
 		CreateFluidsynth(true);
 	}
@@ -141,7 +169,22 @@ bool MidiDecoder::CheckFluidsynth(std::string& status_message) {
 	return works.fluidsynth;
 }
 
+void MidiDecoder::PreloadFluidsynth(bool resample) {
+#if (defined(HAVE_FLUIDSYNTH) || defined(HAVE_FLUIDLITE)) && defined(SUPPORT_STD_SYNCHRONIZATION)
+	sf_load_thread_done.store(false);
+
+	sf_load_thread = std::thread([resample]() {
+		CreateFluidsynthWithLock(resample);
+		sf_load_thread_done.store(true);
+	});
+#endif
+}
+
 void MidiDecoder::ChangeFluidsynthSoundfont(StringView sf_path) {
+#if (defined(HAVE_FLUIDSYNTH) || defined(HAVE_FLUIDLITE)) && defined(SUPPORT_STD_SYNCHRONIZATION)
+	const std::lock_guard<std::mutex> lock(sf_load_mutex);
+#endif
+
 	if (!works.fluidsynth || works.fluidsynth_status.empty()) {
 		// Fluidsynth was not initialized yet or failed, will use the path from the config automatically
 		works.fluidsynth = true;
@@ -173,6 +216,10 @@ void MidiDecoder::Reset() {
 #endif
 
 #if defined(HAVE_FLUIDSYNTH) || defined(HAVE_FLUIDLITE)
+#ifdef SUPPORT_STD_SYNCHRONIZATION
+	const std::lock_guard<std::mutex> lock(sf_load_mutex);
+#endif
+
 	FluidSynthDecoder::ResetState();
 #endif
 }
diff --git a/src/audio_midi.h b/src/audio_midi.h
index 8882cc35..67313c7b 100644
--- a/src/audio_midi.h
+++ b/src/audio_midi.h
@@ -200,6 +200,8 @@ public:
 
 	static std::unique_ptr<AudioDecoderBase> CreateFluidsynth(bool resample);
 
+	static std::unique_ptr<AudioDecoderBase> CreateFluidsynthWithLock(bool resample);
+
 	static std::unique_ptr<AudioDecoderBase> CreateWildMidi(bool resample);
 
 	static std::unique_ptr<AudioDecoderBase> CreateFmMidi(bool resample);
@@ -212,6 +214,13 @@ public:
 	 */
 	static bool CheckFluidsynth(std::string& status_message);
 
+	/**
+	 * Attempts to preload Fluidsynth and soundfont.
+	 * Only works on our major platforms because std::mutex and std::thread are
+	 * required.
+	 */
+	static void PreloadFluidsynth(bool resample);
+
 	static void ChangeFluidsynthSoundfont(StringView sf_path);
 
 	/**
diff --git a/src/output.cpp b/src/output.cpp
index 62606ede..78adc9f8 100644
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -51,6 +51,10 @@
 #include "font.h"
 #include "baseui.h"
 
+#ifdef SUPPORT_STD_SYNCHRONIZATION
+#  include <mutex>
+#endif
+
 using namespace std::chrono_literals;
 
 namespace {
@@ -99,6 +103,10 @@ namespace {
 		std::string msg;
 		LogLevel lvl = {};
 	} last_message;
+
+#ifdef SUPPORT_STD_SYNCHRONIZATION
+	std::recursive_mutex output_mutex;
+#endif
 }
 
 LogLevel Output::GetLogLevel() {
@@ -118,6 +126,10 @@ void Output::IgnorePause(bool const val) {
 }
 
 static void WriteLog(LogLevel lvl, std::string const& msg, Color const& c = Color()) {
+#ifdef SUPPORT_STD_SYNCHRONIZATION
+	std::lock_guard<std::recursive_mutex> lock(output_mutex);
+#endif
+
 #ifdef EMSCRIPTEN
 
 // Allow pretty log output and filtering in browser console
diff --git a/src/player.cpp b/src/player.cpp
index 0107f478..492c1e4e 100644
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -667,6 +667,7 @@ void Player::CreateGameObjects() {
 
 	// Reinit MIDI
 	MidiDecoder::Reset();
+	MidiDecoder::PreloadFluidsynth(true);
 
 	// Load the meta information file.
 	// Note: This should eventually be split across multiple folders as described in Issue #1210
diff --git a/src/system.h b/src/system.h
index 74b8d0b1..83850b3d 100644
--- a/src/system.h
+++ b/src/system.h
@@ -52,6 +52,7 @@
 #  define SUPPORT_JOYSTICK
 #  define SUPPORT_TOUCH
+#  define SUPPORT_STD_SYNCHRONIZATION
 #elif defined(EMSCRIPTEN)
 #  define SUPPORT_MOUSE
 #  define SUPPORT_TOUCH
@@ -77,6 +78,7 @@
 #  define SUPPORT_TOUCH
 #  define SUPPORT_JOYSTICK
+#  define SUPPORT_STD_SYNCHRONIZATION
 #elif defined(__SWITCH__)
 #  define SUPPORT_JOYSTICK
@@ -93,6 +95,7 @@
 #  define SUPPORT_TOUCH
 #  define SUPPORT_JOYSTICK
+#  define SUPPORT_STD_SYNCHRONIZATION
 #endif
 
 #ifdef USE_SDL
```

Ghabry:
This is ready for testing now.

Stuff I will not address here (that is all reasonable, could be moved to another issue):

- Lag when loading soundfont (my attempt to fix it causes other issues so stays disabled for now)

> If I remember correctly, there is a thing on the RPG_RT where the two fonts used are not the same depending on the region/encoding, maybe having something to reverse the fonts used (either as a toggle directly in the menu or automatically depending on the encoding) could be used



> Having the possibility to maybe add an X number of pixels after a character could be useful for some fonts where all of the letters are sticked together, making them unreadable

This requires rewriting parts of the font code because fonts are "shared" so has the possibility to cause issues when the default font is used in Show String Pic.

src/window_settings.h
```diff
@@ -153,6 +163,34 @@ class Window_Settings : public Window_Selectable {
...
+		Japanese_Kanji,
+		Korean,
+		Diacritics
+	};
+
+	EnumConfigParam<SampleText, 8> sample_text{
+		"<Sample Text>", "Text to show in the preview", "", "", SampleText::English,
+		Utils::MakeSvArray("English", "French", "German", "Chinese", "Japanese (Hiragana)", "Japanese (Kanji)", "Korean", "Various Diacritics"),
+		Utils::MakeSvArray("", "", "", "", "", "", "", ""),
+		Utils::MakeSvArray(
```
fdelapena, line 183: Jovencillo emponzoñado de whisky, ¡qué figurota exhibe! (Spanish pangram with `¡`)

fdelapena (approved):

Mimigris:
[FontIssue.zip](https://github.com/EasyRPG/Player/files/13761645/FontIssue.zip)
I have some fonts that cause a crash when attempted to be previewed in the font section of the settings (either a complete crash of the Player or a display of an error Cannot render character), feel free to take a look at it.

Ghabry:
Rebased to have the better spacing. Good for testing.

@Mimigris I know why it crashes. Fixing this 2024, no time anymore xD

Mimigris:
I had a lot of fonts that were not really readable due to spacing issues but now with this commit almost every font I have (if we exclude fonts that are not intended to really be used at this size) are readable 🎉 
Since it now works as intended, there should be no need to have an option to adjust the spacing indeed.

> @Mimigris I know why it crashes. Fixing this 2024, no time anymore xD

No problem, no need to rush, have a nice holiday!

Also, as a bonus, to give an example on a use of the font setting, the feature can be used to play a game in a language where the proper display of the character requires an additional font and that was not set with the game itself: here's how the Thai translation of Ib looks:
EasyRPG Thai Ib fork <-> EasyRPG (use of the default ont) <-> EasyRPG (Tahoma, Font Size 12)
image

Ghabry:
@Mimigris  question: You reported a while ago somewhere else that the font embedded in RPG_RT gives in some cases an "Freetype: Couldn't render glyph error" and closed the Player.

The crash fix for "FontIssue.zip" included a fix for this error (well fix, it loads a fallback glyph now) in case you want to retest this :thinking:

Mimigris:
> @Mimigris question: You reported a while ago somewhere else that the font embedded in RPG_RT gives in some cases an "Freetype: Couldn't render glyph error" and closed the Player.
> 
> The crash fix for "FontIssue.zip" included a fix for this error (well fix, it loads a fallback glyph now) in case you want to retest this 🤔

From what I can see that seems to work as a fix for said issue (#2880), now the broken character is replaced by another, so it works as a fix to avoid the crash :+1: (technically it's not entirely accurate to the RPG_RT since it's not the same character that it is displayed between both but... we're talking about a broken font in any case, so that's not the biggest issue in this case, the font itself would need to be replaced)
screenshot_0

jetrotal:
Tested the soundfont for the first time.
Triggered 2 soundfonts at the same time, don't know exactly why:

image


And would be cool to support the ".soundfont" file format with the sf2. For people, like me, who simply copied soundfont from game folder to soundfont folder.

Mimigris:
> Sorry but after ignoring this for three months I lost a bit the track what was missing here 😓
> 
> Can you help me with telling me again which issues should be fixed before merge?

My first comment on the issue (https://github.com/EasyRPG/Player/pull/3168#issuecomment-1842975058) should still be up to date outside of what you just pushed (what is crossed is already fixed, and the other entries are normally still present)

Ghabry:
> The <Open (Sound)Font directory> should not be present on platforms where it is not possible to open a directory directly (e.g. Wii)

This is now added.

> When Pause when focus lost is disabled, if the window of the game is put in the background, the Mouse Scroll Up/Down keys can still be used if the cursor is over the window of the game and the mouse position is still tracked

That the scrolling works when above the window is normal. Is a feature of the operating system. This works for any application windows: Unfocus your browser window and scroll :) (Windows has this since Windows 10)

Mimigris:
> That the scrolling works when the cursor is above the window is normal. Is a feature of the operating system. This works for any application windows: Unfocus your browser window and scroll :) (Windows has this since Windows 10)
> 
> Do you consider this problematic?

This shouldn't be problematic since you need to specifically have your mouse over the window of the game (so even if in some contexts it could be annoying, it can easily be ignored), so not really, it was more a report of what I found that could be considered as an issue.

Ghabry:
Redid the font size changing: Is now shown next to the font and can be adjusted with left/right. Has the advantage that you finally get a live preview that respects the size :)

grafik

jetrotal:
UI nitpick:
Instead of having "Show FPS" and "Show FPS in window", we could have
Show FPS: [OFF], [Overlay], [Titlebar], [Overlay + Titlebar]

keenanweaver:
Super outside the scope of this PR, but could hardware MIDI support be a potential feature in the future? I have a Roland SC-55 and I think it'd be neat to use! dsda-doom uses the PortMIDI library and it works great with that.

Ghabry:
@Mimigris fluidlite should work now and the error with the native midi reporting as "not working" is resolved (I hope, only tested on Linux).

I think this also fixes the "lag". Before the Native MIDI handler was launched 5 times in a row (side effect of how the settings update...). Now it is only launched once.

fdelapena:
> Super outside the scope of this PR, but could hardware MIDI support be a potential feature in the future? I have a Roland SC-55 and I think it'd be neat to use! dsda-doom uses the PortMIDI library and it works great with that.

Native MIDI is the suitable way for hardware devices support. It is supported on Linux, macOS and Windows. You likely need to make it available to operating system port then should be selectable for output.

Ghabry:
Imo this is now finished.

All pending issues look like "nice to have" so they should be moved in a separate issue.

Have fun testing, hope this doesn't break too much ;)

florianessl:
I'm getting a very weird crash on startup when trying to start a debug build of this branch on Windows:
**"Error: Couldn't create -15x-855638032 image."**
 ('Release' seems to work fine)

Both 'width' & 'height' are pretty much corrupted garbage at this point, with negative values...
image

I'm still actively working on getting some solid C++ experience but for a newbie like myself this is still a little bit too steep of an issue.
But even though I can't make out the exact cause, I could isolate the source of problem to the 'font' member in the window base class. Applying this patch file fixes the crash:
[0001-Remove-member-font-from-Window-base-class-revert-to-.patch](https://github.com/EasyRPG/Player/files/14794987/0001-Remove-member-font-from-Window-base-class-revert-to-.patch)

Ghabry:
@florianessl unfortunately I cannot reproduce this. Did you try a clean rebuild? Does this happen with every game?

I even pointed the Player to the Windows font dir with ``--font-path C:\Windows\Fonts`` and scrolled through the list without any issues :D

screenshot_0

fdelapena (approved):

florianessl:
@Ghabry
I did some more research and found that this crash only happens, when "NewGame=1" is set...
But after doing a complete Clean and Rebuilding, everything seems to work fine now. I'm sorry for the hassle!

jetrotal:
@Ghabry I tried RPG Maker default fonts from https://rpgmaker.net/engines/rt2k3/utilities/1/
Would be possible to have it pixelated the same way it is in RPG_RT?  they will always have the harfbuzz look?

Ghabry:
@jetrotal can you tell me font file and size and a screenshot that shows a comparison where you are _unhappy_ with the result?

jetrotal:
[fonts.zip](https://github.com/EasyRPG/Player/files/14817264/fonts.zip)
I tried both Ms Mincho and MS Gothic (Default RPG Maker fonts) in different sizes. I guess the default one is 12?

image

Ghabry:
The msgothi0.ttc and msmincho.ttc from your link above look better.

jetrotal:
> The msgothi0.ttc and msmincho.ttc from your link above look better.

Oh, you are correct... Sorry about the non-issue I pointed then.
Weird, I remember testing ttc without success before

florianessl:
I should probably mention this blitting issue here, as this PR refactors some of the font handling..
(These screenshots were actually taken with this branch but the exact same issue has previously been observed on 0.8)
https://github.com/EasyRPG/Player/issues/2961#issuecomment-2032839468

Ghabry:
@florianessl it is not perfect imo but give it a try:

grafik

----

But this is now the final final contribution to this PR ;)

fdelapena (approved):

Ghabry:
This is now ready in case somebody wants to do final testing.

----

The font stuff needs some Android Ui support imo but I will do this in another PR together with the LZH archive support. As you know I hate Android dev so I always do everything in one go :sweat_smile:

Mimigris:
Changing the volume of the MIDI currently being played in the settings while having the MIDI output set as FluidSynth or FMmidi will not work entirely, since when reloading the music it seems that the volume set is different (tested with the title screen of Wadanohara and the Great Blue Sea - I suspect that the percentage of the volume is relative compared to the percentage that was set in the options when the music started for those two, = if a music starts while your setting is at 10%, the music almost cannot be heard in the end).

Ghabry:
The MIDI volume issue is fixed now. Forgot that the path where I adjusted the volume is used by all MIDI code, not just MidiOut. The volume does still not perfectly match though. Good enough imo.

Also this took me faaaar too long to figure out (2 hours -_-) but I finally solved this "pop" issue you get when only using MIDI out for BGM and a sound effect ends...

jetrotal:
After reading about the stretch settings in the code, got curious about something:

image

The stretch only works horizontally? 
Shouldn't consider the vertical borders of the window as well?

Ghabry:
This only stretches vertically because it was initially invented for Android. When you stretch in both directions it will look very interesting in Portrait orientation. :)

Ghabry:
Will merge this on Sunday when no further issues arise as this is slowly becoming a blocker for my other PRs.

PR #3178, opened 2023-12-23
libretro: simple support for accessing game variables

celerizer:
Allows libretro frontends to read or poke game variables at runtime

### DIFF
--- a/src/platform/libretro/ui.cpp
+++ b/src/platform/libretro/ui.cpp
@@ -21,6 +21,7 @@
 #include "bitmap.h"
 #include "color.h"
 #include "filefinder.h"
+#include "game_variables.h"
 #include "graphics.h"
 #include "input.h"
 #include "keys.h"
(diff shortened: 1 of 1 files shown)


Ghabry:
Hi, what is the intention of this feature?

We talked about this years ago with retro achievements that exposing random chunks of memory is not a good approach because we are not an emulator.

The variables array can change its location and size while the game is running. We also give no guarantees that the memory layout starts the same forever.

celerizer:
No real intention besides giving an implementation to the stubbed function. The data ptr and size should only be asked for between calls to retro_run so the core and frontend won't be trying to modify it at the same time.

Ghabry:
Sorry but as I argued years ago in https://github.com/libretro/easyrpg-libretro/issues/49 I cannot accept this.

We are not an emulator and we have no fixed memory layout. There is no guarantee that the memory layout of the variables will stay the same forever and there are more useful data structures like e.g. switches, string variables etc.

PR #3180, opened 2023-12-28
Maniacs Feature: Add separate Width/Height parameters for scaling pictures, Picture StringVar, Tint Variable

ToolMan2k:
This commit allows for the use of separate Width/Height parameters for rendering pictures.
You can use this feature with ManiacPatch 211010 using the .scale2 parameter.

It also adds scaling support for "Maniac_ShowStringPic" and a quick fix for failed bitmap assertion.

https://github.com/EasyRPG/Player/assets/26819563/e8a6eb26-f7a9-4267-a4f1-e26e4c5e135c

Sister commit here: https://github.com/EasyRPG/liblcf/pull/473

Added:

- Added support for "scale_width" and "scale_height" for ShowPicture(), MovePicture() and Maniac_ShowStringPicture()
- Maniac_GetPictureInfo returns the separate values of "width" and "height"
- Set both Width and Height to Magnify if one of the following conditions are met:
  - ManiacPatch is not enabled;
  - Using the .scale parameter;
  - Picture Effects are enabled (parity with ManiacPatch, but can be disabled inside the source)
- Maniac_ShowStringPicture now properly scales the picture (both with .scale and .scale2)
- New variables for pictures:
  - scale_width
  - scale_height
  - current_scale_width
  - current_scale_height
  - finished_scale_width
  - finished_scale_height

Fixed:

- Assertion failed when getting the Width/Height of an invalid bitmap (Now returns 0 instead).

### DIFF
--- a/src/sprite_picture.cpp
+++ b/src/sprite_picture.cpp
@@ -112,8 +112,12 @@ void Sprite_Picture::Draw(Bitmap& dst) {
 		SetX(x);
 		SetY(y);
 	}
+
 	SetZoomX(data.current_magnify / 100.0);
 	SetZoomY(data.current_magnify / 100.0);
+	if (Player::IsPatchManiac()) {
+		SetZoomY(data.maniac_current_magnify_height / 100.0);
+	}
 
 	auto sr = GetSrcRect();
 	SetOx(sr.width / 2);
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -2822,7 +2822,12 @@ bool Game_Interpreter::CommandShowPicture(lcf::rpg::EventCommand const& com) { /
 
 	if (param_size > 16 && (Player::IsRPG2k3ECommands() || Player::IsPatchManiac())) {
 		// Handling of RPG2k3 1.12 chunks
-		pic_id = ValueOrVariable(com.parameters[17], pic_id);
+		if (Player::IsPatchManiac()) {
+			pic_id = ValueOrVariableBitfield(com.parameters[17], 0, pic_id);
+			params.name = ToString(CommandStringOrVariableBitfield(com, 17, 2, 30));
+		} else {
+			pic_id = ValueOrVariable(com.parameters[17], pic_id);
+		}
 		if (com.parameters[19] != 0) {
 			int var = 0;
 			if (Main_Data::game_variables->IsValid(com.parameters[19])) {
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -2707,10 +2707,11 @@ namespace PicPointerPatch {
 	}
 
 	static void AdjustParams(Game_Pictures::Params& params) {
-		if (params.magnify > 10000) {
-			int new_magnify = Main_Data::game_variables->Get(params.magnify - 10000);
-			Output::Debug("PicPointer: Zoom {} replaced with {}", params.magnify, new_magnify);
-			params.magnify = new_magnify;
+		if (params.magnify_width > 10000) {
+			int new_magnify = Main_Data::game_variables->Get(params.magnify_width - 10000);
+			Output::Debug("PicPointer: Zoom {} replaced with {}", params.magnify_width, new_magnify);
+			params.magnify_width = new_magnify;
+			params.magnify_height = new_magnify;
 		}
 
 		if (params.top_trans > 10000) {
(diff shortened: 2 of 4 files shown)


fdelapena:
jenkins: ok to test

M3Kat:
Looking at the failed test, we'll need to compile liblcf using a modified fields_easyrpg file.

Ghabry:
The build failures are normal when depending on lcf changes.

You added new easyrpg chunks here. Did you check how maniac patch stores it in the save file? If it isn't something annoying like storing it in some upper bits we usually try to be compatible to mp here.

src/sprite_picture.cpp
```diff
@@ -169,7 +181,9 @@ int Sprite_Picture::GetFrameWidth() const {
 	const auto& data = pic.data;
 
 	auto& bitmap = GetBitmap();
-	assert(bitmap);
+	if (!bitmap) {
```
Ghabry, line 184: We use ``assert`` for "this should never ever happen during normal execution otherwise it is a bug in the code".
Can you explain how you triggered this "impossible" assert?

M3Kat: Playing around with my test project, I can consistently crash the engine whenever I do the following in this order:

1. We modify a X and Y variable
2. We immediately move an anchored picture and set its coordinates to X and Y.

Even if we add an wait period between both commands, it still crashes.
If we invert these steps, it doesn't happen anymore, go figure.

Here's the test project if you want to play around with: 
[PictureTest-assert.zip](https://github.com/EasyRPG/Player/files/13823922/PictureTest-assert.zip)

M3Kat:
> You added new easyrpg chunks here. Did you check how maniac patch stores it in the save file? If it isn't something annoying like storing it in some upper bits we usually try to be compatible to mp here.

This is definitely something that needs more love. I did try to load saves between original RPG_RT and EasyRPG, it does prints about "unknown chunks" inside the save file. Inside my above test project, you can make an instant save file by pressing SHIFT.

Here's the log I've posted on the bug tracker post: 

```
Skipped Chunk 86 (1 byte) in lcf at 1F9 (SaveSystem)
01 
Skipped Chunk 87 (1 byte) in lcf at 1FD (SaveSystem)
04 
Reading Primitive of incorrect size 4 (expected 1) at 29D
SavePicture: Corrupted Chunk 0x06 (size: 4, pos: 0x29d): fixed_to_map : Read 7 bytes! Reseting...
Reading Primitive of incorrect size 3 (expected 1) at 2AD
SavePicture: Corrupted Chunk 0x09 (size: 3, pos: 0x2ad): use_transparent_color : Read 5 bytes! Reseting...
Skipped Chunk 0A (8 byte) in lcf at 2B2 (SavePicture)
00 00 00 00 00 60 73 40 
Skipped Chunk 24 (2 byte) in lcf at 2D0 (SavePicture)
82 36 
Skipped Chunk 0A (8 byte) in lcf at 345 (SavePicture)
00 00 00 00 00 60 73 40 
Skipped Chunk 1E (2 byte) in lcf at 362 (SavePicture)
82 20 
Skipped Chunk 24 (2 byte) in lcf at 37E (SavePicture)
82 36 
Debug: Loading Save Save01.lsd
Skipped Chunk 86 (1 byte) in lcf at 1F9 (SaveSystem)
01 
Skipped Chunk 87 (1 byte) in lcf at 1FD (SaveSystem)
04 
Reading Primitive of incorrect size 4 (expected 1) at 29D
SavePicture: Corrupted Chunk 0x06 (size: 4, pos: 0x29d): fixed_to_map : Read 7 bytes! Reseting...
Reading Primitive of incorrect size 3 (expected 1) at 2AD
SavePicture: Corrupted Chunk 0x09 (size: 3, pos: 0x2ad): use_transparent_color : Read 5 bytes! Reseting...
Skipped Chunk 0A (8 byte) in lcf at 2B2 (SavePicture)
00 00 00 00 00 60 73 40 
Skipped Chunk 24 (2 byte) in lcf at 2D0 (SavePicture)
82 36 
Skipped Chunk 0A (8 byte) in lcf at 345 (SavePicture)
00 00 00 00 00 60 73 40 
Skipped Chunk 1E (2 byte) in lcf at 362 (SavePicture)
82 20 
Skipped Chunk 24 (2 byte) in lcf at 37E (SavePicture)
82 36 
Debug: Savegame version 0 (RPG_RT or EasyRPG Player Pre-0.6.0)
Debug: Loaded Map Map0001.lmu
Debug: Tree: MAP0001
```

M3Kat:
While playing around, jumping between RPG_RT, saving, then reading the savefile using LCF2XML, I can determine the following things:

1. Maniacs uses "current_magnify" and "finish_magnify" for the Width Parameter.
2. When using the height parameter, it adds 8 bytes to the 0A chunk, and  adds between 1 to 4 bytes in chunk 24. 
3. 0A seems to be the current height double, and 24 seems to be the final height integer.

M3Kat:
I'll recompile liblcf with the proper indexes, then I'll let you know.
However, I have a quick question: is it safe to reuse a save chunk twice?
For instance, using the block 07 for both current_magnify and current_width_scale, since they're the same value internally.
Or would we have to change the name for everything, even though width/height is only for Maniac? I would certainly keep magnify as it is for compatibility reasons.

M3Kat:
As per https://github.com/EasyRPG/liblcf/pull/473, I've updated the chunk pointers.

Now I can save and load between Maniacs and EasyRPG, and scaling is kept between the twos.

Ghabry (approved):
redid parts of the patch so fix some issues. Is now fine for me.

fdelapena (approved):

Ghabry:
Jenkins: Test this please

Issue #2719, opened 2022-01-31
Not finding libfmt but libfmt-dev version 5.2.1 installed


Exarkuniv:
Thank you for supporting EasyRPG Player.
Before creating an issue, please verify that your bug still occurs in the "continuous builds" of the Player.
You can get them at the bottom of our download section: https://easyrpg.org/player/downloads

---
Please fill in the following fields before submitting an issue:
#### Player platform:
Linux
Raspberry Pi4/ Buster Debian 10

#### Attach files (as a .zip archive or link them)
 - A savegame next to the problem (if it is not at the beginning of the game) N/A
 - The `easyrpg_log.txt` log file (you can find it in the game directory) N?A

#### Describe the issue in detail and how to reproduce it:

I took over for zerojay who had a script written for have this install on RetroPie and Raspberry's Pis. right not it does not work. it is failing when looking for ``libfmt``, which is installed with ``libfmt-dev version 5.2.1`` it will come up with a prompt for me to install 5.3 or higher, but i know from looking in the issues that you guys it allowed to build with older then 5.3 
i know that it is pulling the least build from GuitHub

Ghabry:
Thanks for the report.

This happens in some build environments and we are not sure why (maybe related to crosscompiling?).

As a workaround you could use CMake as this is supported by fmt 5.2 natively.

Exarkuniv:
thanks for the quick response. ill give that a try and report back

Exarkuniv:
ok now that i have taken a better look at whats going. it is failing to find FMT when i run ``./configure --prefix=/usr --enable-fmmidi=fallback``
i dont even get to running ``cmake``

carstene1ns:
You do not need to run the autoconf stuff when using cmake:
```
# clone Player repo / use release tarball
cmake . -DCMAKE_BUILD_TYPE=Release -DPLAYER_BUILD_LIBLCF=ON
cmake --build .
# move the 'easyrpg-player' executable where you want it
```

Exarkuniv:
that seemed to work better. it did end up getting a error at the end, but im also getting some other things working weird on my end. so im going to redo my load and start from a clean image and ill report back

Exarkuniv:
ok i did a clean image and tried it again i still get a error at the end

```
[ 98%] Linking CXX executable easyrpg-player
/usr/bin/ld: cannot find -lbcm_host
collect2: error: ld returned 1 exit status
[...]

PR #3181, opened 2023-12-31
add wii mixer and logging cleanup

carstene1ns:
tested on wii, linux
untested on windows, macos, android, emscripten, 3ds, psvita, switch

Some notes:
- Wii plays music now directly with DSP, no abstraction over SDL.
- Color log messages are always enabled now, rang did check for a valid terminal (isatty), but I figured nobody redirects player output to file, when there is a logfile. fmt has no support for old terminals on windows, if this is a problem, we can disable colors for these later
- Android logs are cleaner now in release builds
- 3DS should be able to log over network now (untested, copied from switch)


src/string_view.h
```diff
@@ -22,11 +22,6 @@
 #include <lcf/dbstring.h>
 #include <fmt/core.h>
 
-// Needed to allow building with fmt 5, older versions are untested.
```
carstene1ns, line 25: This makes us remove retropie support, which seems still based on buster... #2719
Maybe let it stay until mid/end 2024?

Ghabry: Does Buster support a recent gcc with c++17 support? If not this doesn't matter anyway.

Otherwise: Add a comment so it gets deleted right after 0.8.1 :)

carstene1ns: Version 8.3, which supports most features: https://en.cppreference.com/w/cpp/compiler_support/17

fdelapena (approved):

Ghabry (approved):

Ghabry:
> fmt has no support for old terminals on windows

Well the release build on Windows has no "console subsystem" so you will not see them anyway when you are not programming. xD

PR #3182, opened 2024-01-07
 Maniac Patch : Fix String Vars with ShowPicture

ToolMan2k:
Added the following:
- Check if we're using a string variable with com.parameters[17]
- Set the name parameter to the String Variable

Fixed the following:
- When using a String Var with ShowPicture, it no longers crash the game with picture ID -1

Before: 😿
EasyRPG-Failure

After: 😺
EasyRPG-Winners


jetrotal:
Thanks! 

String Vars is a very recent implementation, I wonder if more commands can break when using it.

ToolMan2k:
Well my implementation was already broken, since I misplaced a "]" character.
Fixed it, now it works!

jetrotal:
@ToolMan2k thx again!
 
One suggestion:
image

Squashing commits like "fixing typos" or "remove debug/comments" always help shortening and maintaning the list of commits

ToolMan2k:
> Squashing commits like "fixing typos" or "remove debug/comments" always help shortening and maintaning the list of commits

Will do!

Ghabry:
I fixed a bug and moved the commits into you other PR. thanks!

PR #3192, opened 2024-02-28
DirectMenuPatch

florianessl:
Implementation for the _DirectMenuPatch_ for RM2k(3) as described in Makerpendiums patch database:
https://dev.makerpendium.de/docs/patch_db/main-en.htm?page=direct_menu

This patch is enabled by adding the following line to the **[Patch]** section in _EasyRPG.ini_
`DirectMenu=[VariableId]`

Additionally the "OpenMainMenu" command is extended to mirror the same functionality. (If both variants are used, scenes set via command parameter take priority)
_1st param_: scene id
_2nd param_: party member id or actor id
_3rd param_: **0**: party member id is used, **1**: db actor id is used

Default var ids for this patch are V3326 & V3327

EDIT: It should be noted that in this implementation scene id 4 (Status) also works for RM2K and not just 2k3 engine type games.

### DIFF
--- a/src/game_config_game.cpp
+++ b/src/game_config_game.cpp
@@ -17,6 +17,7 @@
 
 #include "game_config_game.h"
 #include "cmdline_parser.h"
+#include "directory_tree.h"
 #include "filefinder.h"
 #include "options.h"
 #include "output.h"
@@ -85,7 +86,8 @@ void Game_ConfigGame::LoadFromArgs(CmdlineParser& cp) {
 			patch_common_this_event.Lock(false);
 			patch_key_patch.Lock(false);
 			patch_rpg2k3_commands.Lock(false);
-			patch_anti_lag_switch.Lock(false);
+			patch_anti_lag_switch.Lock(0);
+			patch_direct_menu.Lock(0);
 			patch_override = true;
 			continue;
 		}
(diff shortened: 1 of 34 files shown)


carstene1ns:
jenkins: ok to test

fdelapena (approved):

src/game_config_game.cpp
```diff
@@ -105,6 +106,11 @@ void Game_ConfigGame::LoadFromArgs(CmdlineParser& cp) {
 					patch_key_patch.Set(true);
 				} else if (v == "rpg2k3-cmds" || v == "rpg2k3-commands") {
 					patch_rpg2k3_commands.Set(true);
+				} else if (v.rfind("direct-menu=") == 0) {
+					int num = static_cast<int>(std::strtol(v.substr(12).c_str(), nullptr, 0));
+					if (num > 0) {
+						patch_direct_menu.Set(num);
```
Ghabry, line 112: That looks pretty ugly though not your fault. Maybe is better when we change the entire command line from 
``--patch PATCH1 PATCH2 PATCH3``

to ``--patch-PATCH1`` e.g. ``--patch-maniac``. Then ``--patch-direct-menu 123`` would work :thinking: 

But this can be changed later...

src/game_interpreter_map.cpp
```diff
@@ -636,17 +718,33 @@ bool Game_Interpreter_Map::CommandOpenSaveMenu(lcf::rpg::EventCommand const& /*
...
 		return false;
 	}
 
-	Scene::instance->SetRequestedScene(std::make_shared<Scene_Menu>());
-	++index;
-	return false;
+	int subscreen_id = -1, actor_index = 0;
+	bool is_db_actor = false;
+
+	if (com.parameters.size() > 0) {
```
Ghabry, line 732: are these additional event command parameters part of the direct menu patch? In Makerpendium I only see the mention of variables 3326 and 3327.

florianessl: As stated in the initial post, this part is in addition to the original DirectMenu patch. I guess the proper way would be to add new custom command codes for anything that goes beyond known patches.

Ghabry: right. shouldn't comment so late at night :sweat_smile:

Ghabry:
I will redo the API of these scenes to take an array of actors instead. Your solution works but lets improve it while we are at it (these scenes are around 10 years old, everything a bit dated ^^)

Ghabry:
Rebased and updated the API...

About your extension to the Open Menu event command: I think this should get "ValueOrVariable" support so you can do indirection through it. (otherwise it has less functionality than the patch).

Ghabry:
I will remove this part for now to make the PR-merge ready.

```cpp
	if (Player::HasEasyRpgExtension() && com.parameters.size() > 1) {
		subscreen_id = com.parameters[0];

		if (com.parameters.size() > 1) {
			actor_index = com.parameters[1];
		}
		if (com.parameters.size() > 2 && com.parameters[2]) {
			is_db_actor = true;
		}
	}
```

I'm in favour of this but imo it needs some extensions for the full potential:

- Use ``ValueOrVariable``.
- Instead of a single Actor allow the parameter list: ``actor_mode, size (n), actor_id1, actor_id2, ..., actor_idn``. This is possible now after the API change.

Ghabry (approved):

Ghabry:
Honestly this redirection of the menu invocation is so useful that it must become a real feature later.

carstene1ns (approved):

PR #3201, opened 2024-03-29
CMake: Add WiiU preset

Ghabry:
btw the presets do not really work currently for our CI:

usually we do ``-DCMAKE_PREFIX_PATH="$TOOLCHAIN_DIR;$LIBLCF_DIR"`` but the preset only specifies ``"CMAKE_PREFIX_PATH": "$env{EASYRPG_BUILDSCRIPTS}/wiiu"``. So no way to provide liblcf.

How to solve this? Just add another env-var?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,9 @@ include(PlayerBuildType)
 include(PlayerMisc)
 include(GetGitRevisionDescription)
 
+# Dependencies provided by CMake Presets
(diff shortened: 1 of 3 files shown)


Ghabry:
okay, solved. ``-DCMAKE_PREFIX_PATH`` will still work, the presets inject their path now through an additional variable.

fdelapena (approved):

carstene1ns (approved):

PR #3203, opened 2024-04-01
Support "going up" from the working directory in the game browser

Ghabry:
Currently when starting the Player and the game browser launches you cannot go up from the working directory.

This is now working for the following cases (first commit):

1. Project path is empty (working directory is used)
2. Project Path is an absolute path.
3. When Project path is not absolute the _going up_ will end when the relative path ends. (fixing this edge-case is annoying, so I just gave up for now 🤷 )

No there is no special code for these 3 cases, the implementation just only works correctly for 1 and 2.

----

A bit more involved (second commit): When you can go fully up you will reach the drive letters on Windows.

I implemented this using a new virtual filesystem class ``DriveFilesystem`` and a new filetype ``FilesystemNode`` (which can returns an arbitrary VFS when selected). This filesystem is the parent of ``NativeFilesystem`` so you _reach_ it when selecting ``..`` at the highest location (``c:/``).

It is not attached as a parent when no drives are specified (= disabled under !Windows)

In theory this could be also hardcoded in ``NativeFilesystem`` (when going up from ``C:\`` or ``/`` return a drive list) but we make some assumptions about how the _root_ has to look (e.g. on the 3DS ``romfs:`` is not a valid directory but ``romfs:/`` is. Therefore I decided to add an additional layer here just to not break any of our existing code.

This ``DriveFilesystem`` is flexible enough to also support more _drives_ on other platforms later, implemented is only for Windows as here it is the most useful.

screenshot_2
screenshot_1

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,6 +120,8 @@ add_library(${PROJECT_NAME} OBJECT
 	src/fileext_guesser.h
 	src/filesystem.cpp
 	src/filesystem.h
+	src/filesystem_drive.cpp
+	src/filesystem_drive.h	
 	src/filesystem_lzh.cpp
 	src/filesystem_lzh.h
 	src/filesystem_native.cpp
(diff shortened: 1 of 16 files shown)


Ghabry:
I have to patch this out but it is currently possible to traverse the fake filesystem of emscripten: https://easyrpg.org/play/pr3203/?game=ageag  :D

ghost (approved):

PR #3208, opened 2024-04-12
CommandTeleport (10810) - Allow com.string as map filename

jetrotal:
image

You can teleport to a custom-named map, by filling `com.string`:
```js
@raw 10810, "./Map/test", 0,0,0,3
```
It Even considers custom folder structure.

It still lacks some changes to support:
- .po files from translations
- Manipulate save data.

I have to learn how to implement those.

### DIFF
--- a/src/game_interpreter_map.cpp
+++ b/src/game_interpreter_map.cpp
@@ -475,6 +475,8 @@ bool Game_Interpreter_Map::CommandEnterHeroName(lcf::rpg::EventCommand const& co
 
 bool Game_Interpreter_Map::CommandTeleport(lcf::rpg::EventCommand const& com) { // Code 10810
 																		   // TODO: if in battle return true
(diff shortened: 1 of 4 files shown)


Ghabry:
This is a bit more complex than just this command:

- There is an event to save/recall locations
- There is an event to store escape targets and teleport targets

They must be all updated to support this (+ save data).

Look for ``ReserveTeleport`` to find invocations:

https://github.com/search?q=repo%3AEasyRPG%2FPlayer%20ReserveTeleport&type=code

jetrotal:
> This is a bit more complex than just this command:


I put the filename logic inside of ConstructMapName, since multiple functions related to teleport went through it...
That was what made the command work 🤔

jetrotal:
ok, about stringVars: It's good to support it for feature completion. 
I guess none of the teleport commands uses `com.string`. So, just filling this argument may be enough.

Converting ID entries to 0 if com.string exists seems fair too.

PR #3212, opened 2024-04-20
StringVariable's ToFile - Relax File extension and path

jetrotal:
When saving a file, maniacs forces the File in Text/ folder with .txt extension. 
Through this PR, if '*' is at the end of filename, the file extension is relaxed.

### DIFF
--- a/src/game_strings.cpp
+++ b/src/game_strings.cpp
@@ -183,19 +183,27 @@ StringView Game_Strings::ToFile(Str_Params params, std::string filename, int enc
 	}
 
 	// Maniacs forces the File in Text/ folder with .txt extension
-	// TODO: Maybe relax this?
-	filename = "Text/" + filename + ".txt";
+	filename = "Text/" + filename;
+	
+	// EasyRPG Extension: When "*" is at the end of filename, ".txt" is not appended
+	if (filename.back() == '*') {
+		filename.pop_back();
+	} else {
+		filename += ".txt";
+	}
 
 	auto txt_out = FileFinder::Save().OpenOutputStream(filename);
+	auto txt_dir = FileFinder::GetPathAndFilename(filename).first;
+
 	if (!txt_out) {
-		if (!FileFinder::Save().MakeDirectory("Text", false)) {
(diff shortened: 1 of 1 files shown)


carstene1ns:
Does this open an issue with path traversal: `../../../some/file`?

jetrotal:
I guess not, the default behavior allows leaving the `/text/file.txt` by using `../charset/file.txt` as a file name.

my code, simply moves the text file from `/text/` to the root of the game whenver the filename ends with `*`

Ghabry:
In an ideal world where Maniac Patch and TPC are open source and not obfuscated I would suggest to add a new flag to indicate "absolute path". _End of rant_

-----

This is a bit opinionated but I do not like the ``*``. The typical convention for an _absolute path_ is to prefix it with ``/``. You should be even able to forward such a path directly to ``OpenOutputStream`` without removing the ``slash`` manually.

Can you add a comment that this is an EasyRPG Extension?

(Also please remove trivial comments like ``// Remove the last character``. Makes the code look less ChatGPT-generated.)

src/game_strings.cpp
```diff
@@ -183,13 +183,17 @@ StringView Game_Strings::ToFile(Str_Params params, std::string filename, int enc
 	}
 
 	// Maniacs forces the File in Text/ folder with .txt extension
-	// TODO: Maybe relax this?
-	filename = "Text/" + filename + ".txt";
+	filename = "Text/" + filename;
+	// If "*" is at the end of filename, the file path and extension are relaxed - FEATURE EXCLUSIVE TO EASYRPG
+	if (filename.back() == '*') filename.pop_back();
```
Ghabry, line 188: does this really work as intended? This only relaxes the extension, not the path (still prefixed with ``Text/``.

jetrotal: Oops, you are correct. 
I abandonned the relaxing path idea. 
It would just confuse people, since they can do `/../`to nav out of the text/ folder.

Ghabry (approved):

PR #3214, opened 2024-04-28
New Command 2054 - Get ID By Name

jetrotal:
This command allows you to store the ID of a database asset by searching for its name.

Syntax (TPC):
```js
@raw 2054,
      "",          // Name to search for: "MAP0001", "MyVar", etc.
      0, 0,        // Type of asset (0-18): Actor, Skill, Item, Enemy, Troop, Terrain, Attribute, State, Animation, Tileset, Switch, Variable, String Variable, Common Event, Class, Battler Animation, Map, Map Event, Party Member
      0, 5,        // Variable where the ID will be stored
      0, 1,        // Min range: "0" sets it to the start of the list of assets, any other value specifies the minimum index
      0, 100,      // Max range: "0" sets it to the end of the list of assets, any other value specifies the maximum index
      1, 1         // Usage of string variable instead of the search string, through StringOrVariable()
```

@Ghabry and @carstene1ns commented about hashing the data for better searching and retrieving data by name.
 https://github.com/greg7mdp/parallel-hashmap 
 
 
image

But this may look good enough for a first version (it's fast enough with the 9999 variables limit from vanilla).


Ghabry:
How do you plan to handle translatable strings? Some of them will require a search via the translated string when a translation is loaded and others are not translatable (not exposed by lcftrans).

A string translation must happen here and we actually have a problem: We rewrite all strings on load. Here the string has different meanings based on the context so is the first time where this cannot be used. 🤔

jetrotal:
I forgot about those... Would be hard to have an extra entry called `defaultName` that only exists when dealing with a translated game?

Ghabry:
There is no simple way to get the original because we really overwrite them completely.

You must translate your string and do the lookup. Have to check if we have an API already for this...

Ghabry:
Yes this is solvable, just needs a few more Getters.

What is also interesting (?) is that the command can use String Vars and String Vars are only active when Maniac Patch is on.

I will encounter the same issue soon with my "Language Config" event command which allows altering the active translation and... needs string vars for the getter part :sweat_smile:

jetrotal:
question: would be better to move here  the code from this pr: https://github.com/EasyRPG/Player/pull/3124 ?

Seems better fitted than editing control variables.
also thought about including stuff  like screen size, map size and any other unavaliable data.

PR #3218, opened 2024-05-07
New Command 2056/2057 - Clone/Destroy Map Event

jetrotal:
Command that clones Events from any map inside the current map.

------------------------

#### Syntax (TPC):
```js
@raw 2056,
     "",           // Rename the Target Event [Optional]
      0, 1,        // Source Map's ID.
      0, 2,        // Source Event's ID
      0, 3,        // Target's x coordinate
      0, 4,        // Target's y coordinate
      0, 5,        // Target's Event's ID [Optional] 
                   // 👆 if a value is defined, it will replace an existing event.
```

### DIFF
--- a/src/game_character.h
+++ b/src/game_character.h
@@ -1379,7 +1379,6 @@ inline Game_CharacterDataStorage<T>::Game_CharacterDataStorage(Game_CharacterDat
 template <typename T>
 inline Game_CharacterDataStorage<T>& Game_CharacterDataStorage<T>::operator=(Game_CharacterDataStorage&& o) noexcept
 {
-	static_cast<Game_Character*>(this) = std::move(o);
 	if (this != &o) {
 		_data = std::move(o._data);
 		Game_Character::_data = &this->_data;
--- a/src/game_character.h
+++ b/src/game_character.h
@@ -1379,6 +1379,9 @@ inline Game_CharacterDataStorage<T>::Game_CharacterDataStorage(Game_CharacterDat
 template <typename T>
 inline Game_CharacterDataStorage<T>& Game_CharacterDataStorage<T>::operator=(Game_CharacterDataStorage&& o) noexcept
 {
+	auto* base = static_cast<Game_Character*>(this);
+	*base = std::move(o);
+
 	if (this != &o) {
 		_data = std::move(o._data);
 		Game_Character::_data = &this->_data;
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -810,10 +810,7 @@ bool Game_Interpreter::OnFinishStackFrame() {
 	if (is_base_frame && event_id > 0) {
 		Game_Event* evnt = Game_Map::GetEvent(event_id);
 		if (!evnt) {
-			if (!Player::HasEasyRpgExtensions()) {
-				// Destroy Map Event triggers this sanity check
-				Output::Error("Call stack finished with invalid event id {}. This can be caused by a vehicle teleport?", event_id);
-			}
+			Output::Error("Call stack finished with invalid event id {}. This can be caused by a vehicle teleport?", event_id);
 		} else if (main_flag) {
 			evnt->OnFinishForegroundEvent();
 		}
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -787,8 +787,8 @@ bool Game_Interpreter::ExecuteCommand(lcf::rpg::EventCommand const& com) {
 			return CommandManiacCallCommand(com);
 		case Cmd::EasyRpg_SetInterpreterFlag:
 			return CommandEasyRpgSetInterpreterFlag(com);
-		case static_cast<Cmd>(2056): //EasyRPG_SpawnMapEvent
-			return CommandSpawnMapEvent(com);
+		case static_cast<Cmd>(2056): //EasyRPG_CloneMapEvent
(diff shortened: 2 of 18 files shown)


Ghabry:
This command looks pretty straight forward and simple to review.

The only optimisation I see is that it shouldn't fetch and retranslate the map again when the map ID is the same as the active map. Maybe also consider "map 0" as the current map? (otherwise you need an additional event call to grab the ID).

And maybe a small test game would be good that spawns an event in two ways:

- From the current map
- From a different map

Because I have to add the Web Player Async stuff, so I have directly something for testing...

jetrotal:
Ok, later tonight I'll poke around to fit map0 and mapItself cases.
Would be useful to also have a "destroy Event" command.

Ghabry:
This delete event should be a separate command.

It is also much harder than spawn because the semantics are unclear: what is supposed to happen with the interpreter execution (e.g. when having a parallel process running) of the deleted event when:

- event deletes itself
- event deletes other event

jetrotal:
Project for testing, both from map itself and outside map:
image

[SpawnEventProject.zip](https://github.com/EasyRPG/Player/files/15255695/SpawnEventProject.zip)

jetrotal:
Rebased.

Ghabry:
Added async suspension to CloneMapEvent and (NEW) DestoryMapEvent. Is the same technique we already use for ManiacSave/Load implementation to prevent that inconsistent data is saved.

---

There is still commented out code because it will conflict with a more complex change by #3233 so I didn't wanted to touch this.

---

The semantics for Delete (and Overwrite) are:

When it is _NOT_ a parallel process (= executes on the main interpreter): The deleted event will execute till the end. (this is the same as for Erase Event).

When it is a parallel process: The event execution will end immediately (even when currently inside a common event call) because the interpreter that runs it is destroyed together with the event. (this differs from Erase Event where only the lowest stack frame will end immediately and the rest will happily continue)

src/game_character.h
```diff
@@ -1343,7 +1343,6 @@ inline Game_CharacterDataStorage<T>::Game_CharacterDataStorage(Game_CharacterDat
 template <typename T>
 inline Game_CharacterDataStorage<T>& Game_CharacterDataStorage<T>::operator=(Game_CharacterDataStorage&& o) noexcept
 {
-	static_cast<Game_Character*>(this) = std::move(o);
```
fmatthew5876, line 1346: is that a bugfix?

Ghabry: Oh good point. I wanted to revisit that line... should have added a FIXME instead of just deleting it.

That line fails to build with "the left operand must be an l-value" and I couldn't figure out how to solve it (yet).

src/game_interpreter.cpp
```diff
@@ -847,7 +851,10 @@ bool Game_Interpreter::OnFinishStackFrame() {
 	if (is_base_frame && event_id > 0) {
 		Game_Event* evnt = Game_Map::GetEvent(event_id);
 		if (!evnt) {
-			Output::Error("Call stack finished with invalid event id {}. This can be caused by a vehicle teleport?", event_id);
+			if (!Player::HasEasyRpgExtensions()) {
+				// Destroy Map Event triggers this sanity check
+				Output::Error("Call stack finished with invalid event id {}. This can be caused by a vehicle teleport?", event_id);
+			}
```
fmatthew5876, line 857: Is there a way to maintain this check for all cases except for when the destroymapevent case triggers it?

src/game_character.h
Ghabry: this works. thx

src/game_interpreter.cpp
Ghabry: yes. I set now the base frame event_id on the foreground interpreter to 0 (when it is the deleted event).

Ghabry:
Battle animations can also contain (stale) references to events as noticed in #3255

Updating these references (or cancelling the animation when the event is deleted) must be also handled by the event command.

Ghabry:
Battle Animations are now correctly repointed (event overwritten) or cancelled (event deleted).

fdelapena (approved):

src/game_map.cpp
```diff
@@ -342,42 +342,183 @@ std::unique_ptr<lcf::rpg::Map> Game_Map::loadMapFile(int map_id) {
...
+
+	const lcf::rpg::Event* source_event = FindEventById(source_map->events, src_event_id);
+	if (source_event == nullptr) {
+		Output::Warning("CloneMapEvent: Event ID {} not found on source map {}", src_event_id, src_map_id);
+		return false;
+	}
+
+	lcf::rpg::Event new_event = *source_event;
+	if (target_event_id > 0) {
+		DestroyMapEvent(target_event_id, false);
```
lumiscosity, line 386: this should only run if the target id exists; right now spawning an event to an empty slot makes a wanring appear about the destroy command being called on a nonexistent event

lumiscosity: not sure how to do the code suggestion thing, but i got it working by reusing the check from destroymapevent. that seems pretty inefficient though; what about passing a silent failure flag of some sort?

jetrotal: A simple way of making it less verbose on the screen, 
is using a debug message instead of a warning. So it only appears on the log.

I let a warning due to letting people know if they did something wrong with their events, but dynamic event importers/destroyers may become problematic with those text popups

PR #3219, opened 2024-05-08
Debug Menu extensions: Display strings & interpreter stacks

florianessl:
Adds two new items to the F9 Debug Menu:
- "Strings"
- "Interpreter"

String display supports viewing multiline strings through a special new window "Window_StringView"
Two options can be toggled:
- Automatic linebreaks
- Command Evaluation ( default inserter with \v, \n, \t )

screenshot_160

Viewing the current interpreter stacks
(right side: [current command] / [commands_size])
screenshot_161

### DIFF
--- a/src/window_interpreter.cpp
+++ b/src/window_interpreter.cpp
@@ -45,10 +45,9 @@ Window_Interpreter::~Window_Interpreter() {
 
 }
 
-void Window_Interpreter::SetStackState(bool is_ce, std::string interpreter_desc, lcf::rpg::SaveEventExecState state) {
-	this->interpreter_desc = interpreter_desc;
+void Window_Interpreter::SetStackState(bool is_ce, int owner_evt_id, std::string interpreter_desc, lcf::rpg::SaveEventExecState state) {
+	this->display_item = { is_ce, owner_evt_id, interpreter_desc };
 	this->state = state;
-	this->is_ce = is_ce;
 }
 
 void Window_Interpreter::Refresh() {
@@ -73,12 +72,14 @@ void Window_Interpreter::Refresh() {
 			auto& prev_frame = state.stack[i - 1];
 			auto& com = prev_frame.commands[prev_frame.current_command - 1];
 			if (com.code == 12330) { // CallEvent
-				is_calling_ev_ce = true;
 				if (com.parameters[0] == 0) {
+					is_calling_ev_ce = true;
 					evt_id = com.parameters[1];
 				} else if (com.parameters[0] == 3 && Player::IsPatchManiac()) {
+					is_calling_ev_ce = true;
 					evt_id = Main_Data::game_variables->Get(com.parameters[1]);
 				} else if (com.parameters[0] == 4 && Player::IsPatchManiac()) {
+					is_calling_ev_ce = true;
 					evt_id = Main_Data::game_variables->GetIndirect(com.parameters[1]);
 				}
 			}
@@ -90,7 +91,7 @@ void Window_Interpreter::Refresh() {
 			max_page_id = page_id;
 
 		StackItem item = StackItem();
-		item.is_ce = is_calling_ev_ce || (i == 0 && this->is_ce);
+		item.is_ce = is_calling_ev_ce || (i == 0 && this->display_item.is_ce);
 		item.evt_id = evt_id;
 		item.page_id = page_id;
 		item.name = "";
@@ -147,7 +148,7 @@ void Window_Interpreter::Refresh() {
 }
 
 bool Window_Interpreter::IsValid() {
-	return !interpreter_desc.empty();
+	return !display_item.desc.empty();
 }
 
 void Window_Interpreter::Update() {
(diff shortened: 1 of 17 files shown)


Ghabry:
For the Linux build you must update the Makefile.am file similar to CMakeLists

Ghabry:
Haven't read the entire code yet obviously. Just one thing I noticed: 

You wrote your own wrap code. We already have a word wrapping function Game_Message::WordWrap. Maybe use this one if applicable?

florianessl:
> Haven't read the entire code yet obviously. Just one thing I noticed:
> 
> You wrote your own wrap code. We already have a word wrapping function Game_Message::WordWrap. Maybe use this one if applicable?

I switched the code, so it uses this WordWrapping function now, instead of just cutting of at the exact position a text would overflow.

florianessl:
Rebased & refactored a bit. The ScopedVars branch depends on this, to make it possible to view & edit all frame-local variables for individual interpreter frames. I already moved some of the new code here (New choices window for debug scene) to make things easier to merge.

Considering the necessary changes for the debugging possibilities of the ScopedVars branch, this extension should be feature-complete, so feel free to test this!

Ghabry:
Cool. We do not have too many games that use String Vars online.

This test game here could be suitable: https://easyrpg.org/play/pr3219/?game=string-var&test-play

I think I found a bug: In this game when I go to Interpreter -> Main no cursor appears.

However when I try e.g. uid https://easyrpg.org/play/pr3219/?game=uid&test-play  it works. So I guess only having one interpreter bugs the Ui here?

Ghabry:
The issue is still not completely fixed: When you go in the string-var game to Strings -> Main the cursor is broken the first time. After Cancel -> Decision again it works :shrug:

src/window_interpreter.cpp
```diff
@@ -0,0 +1,181 @@
...
+			auto& prev_frame = state.stack[i - 1];
+			auto& com = prev_frame.commands[prev_frame.current_command - 1];
+			if (com.code == 12330) { // CallEvent
+				is_calling_ev_ce = true;
+				if (com.parameters[0] == 0) {
+					evt_id = com.parameters[1];
+				} else if (com.parameters[0] == 3 && Player::IsPatchManiac()) {
+					evt_id = Main_Data::game_variables->Get(com.parameters[1]);
+				} else if (com.parameters[0] == 4 && Player::IsPatchManiac()) {
+					evt_id = Main_Data::game_variables->GetIndirect(com.parameters[1]);
```
Ghabry, line 73: Are you attempting to figure out the event ID here?

Actually Maniac Patch added some debug fields to the ExexFrame for this purpose: ``maniac_event_info``, ``maniac_event_id``, ``maniac_event_page_id``. We do not have this implemented and you do not have to implement everything but ``maniac_event_id`` could be useful to track the ID across maps because its not reset to 0.

Ghabry (commented):
My bar for Debug Scene changes is quite low as it cannot break code and is - well - for debugging. Imo that code can be merged as is after the two issues are discussed/resolved.

src/window_interpreter.cpp
florianessl: Yeah, that's the idea..., I've noticed that there are several unimplemented fields, but skipped over them because I don't have any implementation details and didn't want to invest the time right now to figure this out. 😬
So this is far from perfect right now.
But yeah, I should probably change the tracked id field there. Completely forgot that the vanilla event_id gets reset.
I'm on holiday right now and won't have much time to look over this. If you want to change something, feel free to do so.

Ghabry: No Problem. Is nothing urgent Is enough when you annotate it with a fixme for now.

florianessl:
I added some "FIXME" comments for stuff that might need some minor rework, fixed some minor display issues & added support for the maniac fields "maniac_event_id" & "maniac_event_page_id".
I couldn't reproduce any more cursor issues with the strings example game though

Here's some example of a parallel process map event calling a CE, calling another map event, calling another CE...
screenshot_165
(Had to highlight this one, because this is some really unsafe stuff were it might be useful for the engine to warn the developer if such bounds are crossed -> on my list...)

Ghabry (approved):
Besides this one nit: LGTM :+1:

src/window_interpreter.cpp
```diff
@@ -54,12 +54,21 @@ void Window_Interpreter::SetStackState(bool is_ce, std::string interpreter_desc,
 void Window_Interpreter::Refresh() {
 	stack_display_items.clear();
 
-	int max_cmd_count = 0;
+	int max_cmd_count = 0, max_evt_id = 10, max_page_id = 0;
 
 	for (int i = state.stack.size() - 1; i >= 0; i--) {
 		int evt_id = state.stack[i].event_id;
+		int page_id = 0;
+		if ((Player::IsPatchManiac() || Player::HasEasyRpgExtensions()) && state.stack[i].maniac_event_id > 0) {
```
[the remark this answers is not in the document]
florianessl: Yup, guessed the same but I stuck with the strict approach. :D 
-> Will fix

PR #3220, opened 2024-05-09
Variable Substitution: Options for enabling substution in command strings & for applying {fmt}-style number formatting

florianessl:
This PR enables two new features:
- Variable substitution in interpreter command strings (default CommandCodeInserter is used -> \v, \n, \t)
- Number formatting using {fmt} 
  just write the format specification after a ':'
  the parser will ignore anything that is not valid fmt syntax (also decimal types & 'precision' are not supported)

These two features require setting new liblcfs fields which are specified as follows:

```
System,easyrpg_var_substitution_in_commands,f,Boolean,0xDC,False,0,0,Enables substitution of '\v' '\n' '\a' inside string parameters of interpreter commands
System,easyrpg_var_substitution_formatting,f,Boolean,0xDD,False,0,0,Enables '{fmt}'-style format specifiers for usage inside variable substitution
```

### Example: Graphics customization
Consider you want to dynamically change an actor graphic or faceset depending on some settings.

With this branch you can just use any pre-existing command like this:

`com.string: "Actor\v[123:02d]"
`

... which evaluates to "Actor01.png", "Actor02.png", ..., "Actor[xx].png" depending on the value of Variable '123'

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -1786,7 +1786,11 @@ int Game_Interpreter::ValueOrVariableBitfield(lcf::rpg::EventCommand const& com,
 }
 
 std::string Game_Interpreter::CommandString(lcf::rpg::EventCommand const& com) {
+#ifdef LIBLCF_STUB_COMSTRING_VARSUBSTITUTION
+	if (!Player::HasEasyRpgExtensions()) {
+#else
 	if (!Player::HasEasyRpgExtensions() || !lcf::Data::system.easyrpg_var_substitution_in_commands) {
+#endif
 		return ToString(com.string);
 	}
 	return PendingMessage::ApplyTextInsertingCommands(ToString(com.string), Player::escape_char, Game_Message::CommandCodeInserter);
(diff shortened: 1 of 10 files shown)


florianessl:
Oh, haven't thought about the reference not working there. Stupid me.
I'l look into how I can improve this without any unnecessary copying, and unit-tests for the formatting were definitely on my mind, I'm just working on several branches right now and  am constantly shifting my focus :D

I'm working on an extensive test suite for my three main branches, which hopefully will cover all sorts of use cases:
"ScopesVars", "MapInitTrigger" & this one. So, no need to go over any of those while this is still a work-in-progress (also, I'm on holiday for two weeks, so it will take some time.)

jetrotal:
1 request + 1 observation: 
- Also allow stringVars on that
- Are there cases where `\` is used to navigate between folders, like `..\Picture\enemies\slime`? would that break anything?

florianessl:
Closed, but will rework some of these features & maybe re-submit at a later time.

PR #3221, opened 2024-05-10
New command "Wait for Single Movement"

Ghabry:
Originally by @jetrotal 

Redid the stuck check.

Also made the "waiting" similiar to how other events do it: E.g. "Wait for Key" will store all the keys it wants and no reread them from the event everytime. This is necessary because the values of variables can change. Storing them on first call avoids this issue.

### DIFF
--- a/src/game_character.cpp
+++ b/src/game_character.cpp
@@ -301,6 +301,7 @@ void Game_Character::UpdateMoveRoute(int32_t& current_index, const lcf::rpg::Mov
 					SetDirection(prev_direction);
 					SetFacing(prev_facing);
 				} else {
+					SetMoveFailureCount(GetMoveFailureCount() + 1);
 					return;
 				}
 			}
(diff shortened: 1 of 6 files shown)


fdelapena (approved):

jetrotal:
this looks good, but is it ok to aprove it while it fails it checks
?

Ghabry:
okay lets continue a bit after a long break...

This command is just great.

Tested it now in detail and it works after merging https://github.com/EasyRPG/liblcf/pull/478

---

Tested with a move failure count of 200 which correctly terminates and writes a 0 in the target variable. Also tested saving and loading. Is correctly restored.

---

Here is a test map that is compatible with "Maniac TestGame" recently merged (but should be its own "TestGame" for EasyRpg exclusive stuff). That work isn't finished so just the map for now.

[Map0017.zip](https://github.com/user-attachments/files/16447630/Map0017.zip)

fdelapena:
jenkins: test this please

PR #3228, opened 2024-05-20
Add nlohmann json, remove picojson

Ghabry:
Is enabled for all platforms because this will be used by a JSON event command (#3215)

Used the opportunity to get rid of picojson :) (one less vendored dep, only dr_wav left xD)


CMakeLists.txt
```diff
@@ -884,6 +883,23 @@ player_find_package(NAME lhasa
 	TARGET LHASA::liblhasa
 )
 
+# json support
+option(PLAYER_WITH_NLOHMANN_JSON "Support processing of JSON files" ON)
```
carstene1ns, line 887: Maybe move the condition to the option instead of package finding? Is not intuitive when the option has no effect.
-> something with set cache force for emscripten

Ghabry:
Quick size benchmark: Moving to nlohmann_json increased the binary size of the WASM file by 34 KB. (at least for the stripped release WASM file, for the debug file it is +1 MB :cold_sweat: )

carstene1ns (approved):
🏙️

PR #3231, opened 2024-05-27
Regression fix of the Scene refactor: Battle skills were not listed correctly

Ghabry:
also cleaned up some compiler warnings (though this concrete weird C++ behaviour was not warned about -_-)


carstene1ns (approved):
Looks good to me.

carstene1ns:
These promotion rules only apply for const references here, would be catched for non-const. This kind of magic is the reason we have `explicit` constructors since c++11....

Issue #1919, opened 2019-10-04
Optimize interpreter with computer gotos


fmatthew5876:
This is a non standard language feature implemented on all compilers except Microsoft. 

- [ ] Merge all 3 `ExecuteCommand` methods into 1
- [ ] Reimplement the switch using computed go-to and measure the performance gain.
- [ ] Provide a fallback for MSVC

https://youtu.be/IAdLwUXRUvg

The speaker claims a 10-20% speed in improvement in Lua and Python interpreter's using this technique.

This could result in 10-40% speedup in Player interpreter execution.

PR #3233, opened 2024-05-31
Interpreter Refactor

florianessl:
- Refactors & restructures some of the logic of my "AntiLag" branch into a new "Caching" namespace.
- Moves some of the common helper function from the interpreter class to a new file "game_interpreter_shared.cpp"
- Adds a new abstract base class for the interpreter: "Game_BaseInterpreterContext"
  -  This class holds the declarations for a few member functions which are used for evaluation logic across different sorts of commands
  - All the moved helper functions (CheckOperator, ValueOrVariable, CommandStringOrVariable & others) are agnostic to the actual implementation of the interpreter itself
- Refactors "ValueOrVariable" & "ValueOrVariableBitfield" into template methods
  - Might save some unecessary branching depending on the context in which they are used

New helper function: "DecodeTargetEvaluationMode"
- Unifies some of the commonly seen logic that was responsible for decoding command arguments into either a single variable id or a set range of variables.
- Used in ControlSwitches, ControlVariables & ControlStrings with different template parameters (these specify the supported variable accessing modes). E.g.: A hypothethical new "ControlVariablesEx" command could just set all of these parameters to true and share the rest of its logic with the original command.
- Used in my ScopedVars branch for these new commands:
  - ControlSwitchesEx
  - ControlVariablesEx
  - ControlScopedSwitches
  - ControlScopedVariables 

This should be most of the easier-to-verify-and-merge changes, my "ScopedVars" depends on.

The current draft for ScopedVars also contains some more changes for the "ControlVariables" & "ConditionalBranch" commands, moving all of the operand switching logic into game_interpreter_control_variables.cpp. Basically I reinvented dispatching tables for this, to share this logic between the different new command spin-offs, but I have yet to confirm with benchmarks that this really a good solution. I don't want to slow any of the existing commands down by even a single cycle. :D  
https://github.com/florianessl/Player/tree/ScopedVars
Best possible way to implement all this switching would likely be a mix of computed go-tos & tail recursion (see #1919 for the suggestion of implementing computed go-tos; but it´s debatable, if any of these changes would create any speed-ups in current-year & aren't already part of common compiler optimizations.)

### DIFF
--- a/src/game_map.h
+++ b/src/game_map.h
@@ -74,14 +74,6 @@ private:
 	bool message = false;
 };
 
-class MapEventCache {
-public:
-	void AddEvent(lcf::rpg::Event& ev);
-
-private:
-	std::vector<int> event_ids;
-};
-
 /**
  * Game_Map namespace
  */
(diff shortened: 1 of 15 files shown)


Ghabry:
Nice refactor / cleanup! Besides ``Refactored several Interpreter related code segments`` this does not look too controversial. Reviewing that one commit will take a while but on first glance I see no obvious issues.

Ghabry:
@florianessl could you attempt to do a rebase? There is one conflict in game_interpreter.

Ghabry (approved):

Ghabry:
Would be good to get this one reviewed and in before the event commands because of conflicts.

Will run our test game through this to find bugs...

src/game_map.h
```diff
@@ -688,13 +680,43 @@ namespace Game_Map {
...
+		public:
+			void AddEvent(lcf::rpg::Event& ev);
+
+		private:
+			std::vector<int> event_ids;
+		};
+
+		using MapEventCacheData_t = std::unordered_map<int, MapEventCache>;
+
+		enum ObservervedVarOps {
```
fdelapena, line 694: Is `Observerved` a typo?

fdelapena (approved):

PR #3234, opened 2024-06-02
WiiU Fixes

Ghabry:
Still lacks polish like e.g. a nice banner but at least audio does not assert anymore and saving of the config works.

----

About this weird teardown in player.cpp (can we do better?)

SDL2 has a ProcUI handle which refers to the current application and for a proper shutdown of the app, SDL2 listens to PROCUI_STATUS_EXITING and then does some further cleanup.

When using the Exit menu item this will not trigger the cleanup automatically because SDL will not receive the shutdown event (but it receives it when using "WiiU Menu" in the home menu).

To make it receive the event I call ``SYSLaunchMenu()``, then pump events once more (via that recursive call, SDL gets the quit event now) and then everything works...

https://github.com/devkitPro/SDL/blob/7b058bd042a7c624c277aaddbca6fb2f6ae5233d/src/video/wiiu/SDL_wiiuvideo.c#L326-L338


carstene1ns:
The `SysLaunchMenu()' call is a good workaround, using this now as well:
https://github.com/carstene1ns/easyrpg-player/blob/956eb50afcb5cbd624ad94d237d2bfa9b4405bed/src/platform/wiiu/main.cpp#L121-L131

And here is the rest of the WIP branch:
https://github.com/EasyRPG/Player/compare/master...carstene1ns:easyrpg-player:feature/wiiu

You can use this network log receiver for debugging: https://gist.github.com/carstene1ns/e38fa10351449fee25cf2cc339c8e043

Ghabry:
woah that UDP logger is awesome. Makes my life much easier xD

Ghabry:
Your code terminates fine when using the exit option but exiting from the WiiU menu crashes in most cases :/

Haven't found a working solution yet.

---

My ugly recursive call on exit is working in both cases

Ghabry:
Shutting down works now properly in all three cases:

- Home Menu
- Player picks Exit
- Output::Error

The only remaining problem is extremly slow opening of ZIP archives, likely some IO buffering issue. But is not a _blocker_ for me.

----   

Had to replace these WHB wrapper functions with calls to ProcUI which is the low level API.

SDL2 will not handle it when it is initialized before launching SDL2. (see code in wiiu_video.c)

Problem is that you cannot query the ProcUI status after it was shutdown (calling the function will crash) so I needed more low level access to differentiate between "Home Menu used" and "Exit option selected (or Output::Error)".

(This also makes ``WHBProcIsRunning`` crash when the shut down was requested, making this function pretty useless for checking whether it is running xD)

---

btw it seems USB Keyboards are not initialized by SDL2/WUT, so cannot use it :(  (used it to input the Wifi password before so should work in theory...)

Ghabry:
IMG_20240603_143132

IMG_20240603_143145

IMG_20240603_143147

resources/wiiu/splash-drc.png
carstene1ns, line 1: Calling @jetrotal for a higher quality version of the drc splash, this one has been upscaled.
Also we could drop the black border of the icon...

Ghabry: The Logo Text indeed needs a bigger version.

For the rest:

There is a PSD file in https://easyrpg.org/downloads/logos/ Gimp can load it but is not fun working with it :sweat_smile: 

For the splash this could be used (?) Has the same metrics as the real logo, just higher res.

splash-drc

Or this:

Unbenannt


Icon:

icon

Unfortunately the EasyRPG text is hard to read here as it is too bright (gray Ui on WiiU)

So Icon with text works better:

icon

Ghabry:
The ICU issue is a simple fix :confetti_ball:  :confetti_ball:   :confetti_ball: 

The patch must be moved before ``cp -rup icu icu-native`` as the tool that requires patching is from the native icu.

https://github.com/EasyRPG/buildscripts/blob/master/wiiu/2_build_toolchain.sh#L45

jetrotal:
icon
image

TV
image

PAD
image


The idea is to promote ourcharsets on the joypad screen, by letting a user search for their favorite character by looking at it.


Here's a psd containing editable assets:
[EasyRP_wiiu_PSDs.zip](https://github.com/user-attachments/files/15546242/EasyRP_wiiu_PSDs.zip)

I recomend opening my psd files over the browser editor https://www.photopea.com/
instead of GIMP.

Ghabry:
Thanks!

The downscale made imo Y look weird because on one side it is one pixel shorter.

Quickly edited it:

grafik

Ghabry:
Somehow the buffering doesn't work and "pubsetbuf" does nothing (which is allowed by the standard xD) so I wrote my own buffering file IO.

Verifying whether the filebuf implementation is correct is a bit tricky. I wrote a test code that does hundreds of random seeks using beg/cur/end mode and reads and compared the result with the normal fstream.

Then executed it multiple times until I got no asserts anymore because of mismatches...

Ghabry:
I will enable this custom IO for the other homebrew targets.

Only did an additional test on the Wii and this gives a huge loading time speedup. So likely also helps on Vita, 3DS and Switch.

src/platform/wiiu/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+#ifdef USE_SDL // SDL might wrap main()
+#  include <SDL.h>
+#endif
+
+#include <coreinit/debug.h>
+#include <coreinit/foreground.h>
+#include <proc_ui/procui.h>
+#include <sysapp/launch.h>
+
+using namespace std::chrono_literals;
```
carstene1ns, line 36: This once could go or let it stay and change in `WiiU_Exit()`.

src/platform/wiiu/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+#else
+#include <whb/log.h>
+#include <whb/log_cafe.h>
+#include <whb/log_module.h>
+#include <whb/log_udp.h>
+
+static bool moduleLogInit = false;
+static bool cafeLogInit = false;
+static bool udpLogInit = false;
+static bool prodUiExited = false;
```
carstene1ns, line 51: This needs to be moved above the `#ifdef NDEBUG` in line 38 and should be changed to `procUiExited`.

src/platform/wiiu/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+}
+
+void WiiU_Exit() {
+	Output::Debug("Shutdown Reason: {}", prodUiExited ? "HOME Menu" : "Player Exit");
+
+    if (!prodUiExited) {
+		// Exit was not through the Home Menu
+		// Manually launch the system menu
+		SYSLaunchMenu();
+		Game_Clock::SleepFor(std::chrono::milliseconds(10));
```
carstene1ns, line 110: Could be `10ms`.

src/platform/wiiu/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+	// Check if wuhb has some files inside or not
+	if(::access("fs:/vol/content/RPG_RT.lmt", F_OK) == 0) {
+		Output::Debug("Running packaged game from wuhb.");
+		std::string cafe_dir = "fs:/vol/content";
+
+		args.push_back("--project-path");
+		args.push_back(cafe_dir);
+
+		//args.push_back("--save-path");
```
carstene1ns, line 141: This should have a `// FIXME: shipped game has nowhere to save` 🐧

src/player.cpp
```diff
@@ -219,6 +221,12 @@ void Player::Run() {
 	// libretro invokes the MainLoop through a retro_run-callback
 #else
 	while (Transition::instance().IsActive() || (Scene::instance && Scene::instance->type != Scene::Null)) {
+#if defined(__WIIU__)
```
carstene1ns, line 224: I had these applet lifetime checks moved from here for the 3DS and switch ports, no real need to add a platform dependency here.
See also https://github.com/EasyRPG/Player/commit/c9b1809898bfb687cd2e775710187887ec146729#diff-10c39dd978ada90e0d090a045c4ae5f51dad010f3b754679e15894c9b93d254b

carstene1ns (changes requested):
Apart from a few nits, works really great now!

Ghabry:
Jenkins: test this please

Addressed the review comments and fixed a few more minor bugs. Is ready now :+1:

fdelapena (approved):

PR #3238, opened 2024-06-10
Make our life on GitHub easier:

carstene1ns:
- Automatically [label pull requests](https://github.com/actions/labeler)
- Hilight some warnings in the output
- Keep actions updated

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,98 @@
+#
+# See https://github.com/actions/labeler
+# and workflows/pr_labels.yml for reference
+#
+
+Building:
+- changed-files:
+  - any-glob-to-any-file:
+    - .github/**
+    - CMakeLists.txt
+    - builds/**
+    - '!builds/android/app/**'
+    - Makefile.am
+    - configure.ac
+
+Documentation:
+- changed-files:
+  - any-glob-to-any-file:
+    - docs/**
+    - '**/*.md'
+    - '**/*.adoc'
+    - src/docmain.h
+    - '**/Doxyfile*'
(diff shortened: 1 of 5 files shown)


carstene1ns:
Tested this by merging to my master branch and issuing a pull request:
https://github.com/carstene1ns/easyrpg-player/pull/1
Labels have been applied.

.github/labeler.yml
```diff
@@ -0,0 +1,91 @@
+
+Building:
+- changed-files:
+  - any-glob-to-any-file: [ '.github/**',
+                            'CMakeLists.txt',
+                            'builds/**',
+                            'Makefile.am',
```
Ghabry, line 7: Not sure if CMakeLists and Makefile.am should be listed here as any PR that adds new files will trigger this label.

Ghabry:
I have no idea how this Yaml stuff works and do not want to learn it. So feel free to merge it when you consider it ready xD

.github/labeler.yml
carstene1ns: Technically each added file changes the building process, but I get what you mean.
Unfortunately these labels can only be changed on file basis or branch basis, so only option would be to apply building label manually then.
I think we should just live with some false positives for now and amend the config later, if needed.

carstene1ns:
emote

carstene1ns:
> My guess for the autolabel as Building is `!builds/android/app/**`. This likely matches when any file outside of android is changed 🤔

Oh right, this makes sense. At least is the only negation, so this can happen with the `Building` label.
Will fix in a bit!

PR #3240, opened 2024-06-18
ttyp0 changes (remove Thai and Hebrew), generator fix

### LINKED ISSUE #2999: ttyp0: Remove Thai and Hebrew glyphs from font
ttyp0 contains glyphs for Thai and Hebrew.

As ttyp0 is a BitmapFont our renderer cannot handle anything complex.

Thai requires some kind of basic shaping to put diacritics correctly: This looks incorrect.

Hebrew is a RTL language: The text must be written backwards to give proper rendering (HarfBuzz does this correctly).

To reduce confusion and bad UX is better to just remove these glyphs completely and tell them to use TTF instead.

rohkea:
This PR does the following changes:

* ttyp0: removes scripts that are broken in the current renderer (Thai and Hebrew) — resolve #2999,
* ~~ttyp0: moves the letter F left 1px, to make sure it lines up in menus (requested in the chat),~~
* generate_cxx_font.rb: fixes the folder (fonts are now in `src/generated`, not in `src`)

~~I'm not sure about the F change. The previous version looked better in running text (but worse in menus), the new version looks better in menus (but worse in running text). Here's before and after:~~

Before: F in the menu is to the right of all the other letters, but Fi and Fo combinations look better

After: F in the menu is aligned with other letters, but Fi has too much space between letters


Ghabry:
Thanks for the fixes. Imo the new ``F`` looks worse (as running text is more common than menus). Maybe shift the ``E`` instead one pixel to the right?

rohkea:
Here is version with `E` shifted to the right: 
screenshot-move-both-e-and-f

And here are some examples of the running text.

Original:
fairy-orig

F shifted to the left:
fairy-f-shifted

E shifted to the right:
fairy-e-shifted

Ghabry:
Okay this _triggers_ now all that want the D and the E ``|`` at the same location.

And we cannot move all upper letters by one to the right, this will look bad in other situations.

This doesn't look solvable without redoing the font... So lets just undo this part.

rohkea:
I've removed the change of the F letter, reverted the original version.

rohkea:
I've checked, the PR doesn't change Shinonome (the case when Ruby version influences the final file, because of encoding issue), so it can be merged (I hope)

carstene1ns (approved):
Looks good to me.
Just one question, why is thai `bin` and the rest `bit`, possible typo?

rohkea:
Renamed `.bin` into `.bit`, thanks for noticing!

fdelapena (approved):

PR #3245, opened 2024-07-14
Maniacs Feature - Pixel scrolling support for CommandPanScreen

ToolMan2k:
This commit allows to pan the camera using X and Y coordinates. 

+ Added support for .relative and .centered options for Absolute Pixel pan. Needed to bitmask parameter[4] in order to check for those.
+ Since Maniac Patch uses doubles for the current pan coordinates & speed, we copy the coordinates as a float in order to accurately round the speed value from double to integer.

Note:
There seems to be a bug on Maniac, when you have a looping map, you scroll up-left into negative coordinates, and you try to pan back to the center, more often than not it will keep panning diagonally indefinitely. Luckily EasyRPG doesn't have this issue, but for parity reasons we could try to reimplement that.

Sister request for https://github.com/EasyRPG/liblcf/pull/482

### DIFF
--- a/src/game_interpreter_map.cpp
+++ b/src/game_interpreter_map.cpp
@@ -608,15 +608,6 @@ bool Game_Interpreter_Map::CommandPanScreen(lcf::rpg::EventCommand const& com) {
 	int speed;
 	bool waiting_pan_screen = false;
 
-	// Maniac has new functions for pixel scrolling, which also have X and Y offsets
-	bool is_maniac = Player::IsPatchManiac();
-	int h;
-	int v;
-	double h_speed;
-	double v_speed;
-	bool centered = false;
-	bool relative = false;
-
 	auto& player = *Main_Data::game_player;
 
 	switch (com.parameters[0]) {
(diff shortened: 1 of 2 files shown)


fdelapena (approved):

Ghabry:
sorry for the slow review process. Played around a bit with it and I think I have to try a newer version of Maniac.

The behaviour in my version when doing a relative scroll makes no sense:

```
@> Scroll Map: Set (px), (10, 20), 1000 * 0.001 px/f, (W), Relative
@> Scroll Map: Set (px), (10, 50), 1000 * 0.001 px/f, (W), Relative
```

In Maniacs this scroll once by (10, 20) and then scrolls once by (0, 30). wtf?!

In your implementation it scrolls by (10, 20), then by (10, 50). Which is imo correct :thinking:

Ghabry:
Jenkins: test this please

Ghabry (approved):
Works for me. @ToolMan2k the only lacking implementation I found is when passing both relative and center. I could not figure out what Maniacs is calculating in this case. When you can solve this problem feel free to open a new PR.

PR #3248, opened 2024-07-24
New Feature - Local Settings

jetrotal:
If you have a copy of config.ini saved at the same folder as the player app, it will load the game settings from it, instead of the global folder.

~~I used `std::filesystem` because couldn't figure another way of extracting the apps absolute path.~~

### DIFF
--- a/src/game_config.cpp
+++ b/src/game_config.cpp
@@ -94,6 +94,9 @@ Game_Config Game_Config::Create(CmdlineParser& cp) {
 	if (!config_path.empty()) {
 		config_file = FileFinder::MakePath(config_path, config_name);
 	}
+	else if (FileFinder::Root().Exists(config_name)) {
+		config_file = ToString(config_name);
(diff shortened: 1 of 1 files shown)


Ghabry:
We do not use ``std::filesystem`` in the Player because the support can be broken on certain platforms.

To get the working directory you can use our VFS layer. Relative paths work here when using ``Root()``.

```cpp
if (FileFinder::Root().Exists(config_name)) {
    config_file = config_name;
}
```

jetrotal:
nice! working as a charm!

I wonder if buttons "Save Local Settings" and "Delete Local Settings" are necessary

Ghabry:
File deletion is not supported yet (through the VFS layer).

Save local settings vs. Save global settings: This is hard to differentiate because there is technically only one config file and the Player doesn't care where it is located.

fdelapena (approved):

PR #3249, opened 2024-07-24
Maniacs Feature - Remove scaling restrictions for pictures

ToolMan2k:
From Maniac Patch version 240423:
```
[RPG_RT.exe]
Support for different magnification rates for special drawing of pictures (rotation, waveform, angle).
```

+ Added check if sprite's width/height is at zero. If so, skip the draw routine (helps avoid a glitchy sprite at the top-left when rotating an infinitely thin picture)
- Removed check for effect mode when setting zoom or width/height.

One thing to note when comparing Maniac and EasyRPG: the former doesn't seem to properly clip the sprite in the corners when rotating the sprite.


fdelapena (approved):

fdelapena:
jenkins: ok to test

Ghabry (approved):

PR #3251, opened 2024-08-11
Fix compilation errors when Resamping is disabled

gameblabla:
If either of these is disabled in a Makefile, then it will result in errors during compiling.

### DIFF
--- a/src/audio_midi.cpp
+++ b/src/audio_midi.cpp
@@ -18,6 +18,7 @@
 // Headers
 #include "audio_midi.h"
 #include "audio_decoder_midi.h"
+#include "audio.h"
 #include "decoder_fluidsynth.h"
 #include "decoder_fmmidi.h"
 #include "decoder_wildmidi.h"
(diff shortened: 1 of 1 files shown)


carstene1ns:
This should not really be necessary, since the used functions have these guards as well.
Do you get the errors from the config stuff?

gameblabla:
I did not try the cmake but i was trying with the old makefile and disabling WildMidi and fluidsynth and i was getting those errors otherwise (when not left enabled). Pretty clear to me looking at the rest of this file in question that this is not the only place where fluidsynth code... is being encapsulated behind ifdefs.

gameblabla:
I checked with Fluidlite and i have not fixed all issues...
```
../../src/audio_midi.cpp:55:13: error: ‘Audio’ was not declared in this scope
   55 |         if (Audio().GetFluidsynthEnabled()) {
```

This happens if the resampler is not enabled but fluidlite or fluidsynth are. Don't merge this yet, i will address this issue.

carstene1ns:
With the moved included file is the rest still needed?

fdelapena (approved):

fdelapena:
jenkins: add to whitelist

PR #3252, opened 2024-08-11
Fix compiling OpenDingux code path with SDL 1.2

gameblabla:
Otherwise, it will complain about undefined references to GetSdlAxis.
As far as i'm aware, this file is only being used if you specifically use SDL 1.2 together with OPENDINGUX flag. (OD builds now use SDL2 so this path is normally never used besides yours truly)


carstene1ns (approved):
Looks good to me.

carstene1ns:
jenkins: test this please

PR #3253, opened 2024-08-11
Refactor: Always check isMessageActive() at the beginning in game_Int…

lychees:


Ghabry (approved):
This only moves local variables closer to their usage loc and removes a useless func so looks good to me.

fdelapena (approved):

fdelapena:
jenkins: ok to test

PR #3259, opened 2024-08-28
Yume2kki fixes

Ghabry:
See commit messages for description.

Thanks to @Desdaemon for figuring this out.

----

@Desdaemon I cannot reproduce the cache key issue because I'm on Linux so it likely uses a different allocator.

Though I'm pretty sure my fix works ;).

Please test if this fixes it for you.


Desdaemon:
Thank you, I'll test this PR locally later today as well.

Desdaemon:
I have confirmed locally that the fix works as intended.

Ghabry:
Thanks! I'll only remove the debug log then this is good to go.

fdelapena (approved):

PR #3261, opened 2024-09-01
Fix issue 3135 (second try) by exhausting the fluidsynth buffer

Ghabry:
@Mimigris  works for me ?!


fdelapena (approved):

Mimigris:
Seems to work for me 👍 
Only thing that may need to be checked is if it causes regressions.

Ghabry:
I think the only case where this can sound slightly different is when a song ends and a new song plays directly then a few ms of the old song that are still in the buffer are lost.

Though our API does not have a way to signal "This is a song change" and stuff like "BGM Stop event followed by BGM play right after" wouldn't be catched by this. So not worth to add.

PR #3282, opened 2024-10-11
Reset UI on WebGL context loss

Desdaemon:
Adds a new API to be called when the `webglcontextrestored` event has been fired by the browser, after restoring the context to the canvas.


fdelapena (approved):

fdelapena:
jenkins: ok to test

PR #3290, opened 2024-11-06
Implemented Maniacs Command 3011: ChangeBattleCommandEx

MakoInfused:
Using the maniacs patch ChangeBaltteCommandEx with something like this:
image

Will now give you the expected behavior in EasyRPG:
image

And for the actors themselves (in my example the row command is turned off):
image

Developers can mix and match these features and all commands are supported. Including the "Win" and "Lose" commands.

### DIFF
--- a/src/game_interpreter_battle.cpp
+++ b/src/game_interpreter_battle.cpp
@@ -31,7 +31,7 @@
 #include "game_map.h"
 #include "spriteset_battle.h"
 #include <cassert>
-#include <scene_battle.h>
+#include "scene_battle.h"
 
 enum BranchBattleSubcommand {
 	eOptionBranchBattleElse = 1
@@ -612,34 +612,31 @@ bool Game_Interpreter_Battle::CommandManiacChangeBattleCommandEx(lcf::rpg::Event
 	}
 
 	// 1 row removed
-	bool actorCommandFlags = com.parameters[0];
+	bool actor_command_flags = com.parameters[0];
 
-	lcf::Data::battlecommands.easyrpg_disable_row_feature = actorCommandFlags;
+	lcf::Data::battlecommands.easyrpg_disable_row_feature = actor_command_flags;
 
 	// 10000 lose added
 	// 01000 win added
 	// 00100 escape removed
 	// 00010 auto removed
 	// 00001 fight removed
-	int partyCommandFlags = com.parameters[1];
+	int party_command_flags = com.parameters[1];
 
 	lcf::Data::system.easyrpg_battle_options.clear();
-	for (size_t i = 0; i < Scene_Battle::BattleOptionType::Lose + 1; i++)
-	{
-		bool partyCommandFlag = partyCommandFlags & (1 << i);
-		bool flagIsSet = i > 2;
+	for (size_t i = 0; i < Scene_Battle::BattleOptionType::Lose + 1; i++) {
+		bool party_command_flag = party_command_flags & (1 << i);
+		bool flag_is_set = i > 2;
 
-		if (partyCommandFlag == flagIsSet)
-		{
+		if (party_command_flag == flag_is_set) {
 			lcf::Data::system.easyrpg_battle_options.push_back(i);
 		}
 	}
 
-	auto& scene = Scene::instance;
-	Scene_Battle* sceneBattle = dynamic_cast<Scene_Battle*>(scene.get());
+	auto* scene_battle = static_cast<Scene_Battle*>(Scene::instance.get());
 
-	if (sceneBattle) {
-		sceneBattle->CreateOptions();
+	if (scene_battle) {
+		scene_battle->CreateOptions();
 	}
 
 	return true;
--- a/src/game_interpreter_battle.cpp
+++ b/src/game_interpreter_battle.cpp
@@ -31,6 +31,7 @@
 #include "game_map.h"
 #include "spriteset_battle.h"
 #include <cassert>
+#include <scene_battle.h>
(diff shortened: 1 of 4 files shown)


src/game_interpreter_battle.cpp
```diff
@@ -605,12 +606,42 @@ bool Game_Interpreter_Battle::CommandManiacControlAtbGauge(lcf::rpg::EventComman
...
+		bool flagIsSet = i > 2;
+
+		if (partyCommandFlag == flagIsSet)
+		{
+			lcf::Data::system.easyrpg_battle_options.push_back(i);
+		}
+	}
+
+	auto& scene = Scene::instance;
+	Scene_Battle* sceneBattle = dynamic_cast<Scene_Battle*>(scene.get());
```
Ghabry, line 639: We have rtti disabled on our homebrew platforms (no ``dynamic_cast`` available).

The typical idiom we use is (has the advantage that it will even work when the battle scene is not top of the stack. Though don't think this can happen here):

```cpp
auto* scene_battle = static_cast<Scene_Battle*>(Scene::Find(Scene::Battle)).get();

// Rest stays the same
```

src/game_interpreter_battle.cpp
```diff
@@ -31,6 +31,7 @@
 #include "game_map.h"
 #include "spriteset_battle.h"
 #include <cassert>
+#include <scene_battle.h>
```
Ghabry, line 34: Use ``""`` because its part of the project

src/game_interpreter_battle.cpp
```diff
@@ -605,12 +606,42 @@ bool Game_Interpreter_Battle::CommandManiacControlAtbGauge(lcf::rpg::EventComman
...
 
-bool Game_Interpreter_Battle::CommandManiacChangeBattleCommandEx(lcf::rpg::EventCommand const&) {
+bool Game_Interpreter_Battle::CommandManiacChangeBattleCommandEx(lcf::rpg::EventCommand const& com) {
 	if (!Player::IsPatchManiac()) {
 		return true;
 	}
 
-	Output::Warning("Maniac Patch: Command ChangeBattleCommandEx not supported");
+	// 1 row removed
+	bool actorCommandFlags = com.parameters[0];
```
Ghabry, line 615: Style:

Variables are usually ``snake_case`` and functions ``UpperPascalCase``.

When it is a lambda function we are a bit indifferent because syntax wise its a mix of a variable and a function :sweat_smile: 

``{`` is usually on the same line as the ``condition``. But this is a bit mixed in our code...

src/scene_battle_rpg2k3.cpp
```diff
@@ -1217,6 +1221,16 @@ Scene_Battle_Rpg2k3::SceneActionReturn Scene_Battle_Rpg2k3::ProcessSceneActionFi
 						Main_Data::game_system->SePlay(Main_Data::game_system->GetSystemSE(Main_Data::game_system->SFX_Buzzer));
 					}
 					break;
+				case Win: // Win
+					for each (Game_Enemy* enemy in Main_Data::game_enemyparty->GetEnemies())
```
Ghabry, line 1225: I think you confused the syntax here (does this compile?)

Syntax should be:

```cpp
for (Game_Enemy* enemy: Main_Data::game_enemyparty->GetEnemies())
```

Ghabry (changes requested):
Thanks for the contribution. The logic of the code itself is fine. I just have style/idiom nits. You could not know them as is your first contribution.

src/game_interpreter_battle.cpp
MakoInfused: Noted!

src/game_interpreter_battle.cpp
MakoInfused: Yeah, I looked a around a bit, noticed it was used in one place, copied it over and started using it without fully understanding the context of how it's used in different places. My bad 😅 I've fixed it now in all the places I could find (not just this specific one you called out).

src/scene_battle_rpg2k3.cpp
MakoInfused: It does, and I saw in your other review you found out it's a Microsoft compiler thing which predates the modern introduction of proper support in the language, it turns out it won't compile on Linux anyway. Oops! Fixed!

src/game_interpreter_battle.cpp
MakoInfused: Good to know. I've ended up changing my cast to static, but keeping the rest the same since we're guaranteed to be inside a battle with this function call. I tested it by calling this function from outside battle and it works as intended (by never getting called). 

I would have taken your suggestion but it resulted in a compiler error and I wasn't sure it was worth spending the time to figure it out:
image

However, if you feel strongly about making the change, just let me know and I can figure out the exact syntax.

Ghabry: Sorry messed up the braces. Correct is ``auto* scene_battle = static_cast<Scene_Battle*>(Scene::Find(Scene::Battle).get());``

But yeah this code is indeed never running outside of the battle as this is code inside the interpreter.

Ghabry:
Jenkins: test this please

@MakoInfused I'll send you an invite for the "Player Team". When you accept it then the PR builders will run automatically...

Ghabry (approved):
The one build failure is just a timeout so LGTM :+1:

fdelapena (approved):

MakoInfused:
Jenkins: retest this please

PR #3291, opened 2024-11-07
Implemented Maniacs Command 3010: ControlATBGauge

MakoInfused:
Using the maniacs patch ControlATBGauge with something like this:
image

Will result in all party members having 1/3rd of a full gauge. 
image

Or using it with 100% ATB for the first party member:
image

Would result in this:
image

Of course developers can use any combination of parameters, I've tested all of them individually with all combinations. 

I've also recreated the functionality from maniacs where when a 0 index is passed in as an actor id, it just ignores the command entirely.

### DIFF
--- a/src/game_interpreter_battle.cpp
+++ b/src/game_interpreter_battle.cpp
@@ -597,12 +597,104 @@ bool Game_Interpreter_Battle::CommandManiacControlBattle(lcf::rpg::EventCommand 
 	return true;
 }
 
-bool Game_Interpreter_Battle::CommandManiacControlAtbGauge(lcf::rpg::EventCommand const&) {
+bool Game_Interpreter_Battle::CommandManiacControlAtbGauge(lcf::rpg::EventCommand const& com) {
 	if (!Player::IsPatchManiac()) {
 		return true;
 	}
 
-	Output::Warning("Maniac Patch: Command ControlAtbGauge not supported");
+	// 0 actor
+	// 1 party member
+	// 2 entire party
+	// 3 troop member
+	// 4 entire troop
+	int targetFlags = com.parameters[0];
+	// 0 constant
+	// 1 variable
+	// 2 variable id
+	int targetReferenceFlags = com.parameters[1];
+	int targetReferenceIdentifier = com.parameters[2];
+	// 0 set
+	// 1 add
+	// 2 sub
+	int operationFlags = com.parameters[3];
+	// 0 value
+	// 1 percentage
+	int operandFlags = com.parameters[4];
+	// 0 constant
+	// 1 variable
+	// 2 variable id
+	int valueReferenceFlags = com.parameters[5];
+	int valueReferenceIdentifier = com.parameters[6];
+
+	auto getVariableOrValue = [](int flags, int identifier) {
+		switch (flags) {
+			case 0:
+				return identifier;
+				break;
+			case 1:
+				return Main_Data::game_variables->Get(identifier);
+				break;
+			case 2:
+				return Main_Data::game_variables->Get(Main_Data::game_variables->Get(identifier));
+				break;
+		}
+	};
+
+	int targetId = getVariableOrValue(targetReferenceFlags, targetReferenceIdentifier);
+
+	auto getAtbValue = [getVariableOrValue, valueReferenceFlags, valueReferenceIdentifier](int flags) {
+		int value = getVariableOrValue(valueReferenceFlags, valueReferenceIdentifier);
+
+		switch (flags) {
+			case 0:
+				return value;
+				break;
+			case 1:
+				return (int) ((double)value / 100 * Game_Battler::GetMaxAtbGauge());
+				break;
+		}
+	};
+
+	auto executeOperation = [getAtbValue, operandFlags](int flags, Game_Battler* battler) {
+		switch (flags) {
+			case 0:
+				battler->SetAtbGauge(getAtbValue(operandFlags));
+				break;
+			case 1:
+				battler->IncrementAtbGauge(getAtbValue(operandFlags));
+				break;
+			case 2:
+				battler->IncrementAtbGauge(-getAtbValue(operandFlags));
+				break;
+		}
+	};
+
+	switch (targetFlags) {
+		case 0:
+			if (targetId > 0) {
+				executeOperation(operationFlags, Main_Data::game_actors->GetActor(targetId));
+			}
+			break;
+		case 1:
+			executeOperation(operationFlags, Main_Data::game_party->GetActor(targetId));
+			break;
+		case 2:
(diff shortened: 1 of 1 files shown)


src/game_interpreter_battle.cpp
```diff
@@ -596,21 +597,143 @@ bool Game_Interpreter_Battle::CommandManiacControlBattle(lcf::rpg::EventCommand
...
+	// 0 value
+	// 1 percentage
+	int operandFlags = com.parameters[4];
+	// 0 constant
+	// 1 variable
+	// 2 variable id
+	int valueReferenceFlags = com.parameters[5];
+	int valueReferenceIdentifier = com.parameters[6];
+
+	auto getVariableOrValue = [](int flags, int identifier) {
```
Ghabry, line 629: We already have a function ``ValueOrVariable`` that does this. Also has the same signature ``(mode, value)``. :)

src/game_interpreter_battle.cpp
```diff
@@ -596,21 +597,143 @@ bool Game_Interpreter_Battle::CommandManiacControlBattle(lcf::rpg::EventCommand
...
 }
 
-bool Game_Interpreter_Battle::CommandManiacControlAtbGauge(lcf::rpg::EventCommand const&) {
+bool Game_Interpreter_Battle::CommandManiacControlAtbGauge(lcf::rpg::EventCommand const& com) {
 	if (!Player::IsPatchManiac()) {
 		return true;
 	}
 
-	Output::Warning("Maniac Patch: Command ControlAtbGauge not supported");
+	// 0 actor
```
Ghabry, line 605: Can you move all the comments after the ``case`` labels? E.g. as done here: https://github.com/EasyRPG/Player/blob/master/src/game_interpreter_control_variables.cpp#L70

(okay in the linked code they are the line after and not on the same line but doesn't matter where you locate them)

src/game_interpreter_battle.cpp
```diff
@@ -596,21 +597,143 @@ bool Game_Interpreter_Battle::CommandManiacControlBattle(lcf::rpg::EventCommand
...
+				executeOperation(operationFlags, member);
+			}
+			break;
+		case 3:
+			if (targetId > 0) {
+				executeOperation(operationFlags, Main_Data::game_enemyparty->GetEnemy(targetId));
+			}
+			break;
+		case 4:
+			for each (Game_Battler* member in Main_Data::game_enemyparty->GetEnemies()) {
```
Ghabry, line 692: Same Syntax with the foreach. I guess this is a C++/CLI extension by Microsoft https://learn.microsoft.com/en-us/cpp/dotnet/for-each-in?view=msvc-170 :roll_eyes:

Ghabry (changes requested):

src/game_interpreter_battle.cpp
MakoInfused: Done! 

Although, I did like the fact that I could read them anywhere from the intellisense before: 
image
Perhaps something to consider, but I'll adhere to the standard regardless :)

src/game_interpreter_battle.cpp
MakoInfused: Updated!

src/game_interpreter_battle.cpp
Ghabry: best would be a local enum in the function and then using these enum-values for the cases (but I will not ask you to do this additional work for us)

fdelapena (approved):

Ghabry:
Please rebase to get rid of the merge commits:

```
git remote add upstream https://github.com/easyrpg/player
git fetch upstream

git rebase upstream/master

Solve merge conflicts (if there are any) 
Then check with git log that the history looks okay

git push --force-with-lease origin
```

Ghabry:
Rebase result looks good 👍

src/game_interpreter_battle.cpp
```diff
@@ -602,95 +602,75 @@ bool Game_Interpreter_Battle::CommandManiacControlAtbGauge(lcf::rpg::EventComman
...
 
 		switch (flags) {
 			case 0:
+				// value
 				return value;
 				break;
 			case 1:
+				// percentage
 				return (int) ((double)value / 100 * Game_Battler::GetMaxAtbGauge());
 				break;
```
Ghabry, line 626: Can you add a ``default: return 0`` case to silence this compiler warning?

You can add that change to the previous commit: ``git commit --amend``, then force push

src/game_interpreter_battle.cpp
```diff
@@ -597,12 +597,84 @@ bool Game_Interpreter_Battle::CommandManiacControlBattle(lcf::rpg::EventCommand
...
+		case 2:
+			// entire party
+			for (Game_Battler* member : Main_Data::game_party->GetActors()) {
+				executeOperation(operation_flags, member);
+			}
+			break;
+		case 3:
+			// troop member
+			if (target_id > 0) {
+				executeOperation(operation_flags, Main_Data::game_enemyparty->GetEnemy(target_id));
```
Ghabry, line 667: ``GetActor`` and ``GetEnemy`` return a ``nullptr`` when the target is missing (crashes).

Can you add sanity checks?

```cpp
auto* actor = Main_Data::game_actors->GetActor(target_id); // <- again for game_actors
if (!actor) {
	Output::Warning("CommandManiacControlAtbGauge: Invalid actor ID {}", target_id);
	return true;
}

auto* enemy = Main_Data::game_enemyparty->GetEnemy(target_id);
if (!enemy) {
	Output::Warning("CommandManiacControlAtbGauge: Invalid enemy ID {}", target_id);
	return true;
}
```

MakoInfused: Thanks for this and great catch! I took your approach as a start, then encapsulated it into a function for reuse, since I have a few calls in this pull request and it will set me up to use in my GetBattleInfo request that I'll post as well.

src/game_interpreter_battle.cpp
```diff
@@ -37,6 +37,18 @@ enum BranchBattleSubcommand {
...
 
+enum TargetType {
+	Actor,
+	Member,
+	Enemy,
+};
+
+static const char* target_text[] = { "actor", "party member", "enemy" };
+
+static const void MissingTargetWarning(char* command_name, TargetType target_type, int target_id) {
```
Ghabry, line 48: ``command_name`` must be ``const char*`` to get rid of the compiler warning. After pushing best is to wait for the Linux runners to finish and look for warnings in the diff. They are usually more pedantic than Visual Studio 😅

Ghabry (approved):
When the warning is fixed looks good to me

src/game_interpreter_battle.cpp
MakoInfused: Good tip, thanks!

PR #3293, opened 2024-11-08
Fix Beloved Rapture Menu issues

### LINKED ISSUE #3128: Small Issues related to Beloved Rapture Esc Menus.
---
#### Name of the game:
Beloved Rapture

#### Player platform:
Windows

https://store.steampowered.com/app/2017620/Beloved_Rapture/

#### Here's a list of issues related to BR Ingame Menus:

- [x] GetSaveInfo fails to set proper transparency on facesets read by it:
image

--------------

- [x] Set Game Options - Picture Limit is not implemented (If not needed, we should just skip it):
image
image

-------------

- [x] Control Variables Expressions - AssignInplace  Is not Supported:
image
image

--------------

- [x] Error alerts that does not appear, on og game ( Guess the dev exploited invalid ID to deal with blank properties? )
image
Something else must be problematic here, the navigation also breaks.

------------

- [x] Game Crashes and Closes after clicking Guides - Quests - 1st item | MAY NOT BE RELATED TO STRVARS
image

- [x] Map scrolls incorrectly because the width is not dividable by 16.
- [x] Items cannot be used
- [x] Equipment is not shown in the list

Ghabry:
The work on the battle events motivated me to fix the non-battle issues :)

What this doesn't fix is the wrong font size, guess they replaced the font in the exe with a different one. So the metrics are different.

Fix #3128

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -463,8 +463,16 @@ BitmapRef Cache::Tile(StringView filename, int tile_id) {
 }
 
 BitmapRef Cache::SpriteEffect(const BitmapRef& src_bitmap, const Rect& rect, bool flip_x, bool flip_y, const Tone& tone, const Color& blend) {
+	std::string id = ToString(src_bitmap->GetId());
+
+	if (id.empty()) {
+		// assert caused too many regressions, use the pointer as the unique key and log instead
+		Output::Debug("Bitmap has no ID. Please report a bug!");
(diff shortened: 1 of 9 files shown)


fdelapena (approved):

Ghabry:
Fixed the "Monster with hue change"-crash and added a workaround to prevent further crashes related to bitmaps without an ID when having Effects applied...

fdelapena (approved):

PR #3294, opened 2024-11-09
Implemented Maniacs Command 3012: GetBattleInfo

MakoInfused:
When we use the Maniacs GetBattleInfo command with settings like these:
image

And we setup our event to display to a picture command like this:
image

We get the exact same result in EasyRPG as we do in the Maniacs Exe:
image

All combination of event commands were thoroughly tested with different combinations, including:

- Target Types: Actor ID, Party Member, Entire Party, Enemy, Entire Enemy.
- Information Types: Parameter Buffs, States, Elements, Others
- Also tested with variables that are outside the defined range: this seems without error in both EasyRPG and Maniacs (surprisingly)

### DIFF
--- a/src/game_interpreter_battle.cpp
+++ b/src/game_interpreter_battle.cpp
@@ -732,12 +732,125 @@ bool Game_Interpreter_Battle::CommandManiacChangeBattleCommandEx(lcf::rpg::Event
 	return true;
 }
 
-bool Game_Interpreter_Battle::CommandManiacGetBattleInfo(lcf::rpg::EventCommand const&) {
+bool Game_Interpreter_Battle::CommandManiacGetBattleInfo(lcf::rpg::EventCommand const& com) {
 	if (!Player::IsPatchManiac()) {
 		return true;
 	}
 
-	Output::Warning("Maniac Patch: Command GetBattleInfo not supported");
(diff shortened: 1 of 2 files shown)


src/game_interpreter_battle.cpp
```diff
@@ -732,12 +732,125 @@ bool Game_Interpreter_Battle::CommandManiacChangeBattleCommandEx(lcf::rpg::Event
...
+			{
+				// elements: size, [...element_id]
+				auto elements = lcf::Data::attributes.size();
+				Main_Data::game_variables->Set(information_identifier, elements);
+				for (size_t i = 0; i < elements; i++)
+				{
+					Main_Data::game_variables->Set(information_identifier + i + 1, battler->GetAttributeRateShift(lcf::Data::attributes.at(i).ID) + 1);
+				}
+				break;
+				break;
```
Ghabry, line 784: This command looks also straight forward 👍. There are two ``break`` here and instead of ``at()`` you can use the index ``[]``. Besides this have no complaints here. Thanks

MakoInfused: Done and done!

Ghabry:
Can you rebase the PR? (same commands as last time should work)

fdelapena (approved):

PR #3303, opened 2024-11-24
String Variables: Operate on Codepoints

Ghabry:
Updated the test game by @enewey to use extended latin characters like é everywhere:

https://easyrpg.org/play/pr3303/?game=string-var

In UTF-8 they require 2 bytes and 1 codepoint.

Correct is using 1 codepoint here (which matches the legacy encoding = 1 byte)

For あ it is 3 bytes UTF-8 or 1 codepoint. When running in shift-jis RPG_RT seems to use a multibyte character so it also reports 1. There is no shift-jis case in the project but I tested this locally.

Sooo we are just extremely lucky here that it _just works_ even though our encoding is completely different :)

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -4823,7 +4823,7 @@ bool Game_Interpreter::CommandManiacControlStrings(lcf::rpg::EventCommand const&
 			int pos = 0;
 			std::string op_string;
 			for (int i = 0; i < 3; i++) {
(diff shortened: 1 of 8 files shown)


fdelapena (approved):

Ghabry:
I handle now the edge cases for out of bounds values.

Also supports left-adjust now when "AdjustMin" value is negative and indexing from the back when the index is negative.

PR #3307, opened 2024-11-27
Implement Maniacs Battle Common Events

MakoInfused:
When using maniacs patch and we have a "Battle Begin" trigger set in the editor like this:
image

In battle we now have it get triggered like this:
image

When we use a "Battle (Parallel)" trigger in the editor like this:
image

In battle it will result in a repeated common event getting triggered every 5 seconds (since that is what my wait time it set to in the above image) like this:
EasyRPG-ManiacsBattleCommonEvents

### DIFF
--- a/src/game_interpreter.h
+++ b/src/game_interpreter.h
@@ -411,6 +411,7 @@ inline void Game_Interpreter::Push(Game_Event* ev, const lcf::rpg::EventPage* pa
 template<InterpreterExecutionType type_ex>
 inline void Game_Interpreter::Push(Game_CommonEvent* ev) {
 	static_assert(type_ex == InterpreterExecutionType::AutoStart || type_ex == InterpreterExecutionType::Parallel
+		|| type_ex == InterpreterExecutionType::BattleStart || type_ex == InterpreterExecutionType::BattleParallel
 		|| type_ex == InterpreterExecutionType::Call || type_ex == InterpreterExecutionType::DeathHandler
 		|| type_ex == InterpreterExecutionType::DebugCall || type_ex == InterpreterExecutionType::ManiacHook, "Unexpected ExecutionType for CommonEvent"
 	);
--- a/src/scene_battle_rpg2k3.cpp
+++ b/src/scene_battle_rpg2k3.cpp
@@ -979,7 +979,7 @@ void Scene_Battle_Rpg2k3::vUpdate() {
 			if (!parallel_interpreter.IsRunning()) {
 				for (auto common_event : battle_parallel_events)
 				{
-					parallel_interpreter.Push<InterpreterExecutionType::Parallel>(common_event);
+					parallel_interpreter.Push<InterpreterExecutionType::BattleParallel>(common_event);
 				}
 			}
 		}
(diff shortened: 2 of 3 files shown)


Ghabry:
The way how you do parallel process updating looks completely different to how it is done on the map.

For the map the common events all have a parallel process directly attached to them:

https://github.com/EasyRPG/Player/blob/d9dd9be210eb0383784a855a770b4e03901cf232/src/game_map.cpp#L1283-L1308

(ignore the async stuff for now)

----

Maybe add ``UpdateBattle`` and ``IsWaitingBattleBackgroundExecution`` to the common events and do it via these functions?

----

What needs to be tested:

On the map parallel processes can be suspended and they will continue where they stopped. Does Maniac Patch do this in battle? To test this run two parallel processes:

PP1 (has run condition S1 = ON):

```
ShowMessage A
Set Switch S2 = ON
Set Switch S1 = OFF
ShowMessage B
```

PP2 (has run condition S2 = ON):

```
ShowMessage C
Set Switch S1 = ON
Set Switch S2 = OFF
ShowMessage D
```

The expected message output is A, C, B. D never shown.

At least when battle pps are like map pps.

----

Also worth testing if they keep the progress _across_ battles:

For this simply do something like

PP1:

```
ShowMessage A
Abort Battle
ShowMessage B
```

When it resets it will always show A. When it remembers the state it will show B when a 2nd battle starts.

MakoInfused:
@Ghabry I tested all of the exact scenarios as you outlined them and here are the results!

> On the map parallel processes can be suspended and they will continue where they stopped. Does Maniac Patch do this in battle?

No. It seems to work in a much more simple/diferent way.

Suspension Test:
**Map PP expected output**: A, C, B. D never shown.
**Maniacs Battle PP actual output**: A, B, C, D.
**My Implementation output**: A, B. C, D is never shown. 
Seems I'll need to fix this is as the events aren't all running in parallel to match Maniacs.

Resume Test:
**Map PP expected output**: When it remembers the state it will show B when a 2nd battle starts.
**Maniacs Battle PP actual output**: It resets and never shows anything except A.
**My Implementation output**: Works the same as Maniacs.

Parallel Test:
**Map PP expected output**: I think the output should be A, B, C, D.
**Maniacs Battle PP actual output**: A, B, C, A, D, C (repeats from beginning)
**My Implementation output**: B, D, A, C (repeated from beginning)
Seems this doesn't work the same either and will also need to be changed to match Maniacs.

Ghabry:
Please don't create merge commits. They result in a mess as you can see here with these >200 commits.

Instead do a rebase:

```
git fetch upstream
git rebase upstream/master
```

During the rebase their will be conflicts. Fix the conflicts and `git commit`, then `git rebase --continue` until the rebase is successful.

(rebasing can be tricky, look for a guide in case you need assistance or ask in the chat)

jetrotal:
Jenkins, test this please
(Do I have those powers?)

korean-guy:
스크린샷 2026-03-26 102816

'Get Battle Info' command fails to output information for enemy number 1 (internal index 0).



스크린샷 2026-03-26 102715
스크린샷 2026-03-26 102725

Upper image = easy rpg
Lower image = rpg_rt(Maniacs)

For states.
easyrpg is 0 and 1, indicating only the presence or absence of states.
rpg_rt(Maniacs) outputs the turn count of the states

For elements.
output results like easyrpg and rpg_rt(Maniacs), but I think it needs to be improved.
Currently, only the presence or absence of elements is displayed.
I think it's better to print out the grades of elements like 'A = 1, B = 2, C = 3, D = 4, E = 5'.

Ghabry:
Thanks for your contribution. I took some parts of it and integrated it more with other Player components so it executes the Parallel Events more correctly.

New attempt is #3545

PR #3308, opened 2024-12-05
New Flag - SkipIndexJSON || --web-skipindexjson

jetrotal:
Skip the validation from index.json when load images (Web Build).

### DIFF
--- a/src/async_handler.cpp
+++ b/src/async_handler.cpp
@@ -358,7 +358,7 @@ void FileRequestAsync::Start() {
 	if (it != file_mapping.end()) {
 		request_path += it->second;
 	} else {
-		if (file_mapping.empty()) {
+		if (file_mapping.empty() || Player::game_config.patch_web_skip_index_json.Get()) {
 			// index.json not fetched yet, fallthrough and fetch
 			request_path += path;
 		} else {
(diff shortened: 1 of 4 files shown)


jetrotal:
After some testing

image
By disabling the json check, I can upload files to `FS` and load them.
Some DL errors appears, due to the confusion between whats inside JSON and what's inside `FS`.

As soon as easyRPG displays an image, its cached and can't be changed.
I wonder if a better approach would be learning what's the difference between loading pictures, 
and loading text from the `save` folder

Ghabry:
not a fan of this flag. Feels like a hack.

Instead how about checking if there is a file in the filesystem and when the file exists simply skip the download instead?

But this proper solution needs more work:

- The AsyncHandler does not know about the file type yet, so cannot execute FindFile yet to figure out the possible extensions
- Needs a solution for translations: They work by clearing the cache and refetching stuff. The files are still in the filesystem. So checking first if they are in the filesystem breaks the fetching of different translations :thinking:
- Export the Invalidate FileSystem Cache function to emscripten so new files become visible
- Create a function for deleting single items from the bitmap cache because you want to swap images...

jetrotal:
@Ghabry I suspect a good system for loading external data already exists in string variable's .TXT files

Why it can load anything through the /Save/ folder? can I make something similar for other files?

Ghabry:
The only reason why loading from ``Save`` works for ``Text`` is that Maniac Patch supports writing of text files and I'm against overwriting game files, so all write operations are redirected to the Save directory. So it must also read from the Save directory.

(of course protection is useless when Gamedir == Savedir)

jetrotal:
A flag like this could work for me:
--Allow-assets-from-Save

jetrotal:
Ditching this feature.
https://github.com/EasyRPG/Player/pull/3478 does a better job at loading external image files.

PR #3309, opened 2024-12-06
Maniacs Feature - Partially implement command GetGameInfo (3021)

ToolMan2k:
This commit partially implements command #3021 from Maniac Patch, which can returns things from screen coordinate/size to screen pixel data.

I've opted to implement most of its features, the reason for these omissions were:

1. Handling the tile/pixel functions would be better suited for someone who delves into these systems;
2. I do not have a good grasps on how the "Command Interpreter State" functions. There is a bit of documentation on the README.txt file in Maniacs;

I have implemented the following commands:
0- Get Map Size: returns the width and height of the map;
2- Get Window Info: returns the width and height of the screen;
5- Get Tileset ID: returns the current tileset ID;
6- Get Actor/Message Face: returns the face string/index for either the current message box or an actor. Dynamic toggles between the current or default sprite;
7- Get Actor/Vehicle/Event Body: returns the sprite string/index for either a map event or an actor. Dynamic toggles between the current or default sprite;
8- Get Screen Position: returns the X and Y coordinates of the camera;
9- Get Screen Shake: returns the current shake value of the screen;
10- Get BGM: returns the current BGM string/parameters.

I've made a few tests and it seems to be stable, however it deserves some extra shaking, in case of differing behavior or crashes.

### DIFF
--- a/src/game_actor.h
+++ b/src/game_actor.h
@@ -244,11 +244,25 @@ public:
 	StringView GetName() const override;
 
 	/**
+	 * Gets actor name from the database.
+	 *
+	 * @return name.
+	 */
+	StringView GetOriginalName() const;
+
+	/**
 	 * Gets actor character sprite filename.
 	 *
 	 * @return character sprite filename.
 	 */
 	StringView GetSpriteName() const override;
+
+	/**
+	 * Gets actor character sprite filename from the database.
+	 *
+	 * @return character sprite filename.
+	 */
+	StringView GetOriginalSpriteName() const;
 
 	/**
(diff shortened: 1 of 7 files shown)


fdelapena:
jenkins: add to whitelist

fdelapena (approved):

Ghabry:
Thanks for the contribution! Looks fine to me. I will only add some utility functions (e.g. so that the "static" lookup can be as simple as the "dynamic" lookup).

Ghabry (approved):

PR #3312, opened 2024-12-17
MoveRoute - Support Variables as Event ID (Maniac patch feature)

jetrotal:
Tiny update based on @MackValentine's code.

In MP, Move Route supports using variables as the target event ID:

```js
//TPC SYNTAX:

v[1] = 2
@ev[v[1]].setAction .freq 8 .unskippable .act {
    .moveDown
}
```

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -3049,7 +3049,9 @@ bool Game_Interpreter::CommandPlayerVisibility(lcf::rpg::EventCommand const& com
 }
 
 bool Game_Interpreter::CommandMoveEvent(lcf::rpg::EventCommand const& com) { // code 11330
-	int event_id = com.parameters[0];
+	int event_id = ValueOrVariableBitfield(com.parameters[2], 2, com.parameters[0]);
+	int repeat = ManiacBitmask(com.parameters[2], 0x1);
+	
 	Game_Character* event = GetCharacter(event_id);
 	if (event != NULL) {
 		// If the event is a vehicle in use, push the commands to the player instead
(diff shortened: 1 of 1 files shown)


src/game_interpreter.cpp
```diff
@@ -3050,6 +3050,15 @@ bool Game_Interpreter::CommandPlayerVisibility(lcf::rpg::EventCommand const& com
...
 bool Game_Interpreter::CommandMoveEvent(lcf::rpg::EventCommand const& com) { // code 11330
 	int event_id = com.parameters[0];
+	int repeat = com.parameters[2];
+	int mode;
+
+	if (Player::IsPatchManiac()) {
+		mode = com.parameters[2] >> 8;
+		repeat = ManiacBitmask(com.parameters[2], 0x1);
+		event_id = ValueOrVariable(mode, com.parameters[0]);
+	}
```
Ghabry, line 3060: It should also work when replace the entire change with:

```cpp
int event_id = ValueOrVariableBitfield(com.parameters[2], 2, com.parameters[0])
int repeat = ManiacBitmask(com.parameters[2], 0x1);
```

Though I still recommend that you test if I'm not wrong here xD.

Reasons (for context):

When No Maniac Patch ``ValueOrVariableBitfield`` ignored the mode, so this is the same as ``int event_id = com.parameters[0];``, when MP is on the mode is ``com.parameters[2] >> 2 * 4``.

When No Maniac Patch ``ManiacBitmask`` only returns the value and ignores the bitmask, so is same as ``int repeat = com.parameters[2]``.

jetrotal: Tested here, and seemed to work with both maniac and vanilla games. Thanks!

Ghabry (approved):

Ghabry (approved):
approve with comment

Ghabry (approved):

fdelapena (approved):

PR #3313, opened 2024-12-18
Erase Event - Erase By ID + Recreate Event Option (Maniac Patch Feature)

jetrotal:
Another smaller update:
Events can be disabled and enabled again through maniacs patch.
 It also allows targeting an event by ID instead of only erasing the caller itself.

```js
//TPC SYNTAX

@ev[2].erase
@wait .input
v[1] = 2
@ev[ v[1] ].return
```

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -3850,11 +3850,21 @@ bool Game_Interpreter::CommandEndLoop(lcf::rpg::EventCommand const& com) { // co
 	return true;
 }
 
-bool Game_Interpreter::CommandEraseEvent(lcf::rpg::EventCommand const& /* com */) { // code 12320
+bool Game_Interpreter::CommandEraseEvent(lcf::rpg::EventCommand const& com) { // code 12320
+	int event_id = 0; // default rm values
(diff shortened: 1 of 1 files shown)


src/game_interpreter.cpp
```diff
@@ -3850,11 +3850,21 @@ bool Game_Interpreter::CommandEndLoop(lcf::rpg::EventCommand const& com) { // co
 	return true;
 }
 
-bool Game_Interpreter::CommandEraseEvent(lcf::rpg::EventCommand const& /* com */) { // code 12320
+bool Game_Interpreter::CommandEraseEvent(lcf::rpg::EventCommand const& com) { // code 12320
+	int event_id = 0; // default rm values
+	bool is_active = 0; // In Vanilla RM event is always itself and it always becomes inactive
+
+	if (Player::IsPatchManiac()) {
```
Ghabry, line 3857: Please also add a check that com.parameters.size() >= 3, otherwise this will read garbage data when Maniac Patch is active and the command has less arguments.

Ghabry:
OT: Hm it looks like the automatic PR labeler is broken after I updated the YAML file -_-

Ghabry (approved):

fdelapena (approved):

PR #3317, opened 2024-12-21
ExeReader: Support 64 Bit executables

Ghabry:
Maniac Patch is now detected when the new exe is used.

----

The difference between 32bit PE and 64 bit PE is actually not _that_ huge when you only care about the resources.

The differences:

- The machine type is 0x8664
- In the optional header the magic at the beginning is 0x20b instead of 0x10b (0x20b is a "PE32+" header, 0x10b is a "PE32" header)
- The data direction offset is different because the PE32+ image file header is larger

The offset calculation was the major change:

Before the resource offset was hardcoded to ``0x88`` (136) which is actually ``size_of_optional_header (24) + size_of_image_file_header (96/112 [PE+]) + 16``.

All the resource format itself is the same. (lucky for us).

----

Tested it with:

Player.exe, both 32 Bit and 64 Bit: PASSED

Maniac exe: PASSED

Various RPG_RT (32 bit): PASSED

### DIFF
--- a/src/exe_reader.cpp
+++ b/src/exe_reader.cpp
@@ -34,19 +34,46 @@ namespace {
 EXEReader::EXEReader(Filesystem_Stream::InputStream core) : corefile(std::move(core)) {
 	// The Incredibly Dumb PE parser (tm)
 	// Extracts data from the resource section for engine detection and can read ExFont.
-	uint32_t ofs = GetU32(0x3C);
+	uint32_t ofs = GetU32(0x3C); // PE header offset
 	uint16_t machine = GetU16(ofs + 4);
-	if (machine != 0x14c) {
-		// FIXME: Newer versions of Maniac Patch can be 64 Bit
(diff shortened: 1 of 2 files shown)


fdelapena (approved):

src/exe_reader.cpp
```diff
@@ -34,19 +34,31 @@ namespace {
...
+		file_info.machine_type = MachineType::Unknown;
 		return;
 	}
 
+	if (machine == 0x14c) {
+		file_info.machine_type = MachineType::i386;
+	} else {
+		file_info.machine_type = MachineType::amd64;
+	}
+
```
carstene1ns, line 51: I think this block could be simplified as a switch/case or just reorder to have the "unsupported" else case last.
Also the comment about 64 bit maniac patch seems resolved now?

Ghabry: yeah that code was bad. Is now a switch-case with error in default.

Added a similar check to the PE/PE+ header magic.

carstene1ns (approved):

PR #3318, opened 2024-12-23
Fix compiler warnings, bump deps etc.

Ghabry:
My final contribution before Christmas to get rid of _all_ compiler warnings reported by g++ and clang with ``Wall, extra, deprecated``.

Can be already reviewed, except:

_Unfortunately_: This is not ready yet because the unit tests that calculate the attack damage fail. The reason is that the tests _never executed_ because ``SUBCASE``s inside a for-loop only executed _once_. Now they run and fail :shrug: .

I think the code itself is correct, just the tests are flawed and nobody noticed as they didn't execute. But fixing this is a bit more involved.


Ghabry:
correction: The PR labeler works. So many colourful labels 😆

Ghabry:
Removed the doctest update for now to make this mergeable. will fix doctest in a different PR.

Ghabry:
uhm will also disable the sanitizers for now. looks like the leak sanitizer makes the check fail...

src/game_battlealgorithm.cpp
```diff
@@ -62,9 +62,12 @@ Game_BattleAlgorithm::AlgorithmBase::AlgorithmBase(Type ty, Game_Battler* source
 	type(ty), source(source), targets(std::move(in_targets))
 {
 	assert(source != nullptr);
+
+#ifndef NDEBUG
```
carstene1ns, line 66: Does it nag about an empty loop here?
Because asserts are no-ops in release builds anyway, so should be optimized away.

Ghabry: It nags about the unused t loop var when the assert is removed.

Ghabry:
good news: Can reenable the sanitizers. The reported memory leak was a real leak (though its a false positive because of some static buffers in liblcf).

The destiny testcase used ``new`` for memory management. :thinking:

fdelapena (approved):

PR #3320, opened 2025-01-02
ChangeSpriteAssociation: Fix OOB reads

### LINKED ISSUE #3314: Changing the CharSet of the actor using a non-relative path doesn't load assets properly and can cause crashes when a battle animation occurs
#### Player platform:

[TestGame-2000.zip](https://github.com/user-attachments/files/18188059/TestGame-2000.zip)

Start the test game and interact with the various NPCs in the room that change the CharSet of your current actor.
When interacting with those on the right of the room, which instead of directly calling an asset in the CharSet directory, call an asset in the following way `../CharSet/nameoftheasset` (I know, this is really dumb to do that here, but it also applies if the CharSet was in another directory, it was just done here this way to not have to duplicate CharSet files), some oddities occur:
- Changing your CharSet to ../CharSet/Chara1 will not display its graphic at all
- Playing a battle animation while in said state of Chara1 will crash the game by displaying an error "Couldn't create ??x?? image" (e.g. 72x-426132992)
- Changing your CharSet to ../CharSet/People2 will be a bit random: sometimes it will display it as intended, sometimes it will display it as transparent for no reason at all
The expected result and what happens in RPG_RT is that calling CharSet files this way works as intended.

Ghabry:
Fix #3314

@Mimigris

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -2112,8 +2112,17 @@ bool Game_Interpreter::CommandChangeSpriteAssociation(lcf::rpg::EventCommand con
 	}
 
 	auto file = ToString(CommandStringOrVariableBitfield(com, 3, 1, 4));
-	int idx = ValueOrVariableBitfield(com, 3, 2, 1);
-	bool transparent = com.parameters[2] != 0;
+
+	int idx = 0;
+	if (com.parameters.size() > 1) {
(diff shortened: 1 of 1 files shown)


fdelapena (approved):

Mimigris:
Seems to work on my side 👍

PR #3328, opened 2025-01-07
Map: initialize screen width and height on SetupCommon

### LINKED ISSUE #3327: The background might become out of sync with the foreground after loading.
Game Name: Ruina 廃都の物語
Platform: Linux, Emscripten
Player Version: [7976386](https://github.com/EasyRPG/Player/commit/7976386dcc3efad39159a71a60b747e3842af3ff)
Issue Description: After saving in the town, restart the app and loading the save, the background become out of sync with the foreground.

edit: this post previously mentioned 59c2c8b, but i realized that might not related.

aidatorajiro:
fixes #3327.
(i don't know how this patch interferes with #3293, but Init() is called only after reset, so not a very destructing change i guess)

### DIFF
--- a/src/game_map.cpp
+++ b/src/game_map.cpp
@@ -153,9 +153,6 @@ int Game_Map::GetMapSaveCount() {
 void Game_Map::Setup(std::unique_ptr<lcf::rpg::Map> map_in) {
 	Dispose();
 
-	screen_width = (Player::screen_width / 16.0) * SCREEN_TILE_SIZE;
-	screen_height = (Player::screen_height / 16.0) * SCREEN_TILE_SIZE;
-
 	map = std::move(map_in);
 
 	SetupCommon();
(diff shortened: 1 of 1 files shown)


Ghabry:
I think I deleted this because I assumed Setup is always called 🤔. Have to check the control flow 😅. Thanks

Ghabry:
@aidatorajiro I checked this now and ``Setup`` is indeed not called when loading from a save (only ``SetupFromSave`` is).

Can you try deleting the

```cpp
screen_width = (Player::screen_width / 16.0) * SCREEN_TILE_SIZE;
screen_height = (Player::screen_height / 16.0) * SCREEN_TILE_SIZE;
```

from ``Init`` and ``Setup`` and instead add these lines at the beginning of ``SetupCommon``.

fdelapena (approved):

fdelapena:
Jenkins: ok to test

Ghabry (approved):
Squashed all commits into one as the intermediate commits were unnecessary here. Otherwise no changes from my side.

PR #3329, opened 2025-01-07
FlashEvent - Maniac Patch Parameters

jetrotal:
Supporting usage of variables when game uses maniac patch
and parameters size is bigger than 7.

```js
//TPC Example
@ev.flash .player .rgbv v[100], v[101], v[102], v[103] .time 100

```


Ghabry:
You can use the other overload of ``ValueOrVariableBitfield`` which does already do the Maniac and size checking:

```cpp
int event_id = VariableOrVariableBitfield(com, 7, 0, 0);
// and so on...
```

See PlayBGM:

https://github.com/EasyRPG/Player/blob/master/src/game_interpreter.cpp#L1978-L1989

Ghabry (approved):

PR #3332, opened 2025-01-16
Wii: Add SDL2 support

Ghabry:
When targeting Wii the possible ``PLAYER_TARGET_PLATFORM`` are now ``SDL2`` (default) and ``SDL1``. SDL1 is still supported for testing purposes (and the code was already written ^^)

For ``PLAYER_AUDIO_BACKEND`` the default is as before ``AESND``, but ``SDL2/SDL1`` can be also selected (I moved the mutex code a bit, hopefully this fixes the random crashes when changing maps).

----

Besides the build system this is not very invasive, only had to change the button mapping and thats all.

_Note_: The CMakeLists changes are quite large because this is already the preparation for the SDL3 support. (initially I had Wii SDL2 and PC SDL3 commit combined)

----

Prebuilt for testing: https://easyrpg.org/downloads/tmp/wii-test.zip  (contains both a SDL2 and a SDL1 build)

- SDL2_Aesnd: The latest build with aesnd audio backend
- SDL2_Audio: The latest build with sdl2audio backend
- SDL1

Before you open SDL2 for the first time delete the file ``/data/easyrpg-player/config.ini`` on your SD card (or when the file is missing just do nothing) or input will be broken.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -555,7 +555,10 @@ elseif(NINTENDO_SWITCH)
 elseif(VITA)
 	set(PLAYER_TARGET_PLATFORM "psvita" CACHE STRING "Platform to compile for.")
 elseif(NINTENDO_WII)
-	set(PLAYER_TARGET_PLATFORM "wii" CACHE STRING "Platform to compile for.")
+	set(PLAYER_TARGET_PLATFORM "SDL2" CACHE STRING "Platform to compile for. Options: SDL2 SDL1")
+	set(PLAYER_AUDIO_BACKEND "AESND" CACHE STRING "Audio system to use. Options: SDL2 SDL1 AESND OFF")
+	set_property(CACHE PLAYER_TARGET_PLATFORM PROPERTY STRINGS SDL2 SDL1)
+	set_property(CACHE PLAYER_AUDIO_BACKEND PROPERTY STRINGS SDL2 SDL1 AESND)
 elseif(NINTENDO_WIIU)
 	set(PLAYER_TARGET_PLATFORM "SDL2" CACHE STRING "Platform to compile for.")
 elseif(AMIGA)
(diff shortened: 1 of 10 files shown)


fdelapena (approved):

carstene1ns:
Since the SDL2 ogc driver uses libaesnd anyway it should not make much difference, because both use our genericaudio class. I have somewhere a custom audio backend like for the 3ds, which can be used when we drop Wii SDL support completely and have a custom video backend (i have some code bitrotting for this as well).

CMakeLists.txt
```diff
@@ -851,8 +897,7 @@ target_link_libraries(${PROJECT_NAME} PIXMAN::PIXMAN)
 
 # Always enable Wine registry support on non-Windows, but not for console ports
 if(NOT CMAKE_SYSTEM_NAME STREQUAL "Windows"
-	AND NOT ${PLAYER_TARGET_PLATFORM} MATCHES "^(psvita|3ds|switch|wii)$"
-	AND NOT NINTENDO_WIIU)
+	AND NOT PLAYER_CONSOLE)
```
carstene1ns, line 900: `PLAYER_CONSOLE` vs. `PLAYER_CONSOLE_PORT`.

src/decoder_oggvorbis.cpp
```diff
@@ -207,7 +207,14 @@ int OggVorbisDecoder::FillBuffer(uint8_t* buffer, int length) {
 #ifdef HAVE_TREMOR
 		read = ov_read(ovf, reinterpret_cast<char*>(buffer + length - to_read), to_read, &section);
 #else
-		read = ov_read(ovf, reinterpret_cast<char*>(buffer + length - to_read), to_read, 0/*LE*/, 2/*16bit*/, 1/*signed*/, &section);
+#  if defined(__WIIU__)
+		// FIXME: This is the endianess of the audio and not of the host but the byteswapping in ov_read does
+		// not sound like it works
+		int byte_order = 1; // BE
```
carstene1ns, line 213: I had this kind of problems when porting as well. Somehow the endianess is broken in the toolchain or the build scripts.

carstene1ns (approved):

carstene1ns:
So, do we merge in half broken state now?
Fortunately only the wine registry is enabled, but the builders need some adaptions.

Ghabry:
Also not happy with this situation... I'll add the ``PLAYER_CONSOLE_PORT`` part to the SDL3 PR.

And I rushed a quickfix for the CI builders (currently still set to SDL1)

PR #3339, opened 2025-01-22
Add SDL3 support

Ghabry:
My last change before xmas is still compiling, so just opening a PR here for it.

~Depends on the Wii SDL2 branch~


CMakeLists.txt
```diff
@@ -720,22 +736,26 @@ else()
...
+		src/platform/sdl/sdl1_audio.cpp
+		src/platform/sdl/sdl1_audio.h)
 elseif(${PLAYER_AUDIO_BACKEND} STREQUAL "AESND")
 	target_sources(${PROJECT_NAME} PRIVATE
 		src/platform/wii/audio.cpp
 		src/platform/wii/audio.h)
 	target_compile_definitions(${PROJECT_NAME} PUBLIC AUDIO_AESND=1)
 elseif(${PLAYER_AUDIO_BACKEND} STREQUAL ${PLAYER_TARGET_PLATFORM})
 
-elseif(${PLAYER_AUDIO_BACKED} STREQUAL "OFF")
```
carstene1ns: At least this typo fixes the 3DS build :D

CMakeLists.txt
```diff
@@ -579,7 +579,10 @@ if(ANDROID AND PLAYER_GRADLE_BUILD)
 	endforeach(f)
 endif()
 
-if(${PLAYER_TARGET_PLATFORM} STREQUAL "SDL3")
+if(${PLAYER_TARGET_PLATFORM} STREQUAL "Unset")
+	get_property(PLAT_PROP CACHE PLAYER_TARGET_PLATFORM PROPERTY STRINGS)
+	message(FATAL_ERROR "Please specify the target platform with -DPLAYER_TARGET_PLATFORM.\nOptions: ${PLAT_PROP}")
```
carstene1ns, line 584: This needs documentation updates.

Ghabry:
Also have to fix the CMake presets. Will add presets for at least SDL3 and SDL2 (if applicable) and a "default" preset where we can use a recommended default.

Since inception of the presets this feature did not improve, so it will be more cartesian product preset generation (hooray).

Ghabry:
This is now finished.

It still uses SDL2 by default as I don't want to get SDL3 up and working right now for all our ports. Too much work.

Also isn't tested as much as SDL2 yet.

src/platform/sdl/sdl2_audio.cpp
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	SetFormat(have.freq, sdl_format_to_format(have.format), have.channels);
+
+	// Start Audio
+	SDL_PauseAudioDevice(audio_dev_id, 0);
+}
+
+Sdl2Audio::~Sdl2Audio() {
+	SDL_CloseAudioDevice(audio_dev_id);
```
Ghabry, line 117: On Arch they replaced sdl2 with sdl2-compat (SDL2 ontop of SDL3). This call aborts now with "Invalid Pointer" (exactly the same reason why I disabled that line in the SDL3 backend)

Ghabry:
Found the related issue: https://github.com/libsdl-org/sdl2-compat/issues/313

``SDL_Quit`` deletes now all the audio stuff. Manually doing ``audio_.reset()`` before we call ``SDL_Quit`` fixes it. :+1:

fdelapena (approved):

PR #3340, opened 2025-01-22
Android: Reload native libraries on resume

Ghabry:
Otherwise gives UnsatisfiedLinkError when the app was killed due to OOM etc.

Is harmless as you can just restart the app and it will work but its the most common crash in the dev console, so lets fix it...

----

to reproduce you can just background the app while in the settings, then ``am kill org.easyrpg.player`` and selecting "EasyRPG folder" will crash.


carstene1ns (approved):
Works for me.

PR #3341, opened 2025-01-23
PROCESS JSON COMMAND - New Features

jetrotal:
The Operations were updated to:

0: GET VALUE
1: SET VALUE
2: GET LENGTH
3: GET KEYS
4: GET VAR TYPE
5:  REMOVE VALUE
6: PUSH INTO ARRAW
7: POP FROM ARRAY
8: CHECK IF VALUE EXISTS

There are also new Flags:
 - Extract data from json path - similar to how stringvars extracts \v[1], \t[2], etc.
 - Prettify json - to make its output easier to read.


Ghabry:
We had this discussion before: Exceptions (``try-catch``) are not enabled on emscripten to reduce binary size.

(and our homebrew platforms but they do not have JSON support enabled so you do not see them failing here)

jetrotal:
Sorry about the confusion, Now the code is way closer to how initially was

jetrotal:
I found some OOB crashes from the older implementation,
When trying to reach `/lists/data/6` from:
```json
"lists": {
            "length": 5,
            "data": [{
                    "id": "MAP TYPE",
                    "length": 7,
                    "values": ["EasyRPG World", "EasyRPG Exterior", "EasyRPG Interior", "EasyRPG Dungeon", "EasyRPG Ship", "Uncanny World", "Danny's House"],
                    "chipset_files": ["EasyRPG_World", "EasyRPG_Exterior", "easyRPG_interiorB", "EasyRPG_Dungeon", "EasyRPG_Ship", "EasyRPG+gfragger_mkt_CCBY", "EasyRPG+kotatsuAkira_Interior_CC0"],
                    "tileset_locations": [1531, 1535, 1534, 1533, 1532, 523, 524],
                    "current": 0
                },
                {
                    "id": "TILES LAYER",
                    "length": 3,
                    "values": ["LOWER", "UPPER", "EVENTS"],
                    "current": 0
                },
                {
                    "id": "DRAW AUTOTILES",
                    "length": 2,
                    "values": ["YES", "NO"],
                    "current": 0
                },
                {
                    "id": "MAP PANORAMA",
                    "length": 4,
                    "values": ["[OFF]", "A", "B", "C"],
                    "current": 0
                },
                {
                    "id": "BACKGROUND MUSIC",
                    "length": 4,
                    "values": ["[OFF]", "A", "B", "C"],
                    "current": 0
                }
            ]
        },
```


I thought about adding some error messages, but those messages could conflict with cases such as:

```json
[[-6,-6,-6,-6,-6,-6,-6,-6,-6,-6],[6,-6,-6,null,-6,-6,-6],[-6,-6,-6,-6,-6,-6,-6,null,null,null,-6,-6],[-6,6,-6,null,-6,-6,-6,-6,-6,-6,-6,-6],[-6,6,6,null,14,14,14,-6,14,14,14,-6,14,14,14,null,14,14,14,6],[6,6,6,null,null,14,6,null,14,null,null,null,14,-6,-6,-6,-6,14]]
```

where I wanted to have null values to fast walk through a list of map tiles.

Ghabry:
To _fix_ (disable) nlohmann for Ubuntu 20.04 you must require a newer version of nlohmann. 3.9.1 appears to work.

In CMakeLists:

```cmake
	player_find_package(NAME nlohmann_json
		VERSION 3.9.1
		CONDITION PLAYER_WITH_NLOHMANN_JSON
		DEFINITION HAVE_NLOHMANN_JSON
		TARGET nlohmann_json::nlohmann_json
		ONLY_CONFIG
	)
```

There are 2 cases where nlohmann_json is used in the file. Please add ``VERSION`` to both.

configure.ac:

```
EP_PKG_CHECK([NLOHMANN_JSON],[nlohmann_json >= 3.9.1],[Support processing of JSON files.])
```

src/game_interpreter.cpp
```diff
@@ -5357,12 +5357,57 @@ bool Game_Interpreter::CommandEasyRpgProcessJson(lcf::rpg::EventCommand const& c
...
+
 	int operation = ValueOrVariable(com.parameters[0], com.parameters[1]);
 	int source_var_id = ValueOrVariable(com.parameters[2], com.parameters[3]);
 	int target_var_type = ValueOrVariable(com.parameters[4], com.parameters[5]);
 	int target_var_id = ValueOrVariable(com.parameters[6], com.parameters[7]);
 
 	std::string json_path = ToString(CommandStringOrVariable(com, 8, 9));
+
+	int extract_data_from_string = com.parameters[10];
+	bool pretty_print = com.parameters[11] == 1;
```
Ghabry, line 5402: ``pretty_print`` could be just a flag in ``parameters[10]``

src/game_interpreter.cpp
```diff
@@ -5377,54 +5422,124 @@ bool Game_Interpreter::CommandEasyRpgProcessJson(lcf::rpg::EventCommand const& c
...
+	}
+	case 3: { // GetKeys operation
+		auto keys = Json_Helper::GetKeys(*json_data, json_path);
+		if (keys) {
+			std::string keys_str;
+			for (size_t i = 0; i < keys->size(); ++i) {
+				if (i > 0) keys_str += ",";
+				keys_str += "\"" + (*keys)[i] + "\"";
+			}
+			std::string json_str = "{ \"keys\": [" + keys_str + "] }";
```
Ghabry, line 5466: It's not necessary to prefix the output with

```
{
   "keys": [....]
}
```

an array is already a valid JSON object, so ``json_str = keys_str`` works.

src/game_interpreter.cpp
```diff
@@ -5377,54 +5422,124 @@ bool Game_Interpreter::CommandEasyRpgProcessJson(lcf::rpg::EventCommand const& c
...
+		}
+		break;
+	}
+	case 4: { // GetType operation
+		auto type = Json_Helper::GetType(*json_data, json_path);
+		if (type) {
+			std::string value;
+			if (target_var_type == 0) {
+				// For switches, true if it's an object or array
+				value = (*type == "object" || *type == "array") ? "1" : "0";
```
Ghabry, line 5477: This doesn't really make sense? Maybe just do not support Switch output for this or just use the same logic as for ``target_var_type == 1`` (always false, except when it exists)

src/json_helper.cpp
```diff
@@ -25,76 +25,376 @@
...
+		if (!json_obj.contains(ptr)) {
+			return {};
+		}
+
+		const json& value = json_obj[ptr];
+		if (value.is_discarded()) {
+			return {};
+		}
+		return value.is_array();
+	}
```
Ghabry, line 197: ``IsObject`` and ``IsArray`` is unused. Can they be removed?

src/json_helper.cpp
```diff
@@ -25,76 +25,376 @@
...
+			return {};
+		}
+
+		json popped = array[array.size() - 1];
+		array.erase(array.size() - 1);
+
+		return GetValueAsString(popped);
+	}
+
+	std::optional<bool> Contains(const nlohmann::ordered_json& json_obj, std::string_view json_path) {
```
Ghabry, line 364: Is their a reason why this uses a custom parser instead of ``json_obj.contains``?

src/game_interpreter.cpp
jetrotal: I added the "Keys" container, improve redability on the events editor:
image

It was becoming really annoying to maintain whenever I stacked multiple similar commands together.

src/game_interpreter.cpp
jetrotal: hm... I guess pretty_print and extract_data can be useful together...

Ghabry:
Another observation:

- The operations ``Get|Length|Keys|Type`` do nothing when the JSON path is missing
- The operations ``Set`` and ``Push`` do not modify the JSON when the source string is empty (Bug?)
- The operation ``Pop`` does not assign and not update the JSON when the popped element is an empty string (Bug?)

Ghabry:
btw to handle the complexity I wrote unit tests to verify the JSON Api (will commit them when the remarks are addressed).

Can confirm that the rest of it appears to work :+1: . Good job.

----

Also found the ``GetPath`` function which looks incomplete (will not find elements when they are integers) and is unused. Guess you started working on this and then decided to not finish it. But this looks useful later so can imo stay this way for now.

Ghabry (approved):

PR #3346, opened 2025-02-01
New Operations for ControlVariables / "ControlVariablesEx"

florianessl:
Some suggested new operations for a custom "**ControlVariablesEx**" command.
(These are currently disabled for the vanilla one, Command Code '2022' has to be used)

### '210': DateTime: Jetrotal's "_GetTime_" command
Cherry Picked & refactored Jetrotal's extension from PR #3124 , so it can be used like any other variable operation.
In the comments I included a suggestion, for how this could be adapted to also allow name-based lookup via strings.

### ~'211': ActiveMapInfo~
~_Width_ & _Height_ are two important infos about the current map I've long desired for ControlVariables.
I've seen that in the meantime, some new Maniac command has been added, that also finally allows access to these, but that seems to be a specialized command that has its own rules..
And with these considered, access to LoopHorizontal / LoopVertical could also prove very useful. (I can think of at least one case, at least)~

### ~'212': InspectMapInfo~
~This command provides access to some information that is stored inside the map tree.
It could still be expanded in functionality, but the most useful one (for my own purposes at least), I think would be the ability, to read Troop information from the current Map.
While I was developing the "_Community Edition_" for Vampires Dawn this was a pretty annoying hindrance, because I wanted to implement an easy way to just switch between normal RNG battles & overworld encounters. I basically had to write a custom code generation tool to be able to do so...~

### '213': MessageSystemInfo
Provides basic info about the system settings for message boxes.

### '214': MessageWindowInfo
Provides some information about a currently active message window.
Another issue I had, while developing a parallel event, that occassionaly displays some popup information. Normally, you haven't got an easy option of knowing if a message window is currently active & if so, where on the screen it is.
You can't really react & redirect your custom drawn picture/strings/etc. to another part of the screen, so that it won't overlap with the system textbox.

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -598,7 +598,7 @@ bool Game_Interpreter::ExecuteCommand(lcf::rpg::EventCommand const& com) {
 		case Cmd::ControlSwitches:
 			return CmdSetup<&Game_Interpreter::CommandControlSwitches, 4>(com);
 		case Cmd::ControlVars:
-			return CmdSetup<&Game_Interpreter::CommandControlVariables, 7>(com);
+			return CmdSetup<&Game_Interpreter::CommandControlVariables<false>, 7>(com);
 		case Cmd::TimerOperation:
 			return CmdSetup<&Game_Interpreter::CommandTimerOperation, 5>(com);
 		case Cmd::ChangeGold:
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -1070,7 +1070,7 @@ bool Game_Interpreter::CommandControlVariables(lcf::rpg::EventCommand const& com
 	int operand = com.parameters[4];
 
 	if constexpr (EasyRpgEx) {
-		if (EP_UNLIKELY(operand >= 200) && !ControlVariables::EasyRpgExCommand(com, value)) {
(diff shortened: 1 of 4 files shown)


florianessl:
Nearly forgot..
Here's a small test project for these new operations:
[ControlVarsEx.zip](https://github.com/user-attachments/files/18630209/ControlVarsEx.zip)

florianessl:
> btw, you spoke about retrieve data based on string, maybe this could work as an replacement for #3214

Yeah, I'll give it some thought when I get back to further developing that feature. (And I'll also consider compatibility/extending the new JSON command as well)
Right now I have to examine my priorities again & focus on some other stuff.

The first (very rough) draft for that feature can be found in my personal repo here:
- **liblcf**: https://github.com/florianessl/liblcf/tree/features/LcfInspection
- **Player**: https://github.com/florianessl/Player/tree/features/LcfInspection

jetrotal:
@florianessl this suggestion could also fit this pr:

Someone on YNO discord asked about a way of detecting if a window is on focus or tabbed out.

They wanted to make some events synched by music and such.

I guess the best way to discover what can influence this variable setting, is the "Pause on Lost Focus" video setting, It's not present on the webplayer, idk why.

PR #3348, opened 2025-02-05
Translation Extensions & Issues Pt 1

### LINKED ISSUE #2945: Feature: Open translations menu via custom command
I played around with the new titlecreen-skipping-Feature (option "NewGame=1") and combined with the support for the LoadMenu & ExitGame commands it looks like its possible to completely replace the default screen of RPG_RT.  (Even the new EasyRPG settings page is accessible via RPG2k3s "Open Video Options" command)
But when using EasyRPGs translation feature, there seems to be no way around using the default title screen as there seems to be no other way to access this menu.
Would it be possible to create a custom command for this?

### LINKED ISSUE #3097: Translation - Allow the title of a game to be translated depending on the language selected
When using the translation feature for a game, the name of the window (only displayed on desktop platforms) is kept as-is, which may be weird when the name of a game is not the same between a language A and a language B.
image

It would need to be changed when a translation is loaded, and adding a field in the meta.ini of a translation for the title could potentially work for that.

### LINKED ISSUE #3213: Display a warning instead of an error when the argument --language {LANGUAGE} is used and that no Language folder exists
#### Player platform:

Using the argument `--language {LANGUAGE}` on a game that doesn't has a Language folder will display an error forcing the game to be closed, would probably be better to change it to a warning similar to what is done for using said command while the language specified doesn't exist.

florianessl:
This PR collects all the minor patches I've done recently on the Player´s translation feature.
Basically I've split my working branch into 3, depending on how deep the changes would go.

Pt.2 would contain some changes to liblcf, to be able to store the selected language setting inside savefiles
Pt.3 is highly experimental & does all the unsafe, live patching of the interpreter state, to be able to switch languages mid-game

And if Pt.3 works out, maybe I'll proceed to experiment with some sort of AI-based auto-translation feature. Let's see.  

### Opening the language selection menu via custom command (Issue #2945):

Previously, the language selection has been hardcoded to be part of the title scene. This was an issue for games that use the "NewGame" flag & implemented their own title screen in-engine.
This PR implements a new scene type for the langue menu and also adds a way for this custom scene to be accessed via either of the currently provided "custom menu" options:
- Maniac patch extension to **_"Open Save menu"_** (Code 11910) -> param[0] = 207 (Patch "Maniac" & "EasyRpg Extensions" have to be active)
- **_"DirectMenu"_** patch -> [Patch variable] = 7

Also, I added support for accessing EasyRPGs "Settings" menu in this manner. This was previously commented out, but can now also be accessed via both of these ways, by using value "206" (Save Menu Cmd) or "6" (DirectMenuPatch)

### Turning of the "Language" menu on title screen.

See feature request here: https://community.easyrpg.org/t/turning-off-the-menu-option-created-by-translation/1277

Two new config options have been added:
- "LanguageInTitle"
- "StartupLangSelect"

"LanguageInTitle" can be used to remove the "Language" options from the title screen entirely, similarly to the previously existing 'SettingsInTitle' config option.

As an alternative, a new way of accessing the Language select scene has been added via the "StartupLangSelect" option. If set, the menu is shown right before the actual title screen of the game is boot up.
Options are: _"Never", "FirstStartup", "Always"._ 
"FirstStartup" will try to look for existing save files & only show the language screen, if none are found.

### Generate only a warning when no Language folder is found (Issue #3213)

As described in the issue, the player would generate an error & thus force the Player to exit, if it was started with the "--language" option for a game where no "Language" folder exists. This would also be an issue for loading up a savegame where the "translation" field is set.

### Translating a game's title (Issue #3097)

A new field in "meta.ini" can be used to translate the game title into different languages. As in the vanilla RPG_RT.ini, the name for this option is "GameTitle".

### Dynamic language switching in-game

In addition to the custom ways of accessing the language menu in-engine, I also added a new option to EasyRPG´s F1 Settings screen. This can't be deactivated at the moment, maybe another config-flag should be added for this? :thinking: 
This doesn't change the way translation is handled in-game (yet), so for the language-switch to be fully active, one has to leave the current map, so that the Player can properly refresh any currently active events.

For a futue PR I'd like to overhaul the way the language-switch works when in-game, so that even currently active interpreter stacks can be fully switched to another translation. I'll probably add a new command too for this, so prepare for some weird test game experiences where the game just randomly switches between languages in unexpected ways. xD

----

Fix #2945 
Fix #3097 
Fix #3213

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -306,6 +306,8 @@ add_library(${PROJECT_NAME} OBJECT
 	src/scene_import.h
 	src/scene_item.cpp
 	src/scene_item.h
+	src/scene_language.cpp
+	src/scene_language.h
 	src/scene_load.cpp
 	src/scene_load.h
 	src/scene_logo.cpp
--- a/Makefile.am
+++ b/Makefile.am
@@ -287,6 +287,8 @@ libeasyrpg_player_a_SOURCES = \
 	src/scene_gameover.h \
 	src/scene_item.cpp \
 	src/scene_item.h \
+	src/scene_language.cpp \
+	src/scene_language.h \
 	src/scene_load.cpp \
 	src/scene_load.h \
 	src/scene_logo.cpp \
--- a/src/game_config.cpp
+++ b/src/game_config.cpp
@@ -664,6 +664,8 @@ void Game_Config::LoadFromStream(Filesystem_Stream::InputStream& is) {
 	player.settings_autosave.FromIni(ini);
 	player.settings_in_title.FromIni(ini);
 	player.settings_in_menu.FromIni(ini);
+	player.lang_select_on_start.FromIni(ini);
+	player.lang_select_in_title.FromIni(ini);
 	player.show_startup_logos.FromIni(ini);
 	player.font1.FromIni(ini);
 	player.font1_size.FromIni(ini);
@@ -751,13 +753,13 @@ void Game_Config::WriteToStream(Filesystem_Stream::OutputStream& os) const {
 	player.settings_autosave.ToIni(os);
 	player.settings_in_title.ToIni(os);
 	player.settings_in_menu.ToIni(os);
+	player.lang_select_on_start.ToIni(os);
+	player.lang_select_in_title.ToIni(os);
 	player.show_startup_logos.ToIni(os);
 	player.font1.ToIni(os);
 	player.font1_size.ToIni(os);
 	player.font2.ToIni(os);
 	player.font2_size.ToIni(os);
-	player.log_enabled.ToIni(os);
-	player.screenshot_scale.ToIni(os);
 
 	os << "\n";
 }
--- a/src/game_config.h
+++ b/src/game_config.h
@@ -58,6 +58,14 @@ namespace ConfigEnum {
 		All
 	};
 
+	enum class StartupLangSelect {
+		Never,
+		/* Shows language screen when no saves are found */
+		FirstStartup,
+		/* Always show the language screen before the title */
+		Always
+	};
+
 	enum class ShowFps {
 		/** Do not show */
 		OFF,
@@ -89,6 +97,12 @@ struct Game_ConfigPlayer {
 	RangeConfigParam<int> font1_size { "Font 1 Size", "", "Player", "Font1Size", 12, 6, 16};
 	PathConfigParam font2 { "Font 2", "The game chooses whether it wants font 1 or 2", "Player", "Font2", "" };
 	RangeConfigParam<int> font2_size { "Font 2 Size", "", "Player", "Font2Size", 12, 6, 16};
+	EnumConfigParam<ConfigEnum::StartupLangSelect, 3> lang_select_on_start{
+		"Startup Language Select", "Show the language select before starting the game", "Player", "StartupLangSelect", ConfigEnum::StartupLangSelect::FirstStartup ,
+		Utils::MakeSvArray("Never", "FirstStartup", "Always"),
+		Utils::MakeSvArray("never", "firststartup", "always"),
+		Utils::MakeSvArray("Never show language menu on start", "Show language menu on first start (when no save files are found)", "Always show the language menu prior to the title screen") };
+	BoolConfigParam lang_select_in_title{ "Show language menu on title screen", "Display language menu item on the title screen", "Player", "LanguageInTitle", true };
 	BoolConfigParam log_enabled{ "Logging", "Write diagnostic messages into a logfile", "Player", "Logging", true };
 	RangeConfigParam<int> screenshot_scale { "Screenshot scaling factor", "Scale screenshots by the given factor", "Player", "ScreenshotScale", 1, 1, 24};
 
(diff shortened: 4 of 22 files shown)


Ghabry:
Good way to test is using the "EasyRPG Dev Simulator" as it has a German translation:

https://easyrpg.org/play/pr3348/?game=strpic

Mimigris:
Could also try with Braingirl using translations from YNOproject, the game skips the title screen so it could work for this use-case. Guess I could some testings on my side too if I have time and don't forget.

Ghabry:
Works for Braingirl.

Notable is that the Window is centered instead of the way the title screen does it.

Maybe could be altered because of aesthetics.

grafik

Mimigris:
> Works for Braingirl.
> 
> Notable is that the Window is centered instead of the way the title screen does it.
> 
> Maybe could be altered because of aesthetics.
> 
> grafik

Would probably work better for a lot of games seeing that the title screen graphic is the basis, though this case is not very much representative since said title screen picture is unused outside of those settings menu in EasyRPG due to the actual title screen being skipped.

florianessl:
I updated the options in **Window_Settings** & possibly (?) also fixed the async issue that caused the Web-Player to fail (I don't have a setup locally to test anything based on Emscripten at the moment but the changed lines might be a cause)

Because of the reliance on the title graphic & different needs, I gave this some thought & came up with some more suggestions for possible customizations. (Maybe we'd should open a ticket for these)

## Suggested customization options for this scene: 

### Defining a background that is to be used for the "Language" scene:
 - This graphic would be picked from the "Title" folder.
 - If not set, then the standard title graphic is still used.
 
### Overriding the "System" graphic to be used for the "Language" scene:
 Depending on the choice of background, a dev might also want to use a different system file for this Scene.

### Placement of the "Choices" box:
 - Some rough options like "Top", "Middle", "Bottom" or rather some exact X/Y placement?
-  Alignment of the window should probably an option. Otherwise, if it defaults to "Center", any time a new language is added, new actual X/Y coordinates would take effect. And depending on the choice of background this might cause undesired results.
  
--> _Changing the placement of the choices window is something that would probably also be useful to finally implement for the standard Title Scene_.
As of now, the standard choices here have been extended with three new options in EasyRPG Player: "Settings", "Import" & "Language" which mess with some title designs. Also there seem to be plenty of patched RPG_RT with changed title screens. 

## Suggested new commands:
**Also, new command suggestions,** so developers can implement their own language selection if they desire to do so:

- ChangeLanguage
- GetLanguageInfo (Output available languages/meta info to vars & strings)

florianessl:
Positioning of the language menu is now reverted back to the default "title screen" positioning.
The original commit for this scene refactoring was quite old, so I didn't realize at first, that I had changed the previous default behavior of this menuing code.

Ghabry:
Here fixes for emscripten. This resolves that the 2nd language switch crashes and that the title screen has a black font after the 2nd switch.

```diff
diff --git a/src/scene_language.cpp b/src/scene_language.cpp
index edbfcfb39..d001cca64 100644
--- a/src/scene_language.cpp
+++ b/src/scene_language.cpp
@@ -40,10 +40,6 @@
 #include "window_settings.h"
 #include <memory>
 
-#ifdef EMSCRIPTEN
-#  include <emscripten.h>
-#endif
-
 Scene_Language::Scene_Language() {
 	Scene::type = Scene::LanguageMenu;
 }
@@ -125,6 +121,11 @@ void Scene_Language::Start() {
 
 
 void Scene_Language::vUpdate() {
+	if (shutdown) {
+		PopOrTitle();
+		return;
+	}
+
 	translate_window->Update();
 	help_window->Update();
 
@@ -144,6 +145,10 @@ void Scene_Language::vUpdate() {
 	}
 }
 
+void Scene_Language::OnTranslationChanged() {
+	Main_Data::game_system->ReloadSystemGraphic();
+}
+
 void Scene_Language::OnTitleSpriteReady(FileRequestResult* result) {
 	BitmapRef bitmapRef = Cache::Title(result->file);
 
@@ -169,8 +174,11 @@ void Scene_Language::ChangeLanguage(const std::string& lang_str) {
 
 	// First change the language
 	Player::translation.SelectLanguage(lang_str);
+	Main_Data::game_system->ReloadSystemGraphic();
 
-	PopOrTitle();
+	// Delay scene shutdown by one frame to allow async requests for language change
+	// and system graphic reload to finish before the scene switches
+	shutdown = true;
 }
 
 void Scene_Language::PopOrTitle() {
diff --git a/src/scene_language.h b/src/scene_language.h
index fd53e4b12..6405de7b3 100644
--- a/src/scene_language.h
+++ b/src/scene_language.h
@@ -45,6 +45,8 @@ public:
 	void Start() override;
 	void vUpdate() override;
 
+	void OnTranslationChanged() override;
+
 private:
 
 	/**
@@ -83,6 +85,7 @@ private:
 	std::unique_ptr<Sprite> title;
 	FileRequestBinding request_id;
 	int input_reset_counter = 0;
+	bool shutdown = false;
 
 	Window_Settings::UiMode mode = Window_Settings::eNone;
 };
diff --git a/src/scene_settings.cpp b/src/scene_settings.cpp
index ecd73dbb1..5d77eb903 100644
--- a/src/scene_settings.cpp
+++ b/src/scene_settings.cpp
@@ -260,6 +260,7 @@ void Scene_Settings::vUpdate() {
 		case Window_Settings::eSave:
 		case Window_Settings::eEnd:
 		case Window_Settings::eAbout:
+		case Window_Settings::eLanguage: // fix compiler warning, not implemented
 			break;
 		case Window_Settings::eMain:
 			UpdateMain();
```

Ghabry (approved):

fdelapena (approved):

PR #3349, opened 2025-02-06
Maniac´s debugging features: GetCommandInterpreterState (GetGameInfo), maniac_event_info & more debugging extensions

florianessl:
First two commits of this PR were already included with my recent PR [:Debug_AssertWay](https://github.com/florianessl/Player/tree/Debug_AssertWay). Basically summed up with refactoring work on existing functionality I implemented in the Interpreter Debug Window that has already been merged for a while; and setting the base with a new Debug namespace inside a new file "game_interpreter_debug.cpp"...

Next up (Commit #3): _"Game_Interpreter_Inspector"_ 
Also some refactoring work. This is to clear out some of the "_friend classes_" I've defined previously. The idea is to only allow access to internal interpreter states via this interface & make it clear that any use of these functions is considered unsafe. (I use this for my new experimental work on trying to translate/patch content of running interpreters. Instead of just declaring "_Translation_" as just another friend class of _Game_Event/Game_CommonEvent/Game_Interpreter/etc_. this approach is used..

Then we get to the new stuff:
I've implemented Maniac's "**Get Interpreter State**" sub-command for "**GetGameInfo**". This required me to fully implement the "**_maniac_event_info_**" field & refactor the way commands are pushed to the interpreter, so that an "_ExecutionType_" (Action, AutoStart, Parallel, etc..) + the type of event are always known & can be included inside the saved execution state.
I then went the extra route & did some more refactoring, so that all the outside calls to "Interpreter.Push" can be verified at compile time via static_assert´s. So if anyone tries to do something weird, it won't compile. 

The rest of the commits concern the Debug Scene`s Interpreter Window again, so the full functionality is restored after the refactoring & so that most of the new available state information is used there.

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -101,9 +101,9 @@ bool Game_Interpreter::IsRunning() const {
 
 // Setup.
 void Game_Interpreter::Push(
+	InterpreterPush push_info,
 	std::vector<lcf::rpg::EventCommand> _list,
 	int event_id,
-	bool started_by_decision_key,
 	int event_page_id
 ) {
 	if (_list.empty()) {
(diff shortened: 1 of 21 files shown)


Ghabry:
Could you fix the two compiler warnings?

src/window_interpreter.cpp
```diff
@@ -153,16 +189,21 @@ void Window_Interpreter::DrawDescriptionLines() {
...
+
+	if (sub_actions.IsVisible()) {
+		sub_actions.Draw(contents, rect);
 	}
 
 	rect = GetItemRect(i++);
 	contents->ClearRect(rect);
 
-	contents->TextDraw(rect.x, rect.y, Font::ColorDefault, "Stack Size: ");
+	auto str_ex_type = std::string(Game_Interpreter_Shared::kExecutionType.tag(static_cast<int>(stack_display_items[0].type_ex)));
```
Ghabry, line 202: Just opened the interpreter debug menu with the game from #3409 and it crashed here because stack_display_items is empty.

Ghabry:
Looks good to me :+1: . Checking this again when the Interpreter Flag PR is done

Ghabry:
the other PR is merged. Can you rebase it on the master branch to get rid of the redundant commits?

src/game_interpreter.cpp
```diff
@@ -114,15 +114,28 @@ void Game_Interpreter::Push(
...
+
 	lcf::rpg::SaveEventExecFrame frame;
 	frame.ID = _state.stack.size() + 1;
 	frame.commands = std::move(_list);
 	frame.current_command = 0;
-	frame.triggered_by_decision_key = started_by_decision_key;
-	frame.event_id = event_id;
+	frame.triggered_by_decision_key = type_ex == ExecutionType::Action;
+	if (type_src == EventType::MapEvent) {
+		frame.event_id = event_id;
```
Ghabry, line 126: is this added if-condition a bugfix?

florianessl: Yup.
"maniac_event_id" should always get initialized with the id of the calling event, even if the source type is a common event.
Previously, the callers of this "Push" function would decide wether to include the event id or just '0' (if it is not a map event). Now this revised "PushInternal" function knows about the source type & can decide for itself how to handle the inputs..

fdelapena (approved):

PR #3355, opened 2025-02-22
Logging: Use one shared logfile

Ghabry:
Instead of putting easyrpg_log.txt in every game directory the logfile is now in "APPDATA/Local" (Windows), "$XDG_STATE_DIR" (Linux/macOS) or "Config dir" (other).

Also disabled logging by default for our homebrew platforms as their were reports (at least for Wii) about FS corruption because our Player is sometimes a bit unstable...


Ghabry:
Bad news everyone: The lack of bug reports over the years is not because of the Player becoming better its because of Android breaking our code since Android 11. :sweat_smile: :sweat_drops:

fdelapena (approved):

fdelapena:
Besides the size_t comparison warnings, looks good to me :+1:

Ghabry:
Right. The goal should be to keep the code warning free before merge :)

src/game_config.cpp
```diff
@@ -260,6 +275,129 @@ Filesystem_Stream::OutputStream Game_Config::GetGlobalConfigFileOutput() {
...
+			} else {
+				home = getenv("HOME");
+				if (home) {
+					path = FileFinder::MakePath(home, ".local/state");
+				}
+			}
+
+			if (!path.empty()) {
+				path = FileFinder::MakePath(path, OUTPUT_FILENAME);
+			}
```
carstene1ns, line 312: missing "if empty" here?

This could potentially remove the `#else` case as well.

src/output.cpp
```diff
@@ -223,34 +189,12 @@ static void HandleErrorOutput(const std::string& err) {
...
-			size_t read = in.gcount();
-			in.Close();
-
-			auto out = FileFinder::Save().OpenOutputStream(OUTPUT_FILENAME, std::ios_base::out);
-			if (out) {
-				out.write(buf, read);
-			}
-		}
+	if (!Game_Config::GetLogFileOutput()) {
+		return;
```
carstene1ns, line 195: This makes no sense to me. I guess the closing brace is wrongly placed.

src/game_config.cpp
```diff
@@ -260,6 +275,129 @@ Filesystem_Stream::OutputStream Game_Config::GetGlobalConfigFileOutput() {
 	return Filesystem_Stream::OutputStream();
 }
 
+Filesystem_Stream::OutputStream& Game_Config::GetLogFileOutput() {
+	if (!Player::player_config.log_enabled.Get()) {
+		static Filesystem_Stream::OutputStream noop_stream;
+		return noop_stream;
```
carstene1ns, line 281: Can we have one `noop_stream` instead of repeating this?

carstene1ns (changes requested):

src/game_config.cpp
Ghabry: wanted to avoid a static variable when it is not used but you are right that this micro optimisation makes no sense. :sweat_smile:

src/output.cpp
Ghabry: right. This is nonsense (and already handled in ``Game_Config::CloseLogFile``

Ghabry:
Thanks! 2 of the 3 were some really trivial mistakes I did...

btw asked Copilot now to review the PR and it found _NONE_ of the flaws. Not even the obvious one in ``Output::Quit`` :facepalm:

carstene1ns (approved):

PR #3356, opened 2025-02-23
Android: Properly handle rotations

Ghabry:
ConfigurationChanged is not sent on all Android versions because of various bugs in the operating system. Instead SurfaceChanged is used which is consistent with what SDL does.

Also Window events are not filtered anymore when in the background. This is required on Android, otherwise rotations are lost.

---

Confirmed by lusciouslover in chat that the rotation issue is fixed.

### DIFF
--- a/builds/android/app/src/main/java/org/easyrpg/player/player/EasyRpgPlayerActivity.java
+++ b/builds/android/app/src/main/java/org/easyrpg/player/player/EasyRpgPlayerActivity.java
@@ -26,6 +26,7 @@ package org.easyrpg.player.player;
 
 import android.app.AlertDialog;
 import android.content.ClipDescription;
+import android.content.Context;
 import android.content.Intent;
 import android.content.res.AssetManager;
 import android.content.res.Configuration;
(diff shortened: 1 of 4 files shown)


fdelapena (approved):

Ghabry:
Here the Android documentation about this btw (with **remarks** by me)

----

On Android 7.0 (API level 24) and higher, Activity recreation only occurs for size-based configuration changes if the size change is significant **(?? What is Significant ??)**. When the system doesn't recreate an Activity due to insufficient size, the system might call [Activity.onConfigurationChanged()](https://developer.android.com/reference/android/app/Activity#onConfigurationChanged(android.content.res.Configuration)) and [View.onConfigurationChanged()](https://developer.android.com/reference/kotlin/android/view/View#onconfigurationchanged) instead.

There are some caveats to observe regarding the Activity and View callbacks when the Activity isn't recreated:

- On Android 11 (API level 30) through Android 13 (API level 33), Activity.onConfigurationChanged() isn't called. **(<-- !!!)**
- There is a known issue where View.onConfigurationChanged() may not be called in some cases on Android 12L (API level 32) and early versions of Android 13 (API level 33). This has since been addressed in later Android 13 releases and Android 14. **(<-- !!!)**

For code that is dependent on listening for size-based configuration changes, we recommend using a utility View with an overridden View.onConfigurationChanged() instead of relying on Activity recreation or Activity.onConfigurationChanged(). **Well yeah except when its broken on Android 13 -_-**

PR #3362, opened 2025-03-04
Implement undocumented mouse binding for the Ineluki patch

florianessl:
While going through different versions of harmony.dll, I found a few special versions of the Ineluki patch.
Most of them were just copy-protected preview versions, which were handed out to some devs, before the key patch was actually released to the public. Those have limited functionality & will refuse to work, if some of the game files do not match the expected values.

Just for documentation purposes, the unique versions I found on rmarchiv's data are:
- 1.2.0.50 **Gekiganger** (Used in the keypatch demo)
- 1.2.0.50 **Kelven** "a" (Used in "Die Bücher Luzifers 2")
- 1.2.0.50 **Kelven** "b" (Used in "Aliminator")
- 1.2.0.50 **Sunny** (Used in "Der Feind")
- 1.2.0.50 **KoA Angel** (Used in "Yoshis Maze Mania")
- 1.2.0.00 **Neo23** (Apparently used in a Moleboxed "Megaman" game by the developer)

The interesting one is the last version, custom-developed for "Neo23". But apparently, another version later surfaced, without the unique "copy-protection" for Neo23, and the same internal compilation date (2004-07-05), which differs by nearly half a year from the most commonly used keypatch version.

This DLL has two additional commands:
- _setMouseAsReturn_
- _setMouseWheelAsKeys_

...which feature a few options for binding mouse keys & the mouse wheel to common input buttons.
Usage of the mouse wheel is already possible in some menus in the Player, but I went & completely implemented the key bindings anyway. (This can be tested in-game in Rutipa's Quest 9 when you have control of the player. Save & restart, the game forgets to deactivate the mouse patch a second time, so the character can now be moved via the mouse wheel..)

This patch is used in the games **Vortex** & both versions of **Rutipa's Quest 9**.

### DIFF
--- a/src/game_ineluki.cpp
+++ b/src/game_ineluki.cpp
@@ -376,7 +376,7 @@ void Game_Ineluki::UpdateKeys() {
 }
 
 void Game_Ineluki::UpdateMouse() {
-#if defined(USE_MOUSE_OR_TOUCH) && defined(SUPPORT_MOUSE_OR_TOUCH)
+#if defined(USE_MOUSE) && defined(SUPPORT_MOUSE)
 	if (Input::IsRawKeyTriggered(Input::Keys::MOUSE_LEFT)) {
 		if ((mouse_decision_binding == MouseReturnMode::Left || mouse_decision_binding == MouseReturnMode::Both)) {
 			Input::SimulateButtonPress(Input::DECISION);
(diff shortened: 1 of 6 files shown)


Ghabry:
hmm I didn't want to add more to 0.8.1 but this looks simple enough to be added :thinking:

Ghabry:
You must guard the inner part of ``UpdateMouse`` inside a 

```
#if defined(USE_MOUSE_OR_TOUCH) && defined(SUPPORT_MOUSE_OR_TOUCH)

#endif
```

The enum values are not available when the support is disabled.

florianessl:
> You must guard the inner part of `UpdateMouse` inside a
> 
> ```
> #if defined(USE_MOUSE_OR_TOUCH) && defined(SUPPORT_MOUSE_OR_TOUCH)
> 
> #endif
> ```
> 
> The enum values are not available when the support is disabled.

Oops, sorry. I went AFK & didn't see that some of the builds did not succeed.

Ghabry (approved):

fdelapena (approved):

Issue #3266, opened 2024-09-02
Support File deletion through VFS 


jetrotal:
Good for Savefiles deletion and other future features.
Cloning a savefile between slots could also be interesting.

florianessl:
_Something to note here, for future compatibility:_
The "**PPCOMP**" tool (PR #3363) already features commands for deleting & copying save files. Though I skipped on the implementation of those two because I couldn't find any examples for games that use them.

Issue #1182, opened 2017-05-24
Patch Support: AEP (Auto Enter Patch)


Ghabry:
Probably best through some kind of database which contains a list of games that use thie patch.

This patch allows skipping the Title Scene and controlling Load and End game (for custom menus) depending on the variable 3350.

When "Stop Parallel Events" is invoked and 3350 == 1 -> Load, 3350 == 2-> End. Else: Normal Stop behaviour.

Patch signature in RPG_RT: "\x90\x90\x90\x90\x90\x90\x8b\x43\x14\x8b\x40\x44\x85\xc0\x90\x90\x90\x90\x90\x90"

adketuri:
While exploring the codebase I implemented this for my personal project. This is not an adequate solution for the task but I figure I'd link this here for reference:

https://github.com/EasyRPG/Player/commit/2e5fcfb67edcb5e29d6a60702da755a2030f2fe6

Ghabry:
Thanks, with an optional opt-in through a RPG_RT.ini setting your patch could be almost used directly.

hellow554:
I would recommend a command line argument as well 🙂

PR #3363, opened 2025-03-05
Support for ppcomp, LS.dat & BetterAEP

florianessl:
Support for the "Compact" version of Cherry's PowerPatch features. This version wasn't really a patch but just an external utility program named "**ppcomp.exe**" (sometimes renamed to _ppcomp.dat_ or _ppcomp.dll_) which provided access to some new functions by accessing the memory of the running game process.

### Detecting the original PowerPatch to give info about compatibility

This is not the full version of the original "**Power Patch**", which goes way deeper (including early Lua support) & barely found use in any games. But I also added a small heuristic check to detect the "Full" version & log a warning about limited compatibility:
- This just check for the existence of a file named "**_powerp.oex_**" in the game's root
- Another commit also checks for "**_warp.dll_**" which was bundled with _PowerMode2003_, another popular patch at this particular time period of RPG2003's lifecycle. _(Reversing & documenting the origins/versions of that patch actually seems quite tricky - more about that in a future issue...)_

### Supported PPCOMP functions
- [x] Quitting & Resetting the game ("QUIT", "RESTART")
- [x] Calling of in-game scenes 
  - [x] "DEBUG"
  - [x] "CALLLOADMENU"
  - [x] "CALLSAVEMENU"
  - [x] "CALLGAMEMENU" (+ setting the cursor position)
  - [x] "CALLTITLESCREEN" (+setting the cursor position)
- [x] Enabling / Disabling the "UnlockPictures" patch ("UNLOCKPICTURES")
- [ ] Load / Save operations
  - [x] Loading from source _[slot-id]_ ("LOAD")
  - [x] Saving to target _[slot-id]_ ("SAVE")
  - [x] Checking existence of Save _[slot id]_ ("CHECKSAVE")
  - [ ] Copying individual save files ("COPYSAVE")
  - [ ] Deleting individual save files ("DELETESAVE") _(Would require an update for FileFinder -> #3266 )_
- [x] GetSaveDateTime
- [x] GetSystemDateTime
- [ ] SimulateKeyPress (_Drafted, but not tested yet - allows for simulating keypresses for a targeted amount of time_)
- [ ] ChangeFunctionKey
- [ ] SetMsgBoxColor
- [ ] PauseGame
- [ ] PauseTimer
- [ ] SetVar

 **NOT PLANNED**: 
 - ~SetGlobalBrightness~ _(Apparently this was only implemented to address a blackscreen issue caused by the original patch)_
 - ~ChangeScene~ _(Sounds a bit too dangerous)_

For the remaining features, I only found instances of "DELETESAVE" & "SIMULATEKEYPRESS" actually being used in some archived games.
-> These are still TODO

### Adjustments to Ineluki KeyPatch implementation
- I refactored handling for script execution, so that the encountered _script.wav_ files aren't solely handled in the audio mechanism inside Game_System, but will be handled inside the interpreter, when a PlaySE command comes across them there. This is, so that it is possible to get some context for the result of executed scripts & to properly set the field "_async_op" in some cases.
- Async requests: While a script file isn't yet available, the PlaySE command will result in a "YieldRepeat" async op.
- Added support for _script.wav_ execution for the "PlayBGM" command. This is rarely seen in games, but is still possible & would have resulted in a warning by the Player when it tried to parse a script INI as Music.
- I also added support for the known program "LS.dat" which was another way many games with custom title screens used to call the Load Scene.

### BetterAEP patch support
While testing this implementation, it quickly became obvious, that many games that were using ppcomp, also relied on BetterAEP, and some used this patches' special functionality to access the Load Scene or even exiting the game via command. (Even though ppcomp also had these options).

So I added support for this patch as well:
`Player.exe --patch-better-aep 3350
`So far I've only seen games that use the default var-id of '3350' for this.
-> Issue #1182

### Some games using ppcomp
- La Leyenda Heroen (https://rmarchiv.de/games/2899)
- Mr. Hat (https://rmarchiv.de/games/183)
- Take It Cheesy (https://rmarchiv.de/games/1355)

### Info
- BetterAEP: https://www.makerpendium.de/index.php/Better_AEP
- Power Patch Compact: https://www.makerpendium.de/index.php?title=Power_Patch_Compact (See PDF download for documentation)
- More info about some obscure ppcomp versions was shared by @CherryDT in this issue: https://github.com/EasyRPG/Player/issues/1181

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -200,6 +200,8 @@ add_library(${PROJECT_NAME} OBJECT
 	src/game_pictures.h
 	src/game_player.cpp
 	src/game_player.h
+	src/game_powerpatch.cpp
+	src/game_powerpatch.h
 	src/game_quit.cpp
 	src/game_quit.h
 	src/game_screen.cpp
--- a/Makefile.am
+++ b/Makefile.am
@@ -176,6 +176,8 @@ libeasyrpg_player_a_SOURCES = \
 	src/game_pictures.h \
 	src/game_player.cpp \
 	src/game_player.h \
+	src/game_powerpatch.cpp \
+	src/game_powerpatch.h \
 	src/game_screen.cpp \
 	src/game_screen.h \
 	src/game_strings.cpp \
--- a/src/async_op.h
+++ b/src/async_op.h
@@ -43,6 +43,7 @@ class AsyncOp {
 			eTerminateBattle,
 			eSave,
 			eLoad,
+			eLoadParallel,
 			eYield,
 			eYieldRepeat,
 			eCloneMapEvent,
@@ -77,6 +78,9 @@ class AsyncOp {
 
 		/** @return a Load async operation */
 		static AsyncOp MakeLoad(int save_slot);
+
+		/** @return a LoadParallel async operation (to be used for patch compatibility) */
+		static AsyncOp MakeLoadParallel(int save_slot);
 
 		/** @return a Yield for one frame to e.g. fetch an important asset */
 		static AsyncOp MakeYield();
@@ -128,7 +132,7 @@ class AsyncOp {
 
 		/**
 		 * @return the desired slot to save or load
-		 * @pre If GetType() is not eSave or eLoad, the return value is undefined.
+		 * @pre If GetType() is not eSave, eLoad, eLoadParallel, the return value is undefined.
 		 **/
 		 int GetSaveSlot() const;
 
@@ -220,7 +224,7 @@ inline int AsyncOp::GetBattleResult() const {
 }
 
 inline int AsyncOp::GetSaveSlot() const {
-	assert(GetType() == eSave || GetType() == eLoad);
+	assert(GetType() == eSave || GetType() == eLoad || GetType() == eLoadParallel);
 	return _args[0];
 }
 
@@ -303,6 +307,10 @@ inline AsyncOp AsyncOp::MakeSave(int save_slot, int save_result_var) {
 
 inline AsyncOp AsyncOp::MakeLoad(int save_slot) {
 	return AsyncOp(eLoad, save_slot);
+}
+
+inline AsyncOp AsyncOp::MakeLoadParallel(int save_slot) {
+	return AsyncOp(eLoadParallel, save_slot);
 }
 
 inline AsyncOp AsyncOp::MakeYield() {
--- a/src/filefinder.cpp
+++ b/src/filefinder.cpp
@@ -399,16 +399,30 @@ bool FileFinder::HasSavegame() {
 int FileFinder::GetSavegames() {
 	auto fs = Save();
 
-	for (int i = 1; i <= 15; i++) {
-		std::stringstream ss;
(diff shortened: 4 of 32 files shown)


Ghabry:
The keystate stuff is already in maniac patch https://github.com/EasyRPG/Player/blob/master/src/maniac_patch.cpp#L674

So this could be unified in the file you proposed

florianessl:
> The keystate stuff is already in maniac patch https://github.com/EasyRPG/Player/blob/master/src/maniac_patch.cpp#L674
> 
> So this could be unified in the file you proposed

Yeah, I already moved that mapping code as part of another branch that's still WIP:
https://github.com/EasyRPG/Player/commit/adb3f378d598009ca33b021ac96e7f429671f0b1
I included a small "FIXME" note in my last PPCOMP commit.

florianessl (commented):
https://github.com/EasyRPG/Player/pull/3363/commits/99c559d073bc59c8e3e0e225086f640ac8426d0f

**Needs some explanation:**
While testing around with the ppcomp debug utility, I first came to the conclusion, that this command is meant to always force the Title Scene, even if something like BetterAEP (also made by Cherry) is used.
At least, I had managed this once in RPG_RT & then promptly added this new force flag to the Player, to ignore the "new-game" option set by other patches/configurations.

But then I couldn't reproduce this at all later. Using this command in a BetterAEP-patched RPG_RT seems to always skip the title screen.

Didn't mean to commit this line at first, but the flag could still be a useful feature.

florianessl:
> hm that the Ineluki script ran async is another bug. Makes no sense to execute a script this way. Though there are almost no games using Ineluki in the web player so was never noticed 🤔

KeyPatch scripts actually do have a setting for "Wait" that is ignored by EasyRPG  but I never checked the details of that one out for myself. So this could maybe lead to behavior changes in some occasions :/

giovannetti-eric:
In case a real user of BetterAEP is useful while this is in progress: *Final Fantasy VII Origin* (French RPG Maker 2003 fan game, RPG_RT 1.0.8.0) depends on it for two things, and both fail silently today.

It drives the patch entirely through the variable, never through Cherry's event commands — codes 5001–5005 appear nowhere in the project. Three writes to variable 3350, each followed by `End Event Processing`:

| where | value | intended | today |
| --- | --- | --- | --- |
| title screen, parallel event | 1 | open the load menu | cursor does nothing |
| in-game custom menu | 2 | exit the game | black screen, quit music, never exits |
| title screen, "new game" branch | 0 | nothing — a plain teleport follows | works |

The third is the useful one for testing: it is the same event and the same code path, and it behaves correctly, so a fix can be checked against a control that already passes.

Screenshots of the title screen under the current continuous build are attached, with the cursor on each of the two entries. Everything drawn works — the cursor moves, the artwork responds — so the gap is purely the action behind the entry, which is what makes it a clean test for this branch.

Happy to test a build against it — the game is not redistributable, but I can reach either behaviour in seconds and report back.

(Unrelated to this patch, I opened #3687 about the same game; mentioning it only so the two are not taken for the same report.)


Map0340


Map0340_continuer


Map0064__10_flat

PR #3365, opened 2025-03-06
Libretro related buildfixes and new presets

Ghabry:
This fixes some flaws encountered while updating the libretro core to the latest version on their buildbot.

They offer a recent m1 builder so I was able to test building a libretro core for iOS and tvOS. It works besides some minor issues that are fixed here.

iOS/tvOS presets are not included (more work needed to make them useful)


carstene1ns (approved):

Ghabry:
Minor update: Small change to the logging callback registration because of a bug in RetroArch... (surprise surprise)

Was also asked to disable our logfile by default as its redundant (RetroArch / other frontends put it onto their own logfile via that log_cb).

fdelapena (approved):

PR #3371, opened 2025-03-16
New Database terms for Maniacs

florianessl:
A few new term fields have been added in the latest version of Maniacs, to provide more customization options for some in-game texts:
- Experience gained
- Level Up
- Items received
- Skill learned

I would have preferred some template strings similar to the ones found in RPG2KE, but they do the job..

I moved those message texts to "battle_commands" which I then renamed to "game_message_terms". The terms code is all over the place & a lot of refactoring should be done here to make localization less of a hassle.
I'm sure this issue came up before & we could use this PR to do also some housekeeping here. 🤔

Maybe move everything related to in-game terms to its own naming scheme? (terms_battle.cpp, terms_actor.cpp, etc.)

New custom fields for _liblcf_:
```
Terms,maniac_item_received_a,f,DBString,0xA1,,0,1,
Terms,maniac_level_up_a,f,DBString,0xA2,,0,1,
Terms,maniac_level_up_b,f,DBString,0xA3,,0,1,
Terms,maniac_level_up_c,f,DBString,0xA4,,0,1,
Terms,maniac_exp_received_a,f,DBString,0xA5,,0,1,
Terms,maniac_skill_learned_a,f,DBString,0xA6,,0,1,
```

maniac_terms1
maniac_terms2

### DIFF
--- a/src/game_message_terms.cpp
+++ b/src/game_message_terms.cpp
@@ -14,15 +14,61 @@
  * You should have received a copy of the GNU General Public License
  * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
  */
-#include "battle_message.h"
+#include "game_message_terms.h"
 #include "player.h"
 #include "utils.h"
 #include "algo.h"
+#include "game_actor.h"
 #include "game_battler.h"
 #include <lcf/rpg/state.h>
 #include <lcf/data.h>
 #include <lcf/reader_util.h>
 #include "feature.h"
+
+namespace ActorMessage {
+
+std::string GetLevelUpMessage(const Game_Actor& actor, int new_level) {
+	std::stringstream ss;
+	if (Player::IsRPG2k3E()) {
+		ss << actor.GetName();
+		ss << " " << lcf::Data::terms.level_up << " ";
+		ss << " " << lcf::Data::terms.level << " " << new_level;
+		return ss.str();
+	} else if (Player::IsRPG2kE()) {
+		ss << new_level;
+		return Utils::ReplacePlaceholders(
+			lcf::Data::terms.level_up,
+			Utils::MakeArray('S', 'V', 'U'),
+			Utils::MakeSvArray(actor.GetName(), ss.str(), lcf::Data::terms.level)
+		);
+	} else {
+		std::string particle, space = "";
+		if (Player::IsCP932()) {
+			particle = "は";
+			space += " ";
+		} else {
+			particle = " ";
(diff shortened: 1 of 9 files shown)


florianessl:
Oops, it wasn't my intention to push this version, which already merges in some stuff not yet in master. I'll rebase & force-push that later.

Another thing I noticed:
_Feature::HasPlaceholders()_ really only affects messages of the battle scene. Makes sense, because the associated database field is called `easyrpg_**battle**_use_rpg2ke_strings`, but the way it is described in code doesn't make it immediately obvious.
I stumbled upon this after finally migrating VDCE to 2k3 & then had these broken placeholder Level-Up messages. I had to change it back to use the non-template way of stitching together strings. :/
The level-up messages are actually used inside & outside of the battle, so this is also a bit inconsistent. But just renaming the flag & make it affect all types of placeholder messages could maybe break the text of some games.

Ghabry:
Jenkins: test this please

src/game_message_terms.cpp
```diff
@@ -552,3 +614,57 @@ std::string GetEscapeStartMessage2k3(const Game_Battler& source) {
...
+
+std::string GetExperienceGainedMessage(int exp) {
+	if (Feature::HasPlaceholders()) {
+		return Utils::ReplacePlaceholders(
+			lcf::Data::terms.exp_received,
+			Utils::MakeArray('V', 'U'),
+			Utils::MakeSvArray(std::to_string(exp), lcf::Data::terms.exp_short)
+		);
+	}
+	std::string space = Player::IsRPG2k3E() ? " " : "";
```
Ghabry, line 628: Is this space here and in itemreceived a bugfix? I cannot see this in the old code.

florianessl: It's found right here in the 2k3 battle scene:
_src/scene_battle_rpg2k3.cpp_
https://github.com/EasyRPG/Player/pull/3371/files/60160e449ffe7bc41495021f852384f7f166261e#diff-815156950c5a8bd7f4c31336e5382d9ce351362c3f7cc8be5b9502091068c204R1857

Ghabry (approved):

Ghabry:
Test if link preview is off now

src/game_message_terms.cpp
Ghabry: A

carstene1ns (approved):
Looks good to me!

Always nice to see refactoring work.

PR #3377, opened 2025-03-22
Clear flash on non-flash pictures

### LINKED ISSUE #3376: Pictures affected by flash don't have their flash fully reset if the flash is interrupted by a map change
#### Player platform:
Windows, 64 bits continuous build of the player.

[Flash Issue Test.zip](https://github.com/user-attachments/files/19397321/Flash.Issue.Test.zip)

Start the testgame as provided. Interact with the sign on the left, then use the warp on the left, and interact with the harpy in front of you. A picture should be displayed, and between RPG_RT and EasyRPG, the picture shouldn't look the same.

RPG_RT:
Image

EasyRPG:
Image

It seems that if a picture is on-screen, set as being affected by flash, and set as being erased on map change, when using a warp while a flash is playing and doesn't get time to finish, the picture ID of the picture doesn't get the values of the flash properly cleared, and as such will be reused if another picture using the same slot ID is displayed and not set as being affected by flash.

lumiscosity:
Fixes #3376.

### DIFF
--- a/src/sprite_picture.cpp
+++ b/src/sprite_picture.cpp
@@ -159,6 +159,8 @@ void Sprite_Picture::Draw(Bitmap& dst) {
 
 	if (data.flags.affected_by_flash) {
 		SetFlashEffect(Main_Data::game_screen->GetFlashColor());
+	} else {
+		SetFlashEffect(Color());
 	}
 
 	SetFlipX((data.easyrpg_flip & lcf::rpg::SavePicture::EasyRpgFlip_x) == lcf::rpg::SavePicture::EasyRpgFlip_x);


carstene1ns:
Jenkins: ok to test

Ghabry:
Thanks! 

Good moment to also test this for other picture features like shake and tone.

Also should be verified with savegame data to see what Rpgrt stores.

Mimigris:
> Good moment to also test this for other picture features like shake and tone.

I didn't tested those a lot, but it doesn't seem they are affected, since the main issue with flash screen is that it gets interrupted on map change, which is not the case for shake and tone.

> Also should be verified with savegame data to see what Rpgrt stores.

When comparing between RPG_RT2k3e and EasyRPG, saving in this specific situation leads to the same thing, being that the picture is stored as not affected by tint and not active (since that's what the whole case is on).
Unrelated, but I've noticed that between RPG_RT2k3e and EasyRPG, the parameters maniac_current_magnify_height and maniac_finish_magnify_height aren't stored with the same value (RPG_RT 100, EasyRPG 50), may need to be checked.

lumiscosity:
tone is set in every code path, but shake isn't - good catch, will add a safeguard for that as well

Ghabry (approved):
Lgtm.

The different values for the scale look unrelated and are maniac. Features.

Ghabry:
Is a single else statement and was tested by at least two people so lets merge this as there isn't much to review here xD

PR #3378, opened 2025-03-22
Support for runtime patches: MonSca, EncounterRandomnessAlert, EXPlus, GuardRevamp

florianessl:
## Detection
_No automatic patch detection yet!_ (Only .INI settings & cmd line args)

This would either require something like my proposed Draft PR #3352 (Extracting patch info from .EXE via known addresses & signatures) or some database of known patch configurations of games (As proposed in Issue #1210)

## Test Game:
[BattlePatches.zip](https://github.com/user-attachments/files/19403624/BattlePatches.zip)

### MonSca:
This patch scales the default battle parameters of an enemy based on the contents of some in-game variables.
(Default: `V[1001] - V[1010]`)

When a switch is set (default: `S[1001]`) to ON, an alternative scaling formula, based on the average party level, is used.

**Default formula:** `val = val * V[...] / 1000`
**Alternative formula:** val = `val * avg_level * V[...] / 1000`

**Variant 'MonScaPlus':**
If set, the variable IDs used for scaling will be offset by the enemy's troop index.
-> `V[base_var_id + troop_index]`

- https://dev.makerpendium.de/docs/patch_db/patches/mon_sca.htm
- **'Plus' Variant**: https://rmarchiv.de/games/24

### Encounter Randomness Alert
This patch skips the normal battle startup logic whenever a random encounter would be triggered.
Instead a switch (default: `S[1018]`) is set to ON and the troop ID is stored into a variable (default: `V[3355]`).

- https://dev.makerpendium.de/docs/patch_db/patches/encounter_alert.htm
- https://dev.makerpendium.de/docs/patch_db/patches/encounter_alert_mepr.htm

### EXPlus:
This patch allows to individually boost the 4 party members' gained experience inside battles by applying an extra percentage based on the values of in-game variables.
(default: `V[3333]` for party member `#1`; the amounts for other party members are read from the subsequent 3 variables)

If the '[+]' option is enabled, a side effect is added to one of the Actor clauses of 'CommandConditionalBranch':
Whenever this command is used to check for the existence of an actor in the current party, the current party slot (1-4) of this actor is set to an in-game variable. (default: `V[3332]`)

- https://dev.makerpendium.de/docs/patch_db/patches/explus.htm
- https://dev.makerpendium.de/docs/patch_db/patches/explus_plus.htm

### GuardRevamp:
This patch changes the way the damage adjustment is calculated whenever the target of an attack is defending.

 Normally this calculation is done by simply dividing the damage in half for normal defense situations, and by quartering it when the target has the 'strong defense' attribute. With 'GuardRevamp' enabled, this is changed to a percentage calculation, allowing for more granular control over the output. The given default values of '50%', and '25%' would provide the same results in most situations.

- https://dev.makerpendium.de/docs/patch_db/patches/guard_revamp.htm

## Notes
Support for all of these patches can be disabled by setting the compiler flag `NO_RUNTIME_PATCHES`.

Some of the handling of command line args has been refactored, to allow setting patch parameters inidividually.
**Example**: 
`Player.exe --patch-monsca -atk 2003 -exp 2027 -plus --patch-guardrevamp -normal 75 -strong 30`
- enables "**MonScaPlus**" for scaling _attack_ & _experience gains_ by reading from the scaling factors from Variables `V[2003]` & `V[2027]` (+ troop_id for this 'Plus' variant)
- enables "**GuardRevamp**" with a factor of 75% for normal defending & 30% for "**strong**" defend.

### DIFF
--- a/src/game_runtime_patches.cpp
+++ b/src/game_runtime_patches.cpp
@@ -1,0 +1,156 @@
+/*
+ * This file is part of EasyRPG Player.
+ *
+ * EasyRPG Player is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * EasyRPG Player is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with EasyRPG Player. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// Headers
+#include "game_runtime_patches.h"
+
+#include "game_config.h"
+#include "game_party.h"
+#include "game_switches.h"
+#include "game_variables.h"
+#include "main_data.h"
+#include "player.h"
+
+namespace RuntimePatches::MonSca {
+#ifndef NO_RUNTIME_PATCHES
+	bool UseLevelBasedFormula() {
+		auto switch_id = Player::game_config.patch_monsca_levelscaling.Get();
+		return switch_id > 0 && Main_Data::game_switches->Get(switch_id);
+	}
+
+	void ApplyScaling(int& val, int mod) {
+		if (mod == 0) {
+			return;
+		}
+		if (UseLevelBasedFormula()) {
+			mod *= Main_Data::game_party->GetAverageLevel();
+		}
+		val *= mod;
+		val /= 1000;
+	}
+#endif
+}
+
+void RuntimePatches::MonSca::ModifyMaxHp(int& val) {
+#ifdef NO_RUNTIME_PATCHES
+	// no-op
+	(void)val;
+	return;
+#endif
+	if (auto var_id = Player::game_config.patch_monsca_maxhp.Get(); var_id > 0) {
+		ApplyScaling(val, Main_Data::game_variables->Get(var_id));
+	}
+}
+
+void RuntimePatches::MonSca::ModifyMaxSp(int& val) {
+#ifdef NO_RUNTIME_PATCHES
+	// no-op
+	(void)val;
+	return;
+#endif
+	if (auto var_id = Player::game_config.patch_monsca_maxsp.Get(); var_id > 0) {
+		ApplyScaling(val, Main_Data::game_variables->Get(var_id));
+	}
+}
+
+void RuntimePatches::MonSca::ModifyAtk(int& val) {
+#ifdef NO_RUNTIME_PATCHES
+	// no-op
+	(void)val;
+	return;
+#endif
+	if (auto var_id = Player::game_config.patch_monsca_atk.Get(); var_id > 0) {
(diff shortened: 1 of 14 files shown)


florianessl:
Ok, this should be ready for review. I also cherry-picked my recent commits that move the handling for "Virtual Key" code mapping because several other PRs & WIP branches depend on these helper functions.

Ghabry:
Looks solid on first glance. 

Only have to do some in-game testing and the extended command line parser. Though in general code looks fine.

Using only one file for these small patches makes sense. Otherwise we have hundreds of small patch files.

Was there a motivation behind this ``NO_RUNTIME_PATCHES``? These patches are not in any "hot" codepaths so this one "is patch on" check doesn't really hurt.

florianessl:
> Looks solid on first glance.
> 
> Only have to do some in-game testing and the extended command line parser. Though in general code looks fine.
> 
> Using only one file for these small patches makes sense. Otherwise we have hundreds of small patch files.
> 
> Was there a motivation behind this `NO_RUNTIME_PATCHES`? These patches are not in any "hot" codepaths so this one "is patch on" check doesn't really hurt.

Yeah, when you are looking at it closer, that flag doesn't make that much of a difference here. The idea is simply, to provide a way of compiling a "clean" Vanilla-compatible version of the Player without much added bloat of patch support code.

One patch that would be in a rather "hot" code path is the PowerMode branch, where patch functionality is triggered inside the routines responsible for changing variable values... (+ some yet-to-be-implemented PM-compatible Keyboard patch made by Cherry, that implements just the 4th Variable register of PowerMode)

src/game_runtime_patches.cpp
```diff
@@ -0,0 +1,307 @@
...
+		if (arg.ArgIsOn()) {
+			for (auto& patch_arg : patch_args) {
+				patch_arg.config_param.Set(patch_arg.default_value);
+			}
+
+			std::string value;
+			if (arg.ParseValue(0, value)) {
+				std::string::iterator it;
+				int i = 0, pos = 0, new_pos;
+				while ((new_pos = value.rfind(',', pos)) >= 0) {
```
Ghabry, line 56: I think that "," part of the command line parser is undocumented in your PR comment? Can you delete this or document it?

Ghabry (approved):
We want to support all the old games so this is moving some obscure stuff forward. Will be useful in combination with some patch database later.

Great PR comment and test game btw.

I added a comment about ``RuntimePatches::MonSca::ModifyMaxSp`` and co going through some ``RuntimePatches::ModifyMaxSp`` function so other patches that do something similar can be delegated in it. But this can be reworked when necessary later.

Ghabry:
that Lint build failure can be ignored. Is a new feature added by c1 for buildsystem rework and not your fault.

You are just triggering the failure because you added something to the CMakeLists but I won't enforce that you fix this xD

carstene1ns:
Yeah, wanted to fix this later.

src/game_runtime_patches.cpp
florianessl: Deleted!

Ghabry:
@florianessl imo this is ready for merge. Can you just rebase it once on the master branch to ensure that nothing broke (compile wise) in the meanwhile?

Ghabry (approved):

fdelapena (approved):

PR #3379, opened 2025-03-22
Change EasyRPG's "SetInterpreterFlag" to make patch overrides temporary

florianessl:
This PR refactors the EasyRPG feature "**SetInterpreterFlag**" (#3123), similar to the way @Ghabry suggested there, by making use of some new state flags that have been merged into _liblcf_.

See _liblcf_ PRs:
- https://github.com/EasyRPG/liblcf/pull/491
- https://github.com/EasyRPG/liblcf/pull/492

Rather than setting the global config options directly as before, a new field "**easyrpg_runtime_flags**" in the **SaveEventExecState**" struct is used, to make these overrides persist across save/load boundaries.

This makes these config overrides apply only to the Interpreter instance where they have been set, and when the end of the base frame is reached, all of the flags are reset automatically. To make this possible, a new global field "active_interpreter_flags" has been introduced, so the Player functions responsible for restricting use to these custom features can access & consider the state of the active interpreter flags for its check.

## Test Project
[RuntimeFlags.zip](https://github.com/user-attachments/files/19403706/RuntimeFlags.zip)

Set **"EasyRPG.ini"** accordingly, whether you want to test the explicit enabling or disabling of certain patches.
- Talking to the ladies at the "**ON**" section should **always** trigger the test code.
- Talking to the ladies at the "**OFF**" section should **never** trigger the test code.
- Talking to the ladies at the "**---**" section: Here it depends on the patches enabled inside _EasyRPG.ini_.

**Note:** 
Some extra test code has been introduced via a commit, to make it possible to easily test DynRPG & Destiny support using a simple log command. (**@easyrpg_output** would always be possible, even without DynRPG enabled)

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -2019,17 +2019,17 @@ bool Game_Interpreter::CommandEndEventProcessing(lcf::rpg::EventCommand const& /
 	return true;
 }
 
-bool Game_Interpreter::CommandComment(const lcf::rpg::EventCommand &com) {
+std::optional<bool> Game_Interpreter::HandleDynRpgScript(const lcf::rpg::EventCommand& com) {
 	if (Player::IsPatchDynRpg() || Player::HasEasyRpgExtensions()) {
 		if (com.string.empty() || com.string[0] != '@') {
 			// Not a DynRPG command
-			return true;
+			return std::nullopt;
 		}
 
 		if (!Player::IsPatchDynRpg() && Player::HasEasyRpgExtensions()) {
 			// Only accept commands starting with @easyrpg_
 			if (!StartsWith(com.string, "@easyrpg_")) {
-				return true;
+				return std::nullopt;
 			}
 		}
 
@@ -2038,6 +2038,7 @@ bool Game_Interpreter::CommandComment(const lcf::rpg::EventCommand &com) {
 		auto& index = frame.current_command;
 
 		std::string command = ToString(com.string);
+
 		// Concat everything that is not another command or a new comment block
 		for (size_t i = index + 1; i < list.size(); ++i) {
 			const auto& cmd = list[i];
@@ -2051,16 +2052,26 @@ bool Game_Interpreter::CommandComment(const lcf::rpg::EventCommand &com) {
 
 		return Main_Data::game_dynrpg->Invoke(command, this);
 	}
+}
 
-
+std::optional<bool> Game_Interpreter::HandleDestinyScript(const lcf::rpg::EventCommand& com) {
 	// DestinyScript
 	if (Player::IsPatchDestiny()) {
 		if (com.string.empty() || com.string[0] != '$') {
 			// Not a DestinyScript
-			return true;
+			return std::nullopt;
 		}
 
 		return Main_Data::game_destiny->Main(GetFrame());
+	}
+}
+
+bool Game_Interpreter::CommandComment(const lcf::rpg::EventCommand &com) {
+	if (auto handled = HandleDynRpgScript(com); handled.has_value()) {
+		return handled.value();
+	}
+	if (auto handled = HandleDestinyScript(com); handled.has_value()) {
+		return handled.value();
 	}
 
 	return true;
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -385,7 +385,7 @@ void Game_Interpreter::Update(bool reset_loop_count) {
 #ifdef ENABLE_DYNAMIC_INTERPRETER_CONFIG
 	Player::active_interpreter_flags = &_state.easyrpg_runtime_flags;
 	auto flags_guard = lcf::makeScopeGuard([]() {
-		Player::active_interpreter_flags = nullptr;
+		Player::active_interpreter_flags = &Player::interpreter_default_flags;
 	});
 #endif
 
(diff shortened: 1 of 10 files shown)


src/game_interpreter.cpp
```diff
@@ -382,6 +382,10 @@ void Game_Interpreter::Update(bool reset_loop_count) {
 		return;
 	}
 
+#ifdef ENABLE_DYNAMIC_INTERPRETER_CONFIG
+	Player::active_interpreter_flags = &_state.easyrpg_runtime_flags;
```
Ghabry, line 386: for the ``= nullptr`` you can use a scope guard:

```cpp
	Player::active_interpreter_flags = &_state.easyrpg_runtime_flags;

	auto flags_guard = lcf::makeScopeGuard([]() {
		Player::active_interpreter_flags = nullptr;
	});
```

This always executes when the scope of the function is left.

Currently this is harmless because we never ``return`` early but when this ever changes this will prevent a potential bug.

src/game_interpreter.h
```diff
@@ -398,4 +405,11 @@ inline AsyncOp Game_Interpreter::GetAsyncOp() const {
 	return _async_op;
 }
 
+#ifdef ENABLE_DYNAMIC_INTERPRETER_CONFIG
+inline void Game_Interpreter::ClearStateRuntimeFlags() {
+	_state.easyrpg_runtime_flags.conf_override_active = false;
+	_state.easyrpg_runtime_flags.flags.fill(false);
```
Ghabry, line 411: clearing the flags shouldn't be necessary here (while we are discussing performance). They are only used when the override is true so you could them in the ``InterpreterFlags`` command when ``override_active == false`` instead.

src/game_interpreter_shared.h
```diff
@@ -103,6 +107,14 @@ namespace Game_Interpreter_Shared {
 	lcf::rpg::MoveCommand DecodeMove(lcf::DBArray<int32_t>::const_iterator& it);
 
 	bool ManiacCheckContinueLoop(int val, int val2, int type, int op);
+
+#ifdef ENABLE_DYNAMIC_INTERPRETER_CONFIG
+	typedef bool lcf::rpg::SaveEventExecState::EasyRpgStateRuntime_Flags::* StateRuntimeFlagRef;
```
Ghabry, line 112: can be modernized to ``using StateRuntimeFlagRef = bool lcf::rpg::SaveEventExecState::EasyRpgStateRuntime_Flags::*``

src/window_interpreter.cpp
```diff
@@ -26,10 +26,55 @@
...
+	bool GlobalIsPatchDestiny() { return Player::game_config.patch_destiny.Get(); }
+	bool GlobalIsPatchCommonThisEvent() { return Player::game_config.patch_common_this_event.Get(); }
+	bool GlobalIsPatchUnlockPics() { return Player::game_config.patch_unlock_pics.Get(); }
+	bool GlobalHasRpg2kBattleSystem() { return lcf::Data::system.easyrpg_use_rpg2k_battle_system; }
+}
+
+struct RuntimeFlagInfo {
+	bool (*config_fn)();
+	Game_Interpreter_Shared::StateRuntimeFlagRef field_on, field_off;
+	char* name;
```
Ghabry, line 47: must be const char* or the more modern ``std::string_view``.

Ghabry (commented):
Thanks for finishing the command implementation.

src/game_interpreter.cpp
florianessl: thx, that's a very useful helper function!

src/game_interpreter_shared.h
florianessl: Ah yes, I used a "using" before, but changed it because I thought it might be the cause of some compiler error I had :D

Ghabry:
Jenkins: test this please

Ghabry:
Guess this one should be before #3349 as is simpler. All of the conflict in the debug scene. (can you rebase it?)

Ghabry:
I'll make some adjustments to this myself during the weekend. So no need for you to fix the warnings etc. (also want to check how the code-gen of your code is as this is all checked alot during execution)

Also from now on we should, before adding further flags, think about whether it even makes sense to have them as the flag is only active while in the interpreter context. (also to save flags, we only have 32, so cannot add all of monsca etc. xD). 

But is in general useful to have some kind of facility that alters the execution behaviour.

florianessl:
> But is in general useful to have some kind of facility that alters the execution behaviour.

**Apropos altering execution behavior:**
I've been testing an alternative approach for variable operations & conditional branches for a while now, which uses some sort of dispatching approach using vtables. This could also be used to swap out individual patch implementations for individual interpreter instances, as every interpreter has its own vtable instance.

It seems pretty stable by now & performs well (played through several larger games to really validate that I didn't break anything..)
https://github.com/florianessl/Player/tree/refactor/VariableOps

All this code stems from me experimenting with my ScopedVars branch last year, to make it possible to use all the pre-existing var operations in other new command variants (eg. "_ControlVariablesEx_" or "_ControlScopedVars_") without additional overhead.

This would be the important part that is responsible for generating the dispatch tables:
https://github.com/florianessl/Player/blob/refactor/VariableOps/src/game_interpreter.cpp#L5735

Ghabry:
Good idea with this dispatch functionality. We really have a problem by now with too much "if ManiacPatch" conditions all around the event code.

Better would be to have a "vanilla 2k" and a "maniac patch" version.

Causes some code duplication but event commands usually don't change except for rare bugfixes and reduces the risk of breaking stuff as has already happened far too often when adding maniac stuff :/.

Also allows more compiler flags like "Only Vanilla game support is enough for me" (could be useful for some of the constrained homebrew platforms).

PR #3383, opened 2025-03-26
Fix MP3 heuristic check

### LINKED ISSUE #3380: Trying to execute a Ineluki script with KeyPatch explicitly disabled, crashes the Player.
Try to execute any KeyPatch `*.script.wav` file while KeyPatch is explicitly disabled -> _CRASH_.

I noticed this while doing tests for PR #3379
-> That's why I moved two of the test NPCs outside the accessible area there. Trying to play a script file will simply crash EasyRPG.

florianessl:
Fixing two issues with the refactored MP3 heuristic check:
- For some files (I mostly noticed this with ModTracker files) the IsMp3() heuristic would now yield wrong positives, because for some reason _mpg123_read_ would immediately result in a status 'DONE without actually reading data.
- Other small non-mp3 files might yield a single error before resulting in a DONE & still pass the allowed err_count of '3'.

Fixes #3380

### DIFF
--- a/src/decoder_mpg123.cpp
+++ b/src/decoder_mpg123.cpp
@@ -230,8 +230,8 @@ bool Mpg123Decoder::IsMp3(Filesystem_Stream::InputStream& stream) {
 	// Read beginning of assumed MP3 file and count errors as an heuristic to detect MP3
 	for (int i = 0; i < 10; ++i) {
 		err = mpg123_read(decoder.handle.get(), buffer, 1024, &done);
-		if (err == MPG123_DONE) {
-			return true;
+		if (err == MPG123_DONE && done != 0) {
+			return err_count == 0;
 		}
 		if (err != MPG123_OK) {
 			err_count += 1;


fdelapena (approved):

carstene1ns (approved):

Ghabry:
I tested this a bit more: The game AE (the last days according to Ezra) has many MOD files. These are still detected as MP3 when libxmp is missing.

What works for me to get them detected as not mp3: Reduce the ``error_count`` to ``< 2`` OR increase the for loop condition to ``i < 20``.

Download: https://rmn.rmarchiv.de/  (Search for Ezra)

carstene1ns:
Unfortunately mod and mp3 have both not a common signature.

Maybe we can look at how others do it  e.g. a media player like mpv or libmagic.

florianessl:
> I tested this a bit more: The game AE (the last days according to Ezra) has many MOD files. These are still detected as MP3 when libxmp is missing.
> 
> What works for me to get them detected as not mp3: Reduce the `error_count` to `< 2` OR increase the for loop condition to `i < 20`.
> 
> Download: https://rmn.rmarchiv.de/ (Search for Ezra)

Oh nice, I haven't added the RMN archive to my test set yet.
I did also do a test-run by deactivating libxmp and running through all non-standard audio files found in the rmarchiv games. Worked perfectly with the two added checks above, but there always seem to be some outliers for these checks that are hard to catch..

PR #3385, opened 2025-03-28
Libretro fixes and one hotfix for YNOnline Users with loooong playtimes

Ghabry:
Guess @Mimigris should also test this a bit.

I only hacked the frame counter to overflow after about two seconds. Works for me (it renders are usual, thats not the RPG_RT behaviour but who cares).

### DIFF
--- a/src/platform/libretro/ui.cpp
+++ b/src/platform/libretro/ui.cpp
@@ -371,7 +371,7 @@ RETRO_API void retro_get_system_info(struct retro_system_info* info) {
 	info->library_name = GAME_TITLE;
 	info->library_version = PLAYER_VERSION_FULL;
 	info->need_fullpath = true;
-	info->valid_extensions = "ldb|zip|easyrpg";
+	info->valid_extensions = "ldb|zip|lzh|easyrpg";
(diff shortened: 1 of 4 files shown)


fdelapena (approved):

Mimigris:
This seems to work well 👍 
I'm not really sure to understand about the overflowing part and the two seconds, does it mean there is a time frame where the water tile issue could be present? If so, a crash could still occur if a map had to be loaded within that timeframe and the data generated was incorrect, though I have not been able to trigger a crash with that using this build.

Ghabry:
I meant I did a test locally where I altered the frame counter in such a way that it will overflow two seconds after selecting "new game" to simulate what happens.

The frame counter overflowing has no negative impact. It could be that for a brief moment (during the overflow) that an animation advances faster but otherwise there will be no negative things happening. That frame counter is never used for any timedelta calculations, so "begin > end" cannot happen.

For all other time measurement we use a different frame counter that always starts at 0 so for this to overflow the Player must really run for 1.5 years without reloading the website :D.

carstene1ns (approved):

Issue #3373, opened 2025-03-19
Android SDK35: Support Edge-To-Edge mode


Ghabry:
Look at the screenshots. Most obvious for the settings scene. The rendering starts at the top of the phone and even draws above the top bar with the icons and time. :interrobang: :interrobang: :interrobang: 

Not reproducable on any other system. Issue just for documentation purposes...

![grafik](https://github.com/user-attachments/assets/652420fb-7f3c-4057-a428-b90fb01c6c0d)

![grafik](https://github.com/user-attachments/assets/e2d940a6-e664-4be5-b176-f5801a83c763)

Reported by M.Martinez via Mail. Thanks.

Ghabry:
This is actually new behaviour in Android 15 (SDK set to 35 or higher): Apps are now "edge-to-edge" this means they have now the entire area of the screen including the system bar and the navigation buttons.

App authors must manually apply the offset so it doesn't render important UI content behind the system UI :facepalm: :facepalm: :facepalm: 

https://developer.android.com/about/versions/15/behavior-changes-15

Ghabry:
Oh and this also breaks the "press back to open sidebar" functionality. Instead pressing back always goes to the previous activity now.

So a "back" while a game is running terminates the game without asking and reopens the game browser and editing layouts becomes impossible because opening the sidebar requires the back button :roll_eyes: 

Will revert to SDK 34. This is still allowed for new apps until 2026.

Fck Google breaking their OS every year with arbitrary new "features".

carstene1ns:
Needs recheck with 0.8.1

Ghabry:
Yes this works after the SDK downgrade. I move this to 0.8.2 and edit the issue title as the problem is now understood.

(and must be fixed within one year because of SDK requirements in the play store -_-)

PR #3388, opened 2025-04-04
Urgent Android fix

Ghabry:
Our App is broken with SDK35 (#3373) now I just hope that Google Play allows this downgrade, otherwise we are kinda fcked :(

Since the last SDL2 update the systembar is green while in-game instead of black. SDL2 calls it "Fixed Immersive Mode". Imo this breaks immersion so I applied a custom theme to revert the behaviour.

Also added a Fullscreen option just because this works now with SDL2 instead. \*g\* (and maybe makes some users happy because the UI is not hidden anymore)

### DIFF
--- a/builds/android/app/build.gradle
+++ b/builds/android/app/build.gradle
@@ -5,9 +5,9 @@ android {
     assetPacks = [":assets"]
     defaultConfig {
         applicationId "org.easyrpg.player"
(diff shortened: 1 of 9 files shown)


carstene1ns (approved):
One can only wonder how broken Android really is.

PR #3397, opened 2025-04-15
Compiler fixes

Ghabry:
Fixes a dumb typo in the Presets breaking RelWithDebInfo (Making it a Debug build >.>).

And Rinnegatamante told me to pass ``-fno-optimize-sibling-calls`` for the Vita build because that optimisation can cause crashes: Can confirm. With that flag set the MainLoop does not randomly crash anymore :shrug:  (that appears to be a general problem on Vita, other homebrew uses the same flag)


fdelapena (approved):

Ghabry:
I must have forgotten to commit the ``RelWithDebInfo`` and the ``CONFIGURATION_TYPES`` stuff. I'm pretty sure I had this already working locally months ago but the code is not in our repo :( (and in liblcf it is indeed in)

carstene1ns (approved):

PR #3398, opened 2025-04-16
Extend the "Strings" view in debug screen to display names of StringVars (+ some debug screen adjustments)

florianessl:
Self-explanatory, but I also included some refactors of the existing var_window, to make it more useful for displaying certain kinds of data that might need more than a single line.
(My ScopedVars branch for example, would also display multiple lines per scoped variable type)

StringVar names have been added with ManiacPatch 241028 & also provide a useful way to detect this major update without having to rely on reading the EXE:
```
if (lcf::Data::maniac_string_variables.size() > 0) {
    mp_version = MP_24;
}
```
(Even if no StringVars are used, at least one list entry will be included in this new LDB field by default)

### Updated Strings view

By default, the view stays as before:
screenshot_206

And by pressing SHIFT, it toggles to the new display mode that displays individual entries over two lines & includes the var name:
screenshot_207

### Another new addition:

Previously, displaying data for Switches/Variables above the ID 9999 would overflow the text:
screenshot_211

This has been adjusted, by shortening the prefix to a single character + changing the way the text is formatted:
screenshot_210

### DIFF
--- a/src/window_varlist.h
+++ b/src/window_varlist.h
@@ -19,9 +19,9 @@
 #define EP_WINDOW_VARLIST_H
 
 // Headers
-#include "window_command.h"
+#include "window_selectable.h"
 
-class Window_VarList : public Window_Command
+class Window_VarList : public Window_Selectable
 {
 public:
 	enum Mode {
@@ -43,8 +43,10 @@ public:
 	 *
 	 * @param commands commands to display.
 	 */
-	Window_VarList(std::vector<std::string> commands);
+	Window_VarList();
 	~Window_VarList() override;
+
+	void Update() override;
 
 	/**
 	 * UpdateList.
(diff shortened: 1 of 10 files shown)


florianessl:
I implemented another new option to make it easier for myself to debug some scripts using the new JSON command:

**Varlist window will display a short indicator, if a string is part of the _json_cache:**
screenshot_217
screenshot_218

**If a string is inside the JSON cache, a new display option "Pretty Print" will be available in the detail window:**
screenshot_215
screenshot_216

src/window_varlist.h
```diff
@@ -43,52 +43,136 @@ class Window_VarList : public Window_Command
...
+		case eMap:
+			return "Mp";
+		case eCommonEvent:
+			return "Ce";
+		case eMapEvent:
+			return "Me";
+		case eString:
+			return "St";
+		default:
+			break;
```
Ghabry, line 136: move the ``assert(false)`` into the ``default:`` followed by ``return {};``. Yeah, compiler is picky here because assert is removed in release builds.

src/window_varlist.h
```diff
@@ -43,52 +43,136 @@ class Window_VarList : public Window_Command
...
+			}
+			break;
+		default:
+			break;
+	}
+	return 10;
+}
+
+constexpr int Window_VarList::GetDigitCount(Mode mode) {
+	return 4;
```
Ghabry, line 155: github action reports ``unused parameter 'mode' [-Wunused-parameter]``

Ghabry (approved):
besides the 2 warnings looks good to me

jetrotal:
Nitpick that may be helpful sorting on the debug menu:

Menu 1 - Game State Menu
```
Save
Load
Goto Map
Open Menu
Move Speed
Full Heal
Battle
Interpreter
```
Menu 2 - Variables & Events Menu
```
Variables
Switches
Strings
Money
Items
Level
Call ComEvent
Call MapEvent
Call BtlEvent
```
This organization puts:

Game flow controls together (Save/Load, navigation, speed)
All variable types together (Variables, Switches, Strings, Money)
All event calls together
Game state items together (Items, Level)

The "Interpreter" is placed in the first menu since it's related to the current game state rather than variables or events.

Ghabry:
can you move this to an issue? This doesn't fit here.

jetrotal:
sure!

carstene1ns (approved):
Looks good to me.

PR #3404, opened 2025-04-23
Fix StringIsAscii by explicitly setting the type to "signed char"

### LINKED ISSUE #3401: Problem with "umlauts" in German Games... (3DS, Switch, Android)
Files (e.g., images) with umlauts in the filename (ä, ü, ß, etc.) will not load in the game.
no difference between unpacked or ZIP archives. what means "deflate compression"?
i tested zips, created with this command
for /d %%i in (*) do "C:\Program Files\7-Zip\7z.exe" a -tzip -mx=9 "%%i.zip" "%%i\*" -scsUTF-8
(without "-mx=...", and with "-mx=0")
it makes no difference. 

everywhere same "bug"... 3ds, switch and android.
only in windows it works for me (+ retroarch).

you could try it with the game "Die Reise ins All" for example.
https://rmarchiv.de/games/50

Gabry said, this problem at nintendo-switch-version is caused by ... :"That's a bug in Nintendos file system driver: It interprets everything as Shift-Jis (Japanese encoding) even when the filesystem has a different encoding."
 **but on Android is the same bug!** 
Windows is ok.
Retroarch on nintendo-switch with easyrpg-core has not this bug. but laggy sound bug^^

I could test easyrpg-player with wii or wiiu too 

greetz

Ghabry:
Turns out "char" is not always signed. Instead it is "implementation defined". In all tests it went unnoticed because x86 is "signed". On ARM it appears to be "unsigned".

This is surprising. For int etc. it is specified to be signed.

Can we just burn the C standard? Who specifies this nonsense?

Fix #3401

### DIFF
--- a/.github/workflows/stable-compilation.yml
+++ b/.github/workflows/stable-compilation.yml
@@ -19,8 +19,8 @@ defaults:
 
 jobs:
   docker:
-    name: ${{ matrix.image }}
-    runs-on: ubuntu-latest
+    name: ${{ matrix.image }} (${{ matrix.os.arch }})
+    runs-on: ${{ matrix.os.name }}
     container:
       image: ${{ matrix.image }}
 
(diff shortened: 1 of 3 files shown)


fdelapena (approved):

florianessl (approved):

Ghabry:
Linux ARM has the same signed-ness behaviour for char (and the compiler even warns about this with the -Wextra we use :/). And just a month ago Github added ARM runners... Will add some of them so this doesn't happen again :).

carstene1ns (approved):
I just hope we do not use all our quota for too much runners :P

Ghabry:
lol, I thought open source projects have infinite quota. :D

PR #3405, opened 2025-04-25
Fix: Two of new 0.8.1 engine options were not written to the config INI

florianessl:
Options "_Logging_" & "_Screenshot scaling factor_" in the "_Engine_" section of the settings menu were never written to the INI file & would thus be reset when the Player is restarted.


Ghabry (approved):
oops :)

carstene1ns (approved):

PR #3406, opened 2025-04-26
New setting options for screenshots: Periodically take auto-screenshots + option for timestamps in file names

florianessl:
### Periodically take automatic screenshots
screenshot_2

screenshot_3

### Timestamps in filenames
This was also requested recently & is pretty easy to implement, so I added this option as well:

screenshot_20250426_140147
If there is a naming conflict (multiple presses of the screenshot key in a single second), file names will still be indexed (starting with "_1.png", rather than "_0.png")
image

### DIFF
--- a/src/game_config.cpp
+++ b/src/game_config.cpp
@@ -63,10 +63,6 @@ void Game_ConfigPlayer::Hide() {
 	font2.SetOptionVisible(false);
 	font2_size.SetOptionVisible(false);
 #endif
-#if !defined(EMSCRIPTEN)
-	automatic_screenshots.SetOptionVisible(false);
-	automatic_screenshots_interval.SetOptionVisible(false);
-#endif
 	if (automatic_screenshots.IsOptionVisible()) {
 		automatic_screenshots_interval.SetLocked(!automatic_screenshots.Get());
 	}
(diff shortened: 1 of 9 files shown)


Ghabry:
This slowly needs a submenu just for screenshots :D

Ghabry:
Open for discussion: Is that screenshot timestamp setting even needed or should it just become the default and replace the old naming?

florianessl:
> Open for discussion: Is that screenshot timestamp setting even needed or should it just become the default and replace the old naming?

I would be fine with changing it to the new timestamped naming scheme by default.
Though, I'd keep the extra new function "GetScreenshotName" & the added preview display on the bottom anyway, because we might want to extend it later when we want to allow the user to configure the storage path for screenshots. (Maybe for a setting to optionally use either project or folder name as part of the filename)

jetrotal:
I've seen different examples of people using the screenshot feature ingames, things like a dynamic battle bg based on map screenshots from dynrpg, or the maniacs screenshot command.

I never tooka look at those tho, to see if they allow reusing the screenshot default filename

Ghabry (commented):
can be approved after adressing the nitpicks

PR #3407, opened 2025-04-28
Support for changed game constants: StatDelimiter & WhiteDragon patches

florianessl:
Mostly picked from PR #3352, with some rebasing & movement of code.
This version has no support for reading individual patched constants from RPG_RT code addresses but is able to heuristically detect both the Italian "WhiteDragon" version of RPG_RT & several variants of the "StatDelimiter" patch. (https://dev.makerpendium.de/docs/patch_db/main-en.htm?page=stat_delimiter)

This also doesn't include the **Battle Stat Split** mentioned in the previous PR.

Plus, for added support, the upper limits for _Enemy HP/SP_ have been removed completely, as those limitations applied only to the RPG Maker's editor & the values could actually be extended to the upper limit of signed integer values.
**But:** If _easyrpg_max_enemy_hp_ or _easyrpg_max_enemy_sp_ are defined in the LDB then those changed limits will still take effect!

### Starting the Player with custom runtime versions
To make it easier to test this feature, I also implemented a new command argument "--engine-path" (See Issue #3166 )

Usage:
`
Player.exe --test-play --engine-path RT_rm2k3_advo_108_en.dat
`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -170,6 +170,8 @@ add_library(${PROJECT_NAME} OBJECT
 	src/game_config.h
 	src/game_config_game.cpp
 	src/game_config_game.h
+	src/game_constants.h
+	src/game_constants.cpp
 	src/game_destiny.cpp
 	src/game_destiny.h
 	src/game_dynrpg.cpp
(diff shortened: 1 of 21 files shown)


Ghabry:
yeah looks also fine to me. Will do it myself as the author is busy.

And I'll change the Constant `namespace` into a `class`. We already have far too much global state everywhere. Better not make it worse :sweat_smile:

PR #3408, opened 2025-04-29
Android: Fix all the crap required since Android 15 (Edge-To-Edge, Back callbacks).

Ghabry:
All this huge pile of disgusting code just to get rid of bad defaults enforced by Google we don't even use. Please don't ask about too many details... It uuuuh works.

(Fun fact: They delayed the opt-out to Android 16 because even Google itself hasn't added it to all their apps yet m( )

For testing you need a Android 15+ device. On olders you can only test if it still works, but all the edge-to-edge stuff etc. is just disabled.

---

Oh and I also added a Dark Mode :). At least something that is useful.

grafik


carstene1ns (approved):
Tested on Android 9, seems to work.

PR #3411, opened 2025-04-29
Several fixes for the CloneMapEvent/DestroyMapEvent commands

### LINKED ISSUE #3393: EasyRPG-exclusive command "DestroyMapEvent": "ThisEvent" not supported
Trying to use this new event command to make it possible to "self destroy" a running event will just cause warnings:
Image

This can of course be circumvented by looking up the current event id via the "_ControlVariables_" command & using the indirect adressing mode of the command, but there are two issues here:

- _ControlVariables_' branching option `Event -> Event ID` is only available if Maniac support is enabled. (Otherwise the result is always zero)
- The current version of the command editor explicitly allows using "This Event" (while also blending out other common special event IDs like Player, Boat, etc.. which wouldn't work. So this was a deliberate choice here.):

Image

florianessl:
Fixes following issues with these EasyRPG-commands:

- Cloned & destroyed events were not considered by the map setup code which is responsible for loading savegame data. This was especially dangerous if a pre-existing map event has been destroyed because the code assumed both the original & the savegame data vectors of events to be both the same size & in the same order, with no skipping of IDs.
- Added check before refresh of "scene": If the **CloneMapEvent** function was called from inside the map setup code, the map scene might not be created yet.
- Fix #3393 
- Sometimes **CloneMapEvent** would write garbage data into the field "_easyrpg_clone_event_id_" due to an invalid pointer (Vector reallocation issue)

### Test project:
[CloneDestroyMapEventTest.zip](https://github.com/user-attachments/files/19967104/CloneDestroyMapEventTest.zip)

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -5656,6 +5656,10 @@ bool Game_Interpreter::CommandEasyRpgDestroyMapEvent(lcf::rpg::EventCommand cons
 
 	int target_event = ValueOrVariable(com.parameters[0], com.parameters[1]);
 
+	if (target_event == 0 || target_event == Game_Character::CharThisEvent) {
+		target_event = GetThisEventId();
+	}
+
 	_async_op = AsyncOp::MakeDestroyMapEvent(target_event);
 
 	return true;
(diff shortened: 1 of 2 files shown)


Ghabry (approved):
The "Clone Event" upon load will likely fail in the web player because of async fetch requirements when the map is different but I don't see a simple solution for this right now :/

Ghabry:
Because this is loading code I don't want that it crashes for the user so I added another range check and removed the assert.

@florianessl could you quickly test this again with some games (and with your Clone Destroy Event stuff xD)

Ghabry:
Okay I think this received enough testing. Gonna merge it.

PR #3416, opened 2025-05-09
Fix #3415 (Access violation in RM2K battle scene)

### LINKED ISSUE #3415: Crash: RM2K battle scene may cause Access Violation due to a removed party member being targeted (Die Reise ins All)
**Game:** _Die Reise ins All_
**Savegame:** [Save07.zip](https://github.com/user-attachments/files/20027990/Save07.zip)
_**Continus Player build**_

In this final boss fight, the character _Heisen_ is temporarily removed during the 3rd round. This is a bit hard to reproduce, because it depends on RNG, but you can use Auto-Battle to quickly spam through Round 1 & 2.
Round 3 will start with some dialog, causing the last party member _Heisen_ being removed. But because the attack targets seem to be calculated prior to the script execution, the boss may have targeted _Heisen's_ party slot with their attack.

The dialog of the 3rd round is over when this message pops up:
Image
Right after, the remaining party members execute their attacks and if Heisen has been targeted previously, EasyRPG Player may crash on trying to execute the enemy's battle action on an invalid target.

florianessl:
Fixes the Player crash described in issue #3415

I haven't looked further if a similar problem could occur in the 2k3 battle scene. But I guess the execution of scripts would look quite different there. (And I remember playing plenty of 2k3 games that switch out characters in the standard battle system & never having a similar issue)

### DIFF
--- a/src/scene_battle_rpg2k.cpp
+++ b/src/scene_battle_rpg2k.cpp
@@ -803,12 +803,6 @@ Scene_Battle_Rpg2k::SceneActionReturn Scene_Battle_Rpg2k::ProcessSceneActionBatt
 		PrepareBattleAction(battler);
 		pending_battle_action = battler->GetBattleAlgorithm();
 
-		if (pending_battle_action == nullptr) {
-			RemoveCurrentAction();
-			SetSceneActionSubState(ePost);
-			return SceneActionReturn::eContinueThisFrame;
-		}
-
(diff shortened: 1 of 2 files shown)


Ghabry:
I created a small battletest myself to test this. Could you check if the patch works properly in Allreise?

Ghabry (approved):

carstene1ns (approved):

PR #3419, opened 2025-05-21
Translation fix: Files inside "Text" folder were not translated

florianessl:
I noticed this after migrating a lot of harcoded strings into json files & trying to translate them..

Files which are fetched via `OpenText()` (Everything inside the _Text_ game folder) would not get translated at all due to their function variant not checking for active translations.

Here's a quick fix...


florianessl:
ok. this alternative approach also works out fine for my test case :thumbsup:

Ghabry (approved):

Ghabry:
Jenkins: test this please

PR #3421, opened 2025-05-24
CMake: Allow disabling (L)GPL libraries and some fonts

### LINKED ISSUE #2500: License related issues
~~Add a new option --nonfree to autotools (details see next checkbox)~~
- [x] Add new option ``-DPLAYER_NONFREE`` to Player. When enabled it disables the following LGPL'ed libraries:
    - FluidLite/Synth
    - mpg123
    - sndfile
    - wildmidi
    - I know that they are LGPL so you are safe when you use "shared libraries" but you know how good people are in reading so it is safer to turn it off by default.
    - Also disable Pixel font WenQuanYi under GPLv2 /w Font Embed Exception (for me it is still unclear whether the document exception applies to us because we convert the font before using it, so this could be a derived work). Btw were we even compatible here? This is GPLv2 not GPLv2 or newer it seems, upps. Would be good to use a new font here or figure out if the exception applies to us.
- [x] Add a new "Scene About" (could be reachable via Scene Config) that contains the following (We are not properly citing our dependencies, must be solved!)
    - About Player (already in Scene GameBrowser
    - Player License information
    - Library licenses, only list the ones enabled via preprocessor macros (1-2 Lines per library and a text "for full license text visit https://easyrpg.org/player/licenses")

see also #167

Ghabry:
This fixes the remaining blocker for the relicensing: Disabling of LGPL libraries and the GPL WQY font.

Adds a new option ``-DPLAYER_NONFREE`` which conveniently disables all (L)GPL dependencies for you so you can be certain as a gamedev that the final result is not under a copyleft license. 

Yes technically you _can_ use LGPL when its a shared library (thats why I kept ALSA enabled) but lets be honest: This is too error prone and most don't know the difference.

The Disabling of WQY and Baekmuk could be also useful for some of our memory constrained devices to save about half a megabyte in size.

Fix #2500

Considering the Makefile will be removed soon I just hardcoded both fonts to enabled.


carstene1ns (approved):

carstene1ns:
jenkins: test this please

Issue #3412, opened 2025-05-03
Incorrect positioning of timer overlay in "Die Reise ins All"


florianessl:
Likeley a timing issue similar to #2432 / #2932

During the liftoff sequence, an in-game timer is set, which should align with the clock sprite on the lower left. But EasyRPG Player shows it on the upper left of the screen.

**_Die Reise ins All_** uses an unmodified **RPG_RT** of Don Miguels _1.07b_ version of _RPG Maker 2000_.

**RPG_RT:**
![Image](https://github.com/user-attachments/assets/9a71780e-fa14-459f-bddb-318a0893ff53)

**EasyRPG**
![Image](https://github.com/user-attachments/assets/70fe9714-d8e8-488b-b007-0fd2affbb8cd)

[Save10.zip](https://github.com/user-attachments/files/20022526/Save10.zip)

florianessl:
I tested around a bit and found out that the positioning defined via "_Control Text Settings_" doesn't affect positioning of the timer at all. 
It seems, that **RPG_RT** actually checks for the **y** parameter of the window & ignores system settings. This is also the case, if the message window isn't active/visible any more.

This simple change should do it:

```diff
diff --git a/src/sprite_timer.cpp b/src/sprite_timer.cpp
--- a/src/sprite_timer.cpp
+++ b/src/sprite_timer.cpp
@@ -23,6 +23,7 @@
 #include "game_party.h"
 #include "game_system.h"
 #include "game_battle.h"
+#include "window_message.h"
 #include <player.h>
 
 Sprite_Timer::Sprite_Timer(int which) :
@@ -90,7 +91,7 @@ void Sprite_Timer::Draw(Bitmap& dst) {
 	if (Game_Battle::IsBattleRunning()) {
 		SetY((Player::screen_height / 3 * 2) - 20);
 	}
-	else if (Game_Message::IsMessageActive() && Game_Message::GetRealPosition() == 0) {
+	else if (Game_Message::GetWindow()->GetY() < 20) {
 		SetY(Player::screen_height - 20 - Player::menu_offset_y);
 	}
 	else {
```

Issue #2932, opened 2023-03-24
"Show Text" in parallel-common events ignores faceset and draws over it if an autorun-ce is also running.


TrunXGIT:
Happens only with text in common events, not with map events.

![grafik](https://user-images.githubusercontent.com/23344806/227584527-74cb3461-5ec8-43da-b6b2-d14ca0a59d11.png)

Tested it in 16:9 and 4:3, happens in both screen modes

EDIT: Noticed, that the first line of text (the name) aligns itself next to the faceset, only the following lines ignore it

Ghabry:
can you upload a savegame?

I don't think the observation "is a common event" is the only problem here. Likely some race condition with some other event. There is no logic difference in handling of messages from CEs

Ghabry:
This could be the same as #2432 where cherry documents a weird face clearing bug in parallel processes

florianessl:
**Very Likely related to this:**
I have a rare timing issue in my game "Urban Purrrge", where the text indentation is suddenly reset, so that the text is drawn over the face graphic:
![Image](https://github.com/user-attachments/assets/fba03304-b097-4ef7-a47d-286ba0e35c1f)
But in the vast majority of cases this prints normally:
![Image](https://github.com/user-attachments/assets/33223497-1155-4b80-a0ad-79badaee5f2e)

Not a common event in this case, but a parallel process running on the map.
To clarify: This message command is actually called by a parallel process, not by the foreground interpreter or an player action. The event for the sheep is set to register Player Touch events & the actual processing of "Decision Logic" is handled by a parallel process that periodically checks the key input & -when the decision key is pressed-, executes the 2nd event page of certain events that might be in front of the player...

Issue #2607, opened 2021-07-31
ShakeScreen on Parallax Background differs from RPG_RT


florianessl:
#### Name of the game:
Battlefield Manka-Dream (Map0100 "Fin #1")

#### Player platform:
Windows

#### Attach files (as a .zip archive or link them)
(Video)
https://drive.google.com/file/d/1iebewL8a9x_79wuS6ABeG_NDUoX3bH2J/view?usp=sharing

#### Describe the issue in detail and how to reproduce it:
The way the ShakeScreen command handles out-of-bounds graphics seems to be ok for both normal maps and pictures. But the result differs for parallax backgrounds.

florianessl:
This drawing difference is also noticable during the liftoff sequence in **_Die Reise ins All_**

(Use the savegame found in Issue #3412 )

**RPG_RT**
![Image](https://github.com/user-attachments/assets/7e1a407a-eac2-44fc-b5a7-2ea27d063f67)

In EasyRPG, this sequence looks a bit weird when combined with the overlayed border sprite:

**EasyRPG**
![Image](https://github.com/user-attachments/assets/b1f7bcf7-89eb-478d-bf1f-168653d2e417)

Issue #2432, opened 2021-01-06
Clearing face on main interpreter event load/end only half-working


CherryDT:
I recently analyzed an issue someone had with RM2k3 where the existence of an autorun event would cause a face graphic set by a parallel process event not to be displayed.

This is because RM clears the face graphic when the main interpreter (for autostart/action-key/touch events, not parallel processes) either loads a new event script or ends execution of an event script (the intention being that faces are scoped to an event, according to the help file - but with faces actually being global, the implementation isn't accurately following that intention). See also: https://www.multimediaxis.de/threads/71-Detail-Wissen-und-Geheimnise-des-RPG-Makers-vorallem-f%C3%BCr-Erfahrene-Profis-lehrreich?p=3407454#post3407454

Since this was pretty obscure and I hadn't known about this before, I decided checking what EasyRPG does there.

I discovered that clearing the face graphic at the end of the event works, but clearing it when loading a new event script into the main interpreter kinda "half-works" - it does clear the face setting, but on the current messagebox the face still gets drawn once, even though the text is already shifted to start at the very left, causing an overlap:

![image](https://user-images.githubusercontent.com/6010463/103777607-6e5adc00-5031-11eb-855d-db447a1a7d23.png)

Expected result as per RPG_RT: Face is not drawn.

Here is the demo project: https://cherryshare.at/f/ElLtXw/Project1.zip

PR #3434, opened 2025-06-11
Visual Fixes

florianessl:
A few fixes for minor visual glitches in the EasyRPG Player.
(As well as some other fixes for warnings MSVC spat out)

### #2432 / #2932 - Face state not emulated correctly for Message Processing in Parallel events because of the global state of game_system
I added another "hacky flag" in Window_Message (See the FIXME there..) to indicate that a text was pushed in the current frame & renamed an existing one ("close_finished_this_frame").
The initial creation of the window contents has been moved from "StartMessageProcessing" to the Update method. Restructuring the code this way allows us to delay this part of the setup, until the main interpreter scripts have been handled.

I modified Cherry's test project a bit, to also verify that the parsing of variables in the message contents behaves the same:
[faceset-parallel.zip](https://github.com/user-attachments/files/20685485/faceset-parallel.zip)

It might make sense to combine all these boolean flags in Window_Message into some enum field (with descriptive names to improve readability), that describes the current state of the window & just remove the "FIXME" comment. Trying to restructure all the code as suggested there would be a major headache & just break other stuff along the way... :thinking: 

This should also fix some other MessageBox issues that are caused by race conditions between interpreter instances. Have to dig up the issue numbers..

### #3412 - Incorrect positioning of timer overlay
I already went into detail in the comments of this issue.

### #2607 - Parallax background graphics would be rendered on the outside area of the screen when ShakeScreen is applied
While a ShakeScreen effect is active, the area outside of the screen boundaries should not be drawn on.
EasyRPG uses a TiledBlit here, effectively repeating the bitmap everywhere. Could maybe be improved, but I just clear the overdrawn areas afterwards, to emulate RPG_RT behavior.

### DIFF
--- a/src/window_shopbuy.cpp
+++ b/src/window_shopbuy.cpp
@@ -67,6 +67,7 @@ void Window_ShopBuy::DrawItem(int index) {
 
 	if (!item) {
 		Output::Warning("Window ShopBuy: Invalid item ID {}", item_id);
+		return;
 	} else {
 		enabled = item->price <= Main_Data::game_party->GetGold() && Main_Data::game_party->GetItemCount(item_id) < Main_Data::game_party->GetMaxItemCount(item_id);
 		price = item->price;
--- a/src/audio_generic.cpp
+++ b/src/audio_generic.cpp
@@ -221,8 +221,9 @@ bool GenericAudio::PlayOnChannel(BgmChannel& chan, Filesystem_Stream::InputStrea
 	chan.paused = true; // Pause channel so the audio thread doesn't work on it
 	chan.stopped = false; // Unstop channel so the audio thread doesn't delete it
 
+	std::string_view name = filestream.GetName();
 	if (!filestream) {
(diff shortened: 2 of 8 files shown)


Ghabry:
About the panorama clearing. There is already code that does the same for the wide-screen hack (for the same reason).

Maybe can be reused? https://github.com/EasyRPG/Player/blob/master/src/screen.cpp

src/window_shopbuy.cpp
```diff
@@ -67,6 +67,7 @@ void Window_ShopBuy::DrawItem(int index) {
 
 	if (!item) {
 		Output::Warning("Window ShopBuy: Invalid item ID {}", item_id);
+		return;
```
carstene1ns: IMO this whole function looks bad.
The item check should be moved to the top and the `price`, `enabled` and `str` variables are either unused or unneeded.

carstene1ns (approved):

PR #3435, opened 2025-06-16
Rewrite ControlGlobalSave

### LINKED ISSUE #3069: Maniac: Global Save - Several issues
#### Player platform:

- [ ] When launching a Maniac patch game and that the global save file exists (Save.lgs), it should be automatically (re)loaded when starting the game, after a game reset or when loading a save, without using any Control Global Save command. Currently this behaviour is incorrect and the global save file is automatically loaded only once when starting the game for the first during a game session.
- [ ] When Opening a Global Save without closing it or Copying to/from the Global Save without Closing the Global Save and then closing the game without resetting or loading a save, the global save file should be created or updated with the content of the global save that was in memory, but currently EasyRPG does not do that.
- [ ] Closing the Global Save using the command in the Maniac Patch should load the content from the Save.lgs file (hence the name...) while currently EasyRPG applies this to the Open Global Save command.
- [ ] Loading the global save using Open Global Save in EasyRPG (not the intended as mentioned above but that's not the issue here) will reload the previous global save file that used to be present instead of loading null values if the Save.lgs file is removed after loading it once.

Since understanding this command is complex, I've listed as Needs feedback. My tests were done with the latest public version of the Maniac Patch 211010.

Ghabry:
Main difference is the logic when opening the save. Also moved it to ManiacPatch file to reduce complexity in the interpreter code.

Header cleanup (was necessary because of weird compile errors due to a Game_String include while moving code around)

Fix #3069

----

@Mimigris  please test this. Should be at least closer than before xD

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -26,20 +26,18 @@
 #include <cassert>
 #include "game_interpreter.h"
 #include "async_handler.h"
-#include "audio.h"
 #include "game_dynrpg.h"
 #include "filefinder.h"
 #include "game_destiny.h"
 #include "game_map.h"
 #include "game_event.h"
-#include "game_enemyparty.h"
-#include "game_ineluki.h"
 #include "game_player.h"
 #include "game_targets.h"
 #include "game_switches.h"
 #include "game_variables.h"
 #include "game_party.h"
 #include "game_actors.h"
+#include "game_strings.h"
 #include "game_system.h"
 #include "game_message.h"
 #include "game_pictures.h"
@@ -70,7 +68,6 @@
 #include "transition.h"
 #include "baseui.h"
 #include "algo.h"
-#include "rand.h"
 
 using namespace Game_Interpreter_Shared;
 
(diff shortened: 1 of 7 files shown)


Mimigris:
Sorry for answering a bit late on this, just started my testing.

From what I can see, the Save.lgs file is not able to be created at the moment using the build of this pull request, only updated, meaning that storing data in a Save.lgs file is impossible unless said file already existed, so this would probably need to be fixed before doing bigger testings due to being at the heart of the command.

Ghabry:
Yeah my bad, the flag that tracked whether the global save is "opened" what only set when the file exists... Is fixed now

Ghabry:
``ResetGameObjects`` was not called when doing it as you described.

I added this call now to ``Continue`` from the game browser. No idea why this wasn't used before, seems to work fine (likely some other bugfix before).

Mimigris:
Outside of the last point mentioned in #3069 , which has been explained in chat why it has not been adressed (tied to the path cache + not something that would be noticeable while playing normally), everything seems good from what I see 👍

fdelapena (approved):

carstene1ns (approved):

PR #3440, opened 2025-06-28
Yume Nikki: Fix mosaic transition and gold window opacity

Ghabry:
@Lt-knb


Ghabry:
@Lt-Knb thx your code pointed in the right direction. Can you test if this works?

Ghabry:
Can you give an example for the 2k3 issues?

Here is a screenshot with engine set to 2k3 (not English/1.10) and Battle Type B with system transparency on and it looks the same in RPG_RT and EasyRPG.

grafik

Ghabry:
Created a new issue #3446 about the missing first frame as this is not exclusive to Mosaic it seems.

Otherwise I consider this now finished... Please check again :) (added randomness)

Ghabry:
Due to the bug the transition skips either the first or the last frame. Changed it now to 2 - 40, so the x41 frame is missing.

PR #3445, opened 2025-06-30
Fix 3436: Consider images with no alpha as Opaque in ComputeImageOpacity.

### LINKED ISSUE #3436: [Wii U] Broken render of some backgrounds in "Ib" game
Some scenes of the Ib game gets rendered incorrectly: Seems texture wasn't properly loaded/converted, and as result, the buffer wasn't clered here:

Image

There are several places with such glitch where I caught:
- The red room that goes somewhere after guillotine
- Violet room prior to the fences trap/mini-labirinth, where Mary gets met first time
- Mary's universe that looks like childish drawings is also broken (because of that, game is almost unbeatable if not play in blind).

Ghabry:
When using pixel format ABGR (prefered on Big Endian) instead of RGBA and the image is considered opaque (no alpha channel), pixman makes the alpha pixels 0x00 instead of 0xFF.

Technically the alpha values are undefined as the image is opaque but our opacity function considered this image as full transparent (because A is 0 everywhere).

This triggered a fast path in the ToneBlit which does nothing when the image is full transparent and causes a "trippy trail" rendering glitch.

Additionally this fixes a glitched Fog effect in the beginning of "Unterwegs in Düsterburg" which only flickered between transparent and full opaque before.

Fix #3436

---

The fix for UiD is cool because that was a long-standing bug where I had no idea where it comes from :)

The ToneBlit rendering bug in IB was not reproducable on the Wii, maybe different pixel format used or somehow different pixman behaviour didn't investigate this further.

### DIFF
--- a/src/bitmap.cpp
+++ b/src/bitmap.cpp
@@ -200,12 +200,16 @@ size_t Bitmap::GetSize() const {
 }
 
 ImageOpacity Bitmap::ComputeImageOpacity() const {
+	if (!GetTransparent()) {
+		return ImageOpacity::Opaque;
+	}
+
 	bool all_opaque = true;
 	bool all_transp = true;
 	bool alpha_1bit = true;
 
 	auto* p = reinterpret_cast<const uint32_t*>(pixels());
-	const auto mask = pixel_format.rgba_to_uint32_t(0, 0, 0, 0xFF);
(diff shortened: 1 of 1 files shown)


Wohlstand:
I tested this on my end, and I confirm the fix just works, the evidence is here: https://github.com/EasyRPG/Player/issues/3436#issuecomment-3020552001

fdelapena (approved):

PR #3454, opened 2025-07-26
Fix build on Arch

lufog:
Fix:
```
easyrpg-player/src/game_interpreter.cpp:4848:28: error: ‘RuntimePatches’ has not been declared
 4848 |                 auto key = RuntimePatches::VirtualKeys::VirtualKeyToInputKey(key_id);
      |                            ^~~~~~~~~~~~~~
```

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -41,6 +41,7 @@
 #include "game_system.h"
 #include "game_message.h"
 #include "game_pictures.h"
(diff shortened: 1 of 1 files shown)


carstene1ns (approved):
Yep, that was introduced yesterday. The fix is already available, but in an unrelated pull request.
So merging this one instead.

Ghabry (approved):
Thanks

PR #3456, opened 2025-07-28
Update dr_wav to 0.14, adapting to API changes

musicinmybrain:
[Version 0.14.0 of `dr_wav` 0.14.0](https://github.com/mackron/dr_libs/releases/tag/wav-0.14.0) was recently released, with a breaking API change that affects EasyRPG Player.

My motivation for this PR is that I maintain the [`dr_libs` package](https://src.fedoraproject.org/rpms/dr_libs) in Fedora, and our [`easyrpg-player` package](https://src.fedoraproject.org/rpms/easyrpg-player) is currently built with the system copy of `dr_wav` – something we [want to do whenever possible](https://docs.fedoraproject.org/en-US/packaging-guidelines/#bundling). I am preparing to update `dr_libs` to provide `dr_flac` 0.13.0, `dr_mp3` 0.7.0, and `dr_wav` 0.14.0, and I need to either patch dependent packages to work with the new versions (this PR is my attempt to do that), adjust them to go back to using bundled copies (with an explicit justification and upstream issue link), or request and maintain a compat package for the previous versions (preferably not indefinitely).

In `dr_wav` 0.14, the function `drwav_init_ex` gains a callback parameter `drwav_tell_proc onTell`. I added a new `GetPosition()` method to `Filesystem_Stream::InputStream` and used it to implement this callback. I think I’ve done this correctly, and tests seem to be passing (assuming I’m running them correctly), but review and testing by someone familiar with the code would certainly be appreciated.

### DIFF
--- a/src/external/dr_wav.h
+++ b/src/external/dr_wav.h
@@ -1537,6 +1537,10 @@ static DRWAV_INLINE int drwav__is_little_endian(void)
     return DRWAV_TRUE;
 #elif defined(__BYTE_ORDER) && defined(__LITTLE_ENDIAN) && __BYTE_ORDER == __LITTLE_ENDIAN
     return DRWAV_TRUE;
+// EasyRPG modification {
+#elif defined(WORDS_BIGENDIAN)
+    return DRWAV_FALSE;
+// }
 #else
     int n = 1;
     return (*(char*)&n) == 1;


carstene1ns:
Jenkins: ok to test

Ghabry (approved):
Thanks! Looks good to me (and made a quick play test to verify).

(I only added a conditional in case a system-wide installed dr_wav is still 0.13.x, but that one is also not relevant for Fedora obviously)

carstene1ns (approved):

musicinmybrain:
> Thanks! Looks good to me (and made a quick play test to verify).
> 
> (I only added a conditional in case a system-wide installed dr_wav is still 0.13.x, but that one is also not relevant for Fedora obviously)

Thanks! This is a reasonable idea, and it is actually slightly helpful for Fedora, because it means that the patch based on this PR, or a future release containing this PR, could be safely merged back to stable releases (currently, Fedora 41/42) even though the breaking `dr_wav` update will only be shipped in Fedora Rawhide and released in Fedora F3 and later.

----

Looking at https://github.com/EasyRPG/Player/pull/3456/commits/934524389c62421cee78a4f14a9e1ed403cb3e7f#diff-665f37409ef2960f9a676f6f9b75ae6423071d6d310b3d2ed8ba99f614180bebR1540,

```c
// EasyRPG modification {
#elif defined(WORDS_BIGENDIAN)
    return DRWAV_FALSE;
// }
```

it seems like this plumbs in the build system’s endianness check,

https://github.com/EasyRPG/Player/blob/14f950712b0803ef7133e5d0bc9aa768bdb84599/CMakeLists.txt#L891-L902

but in the case where we are using a system copy of `dr_wav.h` without the “EasyRPG modification,” we just fall back to

```c
    int n = 1;
    return (*(char*)&n) == 1;
```

which should give the same result, potentially at the cost of a few extra instructions, or possibly optimized to the same machine code as `return DRWAV_FALSE`. Therefore, there should be no practical disadvantage to using a system copy without this modification. Am I missing anything?

PR #3458, opened 2025-07-31
Fix --sound-volume incorrectly setting music volume

Pri0r:
As the title says, the `--sound-volume` argument was actually setting the music volume.


Ghabry:
Jenkins: test this please

fdelapena (approved):

PR #3466, opened 2025-08-16
Enable PS4 support

kluxi1:
Added code to support EasyRPG on PS4


resources/ps4/sce_module/libSceFios2.prx
Ghabry, line 1: Is the inclusion of the libSceFios2.prx, libc.prx and right.sprx necessary?

Are they not part of the ps4 toolchain?

src/game_player.cpp
```diff
@@ -903,9 +915,14 @@ void Game_Player::UpdatePan() {
...
 		dy = Utils::RoundTo<double>(std::abs(maniac_pan_current_y)) == std::ceil(std::abs(maniac_pan_current_y)) ? static_cast<int>(std::floor(dy2)) : static_cast<int>(std::ceil(dy2));
 	} else {
+		#ifndef __PS4__
 		dx = std::min(step, std::abs(pan_remain_x));
 		dy = std::min(step, std::abs(pan_remain_y));
-
+		#else
+		dx = min(step, std::abs(pan_remain_x));
+		dy = min(step, std::abs(pan_remain_y));
+		#endif
```
Ghabry, line 924: I don't really like these ifdefs here (and I wonder why it only fails in this file).

Can you provide the build error that happens without these ifdefs? Maybe can be fixed in a better way.

src/system.h
```diff
@@ -91,6 +91,11 @@
 #  define SUPPORT_MOUSE
 #  define SUPPORT_JOYSTICK
 #  define SUPPORT_JOYSTICK_AXIS
+#elif defined(__PS4__)
+#  define SUPPORT_ZOOM
+#  define SUPPORT_JOYSTICK
+#  define SUPPORT_JOYSTICK_AXIS
+#  define SUPPORT_FILE_BROWSER
```
Ghabry, line 98: The `SUPPORT_FILE_BROWSER` is for opening file open dialog in the soundfont and the font settings. Can you remove the define as this will only work on desktop platforms (or does the PS4 support anything like this?)

Ghabry (commented):
Thanks for your contribution. Very useful. Have no experience about PS4 dev so only some generic feedback.

CMakeLists.txt
```diff
@@ -1422,6 +1424,11 @@ elseif(PLAYER_CONSOLE_PORT AND NOT PLAYER_TARGET_PLATFORM STREQUAL "libretro")
 			FILE "${CMAKE_CURRENT_SOURCE_DIR}/resources/psvita/touch_ui.png" touch_ui.png)
 		install(FILES ${CMAKE_CURRENT_BINARY_DIR}/EasyRPG-Player.vpk
 			DESTINATION . COMPONENT psvita)
+	elseif(PS4)
+		add_executable(easyrpg-player "src/platform/ps4/main.cpp")
+		target_link_libraries(easyrpg-player ${PROJECT_NAME} bz2 SDL2 SceSysmodule ScePigletv2VSH SceUserService SceAudioOut SceSystemService ScePad samplerate freetype)
```
carstene1ns, line 1429: Where does the need for bzip support come from, I remember freetype maybe, but these should be included by the dependency resolution already?
Or is freetype provided by the host toolchain somehow (because of that special include)?

src/platform/ps4/main.cpp
```diff
@@ -0,0 +1,28 @@
+#include "player.h"
+#include "output.h"
+
+static void log(LogLevel lvl, std::string const& message, LogCallbackUserData userdata)
+{
+
```
carstene1ns, line 6: If there is no logging method, you can simply remove this...

src/platform/sdl/sdl2_audio.cpp
```diff
@@ -22,9 +22,15 @@
 #include <cassert>
 #include <cstdint>
 #include <chrono>
+#ifndef __PS4__
 #include <SDL.h>
 #include <SDL_audio.h>
 #include <SDL_version.h>
+#else
```
carstene1ns, line 29: How old is the SDL2 configuration for PS4? Likely this is not needed, since the paths are provided.

src/platform/sdl/sdl2_ui.cpp
```diff
@@ -1317,6 +1319,15 @@ void Sdl2Ui::vGetConfig(Game_ConfigVideo& cfg) const {
 	// WiiU always pauses apps in the background
 	cfg.pause_when_focus_lost.SetOptionVisible(false);
 #endif
+
```
carstene1ns, line 1322: `#elif`

src/platform/ps4/main.cpp
```diff
@@ -0,0 +1,28 @@
...
+	args.push_back("--project-path");
+	args.push_back("/data/easyrpg-player");
+
+	Output::SetLogCallback(log, nullptr);
+	
+	Player::Init(std::move(args));
+	
+	Player::Run();
+	
+	while(1)
```
carstene1ns, line 22: I am not sure why you want to avoid exiting the player. Does the system crash or is returning to the menu not possible?

This "hang" and the changes in the gamebrowser are a bit intrusive I think.

Maybe we could show a fullscreen message here on how to reset the system or other instructions instead.

src/platform/ps4/main.cpp
carstene1ns: I repeat: remove the function if unused.

CMakeLists.txt
kluxi1: The freetype library provided by the host toolchain depends on bz2. But the dependency resolution does not automatically include the bz2 library, so I have to add bz2 manually. 
I removed the other libraries from the target_link_libraries list since they are already added by the dependency resolution.

src/game_player.cpp
kluxi1: Hi Ghabry, 
Can you give me feedback on commit [8bbb766](https://github.com/EasyRPG/Player/pull/3466/commits/8bbb766a61ae7663ba469f6ac15a8d0ca55b95e5) please?

carstene1ns:
Waiting for another feedback from @Ghabry, but looks okay for me now.

Ghabry (approved):
Looks also fine to me.

Could you do me a favor:

Could you (with a camera) take some photos of playing games on the PS4 and a video recording of launching a game (e.g. Ib) on a PS4 and playing it for a short time.

I collect this media for our release note blog posts.

carstene1ns (approved):

carstene1ns:
jenkins: test this please

PR #3474, opened 2025-09-15
New Feature - Play Music From Timestamp

jetrotal:
Very dirty implementation of play music from timestamp.

TPC Syntax:
```cpp
@msg.choice {
    .case "MEMORIZE + TIMESTAMP" {
        @comment "memorize bgm"
        @cmd 11530, "", [0, 10] // [tick_variabletype, tick_variablevalue]
        
    }
    .case "PLAY MEMORIZED W TIMESTAMP" {
        @bgm.stop
        @comment "play memorized bgm"
        @cmd 11540, "", [0, 10] // [tick_variabletype, tick_variablevalue]
        
    }
    .cancel bl {
        
    }
}

```

I runs correctly with wav files. 

MIDI seems to have a glitch: Every time The music resets, the latest tick from the music is understood as 0:
 - first time playing a midi: 0 is 0
- if I reset the song at 9s: 9 is the new 0.
- If I reset the song again: 18s is the new 0.

----------------

(Oh boy, I can see this becoming another feature to never be aproved)

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -24,6 +24,7 @@
 #include <sstream>
 #include <string>
 #include <cassert>
+#include "audio.h"
 #include "game_interpreter.h"
 #include "async_handler.h"
 #include "game_dynrpg.h"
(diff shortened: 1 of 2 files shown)


jetrotal:
My first thought was to base this on the tick metrics the engine uses as a variable option, since the engine has it coded already and can be useful if someone tries something different, like dynamic music changes with musical changes.

That's why I chose that part of the code. It's a shame midi gets the tick value in such broken way.

PR #3479, opened 2025-10-10
Maniacs text rendering smaller updates

jetrotal:
Support For some missing maniacs text rendering tags:
- more colors: colors limit can go beyond `\c[19]`
- index mode for exFont icons: `$[n]` or `$[\v[n]]`
- array tags for both color and exfont: you can pick up with tags such as `\c[x,y] or $[x,y]`

image

```cpp
//TPC SYNTAX:
v[1] = 31
@msg.show "\c[19]AAA\c[\v[1]]AAA\c[3,3]AAA 
$a $[A] $[1] $[\v[1]] $[1,4]
"
```
System
exfont

I noticed some duplicated code when formating strings in game_message and game_windows... But I won't mess with those, because every time I propose structural changes my PR goes to limbo to never be aproved.

### DIFF
--- a/src/font.cpp
+++ b/src/font.cpp
@@ -939,12 +939,10 @@ Font::GlyphRet ExFont::vRender(char32_t glyph) const {
 	auto exfont = Cache::Exfont();
 	bm->Clear();
 
-	Rect rect(0, 0, 0, 0);
-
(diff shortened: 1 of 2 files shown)


jetrotal:
Just tested it with my batch of commands, seems to be working as intended.

Issue #3446, opened 2025-07-02
Transition: First frame not rendered


Ghabry:
Noticed this while implementing the proper version of the Mosaic transition:

It starts with "current_frame = 1" because Update is called before Draw.

Lt-knb:
Now that comparisons are easy, I found two more issues in Wave:
The vertical movement (phase) is wrong in half of the frames, due to the integer truncation in percentage.
In WaveOut, the horizontal movement (depth) is 30 fps, for the same reason

PR #3482, opened 2025-10-17
Fix Mosaic and Wave, and other transition issues

Lt-knb:
See #3446

Fix the following:
- First frame not rendered because Update is called before Draw.
- Duration is wrong (too short).
- Stretching (bottom right extension) in Mosaic, due to the random offset range being incorrect.
- Background trailing in Wave due to blitting "nothing".
- Depth (horizontal movement) is 30 fps in Wave because Percentage is integer (truncation).
Phase (vertical movement) is wrong in half of the frames for the same reason.
- Blind visibly starts 5 frames late, and pacing is 4+6 instead of 5.
- Stripes are 30 fps like Depth.
- The scale is wrong in Zoom.
- Most of the other transitions (like Scroll and Combine/Division) are "choppy" (but not 30 fps like Stripes).

All transitions are frame-based now.
Percentage was removed

### DIFF
--- a/src/transition.cpp
+++ b/src/transition.cpp
@@ -211,7 +211,7 @@ void Transition::Draw(Bitmap& dst) {
 	case TransitionFadeIn:
 	case TransitionFadeOut:
 		dst.Blit(0, 0, *screen1, screen1->GetRect(), 255);
-		dst.Blit(0, 0, *screen2, screen2->GetRect(), 255 * (current_frame + 1) / (total_frames - 2);
+		dst.Blit(0, 0, *screen2, screen2->GetRect(), 255 * (current_frame + 1) / (total_frames - 2));
 		break;
 	case TransitionRandomBlocks:
 	case TransitionRandomBlocksDown:


Ghabry:
@Lt-knb is their any work left to do here or is this ready?

Lt-knb:
There's one remaining issue that affects all transitions.
There's 3 extra frames in Hide+Show:
A black one after Hide, a "normal" one after Show, and one more black frame between Hide and Show.
Those are not there in RPG_RT.
So all transitions have a black flicker in the middle right now, even Wave and Mosaic.
What's causing those frames to appear, and how do we get rid of them?

Ghabry:
Imo, even though it is not 100% perfect this could be already reviewed and merged.

It fixes alot of stuff already and the "unsolved" problems are not regressions (they are also present in the upstream easyrpg-player build) 

---

Main motivation for "lets merge it" is that I have no idea how to get rid of the black frame. xD

The black flicker is another race condition it seems.

In `graphics.cpp` the transition is considered for one frame as `IsErasedNotActive` so the screen is cleared.

```cpp
if (transition.IsActive()) {
	min_z = transition.GetZ();
} else if (transition.IsErasedNotActive()) {
	min_z = transition.GetZ() + 1;
	dst.Clear();
}
```

Not sure how to correctly fix this currently.

Lt-knb:
Alright, the black flicker deserves a separate PR then.
Marked as ready

PR #3484, opened 2025-10-24
NEW FEATURE - GIF image support with animated bitmap handling

jetrotal:
EasyRPG is now able to render gif animations.
Based on libnsgif used in mkxp-z:  https://github.com/netsurf-plan9/libnsgif (MIT license).

gif for quick testing:
heroes

This PR Introduces GIF image decoding and animation support by adding new source files for GIF and LZW decoding, animated bitmap management, and GIF image handling. Updates build scripts to include new sources, extends Bitmap creation to detect and load GIFs as AnimatedBitmap, and registers '.gif' as a supported image type. Also ensures proper virtual destructor in Bitmap and integrates animated bitmap update logic.

I tried to inject the code very early on the image pipeline, to affect every image source.

-----

Known hiccups from current approach: Picture effects, such as flip or overlay color mode locks the gif into a single frame, I'm not that aware about the entire pipeline to solve that...

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,8 @@ EasyRPG Player makes use of the following 3rd party software:
   (Yoshio Uno), provided under the (3-clause) BSD license
 * [dr_wav] WAV audio loader and writer - Copyright (c) David Reid, provided
   under public domain or MIT-0
+* [libnsgif] for progressive animated GIF decoding - Copyright (c) 2004-2022 The
+  NetSurf Browser Project, provided under the MIT license
 
 ### 3rd party resources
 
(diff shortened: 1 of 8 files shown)


Ghabry:
When I attempted implementing APNG (animated PNG) or AseSprite animation stuff I was stuck at exactly the same problem: The bitmap is duplicated via a blit and this causes the animation to be lost. :)

The bitmap API is unfortunately far too complicated (sooooo many methods) and overwriting all of them is alot of work (and error prone as most overloads would be the same and just lame forwarders).

PR #3487, opened 2025-10-30
Maniacs Patch - Partial Implementation of game option commands

jetrotal:
Added support for various Maniac Patch runtime engine settings via a new command handler in Game_Interpreter. Introduced new Game_System methods for pause screen when focus lost, fatal mix settings, fast-forward text, frame skip mode and message mouse disabling. Updated Player and Window_Message to respect these options, including frame skipping and fast-forwarding message text.

missing features are: battle origin, and message face size.


Ghabry:
Can confirm that this doesn't persist in the savedata. It isn't even reset when returning to the Title (imo this should be changed).

Ghabry:
Correction. Only persisted are "Size of Face" and "Position of Battle Screen".

SaveSystem:

- 0x40: Face Width
- 0x41: Face Height
- 0x85: Battle Screen Position

Ghabry:
@jetrotal this should work now

Most notable change to your code is that RSHIFT is now a key bind to fast forward text as this feature is cool in general.

And the frame rate handling is now setting the Game FPS directly instead of doing some fast forward multiplier trickery (which conflicted with normal fast forward)

Mimigris:
The FPS changes are currently a bit broken:
- The FPS counter with V-Sync ON displays an incorrect count
screenshot_20260804_183557

- Setting V-Sync to OFF with a defined limit (e.g. 60) seems to run worse than before
- The line used when restarting the game as its number of seconds displayed derived from the FPS function, so if you have different TPS than 60 the display is incorrect
screenshot_20260804_183154

Ghabry:
Okay that is weird. I tested all this stuff when implementing the feature 😅. Will check again on Windows

TPS is (game) ticks per second btw

Ghabry:
Yeah there was a bug: I had frameskip enabled by default.

To make this niche feature easier to implement the frameskip is relative to the VSYNC or the FPS-Limit rate not the game updates. (when you have a 160 HZ Display you probably wont see the skipped frames). In general kinda dumb feature? We already frameskip automatically when rendering is slow :shrug: 

The TPS (FPS option in Fatal Mix) is now limited to 10 < fps < 500 as everything higher is unplayable crap. The reset counters and the message overlay use now independent timing so they should be the same independent of what kind of TPS is used.

Ghabry:
Tested this now at https://easyrpg.org/play/pr3487/?game=TestGame-Maniac&start-map-id=12&start-position=1%209 and imo this all works? @Mimigris  (except Picture Limit which I don't consider worth adding. There are already other ways to delete pictures)

Mimigris:
> Tested this now at https://easyrpg.org/play/pr3487/?game=TestGame-Maniac&start-map-id=12&start-position=1%209 and imo this all works? @Mimigris (except Picture Limit which I don't consider worth adding. There are already other ways to delete pictures)

The only issues I had with the command were tied to the FPS feature and all those points seem to have been tackled, so looks good to me 👍 

I agree for the picture part, does not make sense to support for the Player on its own as picture IDs are already handled dynamically, so unless a game is seen using it to delete pictures which leads to issues, no point in having it.

PR #3491, opened 2025-11-16
Add config option allowing map loader to prefer RPG Maker files

ianling:
Adds a new boolean to the player config: `PreferEasyRpgMapFiles` (default: `true`)

This makes the map loader attempt to load EasyRPG map files first, and fall back to RPG Maker map files on errors. Since it defaults to `true`, this PR retains the existing behavior of the EasyRPG Player, which attempts to load EasyRPG map files first.

I just wanted a way to turn this behavior off in my game, as it creates issues with my VCS setup (essentially, a pre-commit hook that runs lcf2xml on all the files in the project).

QA steps I tried:

* New option set to the default of `1` (true) in config:
  * .emu and .lmu present -> Player loads .emu ✅ 
  * .emu present, .lmu not present -> Player loads .emu ✅ 
  * .emu not present, .lmu present -> Player loads .lmu ✅ 
  * .emu and .lmu not present -> Player crashes with expected error (`Error: Loading of Map Map0001.lmu failed.` -- notice it loaded the .lmu last) ✅ 
* New option set to `0` (false) in config:
  * .emu and .lmu present -> Player loads .lmu ✅ 
  * .emu present, .lmu not present -> Player loads .emu ✅ 
  * .emu not present, .lmu present -> Player loads .lmu ✅ 
  * .emu and .lmu not present -> Player crashes with expected error (`Error: Loading of Map Map0001.emu failed.` -- notice it loaded the .emu last) ✅ 
* New option missing from config uses default value (`1`) ✅

### DIFF
--- a/src/game_config.cpp
+++ b/src/game_config.cpp
@@ -683,6 +683,7 @@ void Game_Config::LoadFromStream(Filesystem_Stream::InputStream& is) {
 	player.screenshot_timestamp.FromIni(ini);
 	player.automatic_screenshots.FromIni(ini);
 	player.automatic_screenshots_interval.FromIni(ini);
+	player.prefer_easyrpg_map_files.FromIni(ini);
 }
 
 void Game_Config::WriteToStream(Filesystem_Stream::OutputStream& os) const {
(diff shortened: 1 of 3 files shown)


ianling:
Done

PR #3498, opened 2025-12-04
Maniacs Patch - Save Image command

jetrotal:
Implements the CommandManiacSaveImage function, allowing saving of screen or picture images to disk with options for dynamic effects and opacity. Registers the new command in ExecuteCommand and updates the header file with its declaration.

### Maniac Patch: Image Saving Support

* Added `CommandManiacSaveImage` to `Game_Interpreter`, which enables saving either the current screen or a specified picture to a PNG file. The command supports dynamic mode (applying color tone, flash, flip effects) and an option to force opaque pixels.
* Registered the new Maniac event command (`Maniac_SaveImage`, command ID 3026) in the interpreter's command dispatcher, allowing it to be invoked from events.
* Declared the new command function in the `Game_Interpreter` class header (`game_interpreter.h`).

### Supporting Changes

* Included the `cache.h` header in `game_interpreter.cpp` to support bitmap effect processing for dynamic image saving.

### DIFF
--- a/src/baseui.cpp
+++ b/src/baseui.cpp
@@ -63,9 +63,7 @@ BaseUi::BaseUi(const Game_Config& cfg)
 }
 
 BitmapRef BaseUi::CaptureScreen() {
-	BitmapRef capture = Bitmap::Create(main_surface->width(), main_surface->height(), false);
-	capture->BlitFast(0, 0, *main_surface, main_surface->GetRect(), Opacity::Opaque());
-	return capture;
+	return Bitmap::Create(*main_surface, main_surface->GetRect(), false);
 }
 
 void BaseUi::CleanDisplay() {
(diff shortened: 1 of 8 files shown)


jetrotal:
TODO:
- Properly export visible sprite instead of complete image sheet.

Ghabry:
Could you create a simple test case for me that has these things in it:

- Save entire screen
- Save a single image (you can use `320_240.png` from the testgame) with an active spritesheet in all 4 combinations (with effects/without effects + opaque / non-opaque)

(btw wow `static` vs `dynamic` is such a awful way to name this. Why not apply effects))

---

To ensure this works when the save directory is redirected (e.g. web player or game inside a ZIP) ``FileFinder::OpenImage`` requires `open_generic_with_fallback`.

----

Also saving of a single picture looks pretty complicated. Will see if I can simplify this.

Ghabry (approved):

Ghabry (approved):
simplified the logic in the command and implemented RGBA saving. Matches now exactly what Maniacs does

carstene1ns (approved):

PR #3499, opened 2025-12-05
Maniacs Patch - Zoom Command

jetrotal:
Implements Maniac Zoom support, allowing zooming of specific map layers with configurable center, rate, duration, and layer Itself. 

- Adds new interpreter command: 3032
- zoom state management in Game_Screen: SetZoom() and UpdateZoom()
- and rendering logic in Graphics to apply zoom transforms to selected layers:  GetZForManiacLayer() - It has a weird reordenation of those entries, Graphics::LocalDraw()

-----------------------

Right now lacks support on savefiles.

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -4402,6 +4402,7 @@ bool Game_Interpreter::CommandManiacZoom(lcf::rpg::EventCommand const& com) {
 		SetupWaitFrames(duration);
 	}
 
+	Output::Debug("Maniac Zoom: CenterX {}, CenterY {}, Rate {}, Duration {}, Layer {}", center_x, center_y, rate, duration, layer);
 	return true;
 }
 
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -812,6 +812,8 @@ bool Game_Interpreter::ExecuteCommand(lcf::rpg::EventCommand const& com) {
 			return CmdSetup<&Game_Interpreter::CommandManiacGetGameInfo, 8>(com);
 		case Cmd::EasyRpg_SetInterpreterFlag:
 			return CmdSetup<&Game_Interpreter::CommandEasyRpgSetInterpreterFlag, 2>(com);
+		case static_cast<Cmd>(3032): // Maniac_Zoom
+			return CmdSetup<&Game_Interpreter_Map::CommandManiacZoom, 7>(com);
 		case Cmd::EasyRpg_ProcessJson:
 			return CmdSetup<&Game_Interpreter::CommandEasyRpgProcessJson, 8>(com);
 		case Cmd::EasyRpg_CloneMapEvent:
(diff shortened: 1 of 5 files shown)


Ghabry:
While I appreciate all your contributions and I just havnt found time to look at all of them I have to criticize the AI usage especially in the commit messages and the pull request texts. 

These texts are verbose and unfun to read.

Everything after **Maniac Zoom feature implementation** could be removed and just a remark "see commit message" added which mentions the most relevant parts. (or just copy-paste the commit message into the PR description. Don't reword anything, is a waste of time).

- Added a new interpreter command Maniac_Zoom (CommandManiacZoom) that parses parameters and triggers zoom effects, including handling for cancellation/reset and wait behavior. _This could be briefly mentioned in the commit message. "parses paremeters" is a useless information, all commands have to do this. About cancellation/reset and wait behaviour: Dont think its necessary to list all the detailed features it has._
- Implemented zoom state management in Game_Screen, including current and target zoom positions/rates, duration, layer selection, and update logic for smooth interpolation. _That is useful but imo too detailed. Is enough to mention that the zoom state is saved in Game_Screen where the necessary fields were added. Btw: I don't see this state preserved in the save file? What is Maniacs doing here?_
- Integrated Maniac Zoom into the graphics rendering pipeline: affected layers are rendered to an intermediate buffer and zoomed using the specified parameters, while other layers are drawn normally on top. This ensures the zoom effect is correctly applied to the intended layer group. _Could be shortened: Modified the rendering code to handle the zoom feature. Everything below a certain layer is drawn zoomed, the layers above are drawn normally on top. Btw why is drawing in an intermediate buffer necessary here? It should also work without._
- Added utility for mapping Maniac Zoom layers to drawable Z-order, ensuring correct rendering boundaries for zoomed content. _Here I wonder more about the why instead of the what. Why was this function necessary? Don't we already have layer mapping code? What is Maniacs doing different here?_
-  Updated includes and initialization to support new features and ensure proper access to game screen and main data. _This information is unnecessary. Its something you always have to do_

----

Same for the commit message. Make the texts short bullet points. Keep it to a bare minimum (except when you think that a feature is so complecated it is worth a longer explanation)

Remove everything that is redundant like this last sentence:

> This enables compatibility with Maniacs Patch zoom event commands.

jetrotal:
removed all the automatic text from the commit description. I was simply tired after fighting the code all night and let copilot autofill it.

src/graphics.cpp
```diff
@@ -118,9 +122,86 @@ void Graphics::Draw(Bitmap& dst) {
 	LocalDraw(dst, min_z, max_z);
 }
 
+static Drawable::Z_t GetZForManiacLayer(int layer) {
+	if (layer <= 0) return std::numeric_limits<Drawable::Z_t>::min();
+	if (layer >= 10) return std::numeric_limits<Drawable::Z_t>::max();
+
+	// Layer 9 (Windows) includes Message Text (Overlay)
+	if (layer == 9) return Priority_Overlay;
```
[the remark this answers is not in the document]
jetrotal: debug overlay and "reseting" messages were affected. fps overlay wasn't. Those are now solved.

That made me want to modify this command, to allow messing with menus as well, So we could customize menus that get too small in resolutions bigger than rm2k3 default.

src/game_interpreter.cpp
```diff
@@ -812,6 +812,8 @@ bool Game_Interpreter::ExecuteCommand(lcf::rpg::EventCommand const& com) {
 			return CmdSetup<&Game_Interpreter::CommandManiacGetGameInfo, 8>(com);
 		case Cmd::EasyRpg_SetInterpreterFlag:
 			return CmdSetup<&Game_Interpreter::CommandEasyRpgSetInterpreterFlag, 2>(com);
+		case static_cast<Cmd>(3032): // Maniac_Zoom
```
ianling: Should open a PR to liblcf to add this command to the enum: https://github.com/EasyRPG/liblcf/blob/master/generator/csv/enums_easyrpg.csv#L14

ianling: Ah my bad, if there is more than one missing, I would say don't worry about it as a blocker on this PR, but someone should clean up the missing ones

ianling (approved):

Ghabry:
setting this to 0.9: The Zoom conflicts with the hardware renderer and to prevent some massive headache on my side this will be added after SDL_GPU is merged (also will give a nice opportunity to add Zoom Out)

Ghabry:
Jenkins: test this please

Mimigris:
Checked a bit the command and compared it with Maniac:
- It seems the layer 9 and layer 10 of the command got mixed together (not that surprising, as the command in the editor says that the layer system is the same than for pictures while it is actually shifted by 1, which can lead to confusion). Layer 9 zoom should affect the message window, pictures of layer 8, + everything below, while layer 10 zoom should affect in addition the timer and pictures of layer 9, meaning everything is zoomed.
- A zoom layer of 11 or more should be the same as a zoom layer 0 (no visible zoom).
- The Center X and Center Y parameters are incorrectly handled: for Maniac, those points are those that represent the coordinates of the center of the screen that can be zoomed in, which implies that it must be at the center enough to be zoomed in, and otherwise the visible zoom will be on the nearest center of the screen possible. Here, by zooming at 150% on the coordinates 0, 0, we can see that Maniac does not display the top and left part of the screen due to zooming more at the center, while EasyRPG can display them.
image


Also for reference, some Maniac edge-cases for the Zoom command handling:
- The zoom range is 101% ~ 24000%, other values will be the same as a 100% zoom (on our side we also use 0 ~ 99% for zooming out)
- If Center X or Center Y has a negative value, zooming with a scale of 302% or more will consider said Center value to be the one of the center of the screen rather than the one set

PR #3500, opened 2025-12-06
Maniacs Patch - ChangePictureId command + Pictures Inspector

jetrotal:
Had to put those two together because I needed one to test another

### Change Picture command:
Added support for the Maniac Patch ChangePictureId event command, enabling move, swap, and slide operations on picture IDs with error handling and sprite refresh logic. This replaces the previous stub implementation and ensures proper picture management for Maniac Patch compatibility.

Map for testing the command:
[Map0003.zip](https://github.com/user-attachments/files/23979021/Map0003.zip)

### Pictures Inspector:
Introduces a new debug scene and windows for viewing and inspecting active pictures and string windows. Updates scene and debug menu logic to support the new 'DebugPicture' scene, adds menu option, and implements detailed property display for both image and string pictures.

image


Ghabry:
The code actually looks fine to me. Only:

1. integrated the Picture Inspector better in the existing Debug UI
2. extracted these complex lambda functions into methods in the Picture class

Only have to test it *g*

Ghabry:
Testing some out of bounds stuff (as this is always the interesting part xD) I noticed some weird behaviours depending on whether src or dest are out of bounds.

Will document it when I fully understood it.

So this will take a bit longer to get everything right... Unexpected problems :cry:

Ghabry:
Works including all my tests concerning out of bounds and overlapping ranges.

Honestly this took again far too long to "review" (well, rewrite).

Turns out most of the code in Swap was unnecessary. It's enough to swap the two elements in the vector and then to adjust the IDs.

Move can be simplified by just doing a swap followed by deleting the one swapped with.

In general looking at the original code the AI is too afraid to just invent new API. Instead it hacks around by doing lots of function calls.

Ghabry (approved):
imo this is ok now

not reviewing the debug scene. Considering this is debug code I don't really care how its written (based on my tests it works)

Mimigris:
Checked a bit the Picture Inspector (will do later on ChangePictureID as it requires playing with the command), only thing I have to say so far is that the empty filename picture bug from RPG_RT that we partially support will not be correctly listed (listed in the list as undefined even if the picture still exists and is displayed).
As this only affects the wallpaper computer screen of Yume 2kki on its picture of ID 2 and the TestGame-2000, it may not be that big of a deal, but was still worth mentioning.

Mimigris:
Okay, finished testing the command I think.
The only issue I found is that the picture layer is not swapped along the picture.

More of a feedback on something else, but for the picture inspector, when the using the spritesheet feature, the first cell is displayed as 0, wouldn't it make more sense to have it be displayed as 1, as it seems to be how it is referred in RPG Maker? Or is it how it is internally referenced in the Player, and thus was simpler to implement this way?

screenshot_20260731_223504

Ghabry:
Internally the cell number is 0-based but the editor uses 1-based. 0 is usually more convenient for coding but non-coding humans are used to start counting with 1.

As this is user-facing it should be 1-based. I changed it.

Also renamed "Chroma" to "Transparent" as this is the term used everywhere. Also to nit-pick Chroma is not really the correct term here as this refers to "Chroma Keying" which is about background removal (typical application is a Green Screen).

Also removed the "GetPicInfo" commit concerning missing pictures. The SetPixel PR has a better fix and this will only cause conflicts.

Ghabry:
It feels abit like a waste of time as this command feels so useless but all test cases pass...

PR #3501, opened 2025-12-07
Maniacs Patch - SetPicturePixel Command + GetPictureInfo Update

jetrotal:
Another one that must come in pairs:

### SetPicturePixel Command

Implements the CommandManiacSetPicturePixel function to allow direct pixel manipulation of game pictures via event commands. Handles bitmap uniqueness, format conversion, and window picture detachment to ensure safe editing. Updates command dispatch logic and header declaration.

--------------------

### Update - GetPictureInfo (pixel data extraction)

Adds support for extracting raw pixel data from pictures, including window-type pictures with forced refresh. Optimizes out-of-bounds handling to preserve variable values and ensures only 32-bit bitmaps are processed. Existing logic for info types 0, 1, and 2 remains unchanged.

------------------------

TODO: 
- ~~Save persistence~~
- Few Quirks that I could not figure out, e.g.: 
      I assumed some opacity changes between maniacs and easyRPG were due to out of bounds color picking. But I was wrong, someone smarter than me could help me with that.

[Map0003.zip](https://github.com/user-attachments/files/24009634/Map0003.zip)


jetrotal:
Also supporting game Info pixels, test map:
[Map0003.zip](https://github.com/user-attachments/files/24064127/Map0003.zip)

Ghabry (approved):
To the surprise of noone there was no feedback from the community or the examples were such a brainfck that they needed 10 other features we dont support yet.

So the only two options we have are:

- Keep this in limbo forever
- yolo merge and then wait for people complaining in a few days because they found a bug in the feature

Mimigris:
> To the surprise of noone there was no feedback from the community or the examples were such a brainfck that they needed 10 other features we dont support yet.
> 
> So the only two options we have are:
> 
>     * Keep this in limbo forever
> 
>     * yolo merge and then wait for people complaining in a few days because they found a bug in the feature

I could try to find time during the week to test the commands if needed, once I'm done working on some other stuffs.

Is the SetPicturePixel command the one called `Edit  Picture(Tile)` in the editor, or is it another? As I'm not entirely up to date with the recent Maniac versions, that the vocabulary is sometimes not consistent and that some commands are locked behind TPC which forces looking for documentation, I'm not really familiar with that.

-------------------

About lacking feedback from the community on such pull requests, I think there are several things:
- More specific to this case, but due to being a Maniac patch feature, it can be hard to even understand how the command can be set and can be used (see my previous comment as an example).
- Just opening a pull request for a feature on GitHub does not attract a lot of people, as the only people that will become aware of this pull request are 1) people on Matrix or Discord, active or not, when backreading or reading messages, 2) people on IRC, if they are currently active when the pull request is created or updated, 3) people randomly browsing the pull requests of the GitHub, and 4) people that are subscribed to changes made to the EasyRPG Player repo. All of these options are limited, as only active people in the community will find it, and if there was a request for testers that was listed inside the pull request, it goes unseen unless you manually open the page.
- The fact you can test the pull requests is hard to see for a random user not familiar with GitHub, as you need to click on the "All checks have passed" button to reveal a list of platforms that the pull request was built for, and it is the only indication of that. This excludes de facto everyone not familiar enough with GitHub and with how the EasyRPG Player repository builds its versions for that.
- Lastly, even if a pull request is opened and is ready, it can be hard to really know if it is in a ready for testers state: there are a lot of different pull requests on the repository these days, all opened not set to WIP, some still requiring reviews for the idea if valid and some not, some made by AI that are sometimes held by glue meaning you do not even know if it is a pull request that will survive the basic review phase, making it harder for someone that wants to test to know if it is worth the effort or not.

If I had suggestions on what could be done to improve that, I think calls for testers could potentially be done:
- In the EasyRPG chat, when a feature is ready for testing; perhaps having a "wanted board" channel to list that users could be wanted for testing selected pull requests or features
- In nearby related Discord servers, such as the easyrpg channel of the RPG Maker 2003 Discord server or in the Maniac patch server (for the latter, if it is for a Maniac patch feature and that they see no issue with doing that over there)
- On our social networks (Bluesky, Fediverse, Twitter); as this last option aims at a broader audience, though, it may be more limited in scope (e.g. lot of Android users which may only be able to use the Player but not the editor), so it may only be doable for global changes that are more accessible for random users (doing tests this way would also require telling players where feedback can be given)

Perhaps having a tag for pull requests looking for testers could also be a thing, if it can help.

Mimigris:
Technically unrelated to this pull request, but the test case created by Jetrotal for testing this feature includes a Show String Picture command that displays the text `..\charset\actor1`, which makes the Player crash. Should this be adressed?

Mimigris:
It seems that using the command OutputImage with a string picture is now broken in this version and makes the Player crash.

Mimigris:
Behaviour of Get Picture Pixel Info for pictures that do not exist/no longer exist:

Maniac:
- Picture that has never existed: returns 0 for the pixels extracted
- Picture that has once existed but has been deleted: crashes

The Player currently crashes in both cases, may want to sanitise this to either throw a warning or consider the pixels returned to be 0 in both cases.

Mimigris:
All of the tests present in the test map sent by Jetrotal involved writing pixels over a string picture: none were conducted over normal pictures, and it seems that writing pixels on normal pictures currently does not work.
Between that and the various crashes involved, this pull request requires more work before being ready and being able to be fully tested.

Ghabry:
That reading from invalid picture seems to vary based on the MP version. For me it just does nothing in both cases. So lets ignore this...

Ghabry:
Fixed all these embarrassing crashes. Thanks for testing.

Normal picture editing one was also dumb: Forgot to assign the new bitmap xD

Mimigris:
At a quick glance, I still can't manage to make the feature work correctly with a picture, at least with the code I had set.
I had made a small test to run and edit a picture using a colour picker (see below), and it still seems to render as black in the Player.
image

You can find the program below if it can help (I'll try to go back to smaller tests/less convoluted tests later on to see how it works more simply):
[EditPictureTest.zip](https://github.com/user-attachments/files/30598966/EditPictureTest.zip)

Mimigris:
I'll check a bit more those commands tomorrow, but for now, from what I can see that makes my script not work:
- The spritesheet feature is ignored for the coordinates drawn (e.g. if you use the spritesheet feature to only keep the right part of the screen, the writing still starts on the inexisting top left in the Player, which is not the case in Maniac as it only writes on the existing part of the picture)
- For some reason the command seems to only work once per picture which makes it hard to use

(I'll also need to check how the alpha value works, as I think I did not understood that when making my example so the alpha value system may just not work)

Mimigris:
Output Image also seems to fail (no output done) if the picture was edited.

Mimigris:
Not fully tied to this pull request (can open a different issue if needed), but currently Output Picture also fails if done on a picture that was affected by the empty filename exploit.
Maniac seems to have gotten rid of the empty filename exploit, instead deleting the picture ID if called, so the empty filename part could also be disabled if done in a Maniac enabled game.

Ghabry:
This should work now but what I can already see in your test is that we handle the alpha value different: I consider the alpha value the transparency. Maniacs does nothing?!

Btw is the picture you load for testing with active transparency or opaque? Maybe this makes a difference?

Also I'm pretty sure that there are some cache invalidation problems: When a Picture has an effect e.g. Tone and you switch between has Tone and no Tone its possible that it operates on the wrong image but this needs checking :sweat_smile:

Mimigris:
> This should work now but what I can already see in your test is that we handle the alpha value different: I consider the alpha value the transparency. Maniacs does nothing?!

I'll investigate that later on to understand how it works, as I'm also struggling to understand how Maniac handles it so far.

> Btw is the picture you load for testing with active transparency or opaque? Maybe this makes a difference?

It has the property `Use Transparent Colour` ticked; at a first glance it seems Maniac just ignores the alpha value if it is not ticked, though this requires a bit more testing first.

Mimigris:
Did some tests with Get Picture Info: Pixel to familiarise with it in the meantime:
- Currently we lack all support on the effects flags (we always consider Get Picture Info: Pixel to not retrieve the effects; if ticked, it seems that Maniac takes into account colours changed led by Picture Tint and Screen Tint)
- Index 0 on a non-transparent picture should retrieve "no effect" and "effects" in the same way than "ignore alpha" and "effects, ignore alpha" rather than getting different values
- Index 0 on a transparent picture should retrieve the value in the same way than if ignore alpha was used normally on the colour, for all modes (currently it retrieves 0; it's the case you mentioned earlier due to library limitations)

Mimigris:
From what I understand of the Alpha value when editing a picture:
- If Alpha is set to 0 and that the picture has the transparent colour property, a transparent pixel is written (transparent as-in same as a index 0 pixel)
- If Alpha != 0 or that the picture does not has the transparent colour property, a normal pixel is written

In both cases, the pixel written keeps the written AARRGGBB value, even if storing its complete alpha value or colour data does not make sense for the edited pixel.

Ghabry:
Ah so this is probably the same behavior as when showing pictures: try loading a picture with semi transparent pixel. Easyrpg will correctly show them. Maniac considers them opaque

Mimigris:
Attempting to edit a picture affected by a picture tint seems to only work once (the picture needs to be set again to work with the command).

Ghabry:
Okay I'll fix all the tint stuff and then this should be good enough. Thanks

The bogus alpha blending is too costly to implement so I'll ignore this for now.
Could be emulated almost for free in the GPU branch though.

Mimigris:
Just to document out of range Edit Picture Maniac properties if needed at some point:
- If the X and/or Y parameter is negative, Maniac will not start writing from the negative coordinate but from the coordinate 0 of said coordinate, first writing the value 0 to the pixels of the area to write, then writing the actual pixels to write (actual pixels to write starting from their expected coordinate)
- If the X or Y parameter is over the picture size, as the writing will be out of the picture it will not matter
- If the width or height parameter is null or negative, there will be no area to draw -> nothing to do

Not much to see on Get Picture Info: Pixel, meanwhile, requesting pixel data from an out of range area will just not alter variables.

Mimigris:
Random weird Maniac behaviour, while editing pixels of a picture to use an alpha value of 1~127 is identical to using a negative alpha value, outputting them to a picture file will render them transparent (so difference between display and export).

Another thing just on our side, again against the empty filename feature, editing a picture that was affected by the empty filename feature does not work.

With that I think I tested everything I could think of for now with the feature.

Mimigris:
Hmmm, went on to test but this pull request currently breaks the command Get Picture Info: with the parameter Current Position, Current Size or the parameter Target Position, Target Size, retrieving Width and Height return 0 instead of the expected value, while for X, Y, Left, Top, Right, and Bottom, the parameters X, Left and Right return Center X, while the others return Center Y.

Ghabry:
Modifying pictures with effects should work now.

The Get Pic Info not working for effects is an oversight. Its just not implemented at all.

> Hmmm, went on to test but this pull request currently breaks the command Get Picture Info: with the parameter Current Position, Current Size or the parameter Target Position, Target Size, retrieving Width and Height return 0 instead of the expected value, while for X, Y, Left, Top, Right, and Bottom, the parameters X, Left and Right return Center X, while the others return Center Y.

Good catch. Somehow a bit of code was moved around a broke it.

Ghabry:
~~Retrieving the pixels with effects applied in GetPicInfo is now also working. This seems to be broking in Maniacs? I always get a black image back.~~

~~Well I implemented it in a way that makes sense for me. (For testing I grabbed pixel from a Tone modified image and then copied them over in a unmodified image)~~

Azure told me that you must "Wait 0" after Show Picture to get the effect applied.

> The moddified Bitmap is refreshed after Executing Event so it's not bug

> There's simply no data on the frame right after you use show picture 

Well I would object that it _is_ a bug because the behavior is not intuitive but okay.

---

(btw copying data from transparent areas behaves different in Maniac vs. EasyRPG. As already explained I will not fix this right now due to technical reasons)

Mimigris:
> Azure told me that you must "Wait 0" after Show Picture to get the effect applied.
> 
> > The moddified Bitmap is refreshed after Executing Event so it's not bug
> 
> > There's simply no data on the frame right after you use show picture
> 
> Well I would object that it _is_ a bug because the behavior is not intuitive but okay.

When I was pestering yesterday that the command didn't seemed to work, I was facing the exact same issue you mentioned here, as I was constantly asking while a picture was being created (so was too early to be able to retrieve the pixel).

Maniac would truly need to have an official complete documentation at some point to explain how commands *actually* work and should be used...

---

I tested now with the updated build, and while the Get Picture Info features now works, some of the Get Picture Info ones may be broken (I suppose it is tied to the edits you were making to implement effects), as I now encounter crashes.

Feel free to use this test case to test Get Picture Info, that's one I'm using and what currently crashes:
[EditPictureTest2.zip](https://github.com/user-attachments/files/30636244/EditPictureTest2.zip)

Ghabry:
The crash was because of a sanity check we do: Computing effects when no effect is active will abort the Player on purpose. It was never intended that this codepath is taken from the outside.

I refactored the Sprite code now to make this less error prone (and to grab less into implementation details which doesn't work properly anyway as we just noticed)

---

Modified your map to extract the RGBA values seperately. Also found a bug in our implementation of bitshift right (>>) function I fixed now:

[Map0025.zip](https://github.com/user-attachments/files/30657896/Map0025.zip)

---

What I usually see now is slight inaccuracies in the RGBA values (due to algorithmic differences in how the effect value is calculated).

Well and the extracting of values when alpha is used.

I don't consider this fixable in a reasonable time.

Mimigris:
> Modified your map to extract the RGBA values seperately. Also found a bug in our implementation of bitshift right (>>) function I fixed now:
> 
> [Map0025.zip](https://github.com/user-attachments/files/30657896/Map0025.zip)

While the script edit you made seems to work to retrieve values, it required some adjustments for the string picture display:
- The no effect values for RGB pointed to incorrect variables
- The display of the red and blue values were all reversed

Overall it seems to work greatly, and the commands seem to work properly outside of the cases mentioned (which should hopefully not matter much as long as developers using those commands only use those commands to read and write without doing any exact value check).

Ghabry:
Right the order is BGRA, not RGBA so the labels are wrong 😅

Ghabry:
Was there anything I forgot here? Otherwise would merge this now.

I only see the incorrect RGBA stuff but this was just visual in the test game so doesnt affect the PR.

Mimigris:
> Was there anything I forgot here?

From what I can tell, we have an incorrect handling of the transparency for Edit Picture: for Maniac, if the Alpha value is not ignored, set to 0, and the picture allows transparency, then it will write transparency, and otherwise will write a normal colour. Not being 1:1 with the range is not necessarily a big deal, as the alpha values -1 and 0 (which are the values you can get from the commands Get Picture Info and Get Game Info) are correctly handled, so as long as games use those values there will be no discrepancy, but as we ignore if a picture uses transparency to apply the parameter, it could more easily lead to inconsistencies during the application.

The other inaccuracies outside of that are just those mentioned previously (differently handling of picture transparency for Get Picture Info: Pixel, different pixels retrieved when a tint is applied), but as the graphical pixels retrieved will be close or identical, and that those would not be trivial to solve, I do not think it will cause a lot of issues.

Ghabry:
Also rechecked this now: Yeah except for these minor inaccuracies with the blend modes only transparency stuff appears to be broken due to how the renderer is implemented. Best here is opening a new issue about the wrong transparency handling with these commands and then lets merge this one :)

PR #3502, opened 2025-12-07
Maniacs Patch - Rewrite Map - AB Autotiling Support

jetrotal:
I guess this completes the last feature from Rewrite Map Commap.

Introduces IsWaterTile and IsDeepWaterTile helpers and implements specialized autotiling logic for water and deep water tiles in RecalculateAutotile. This enables more accurate coastline and deep water boundary rendering, including diagonal and cardinal neighbor checks for seamless water transitions.

--------------------------------------

image
image

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -4932,7 +4932,7 @@ bool Game_Interpreter::CommandManiacRewriteMap(lcf::rpg::EventCommand const& com
 	}
 
 	int mode = com.parameters[0];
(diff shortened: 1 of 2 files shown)


ghost (approved):
Looks good to me. It could be nice adding comments or function names on parts are from patches or maniac patch specific imo.

Ghabry (approved):
I will accept it because it apparently works and is a nice feature 

Don't have enough expertise in this field to reason about the code. 

Will recheck this code when I implement the QML Map code for the editor.

PR #3504, opened 2025-12-09
Maniacs Patch Commands - Missing simple variable parameters support

jetrotal:
Enhanced several event command handlers to support PatchManiac-specific variable and bitfield-based parameters. This allows dynamic values for BGM, SFX, system graphics, event location trading, map tileset changes, and parallax background changes when the PatchManiac patch is active.


Ghabry (approved):
Finally came around to test this... Found one bug that I just fixed and is working correct now for all the cases.

PR #3507, opened 2025-12-15
Fix CMake Linter error

Ghabry:


carstene1ns:
The real problem is here, that GitHub provides no direct way to have the linter say "there are recommendations or possible problems" without having the workflow fail.
We could let it succeed and nobody cares about the log or we could add a script to mess with the api to have the output changed to "warning".

PR #3510, opened 2025-12-31
[WideScreen] Various fixes related with FakeResolution Mode (Pictures/Panoramas' position & Camera Scrolling)

### LINKED ISSUE #3342: [Wide screen] Bug with "Change Parallax Back"
- The project use the resolution 416x240 (16/9), with "FakeResolution=1" flag
- A map has a size of 26x15 (adapted to 16/9 ratio) and use a 320x240px Background
- When using the "Change Parallax Background" command with a 416x240 Background, the background is displayed incorrectly

The problem seems to be linked to the change of Background size and the FakeResolution feature.

Here is a project isolating this issue. Talk to Alex to change the background.

[Project Change Parallax Back bug.zip](https://github.com/user-attachments/files/18547615/Project.Change.Parallax.Back.bug.zip)

Image

### LINKED ISSUE #3509: [Wide screen] Panorama displayed with the wrong offset
EasyRPG display incorrectly the Panorama with this setup:
- Use of a custom resolution (416x240) with the FakeResolution flag (the problem disappears without the flag)
- A map of 23x12 using a Panorama of the same size than the map (RPG Maker even display the map)

Image

On EasyRPG:
Image

Here is a project isolating the bug:
[test_map_easyrpg_report_2.zip](https://github.com/user-attachments/files/24372064/test_map_easyrpg_report_2.zip)

BlisterB:
This PR will address problems of Panoramas/Pictures not being correctly displayed when using Custom Resolution and the FakeResolution mode.
Problems occur when displaying a map smaller than the screen. With the FakeResolution mode, the map is centered (and black bar are displayed around the screen.
But Panoramas and pictures offset are not calculated correctly, resulting in wrong positioning.

I discovered that the root cause was that the code was assuming that small maps were always of the size of 320x240, therefore maps with size larger than 320x240 but smaller than the Custom Resolution were not handled correctly.

Please do not merge yet, the issue related to Panorama should be fixed, but I discovered that we have a similar problem with Pictures.

As mentioned in #3509, Aëdemphia is a great game to test this PR.

This also fixes #3342 .

Fix #3509

### DIFF
--- a/src/game_map.cpp
+++ b/src/game_map.cpp
@@ -2051,9 +2051,6 @@ void Game_Map::Caching::MapEventCache::RemoveEvent(const lcf::rpg::Event& ev) {
 namespace {
 	int parallax_width;
 	int parallax_height;
-
-	bool parallax_fake_x;
-	bool parallax_fake_y;
 }
 
 /* Helper function to get the current parallax parameters. If the default
@@ -2157,8 +2154,6 @@ void Game_Map::Parallax::ResetPositionX() {
 		return;
 	}
 
-	parallax_fake_x = false;
-
 	if (!params.scroll_horz && !LoopHorizontal()) {
 		// What is the width of the panorama to display on screen?
 		int pan_screen_width = Player::screen_width;
@@ -2185,10 +2180,7 @@ void Game_Map::Parallax::ResetPositionX() {
 			}
 		} else {
 			panorama.pan_x = 0;
-			parallax_fake_x = true;
 		}
-	} else {
-		parallax_fake_x = true;
 	}
 }
 
@@ -2198,8 +2190,6 @@ void Game_Map::Parallax::ResetPositionY() {
 	if (params.name.empty()) {
 		return;
 	}
-
-	parallax_fake_y = false;
 
 	if (!params.scroll_vert && !Game_Map::LoopVertical()) {
 		// What is the height of the panorama to display on screen?
@@ -2222,10 +2212,7 @@ void Game_Map::Parallax::ResetPositionY() {
 			SetPositionY(pv);
 		} else {
 			panorama.pan_y = 0;
-			parallax_fake_y = true;
 		}
-	} else {
-		parallax_fake_y = true;
 	}
 }
 


BlisterB:
[ca660b6](https://github.com/EasyRPG/Player/pull/3510/commits/ca660b6527d6a83866b180c19176fc9ab4b009e5) Fixes the calculation of Picture's offset (the fix is similar to [e073662](https://github.com/EasyRPG/Player/pull/3510/commits/e073662a39412d26ae70d106198115c3e4dfa71c) for Panoramas).

At first it didn't solve issues in Aëdemphia for some map smaller than the screen.
I discovered that the game uses the famous trick for centering pictures scrolling with the map. It retrieves the screen X/Y position of an event with "Control Variable" command, and place the picture accordingly to the result.

image

I discovered that when using FakeResolution mode, the result was wrong, ex. with a map of 20 horizontal tiles:
- With the standard resolution of 320x240, the result was 160 for ScreenX
- With the widescreen resolution of 416x240, the result was 112 (160 - the black border width of 48)

Turns out `ControlVariables::Event()` was applying an offset to the result in case of FakeResolution mode, this offset seems unnecessary because in FakeResolution mode, we want to keep the same value than with standard resolution.

[80c4715](https://github.com/EasyRPG/Player/pull/3510/commits/80c471581580e856158ce8a7b844acfbcf4d1bad) fixes this issue by removing this offset applying to the screen X/Y position when using FakeResolution mode.

Interior are now fixed in Aëdemphia, here is a video of the first town of the city with this PR:
https://www.youtube.com/watch?v=BKPsRmbDW0g

BlisterB:
In Aëdemphia, the waterfall map using parallax effect is totally broken (both Panorama and Picture).

In 320x240: OK
[Ae Paralax 320x240.webm](https://github.com/user-attachments/assets/1fb2e86b-2412-4d70-b923-4c74032f22b8)

In Ultra-wide: KO
[Ae Paralaxe Ultra Wide KO.webm](https://github.com/user-attachments/assets/d7d6375f-fdad-496b-94d7-92c9f8cdbd8f)

BlisterB:
@Ghabry  I think we are good for this PR 👍

BlisterB:
[c60093d](https://github.com/EasyRPG/Player/pull/3510/commits/c60093df1eeeba8c6aedd19005f878e8368267e4) Fixes the starting position of the camera : 
- when the hero is on a looping map
- using the FakeResolution mode 
- with a map too small for the screen

It applies an offset corresponding to the black border's size.

Fix one of the two problems from #3511 
I still need to fix the camera not centering to the player while they move.
 
image

PR #3519, opened 2026-02-17
Add IgnoreAbsentBattleAnimationFlash to Player config

ianling:
It is common for developers to use the `Flash Event` command rapidly in order to tint the player's sprite.

If a developer uses the `Show Animation` command at the same time, the animation will override the flash from `Flash Event`, even if the animation does not contain any flashes. This behavior is present in both RPG Maker 2003 and EasyRPG.

In my game for example, I am using Flash Event to darken the player sprite in a dark cave map. Also, when the player equips any item anywhere in the game, I play an animation using Show Animation. If the player equips an item in the cave map, despite the animation not containing any flashing, the Flash Event command appears to stop working for the duration of the animation. This causes the player sprite to appear untinted briefly.

----

This PR adds a new option to the Player config: `IgnoreAbsentBattleAnimationFlash` (default: false -- retains prior behavior)

When `true`, battle animations that do not contain flashes will not override flashes from other sources, including Flash Event. Battle animations that do contain flashes still behave as before.

If this isn't the kind of change you want to pull in, feel free to reject, I just needed this for my game and thought others might run into the same issue, as tinting the player is fairly common in Yume Nikki fan games.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /easyrpg-player
 /Player.app
+/EasyRPG Player.app
 *.a
 *.elf
 *.o
@@ -8,6 +9,7 @@
 *.bak
 
 # cmake
+.cmake/
 /Makefile
 CMakeFiles/
 CMakeScripts/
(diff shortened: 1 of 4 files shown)


Ghabry:
I don't think such stuff should be a config option as this is highly game specific.

That's (maybe) a good candidate for a database chunk or an additional flag for the battle animation event command. Closing this one for now. We can add this later as a per-game option when the editor is advanced enough :).

PR #3520, opened 2026-02-23
Particle effects Version 1 & 2

LizardKing777:
Both function independently of one another, so if a game was released using either version it should work flawlessly.


LizardKing777:
As far as I know, the two versions are basically independent of one another, so probably keeping the code separate for now is probably the pragmatic way to go. I added the license info back and got V1 to sync with the acceleration (they were keeping the same speed and calling at the accelerated rate, which was creating a neat but undesirable effect. Basically moving the particles from "Draw" to "Update" - but it's causing a "Vector Substack Out of Range" overload on the V2 particle effects, so I'm going to have to play around with that one a bit more.

Ghabry:
Is the particles V2 based on my deep8 code from here? https://github.com/DeepIA8/easyrpg-player-deep8/blob/latest2/src/dynrpg_particle.cpp

(Just asking as the simplest way to review it would me doing a diff so I know what I wrote myself)

LizardKing777:
Originally it was but it's pretty much been completely re-written.  Using a compare tool it appears 328 lines were taken out and 258 were added. So I would treat it as its own code.

Ghabry:
Could you link the kazesui dynrpg particle v1 source code? It looks like you "ported" it based on the dynrpg plugin but no idea where v1 can be found (I only have the V2 code)

LizardKing777:
[ParticleEffectsPlugin1.zip](https://github.com/user-attachments/files/27294039/ParticleEffectsPlugin1.zip)

No idea how long this has been sitting in my RPG Maker folder, but here it is!

Ghabry:
I cannot see the plugins source code in this folder. That's a game with a DLL

LizardKing777:
[Source.zip](https://github.com/user-attachments/files/27305370/Source.zip)

Sorry, wrong zip. Try this one.

LizardKing777:
Ugh, I forgot to update the CMakeLists.txt file to include the new .cpp and header files. Hopefully this compiles now.

PR #3522, opened 2026-02-23
SDL Android: Bump to 2.32.10 and preparations for SDL3

Ghabry:
SDL2 is almost EOL so this patch was really hard to apply ;)

btw unrelated: There is something broken in libogc, Wii builder gives errors


Ghabry:
updated to latest gradle wrapper jar.

Android 9 is also the oldest device around for testing here.

I used an emulator to test Android 6, Android 10, 15 and 16.

carstene1ns (approved):
Works on my device.

PR #3524, opened 2026-03-13
Transition: Fix crash for "Random Block" (regression)

Ghabry:
Making all the transitions one frame longer causes the random block transition to index out of bounds because it uses a fixed array size.

As there are still some flaws in the transition code I added a safeguard for now.


carstene1ns (approved):
Feels a bit hacky, but we still need to fix parts here, so good for now ^^

PR #3532, opened 2026-04-07
Windows & emscripten: Fix SDL3 build

Ghabry:
Also the "disable rounding of windows" flag was lost when leaving fullscreen mode.

Fixed this also for SDL2


Ghabry:
Also added some fixes for emscripten now. Though I would stick to SDL2 for now. SDL3 currently overwrites the canvas size with the one from `SDL_CreateWindow` (so you get a 320x240 canvas). I couldn't find a good way to workaround this and fullscreen is broken.

There is a pending PR which appears to improve this behaviour. So will likely be fixed with 3.6 https://github.com/libsdl-org/SDL/pull/14690

---

The pre/post.js fixes are required by a new emscripten version.

carstene1ns (approved):

PR #3537, opened 2026-04-21
Pixel Movement

LizardKing777:
Working with all normal functions  - including Pathfinding


src/game_character.cpp
```diff
@@ -36,1199 +37,1838 @@
...
-
-	this->UpdateAnimation();
+    if (!IsActive() || IsProcessed()) {
+        return;
+    }
+    SetProcessed(true);
+
+
+//	if (true) {
+    if (Player::game_config.allow_pixel_movement.Get()){        // TODO - PIXELMOVE
```
Ghabry: You messed up the whitespace. All your changes replace TAB with Space. No idea how this happened as we have a ".editorconfig" file and most programs respect this now.

lumiscosity: probably qt creator :p

LizardKing777: I mainly use CodeBlocks. Is it just game_character.cpp that has this problem or are there more?

PR #3538, opened 2026-04-22
Pathfinder Command: fix GCC 15 -Wnonnull warning 

staticssleever668:
`event_id_ignore_list.begin()` pointed at an empty vector. I don't have a game to test this with, but it looks like it should not have worked before.

### DIFF
--- a/src/game_interpreter_map.cpp
+++ b/src/game_interpreter_map.cpp
@@ -945,7 +945,7 @@ bool Game_Interpreter_Map::CommandEasyRpgPathfinder(lcf::rpg::EventCommand const
 		int var = ValueOrVariable(com.parameters[11], com.parameters[12]);
 		int num_events_ids = Main_Data::game_variables->Get(var);
 		auto lst = Main_Data::game_variables->GetRange(var + 1, num_events_ids);
-		std::copy(lst.begin(), lst.end(), event_id_ignore_list.begin());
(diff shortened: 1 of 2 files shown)


Ghabry:
uhm thanks for your contribution. I'm curious and will check what happens in the upstream code right now first o_O

Ghabry:
Jenkins: test this please

Ghabry:
Yeah this does indeed crash.

Unfortunately your assignment does not compile. We have some platforms (notable 3DS) where `int32_t` != `int`

https://ci.easyrpg.org/job/player-3ds-pr/592/console

---

(Besides, but thats something for me to fix, I noticed that in general that code doesn't do what is intended. When you e.g. pass in "Variable 5" this should start reading the array from location 5. Instead it dereferences the Variable and then starts the array their.)

staticssleever668:
Replaced assignment with `std::back_inserter` which happily allows narrowing conversions as demonstrated by this snippet:
```cpp
	std::vector<int8_t> into;
	std::vector<int64_t> from;
	std::copy(from.begin(), from.end(), std::back_inserter(into));
```
I don't know whether checking that the value fits makes sense in this context. I assume `int` on 3DS is 16 bits-big if not 32.

Ghabry:
Jenkins: Test this please

Actually it is 32 Bit. Its just on most systems `int32_t == int` but on 3DS it is `int32_t == long int` :sweat_smile:

Ghabry (approved):

PR #3539, opened 2026-04-24
Return an optional instead of casting -1 into an enum

staticssleever668:
For enums without a fixed underlying type, casting into them from numbers outside of their enumeration values is undefined behavior. Fixes running tests with UBSan: `runtime error: load of value 4294967295, which is not a valid value for type 'Trigger'`.


Ghabry:
Jenkins: test this please

Ghabry (approved):

PR #3541, opened 2026-04-30
Fix issues reported by zoeontheloose + Undefined Behaviour in ControlSwitches

Ghabry:
Various small stuff:

- Touched extremely old code Control Switch code (I tested in RPG_RT and the new behaviour seems more accurate)
- "Event exists" for Conditional Branch (very useful in combination with out Clone Event command)
- String ToNum always uses decimal now
- Var support for ShowStringPic sys/font name. This was a bit annoying to add as had to figure out how the numbers are encoded...

Most were reported by zoeontheloose in our community because is shadowbanned on GitHub: https://community.easyrpg.org/t/various-small-bug-reports-regarding-maniacs-features/1691

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -1076,16 +1076,16 @@ bool Game_Interpreter::CommandControlSwitches(lcf::rpg::EventCommand const& com)
 		int val = com.parameters[3];
 
 		if (start == end) {
-			if (val < 2) {
+			if (val == 0 || val == 1) {
 				Main_Data::game_switches->Set(start, val == 0);
-			} else {
+			} else if (val == 2) {
 				Main_Data::game_switches->Flip(start);
 			}
 			Game_Map::SetNeedRefreshForSwitchChange(start);
 		} else {
-			if (val < 2) {
+			if (val == 0 || val == 1) {
 				Main_Data::game_switches->SetRange(start, end, val == 0);
-			} else {
+			} else if (val == 2) {
 				Main_Data::game_switches->FlipRange(start, end);
 			}
 			Game_Map::SetNeedRefresh(true);
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -4680,7 +4680,31 @@ bool Game_Interpreter::CommandManiacShowStringPicture(lcf::rpg::EventCommand con
 	}
 
 	params.system_name = components[2];
+	uint32_t var_id = 0;
+	auto mode = delims[1] - 1;
+	if (mode > 0) {
+		if (!ManiacPatch::DecodeStringToInt(params.system_name, var_id)) {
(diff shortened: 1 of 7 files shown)


src/maniac_patch.cpp
```diff
@@ -834,6 +834,53 @@ std::string_view ManiacPatch::GetLcfDescription(int data_type, int id, bool is_d
...
+
+	For all other characters c with index i the value is:
+	(c - a + 1) << (i * 5)
+
+	Due to integer overflow the max string length is 6.
+	*/
+
+	out = 0;
+
+	if (!(str.size() > 0 && str.size() <= 6)) {
```
carstene1ns: `str.empty()||size>6`

src/maniac_patch.cpp
```diff
@@ -834,6 +834,53 @@ std::string_view ManiacPatch::GetLcfDescription(int data_type, int id, bool is_d
...
+		return value >= 0 && value < 32;
+	};
+
+	int32_t result = (str.back() - 'A' + 1);
+
+	if (!in_range(result)) {
+		return false;
+	}
+
+	out = static_cast<uint32_t>(result);
```
carstene1ns: The casting here is a bit ugly, `result` variable is not really needed, when doing all unsigned. `!in_range` could be simplified to `>=32` as well then.

src/maniac_patch.cpp
```diff
@@ -834,6 +834,53 @@ std::string_view ManiacPatch::GetLcfDescription(int data_type, int id, bool is_d
...
+
+		if (!in_range(result)) {
+			return false;
+		}
+
+		int chidx = str.size() - i;
+		out += static_cast<uint32_t>(result << (chidx * 5));
+	}
+
+	return out;
```
carstene1ns: what if `out` is 0, is this not valid?

src/game_interpreter.cpp
```diff
@@ -1076,16 +1076,16 @@ bool Game_Interpreter::CommandControlSwitches(lcf::rpg::EventCommand const& com)
 		int val = com.parameters[3];
 
 		if (start == end) {
-			if (val < 2) {
+			if (val == 0 || val == 1) {
```
carstene1ns: Maybe any log entry/assert for `val>2`?

carstene1ns (approved):

src/maniac_patch.cpp
Ghabry: oh this should be of course return true :D   (0 is valid and does nothing as 0 is not a valid variable index)

Ghabry:
thx @carstene1ns addressed your comments.

(the macOS failure is unrelated, thats just macOS breaking the compiler everytime an update is installed :facepalm: )

carstene1ns (approved):

PR #3542, opened 2026-05-03
Encounter rate: Fix off-by-one error

### LINKED ISSUE #2903: The Encounter Rate seems to not be accurate (e.g. with an Encounter Rate of 1 step, you can not trigger a battle while walking, while this is not the case in RPG_RT)
#### Name of the game:
EasyRPG Test Game 2003.

#### Player platform:
Windows, latest continuous build of EasyRPG Player.

The Encounter Rate does not seem to be accurate compared to the original RPG_RT (for example with an Encounter Rate of 1 step, you can not trigger a battle while walking, while this is not possible in RPG_RT).

Ghabry:
Honestly not sure if this is 100% correct but it "feels" at least closer to RPG2k3 now when testing.

What matches:
- With EncRate 1 you get a forced encounter (before it was 2)
- With 2 a forced after 2 (before it was 3)

Higher numbers are harder to test due to randomness involved.

Fix #2903

### DIFF
--- a/src/game_player.cpp
+++ b/src/game_player.cpp
@@ -752,7 +752,7 @@ void Game_Player::UpdateEncounterSteps() {
 	const auto ratio = GetTotalEncounterRate() / encounter_steps;
 
 	auto& idx = last_encounter_idx;
-	while (ratio > enc_table[idx+1].ratio) {
+	while (ratio >= enc_table[idx+1].ratio) {
 		++idx;
 	}
 	const auto& row = enc_table[idx];


Mimigris:
Just tested and it indeed seems closer to RPG_RT now 👍 
Once again, it's a bit hard to truly compare due to the RNG involved, but it at seems more in line at the very least.
(only thing I noticed when checking is that negative encounter rates aren't handled in the same way between both engines, but as there is no real purpose in doing so and that the default engine does not allow you in any way to do it, it's not like it is a big deal)

carstene1ns (approved):

PR #3543, opened 2026-05-08
SDL3: Make it the default, further Emscripten & Android fixes

Ghabry:
~~Draft as the Android fixes are missing *cough* - working on it.~~

Though already good enough for testing Emscripten and other platforms if anything breaks.


Ghabry:
guess for the github runners I have to set them to SDL2.

Looking at repology the first Debian/Ubuntu with a new enough SDL3 (3.4) are Debian 14 and Ubuntu 26.04 :sweat_smile:

src/platform/sdl/sdl3_ui.cpp
```diff
@@ -1174,13 +1212,13 @@ int FilterUntilFocus(const SDL_Event* evnt) {
 
 void Sdl3Ui::vGetConfig(Game_ConfigVideo& cfg) const {
 #ifdef __EMSCRIPTEN__
-	cfg.renderer.Lock("SDL2 (Software, Emscripten)");
+	cfg.renderer.Lock("SDL3 (Software, Emscripten)");
 #elif defined(__wii__)
-	cfg.renderer.Lock("SDL2 (Software, Wii)");
+	cfg.renderer.Lock("SDL3 (Software, Wii)");
```
carstene1ns, line 1221: I guess the Wii (U) stuff here is not needed the next 10 years^^

Ghabry:
Figured out the Android crash with AI assistance... I noticed that it isn't really crashing, just terminating. So I dumped the Activity Stack and noticed that after the first game shutdown the Activity Stack is empty (no GameBrowser).

Feeded this information into Gemini and the answer was that I must set `SDL_ANDROID_ALLOW_RECREATE_ACTIVITY=1` to prevent SDL from calling System.exit().

Yeah that is a new SDL3 feature to get a clean environment on every restart. Disabling this behaviour indeed fixes the "crash" and it runs fine.

Ghabry:
Also updated the bounding box calculation of the game window so it's possible now to create Portrait Mode games :D.

(I remember a user telling us this didn't work a while ago)

Screenshot_2026-05-15-15-57-15-588_org easyrpg player

src/platform/sdl/sdl3_ui.cpp
Ghabry: probably :D. If you want I can remove this for now.

Ghabry:
Everything should work now at least not worse than with SDL2 :D

carstene1ns:
```
sdl3_ui.cpp:988:41: warning: comparison of integer expressions of different signedness: ‘const int’ and ‘SDL_FingerID’ {aka ‘long unsigned int’} [-Wsign-compare]
  988 |                         return input.id == evnt.tfinger.fingerID;
      |                                ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
```

Type was signed in SDL2, now unsigned in SDL3. In general it seems fragile to store a 64 bit integer in an int in our struct, since documentation says:

> This ID is valid for the time the finger (stylus, etc) is touching and will be unique for all fingers currently in contact, so this ID tracks the lifetime of a single continuous touch. **This value may represent an index, a pointer, or some other unique ID, depending on the platform.**
>
>The value 0 is an invalid ID.

I doubt this will bite us soon, but this could create some nice bugs, when there is a platform that uses some weird ID or a user touched 2147483647 times... :D

But there is no easy solution, for example use also 64 bits or hash the id.

Ghabry:
hmpf maybe we should just make `-Wall -Wextra` the default so I don't miss the additional warnings when I forget to turn them on :/

Valid concern though. I will fix this :)

carstene1ns (approved):

Ghabry:
This broke our Flatpak because they only ship SDL3 3.2 ...

Guess easiest solution is forcing SDL2 for now and hope they bump SDL3 sooner or later. >.>

PR #3544, opened 2026-05-12
Interpreter: Implement Maniac SetGameOption operation 1 (variable count)

luadebug:
maniac: Fix PicPointer variable lookup in CommandShowPicture

The IsValid guard was unnecessary since Get() already returns 0 safely
for out-of-bounds variable IDs. Removing it also fixes the case where
the variable array hasn't grown yet to the requested index.

SetGameOption operation 1 is now a no-op since the Player auto-expands
the variable array on Set(), making pre-allocation unnecessary.

Remove the redundant IsValid guard before Get() in CommandShowPicture's PicPointer handling — Get() already returns 0 safely for any variable ID. Implement SetGameOption operation 1 as a no-op since the variable array auto-expands on write.

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -2818,10 +2818,7 @@ bool Game_Interpreter::CommandShowPicture(lcf::rpg::EventCommand const& com) { /
 			pic_id = ValueOrVariable(com.parameters[17], pic_id);
 		}
 		if (com.parameters[19] != 0) {
-			int var = 0;
-			if (Main_Data::game_variables->IsValid(com.parameters[19])) {
-				var = Main_Data::game_variables->Get(com.parameters[19]);
-			}
+			int var = Main_Data::game_variables->Get(com.parameters[19]);
 			params.name = PicPointerPatch::ReplaceName(params.name, var, com.parameters[18]);
 		}
 
@@ -5011,6 +5008,7 @@ bool Game_Interpreter::CommandManiacSetGameOption(lcf::rpg::EventCommand const& 
 	//int value = ValueOrVariable(com.parameters[0], com.parameters[2]);
 
 	switch (operation) {
+		case 1: // Set variable count (noop, Player auto-expands the variable array on access)
 		case 2: // Change Picture Limit (noop, we support arbitrary amount of pictures)
 			break;
 		default:


Ghabry:
Please don't use copilot or whatever AI that is to write these PR descriptions. The tools are extremely verbose and a human has to read this...

Thanks for implementing however I wonder why this call to `SetLowerLimits` is necessary. I don't think there is a way to query this value and our Player just expands the variable array so it should just work.

The only non-debug usage of `IsValid` I could find is

```
        if (com.parameters[19] != 0) {
            int var = 0;
            if (Main_Data::game_variables->IsValid(com.parameters[19])) {
                var = Main_Data::game_variables->Get(com.parameters[19]);
            }
            params.name = PicPointerPatch::ReplaceName(params.name, var, com.parameters[18]);
        }
```

in `CommandShowPicture`  so maybe the real bug is this `IsValid` call that shouldn't be necessary their?

Is that the cause of the bug?

luadebug:
> Please don't use copilot or whatever AI that is to write these PR descriptions. The tools are extremely verbose and a human has to read this...
> 
> Thanks for implementing however I wonder why this call to `SetLowerLimits` is necessary. I don't think there is a way to query this value and our Player just expands the variable array so it should just work.
> 
> The only non-debug usage of `IsValid` I could find is
> 
> ```
>         if (com.parameters[19] != 0) {
>             int var = 0;
>             if (Main_Data::game_variables->IsValid(com.parameters[19])) {
>                 var = Main_Data::game_variables->Get(com.parameters[19]);
>             }
>             params.name = PicPointerPatch::ReplaceName(params.name, var, com.parameters[18]);
>         }
> ```
> 
> in `CommandShowPicture` so maybe the real bug is this `IsValid` call that shouldn't be necessary their?
> 
> Is that the cause of the bug?

I think that is the cause. You can validate against https://freegame-mugen.jp/roleplaying/game_9561.html
Before change game stuck on main menu next to start game. After change its possible to select new game and play this game.

Although I am a bit confused how to escape this menu...? I can't save the game after boss fight so :/
image
[Save01.lsd.tar.gz](https://github.com/user-attachments/files/27647446/Save01.lsd.tar.gz)

jetrotal:
Are those covered in https://github.com/EasyRPG/Player/pull/3487 ?

Ghabry:
We asked the Maniac devs and they told us that this _Set variable count_ feature does not exist. So you made this up. I'll close this PR for that reason.

I added the suggested `IsValid` change to #3541 , so this one isn't lost.

luadebug:
I really lack of idea what to do and where to search for solution. Maybe a culprit is that during change of team in Baton menu should affect SPD but its always 0 no matter how I change Baton formation of a team. Besides I already tried building Player with changes 3487 + 3541.

image

I noticed it keeps spamming
Debug: PicPointer: Top transparency 35187 replaced with 0
Debug: PicPointer: Top transparency 35189 replaced with 0
Debug: PicPointer: Top transparency 35192 replaced with 0
Debug: PicPointer: Top transparency 35194 replaced with 0
Debug: PicPointer: Top transparency 35197 replaced with 0
Debug: PicPointer: Top transparency 35199 replaced with 0
Debug: PicPointer: Top transparency 35202 replaced with 0
Debug: PicPointer: Top transparency 35204 replaced with 0
Debug: PicPointer: Top transparency 35207 replaced with 0
Debug: PicPointer: Top transparency 35209 replaced with 0
But that is not much idea of what is going on there. As I am still unable to leave Baton formation menu.
I tried to gather logs
```
Debug: Invalid write var[-13] = 0!
Debug: Invalid write var[0] = 0!
Debug: Invalid write var[-12] = 0!
Debug: Invalid write var[-13] = -20!
Debug: Invalid write var[0] = 0!
Debug: Invalid write var[-12] = 0!
Debug: Invalid write var[-13] = -40!
Debug: lcf (Debug): Skipped Chunk 00 (0 byte) in lcf at 14 (CommandManiacControlGlobalSave)
Debug: Loaded Map Map0438.lmu
Debug: Tree: Title < Common / System
Debug: lcf (Debug): Skipped Chunk 00 (0 byte) in lcf at 14 (CommandManiacControlGlobalSave)
Debug: Invalid read sw[11]!
Debug: Invalid read sw[12]!
Debug: Invalid read sw[13]!
Debug: Invalid read sw[21]!
Debug: Loading Save Save01.lsd
Debug: Savegame version 810 (EasyRPG Player 0.8.1)
Debug: Loaded Map Map0048.lmu
Debug: Tree: Initial Map < Demon Cave < Crane < Prologue
Debug: Loaded Map Map0007.lmu
Debug: Tree: Map 2 < Demon Cave < Crane < Prologue
Debug: Loaded Map Map0445.lmu
Debug: Tree: Map 3 < Demon Cave < Crane < Prologue
Debug: Loaded Map Map0444.lmu
Debug: Tree: Tia < Map 2 < Demon Cave < Crane < Prologue
Debug: Starting battle 301 (ティーア,飛魔x2): algos=(RPG_RT/RPG_RT)
Debug: Loaded Map Map0446.lmu
Debug: Tree: After Tia < Map 2 < Demon Cave < Crane < Prologue
Debug: Loaded Map Map0009.lmu
Debug: Tree: Castle Normal Structure < Lord's Castle < Crane < Prologue
```

PR #3545, opened 2026-05-16
Implement Maniacs Battle Common Events (Second Try)

Ghabry:
See #3307 with the difference that it hooks directly into our `Game_CommonEvent` class to make the parallel events execute as intended.

Imo this is closer than #3307 but still not perfect, especially when the Parallel events show messages. When no messages are shown it appears to be... okay (?)

---

**I need testers for this!** I don't really know of alot of games that use this Battle stuff and are not 100 million lines of event voodoo. So this is hard to test.

**Your task**: Throw this Player PR at games that use this feature and _even better_ figure out where it fails in the event code.

When I don't get enough useful reports this PR will never be merged as I won't integrate untested code into the Player. Thank you for your attention.

---

Also noticed a bug in the Get ATB function: It must return the percentage

### DIFF
--- a/src/game_battle.cpp
+++ b/src/game_battle.cpp
@@ -20,6 +20,7 @@
 #include <lcf/data.h>
 #include "player.h"
 #include "game_actors.h"
+#include "game_commonevent.h"
 #include "game_enemyparty.h"
 #include "game_message.h"
 #include "game_party.h"
@@ -43,6 +44,8 @@ namespace Game_Battle {
 	std::string background_name;
 
 	std::unique_ptr<Game_Interpreter_Battle> interpreter;
+	std::vector<Game_CommonEvent> common_events;
+
 	/** Contains battle related sprites */
 	std::unique_ptr<Spriteset_Battle> spriteset;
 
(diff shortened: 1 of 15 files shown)


jetrotal:
I wonder if gobble fantasy uses it:

https://dogs-understand.itch.io/gobble-fantasy-2003

image


From what I read, it uses maniacs to force atb to work as a turn based battle

Ghabry:
Nope:

```
> grep trigger RPG_RT.edb | sort | uniq -c
      4     <trigger>3</trigger>
     53     <trigger>4</trigger>
    217     <trigger>5</trigger>
```

Start is 6, Battle Parallel is 7

korean-guy:
When I simply tested it, it seems to behave the same as #3307 .
I haven't checked it in #3307 , but I tested the ATB gauge in this PR.
If member[1].gauge is output as a variable value, it is output as a percentage,
When the ATB value is obtained from the Control Battle command, it is output as a numerical value, not as a percentage.
I don't know if this was meant to be.

PR #3546, opened 2026-05-17
Maniac Patch: self-variables, scoped modes 5/6, BreakLoop fix, negative IDs in messages

luadebug:
## Summary

Four fixes for Maniac Patch behaviour, discovered while running *SOLDIERS -DesireWing-* (RPG Maker 2003 + Maniac Patch). The game uses all four features heavily in its Baton formation menu.

### 1. Self-variables (negative variable IDs -1..-64)

Maniac Patch defines variable IDs -1 through -64 as "self-variables" — per-context scratch storage separate from the global variable table. Adds a `_self_vars[64]` array to `Game_Variables` with `Get()` and `SetOp()` support for negative IDs, and suppresses spurious out-of-range warnings for valid negative IDs in `ShouldWarn()`.

### 2. Scoped variable modes 5 and 6 in `DecodeTargetEvaluationMode`

Mode 5 (`ScopedSingle`) and mode 6 (`ScopedRange`) address self-variables by a 1-based positive index in `com.parameters[1]` that maps to a negative ID: `params[1] = 1` → variable ID `-1`, etc. These modes were silently falling through to the `default:` branch (id_0 = 0).

### 3. BreakLoop — find actual enclosing loop by scanning backward

The previous Maniac Patch path used `SkipToNextConditional({EndLoop}, indent - 1)`, which breaks when `BreakLoop` is nested inside conditionals within a loop. Example: `BreakLoop` at indent 3 (inside two levels of `if` inside an infinite loop at indent 0) was using `indent - 1 = 2` to search for `EndLoop`, finding the wrong target.

Fix: scan backward from `BreakLoop` to find the innermost enclosing `Loop` command (tracking `EndLoop`/`Loop` pairs to skip already-closed nested loops), then scan forward to its `EndLoop`. The vanilla RPG_RT bug-emulation path is unchanged.

### 4. `ParseParam`: support `\v[-N]` negative variable IDs in message escape sequences

`ParseParam()` treated a leading `-` inside brackets as a non-digit, setting `stop_parsing = true` and returning 0. Adds an `is_negative` flag that consumes the `-` before the digit loop and negates the parsed value at the end. Affects all escape sequences that call `ParseParam`: `\v[N]`, `\t[N]`, `\c[N]`, `\s[N]`, `\n[N]`.

## References

- Maniac Patch official documentation: https://bingshan1024.github.io/steam2003_maniacs/
- EasyRPG Maniac Patch metabug (feature tracking): https://github.com/EasyRPG/Player/issues/1818
- Prior Maniac loop/break-loop work: https://github.com/EasyRPG/Player/pull/2649

## Test case

Game: *SOLDIERS -DesireWing-* (RPG Maker 2003 + Maniac Patch)
- Baton formation menu (CE 219) uses `\v[-1]` in `ShowStringPicture` to display character SPD from self-variables → previously always showed 0
- Cancel key exit from the menu uses `BreakLoop` at indent 2 inside a `ConditionalBranch` at indent 1 inside an infinite loop at indent 0 → previously could not exit the menu

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -3882,9 +3882,43 @@ bool Game_Interpreter::CommandBreakLoop(lcf::rpg::EventCommand const& /* com */)
 
 	// BreakLoop will jump to the end of the event if there is no loop.
 
-	bool has_bug = !Player::IsPatchManiac();
-	if (!has_bug) {
-		SkipToNextConditional({ Cmd::EndLoop }, list[index].indent - 1);
+	if (Player::IsPatchManiac()) {
+		// Maniac Patch: find the actual innermost enclosing loop by scanning backward.
+		// This correctly handles BreakLoop inside conditionals nested within a loop —
+		// e.g. BreakLoop at indent=2 inside an if (indent=1) inside an infinite loop
+		// (indent=0) must break the outer loop, not some unrelated inner loop that
+		// happens to appear later in the same loop body.
+		int break_indent = list[index].indent;
+		int target_indent = -1;
+
+		// endloop_depth[M] counts how many EndLoops at indent=M we have seen going
+		// backward; a matching Loop cancels one out rather than being the enclosing loop.
+		constexpr int kMaxIndent = 20;
+		int endloop_depth[kMaxIndent] = {};
+
+		for (int i = index - 1; i >= 0; --i) {
+			const auto& c = list[i];
+			if (c.indent >= break_indent) continue;
+			auto cc = static_cast<Cmd>(c.code);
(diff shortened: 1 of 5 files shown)


Ghabry:
I will directly tell you that we will not accept a contribution of such a complex feature from an unknown "developer". Especially not one who sends us vibecoded stuff.

Your AI already made something up and wasted our time with the much simpler Set Game Option contribution.

Just looking for 1 minute at it this patch already makes no sense: The scoped variables are per event and not some "negative index global scratch buffer".

---

I give you a warning because of spamming. Next AI slop contribution is a ban.

PR #3556, opened 2026-06-15
Emscripten: Rework touch input and shell layout

johannschopplich:
This PR brings the web-shell improvements from my production setup on [realtroll.de](https://realtroll.de/) – touch and layout work that's been running there a while.

Per review, reduced to the shell rework itself: the integer scaling toggle is gone (covered natively by the settings scene), and the save manager moved to a follow-up PR based on this one.

**What's in:**

- **Touch controls** – rewritten on Pointer Events: a responsive grid (d-pad / screen / action buttons) with safe-area insets, both orientations, replacing the floating corner buttons. Controls dim when idle and light back up on input.
- **Wake Lock** – re-acquired on `visibilitychange` so the screen doesn't sleep mid-session.
- **Fullscreen button** – toggles enter/exit and reflects its state; hidden where unsupported (iOS Safari).
- **Boot errors** – show a message instead of a silently hanging loader (e.g. stale cached `.wasm`).
- UI icons consolidated into one inline SVG sprite.
- Dropped the legacy `scanGamePads` poll (`ongamepadconnected` predates WASM in every browser).

### DIFF
--- a/resources/emscripten/emscripten-shell.html
+++ b/resources/emscripten/emscripten-shell.html
@@ -1,365 +1,560 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="utf-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1.0">
-  <title>EasyRPG Player</title>
-  <style>
-    :root {
-      --color-gray: hsl(0, 0%, 55%);
-      --controls-size: 10vh;
-    }
-
-    @media (orientation: landscape) {
+  <head>
+    <meta charset="utf-8" />
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1.0, viewport-fit=cover"
+    />
+    <meta name="color-scheme" content="dark" />
+    <meta name="theme-color" content="#000000" />
+    <title>EasyRPG Player</title>
+    <style>
       :root {
-        --controls-size: 20vh;
+        --color-gray: hsl(0 0% 55%);
+        --controls-size: clamp(3.5rem, 20vmin, 5.5rem);
+        --controls-opacity: 0.4;
+        --controls-fade: 80ms;
       }
-    }
 
-    html {
-      touch-action: none;
-    }
+      html {
+        touch-action: none;
+      }
 
-    body {
-      font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, sans-serif, "Apple Color Emoji", "Segoe UI Emoji";
-      margin: 0;
-      color: white;
-      background: black;
-    }
+      body {
+        margin: 0;
+        font-family: system-ui, sans-serif;
+        color: white;
+        background: black;
+        -webkit-touch-callout: none;
+        -webkit-user-select: none;
+        user-select: none;
+      }
 
-    .unselectable {
-      -webkit-touch-callout: none;
-      -webkit-user-select: none;
-      user-select: none;
-    }
+      #boot {
+        position: fixed;
+        inset: 0;
+        z-index: 1;
+        display: grid;
+        place-content: center;
+        justify-items: center;
+        font-family: ui-monospace, monospace;
+        text-align: center;
+        background: black;
+        pointer-events: none;
+        transition:
+          opacity 300ms ease,
+          visibility 300ms ease;
+      }
 
-    #status {
-      font-size: 1.5rem;
-      color: var(--color-gray);
-      text-align: center;
-    }
+      #boot.done {
+        visibility: hidden;
+        opacity: 0;
+      }
 
-    #controls {
-      position: relative;
-      text-align: right;
-      z-index: 10;
-    }
+      #status {
+        font-size: 0.875rem;
+        color: var(--color-gray);
+      }
 
-    #controls button {
-      -webkit-appearance: button;
-      appearance: button;
-      display: inline-flex;
-      background: transparent;
-      border: 0;
-      color: white;
-      font-family: inherit;
-      font-size: 1em;
-      line-height: inherit;
-      cursor: pointer;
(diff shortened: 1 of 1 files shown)


Ghabry:
Jenkins: Test this please

Thanks! Have to take a look at this :)

johannschopplich:
Pushed a follow-up batch from the same setup: a save manager! Fits nicely into the design.

Updated the description above accordingly. 🙂

jetrotal:
Hello, since you are working on the touch screen buttons for the weblpayer, maybe consider taking a look at this proposal from a user called Bilou Master Joke: https://codepen.io/editor/BilouMaster/pen/019edf09-20e1-73a0-8efa-3ccb9a832518

He made a bunch of suggestions to both support the complete vanilla keyboard, and some quality of life usage of buttons.

johannschopplich:
@jetrotal Thanks for the input! Honestly, it feels a bit rough, but it's a draft anyway. Worth digging into for a future iteration after this PR. (It already touches a lot of parts of the shell.)

BilouMaster:
Ah yes, that's perfect because I've been doing the design only like an empty shell and not finished/merged with the real player yet, so I'll also take inspiration of your work then! :D

You can check the codepen Jetrotal shared (most recent take) or my old previous player which had better layout too but not commutable/customisable yet
 https://biloumaster.fr/jeux/bogossstory

I like your buttons for importing save files or toggle between perfect pixel/not perfect, that's a good point but they are small and too close to each other, it would likely need a bit scale up + more margin but 👍

johannschopplich:
@BilouMaster Love the input. 🙏 Yeah sure! Maybe we can collaborate on the implementation. My take: Giving users too much customization options clutters interfaces. I prefer a unified polished interface in general. But that's personal preference, eh.

On the buttons – good catch. The icon controls had too little spacing between them (due to my buttons control unification in this PR). I've improved every tap target and given the control row more room.

BilouMaster:
The customisation costs nothing and I'm planning to set the perfect ratio by default so the customisation will just be a fallback for accessibility and non anticipated cases like non conventional devices, I feel it's safer to add some free settings beside the way we think it's good.

The swap Enter/Echap have been demanded many times for example 

Sure we can collaborate on this, I will clone your branch (or the merged in official if the pr is accepted) when I'll have time to re-work on this, I've had many other things to do lastly :)

Let's build the perfect controller together, this big beauty of a project needs it :D

Ghabry:
First thanks for the contribution. Some remarks:

Emscripten: Add pixel-perfect integer scaling toggle: this is already supported native in the settings scene. Maybe make the button open the settings scene instead?

Emscripten: Fix canvas scaling on load and orientation change: Have to check if SDL3 still needs this


Emscripten: Add save manager, fullscreen toggle, and SVG icon sprite: honestly for this one and the follow up commits I would prefer a 2nd PR (based on this one). This is quite complex and I don't really like that hooking of the private API instead of impending this properly in pre/post.js. The scaling factor of the screenshot is also available as a setting in the settings scene btw.

johannschopplich:
Thanks for the review! I agree, the PR mixed concerns. Refactored:

- **Integer scaling toggle**: gone. You're right, the settings scene covers this.
- **Canvas scaling on load/orientation change**: the `!important` sizing workaround is already on master; this PR merely re-expresses it for the new layout. Happy to drop it everywhere if SDL3 no longer needs it.
- **Save manager**: split into a follow-up PR.

Ghabry:
Jenkins: test this please

Thanks this looks more reasonable for the PR now. Will test it a bit tomorrow - also regarding SDL3 behaviour - and then this should be good to go.

johannschopplich:
Thanks a lot! Looking forward to your tests. 🙂 

Follow-up created (based on this one): https://github.com/EasyRPG/Player/pull/3643

Ghabry:
Sorry, this was much longer than _tomorrow_.

There is one weird bug on my Phone with Firefox: When I toggle fullscreen and then leave fullscreen it adds a black border left and right of the frame.

When I then click around randomly the border becomes larger once.

This is repeated everytime I do fullscreen->window.

https://github.com/user-attachments/assets/d3d5fc2e-6218-4a02-ba40-6ff4e03bd47c

https://easyrpg.org/play/pr3556/

resources/emscripten/emscripten-shell.html
```diff
@@ -1,365 +1,581 @@
...
+      let player;
+
+      // Launch the Player; it reads the game from the query string itself
+      window.addEventListener("load", async () => {
+        try {
+          player = await createEasyRpgPlayer({
+            game: undefined,
+            saveFs: undefined,
+          });
+          player.initApi();
```
Ghabry, line 377: You renamed `easyrpgPlayer` to `player` and its not a global variable anymore. I cannot access that variable through the webbrowser now.

In the master code its possible to use `easyrpgPlayer` (e.g. `easyrpgPlayer.Module.FS`) so the website can interact with the instance.

johannschopplich: Oops, good catch, the module script scoped the instance away. It's now assigned to `window.easyrpgPlayer` again in https://github.com/EasyRPG/Player/pull/3556/commits/8822897847fd189baabb423273e05bc1f9467f19.

johannschopplich:
@Ghabry No worries. Thanks for testing! 

The bug yo've encounted probably comes from SDL3 re-reading the canvas size only on window `resize`. https://github.com/EasyRPG/Player/pull/3556/commits/64f4124bcf4057af8c3537b8817fe339f3c86b39 observes the canvas and dispatches a `resize` once its box changes. 

Could you retry on your phone/emulator? I can't reproduce the exact Firefox mobile timing locally.

PR #3581, opened 2026-06-20
Update doc

Mimigris:
Pictures being allowed to move while a message is displayed is also a feature of the English version of RPG Maker 2000 (already supported by the Player), so the flag description was incomplete.


Mimigris:
Updated the pull request to also update the documentation to the best of my knowledge; I'm not so sure about the exact line formatting nor the command line argument for MonSca and EXPlus so this will need proofreading.

PR #3587, opened 2026-06-30
Larger Sprites collide properly

LizardKing777:
When using larger sprites with "$" in the title, the game still uses the single 16x16 block for the event's collision, so you would have to line up extra events to create realistic collision. This fixes that by reading the size of the sprite drawn on the screen and adding additional collision boxes accordingly. So a sprite which is three tiles wide now occupies three tiles of game space.


Mimigris:
This may require discussion, as some games may rely on this feature and this change could make them incompatible depending on how the game is designed (I know two games using this feature, DEEP 8, which uses a custom EasyRPG version so already not fully compatible, and Collective Unconscious; no clue if it changes things for them).

I was wondering, does this hitbox modification changes anything if the $ CharSet is applied to a player or vehicle, or does it only affect regular events?

LizardKing777:
I haven't tested it with vehicles yet, but when the player is a larger sprite they adopt the larger collision. To ensure compatibility with existing games, I'll make it an optional flag to enable this in the next update (similar to the allow_pixel_movement flag) so it doesn't break anything. Thanks for the feedback!

lumiscosity:
i wonder if there could be a filename-triggered opt-in for this instead? it'd be nice to mix the classic behavior with this.

jetrotal:
I would preffer a new command to do this:
Resize Event hitbox:
```js
Param0 = evt_id_type //  int or var or nested_var
Param1 = evt_id_value //  int

Param2 = topLeft_corner_type
Param3 = topLeft_corner_value

Param4 = topRight_corner_type
Param5 = topRight_corner_value

Param4 = bottomRight_corner_type
Param5 = bottomRight_corner_value

Param4 = bottomLeft_corner_type
Param5 = bottomLeft_corner_value


```

This would allow a dev to customize the hit area rectangle shape of an event, allowing for precise barriers or large teleport areas with single event.

Mimigris:
Personally, I think a resize hitbox command is a bit unrelated to the $ CharSet feature on its own (no direct dependency between the two, can be or not be used with $ CharSets), and would lack the simplicity of the $ CharSet command from the way it is described, so if a command like that were to be implemented, I'm not sure if it would have to be done here.

Ghabry:
The idea is good and should be generalized for sure. Multi tile events is something useful for the editor.

PR #3602, opened 2026-07-06
Event Animation Frame Cycle Control

LizardKing777:
This adds three comment commands: @easyrpg_get_event_frame E, V (stores current frame in a variable); @easyrpg_set_event_frame E, # (forces an event to a specific animation frame); and @easyrpg_event_anim_reset E (resets an event to the vanilla Frame Cycle) — E = Event, V = Variable, # = Numerical parameter (or variable pointer.) 

This gives the user more control over how an Event Sprite appears at runtime - in the case of the player the user can lock the sprite in the 0 or 2 frame (left or right step) if using a special pose (say nodding your head, blinking, surprise, etc). This was already possible with an Event using extra pages and switches or variables - but now you can do the same without all that.

With this, it is also easy to shuffle between charset sprites and have more than 3 frame walking animations by manually advancing the animation and  changing the charset once the sprite has reached frame "3" (which is reset to the middle/rest frame.) There may be other uses for this.


LizardKing777:
There's probably a whitespace issue with this one too, so I'm planning to do a re-do of the code and see if I can get rid of that.

PR #3607, opened 2026-07-10
Look for map files in a "Map" subfolder, with root fallback

THandke:
## Summary
- Player currently only looks for `MapNNNN.lmu`/`.emu` files directly in the game's root folder. This adds support for storing them in a `Map` subfolder instead, for a tidier project layout.
- Lookup prefers `Map/` but **falls back to the root directory** when a map isn't found there, so the large existing library of RPG Maker 2000/2003 games (which universally keep maps in the root) keeps working unmodified.
- Touches map loading (`Game_Map::LoadMapFile`), async map prefetching (`Game_Map::RequestMap`, relevant for Emscripten), the start-map existence check that triggers non-standard-extension guessing (`DefaultLmuStartFileExists`), and the non-standard LMU extension guesser (`FileExtGuesser::GuessAndAddLmuExtension`).

## Why no RPG_RT.lmt (TreeMap) changes
`RPG_RT.lmt` only stores map IDs, names, hierarchy and per-map metadata — never file paths. The ID-to-filename mapping (`MapNNNN.lmu`) is a pure Player-side convention, so `lcf::LMT_Reader` parsing needed no changes; only the file-lookup logic that turns a map ID into an on-disk path was updated.

## Reviewer notes
- New shared helper `Game_Map::FindMapFile()` centralizes the "try `Map/`, then fall back to root" logic so `LoadMapFile`, `RequestMap`, and `DefaultLmuStartFileExists` stay consistent.
- Probing a nonexistent `Map/` directory is cheap and silent: `DirectoryTree::FindFile` caches missing-directory lookups and defaults `file_not_found_warning` to `false` (existing pattern used elsewhere in this codebase for the same reason).
- I was not able to compile or run the test suite in the environment this was authored in (no C++ toolchain available). Changes were verified by careful manual/static review of signatures, includes, and call sites, but a local build + `ctest` before merging is strongly recommended.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/async_handler.cpp
+++ b/src/async_handler.cpp
@@ -243,6 +243,12 @@ FileRequestAsync* AsyncHandler::RequestFile(std::string_view file_name) {
 	return RequestFile(".", file_name);
 }
 
+FileRequestAsync* AsyncHandler::RequestFile(std::string_view preferred_dir, std::string_view fallback_dir, std::string_view file_name) {
+	auto* request = RequestFile(preferred_dir, file_name);
+	request->SetFallbackDirectory(fallback_dir);
+	return request;
+}
+
 bool AsyncHandler::IsFilePending(bool important, bool graphic) {
 	for (auto& ap: async_requests) {
 		FileRequestAsync& request = ap.second;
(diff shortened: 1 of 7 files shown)


carstene1ns:
jenkins: ok to test

Ghabry:
Hi.

You didn't set an author for your git commit. It is just "Your Name".

For consistency with other folders it should be called "Map" instead of "Maps".

The logic in RequestMap for the web player will not work this way: When the map was not requested it is not available in the file system yet so the condition will be always false. Instead this needs additional logic in the AsyncHandler class to fetch the correct file depending on its existance.

THandke:
@Ghabry Thanks, updated the author name and changed the folder name to Map. The new logic in AsyncHandler should now handle maps correctly when requested by the web player.

PR #3610, opened 2026-07-12
update translations

Ghabry:
After updating weblate to the latest version something is broken with the repo sync.

Let's merge this PR and then force reset the downstream repo...


carstene1ns (approved):
Well, i can't read this anyway. :)

PR #3614, opened 2026-07-14
New Command 2059 - Game Loader Command

jetrotal:
This command allows loading another game hosted in the same folder of current game. Useful for having a game split into chapters, or gamejams.

Syntax:

```js
@raw 2059, "[path and flags]",
0,1, // [0] Exit Behavior Type // [1] Exit Behavior Value (0: Exit, 1: Parent Title, 2: Parent Map/Savestate)
2,3 // [2] Target String Type // [3] Target String Value (Ignored for 0, otherwise ID of String Var)

```

example:
```js
@raw 2059, "dons_adventure", 0, 2 // loads dons_adventure, and return back to parent_game's previous map when closing it.

@raw 2059, """dons adventure"" --language EN", 1, 100 // when having space or using flag arguments "" "" is needed. also when leaving child game, what's happen is decided by variable 100

```

A test project that uses it:
[demo_loader.zip](https://github.com/user-attachments/files/30019882/demo_loader.zip)

### DIFF
--- a/src/game_interpreter.cpp
+++ b/src/game_interpreter.cpp
@@ -825,6 +825,8 @@ bool Game_Interpreter::ExecuteCommand(lcf::rpg::EventCommand const& com) {
 			return CmdSetup<&Game_Interpreter::CommandEasyRpgCloneMapEvent, 10>(com);
 		case Cmd::EasyRpg_DestroyMapEvent:
 			return CmdSetup<&Game_Interpreter::CommandEasyRpgDestroyMapEvent, 2>(com);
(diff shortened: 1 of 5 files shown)


Ghabry:
Honestly weird feature and the web player will hate it. 😅

Anyway. What the code nicely shows is our awful amount of global state.

There are similar hacky approaches for the game browser and the map savegame load

jetrotal:
web player could simply have a reload the entire page with different params
thought that would made the experience worse due to the reload routines.

PR #3631, opened 2026-07-19
Larger events V3

LizardKing777:
Resolved whitespace issues, made the new code a branch so the original code is maintained and made sure to keep CheckOrMakeWayEx in it's original state and put the new code as a separate branch so nothing should be at risk of breaking.


src/game_map.cpp
```diff
@@ -816,6 +816,8 @@ bool Game_Map::CheckWay(const Game_Character& self,
 		ignore_some_events_by_id, false
 	);
 }
+
+/*    OLD VERSION PRE-LARGER EVENTS
```
Ghabry, line 819: Just overwrite the old version with the new version. Do not comment it out.

Ghabry (changes requested):
The commit list is a mess. Imo the only way to save it is to squash your 8 commits into a single commit.

Also you reverted code by us. Make sure you are always using the latest upstream code: 

```
git fetch upstream
git rebase upstream/master
```

src/game_map.cpp
LizardKing777: Would it make more sense to just create a new branch and pull request that one?

PR #3632, opened 2026-07-20
Larger Events V4

LizardKing777:
modified:   src/game_character.cpp
	modified:   src/game_character.h
	modified:   src/game_event.cpp
	modified:   src/game_map.cpp
	modified:   src/game_player.cpp

Fresh build off the latest upstream source. Adds checks Event triggering and Vehicle boarding/unboarding code to ensure new code is only executing when {_X} and/or {!Y} strings are detected.


src/game_player.cpp
```diff
@@ -565,52 +714,106 @@ bool Game_Player::GetOffVehicle() {
...
-	}
+	if (!vehicle) return false;
 
 	if (InAirship()) {
-		if (vehicle->IsAscendingOrDescending()) {
-			return false;
-		}
-
-		// Note: RPG_RT ignores the lock_facing flag here!
+		if (vehicle->IsAscendingOrDescending()) return false;
```
Ghabry: This is slowly getting at a reviewable stage.

Your commit contains various unrelated changes. Carefully look through your commit and fix them. 

Imo this here is a good example: it changes the formatting and removes a comment.

PR #3660, opened 2026-08-11
libretro: increase ANDROID_API to 21 for android builds

cscd98:
https://github.com/libretro/easyrpg-libretro/issues/63

We now use NDK 29 on the libretro buildbot which does not support API level 16.

Hopefully fixes this:

```
Preparing ARMeabi-v7a toolchain **** Building zlib **** CMake Error at /opt/android-sdk-linux/cmake/3.30.3/share/cmake-3.30/Modules/Platform/Android-Determine.cmake:503 (message): Android: The API level 16 is not supported by the NDK. Choose one in the range of [21, 35].
```

There might be further issues after this.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -850,7 +850,7 @@
 			"name": "android-armeabi-v7a-parent",
 			"cacheVariables": {
 				"CMAKE_ANDROID_ARCH_ABI": "armeabi-v7a",
-				"CMAKE_ANDROID_API": "16"
+				"CMAKE_ANDROID_API": "21"
 			},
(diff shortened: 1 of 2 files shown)


carstene1ns:
Jenkins: test this please.


We use API 21 for our buildscripts/android port as well[^1], so should likely work without further problems.

[^1]: https://github.com/EasyRPG/buildscripts/blob/84cf2f449a0d1c09f17d2bbf3d28f4f573aa718a/android/2_build_toolchain.sh#L48

Ghabry (approved):

Issue #3427, opened 2025-05-31
Player only loading smallest size of bitmap font


maxymax13:
Game: [OFF (v1.6.1)](https://www.rpg-maker.fr/jeux-425-off.html)

Steps to reproduce:
1. Move `OFF\Profont.fon` to `OFF\Font\Font.fon`
2. Intended font loads but everything is small

RPG_RT:
![Image](https://github.com/user-attachments/assets/597fa889-efd6-4ff1-913b-eb25ac04bc2f)
![Image](https://github.com/user-attachments/assets/7ecd7a0d-5f2f-47fb-a69d-94081249b158)
![Image](https://github.com/user-attachments/assets/31bae33b-c328-4656-a8c6-75248c78875f)

EasyRPG:
![Image](https://github.com/user-attachments/assets/f2a7f72d-2a37-4591-b30e-887ef50fc715)
![Image](https://github.com/user-attachments/assets/0a7c2222-782d-4119-a7a7-f5d7f8f7e0a7)
![Image](https://github.com/user-attachments/assets/63fa18d8-246e-47b7-8743-352231dd4e09)

Manually changing the font through the F1 menu instead, all sizes show the smallest size and only change the line height
![Image](https://github.com/user-attachments/assets/d07fdc09-1653-4ddb-a49d-a55cf4e3484f)

Tested on latest continuous build (Windows x86 and x86_64)

PR #3676, opened 2026-08-31
FTFont: Pick closest matching for Bitmap Fonts and disable hinting (Issue 3427)

Ghabry:
Pick closest fixes the mentioned issue.

The rest is for getting closer to GDI behaviour on Windows: Main culprit was the FT autohinter. With it disabled we are metric wise almost the same. (this doesn't affect TrueType Auto Hinting if available inside the font, which will be still used).

Though Maniacs in Wine looks completely off. Though can be likely altered in the Wine registry similar to how ClearType can be turned off.

(Before that patch WineManiac vs. EasyRPG was closer but imo the goal is to be closer metric wise to the Windows version ^^).

Test Map with fonts included:

[Archiv.zip](https://github.com/user-attachments/files/31649324/Archiv.zip)

Screenshots: The message box shows x/y/w/h per Pic. As you can see **the metrics match!!!**. Would never have expected that this will happen someday considering its GDI vs. FreeType but here we are. (the rendering is different in some parts but honestly both versions look awful so it doesn't matter xD)

With `Blank Font.ttf` by Gorurinbro. Top Left and Bottom Center are Bitmap Glyphs embedded in the Font thats why they look so good.

grafik

With Pro Font from #3427: Unscalable Font: Picks the closest matching.

grafik

### DIFF
--- a/src/font.cpp
+++ b/src/font.cpp
@@ -481,8 +481,10 @@ Rect FTFont::vGetSize(char32_t glyph) const {
 	FT_GlyphSlot slot = face->glyph;
 
 	Point advance;
-	advance.x = Utils::RoundTo<int>(slot->advance.x / 64.0);
-	advance.y = Utils::RoundTo<int>(slot->advance.y / 64.0);
+	// TrueType default rounding is "Round to Grid" (RTG) which matches the
+	// behaviour of std::round
+	advance.x = std::round(slot->advance.x / 64.0);
+	advance.y = std::round(slot->advance.y / 64.0);
 
 	if (EP_UNLIKELY(rm2000_workaround)) {
 		advance.x = 6;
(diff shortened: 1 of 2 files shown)


src/font.cpp
```diff
@@ -149,6 +149,25 @@ namespace {
 #ifdef HAVE_FREETYPE
 	FT_Library library = nullptr;
 
+	static bool IsBetterFontSize(int candidate, int current_best, int target_size) {
+		// When the font is not scalable this helper function is used to find
+		// the closest smaller font height. An exception is when their is no
+		// smaller in which cases it picks the minimum.
```
carstene1ns:
```suggestion
		// When the font is not scalable, this helper function is used to find
		
		// the closest smaller font height. When there is no smaller height,
		
		// it picks the minimum.
```

carstene1ns (approved):

carstene1ns:
Time to remove debain 11 and revamp cmake config for 3.22+...
https://endoflife.date/debian

PR #3685, opened 2026-09-20
Add tracker module file types to filefinder

RPGoogles:
Adds functionality so the player can actually find tracker module files(.mod, .xm, .s3m and .it) for Music and Sound.


src/filefinder.h
```diff
@@ -41,9 +41,9 @@
 namespace FileFinder {
 	constexpr const auto IMG_TYPES = Utils::MakeSvArray(".bmp", ".png", ".xyz");
 	constexpr const auto MUSIC_TYPES = Utils::MakeSvArray(
-			".opus", ".oga", ".ogg", ".wav", ".mid", ".midi", ".mp3", ".wma");
+			".opus", ".oga", ".ogg", ".wav", ".mid", ".midi", ".mp3", ".wma", ".mod", ".xm", ".s3m", ".it");
 	constexpr const auto SOUND_TYPES = Utils::MakeSvArray(
-			".opus", ".oga", ".ogg", ".wav", ".mp3", ".wma");
+			".opus", ".oga", ".ogg", ".wav", ".mp3", ".wma", ".mod", ".xm", ".s3m", ".it");
```
carstene1ns: IMO these should be music-only.

